Issue #4, opened 2018-06-22
trajectory controller tests fail


Karsten1987:
The trajectory controller tests currently fails with the change to the ROS2 master time/clock interface.

PR #7, opened 2019-08-08
Dashing update

Karsten1987:
connects https://github.com/ros-controls/ros2_control/issues/18

### DIFF
--- a/ros_controllers/src/joint_trajectory_controller.cpp
+++ b/ros_controllers/src/joint_trajectory_controller.cpp
@@ -226,7 +226,7 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous
   // TODO(karsten1987): create subscriber with subscription deactivated
   joint_command_subscriber_ =
     lifecycle_node_->create_subscription<trajectory_msgs::msg::JointTrajectory>(
-    "~/joint_trajectory", callback);
+    "~/joint_trajectory", rclcpp::SystemDefaultsQoS(), callback);
 
   // TODO(karsten1987): no lifecyle for subscriber yet
   // joint_command_subscriber_->on_activate();
--- a/ros_controllers/src/joint_trajectory_controller.cpp
+++ b/ros_controllers/src/joint_trajectory_controller.cpp
@@ -22,11 +22,16 @@
 #include <vector>
 
 #include "builtin_interfaces/msg/time.hpp"
+
 #include "lifecycle_msgs/msg/transition.hpp"
 #include "lifecycle_msgs/msg/state.hpp"
+
 #include "rclcpp/time.hpp"
+
 #include "rclcpp_lifecycle/state.hpp"
+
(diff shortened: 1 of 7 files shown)


ros_controllers/src/joint_trajectory_controller.cpp
```diff
@@ -22,133 +22,126 @@
...
 
 using namespace std::chrono_literals;
+using controller_interface::CONTROLLER_INTERFACE_RET_SUCCESS;
+using lifecycle_msgs::msg::State;
 
-namespace
-{
+JointTrajectoryController::JointTrajectoryController()
+: controller_interface::ControllerInterface(),
+  joint_names_({}),
```
piraka9011: This could just be `joint_names_{}` in `joint_trajectory_controller.hpp` (along with `write_op_names_{}`).

ros_controllers/test/test_trajectory_controller.cpp
```diff
@@ -195,47 +195,47 @@ TEST_F(TestTrajectoryController, configuration) {
...
-  test_robot->write();
-
-  // change in hw position
-  EXPECT_EQ(3.3, test_robot->pos1);
-  EXPECT_EQ(4.4, test_robot->pos2);
-  EXPECT_EQ(5.5, test_robot->pos3);
-
-  executor.cancel();
-}
+// TEST_F(TestTrajectoryController, activation) {
```
piraka9011: Any reason for removing this test?

Karsten1987: see #4

piraka9011: I saw the [TODO](https://github.com/ros-controls/ros2_controllers/blob/4322e8d0031f71d6e207fa8de5a48bb189615684/ros_controllers/src/trajectory.cpp#L27). I can create a PR and update when this lands then.

ros_controllers/src/joint_trajectory_controller.cpp
Karsten1987: we don't have any default initializers within this code base, so I would place these style changes for another set of PRs.

Karsten1987:
@piraka9011 @bmagyar this is up for review. 

We have to land these changes in order to make dashing work and branch off to master correctly.

Karsten1987:
@piraka9011 how come that all these commits are already on master? I don't see a PR attached to this? Do you push all the commits of this PR onto master?

piraka9011:
@Karsten1987 It looks like when making the master branch I created it off `dashing_update` and not `dashing` so that CI passes.
We can rebase master on to `dashing` instead and work from there.

Karsten1987:
the problem with this is that we have duplicate commits now.
Also, the current master seems to be broken: https://github.com/ros-controls/ros2_control/issues/42

Karsten1987:
CI came back green, merging.

PR #17, opened 2020-02-12
add differential drive controller

brightenlee:
The ros1 diff package was ported except for the urdf parser part.
I'm testing the ported package using my robot.
Maybe there is a lot to be fixed. Please check these.


bmagyar:
Well that escalated quickly :) I was about to write that I think we can base our "port" on your version, tests are much needed though

brightenlee:
I'll merge them again and request it.

PR #25, opened 2020-02-25
Add state_publish_rate to JointTrajectoryController

ddengster:
As per title.
Has additional dependency on ros-controls/realtime_tools, dashing-devel branch.

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -18,6 +18,8 @@ find_package(rclcpp REQUIRED)
 find_package(rclcpp_lifecycle REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(trajectory_msgs REQUIRED)
+find_package(realtime_tools REQUIRED)
+find_package(control_msgs REQUIRED)
 
 add_library(joint_trajectory_controller SHARED
   src/joint_trajectory_controller.cpp
@@ -33,6 +35,8 @@ ament_target_dependencies(joint_trajectory_controller
   "rclcpp_lifecycle"
   "rcutils"
   "trajectory_msgs"
+  "realtime_tools"
+  "control_msgs"
 )
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -12,14 +12,14 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(controller_interface REQUIRED)
+find_package(control_msgs REQUIRED)
 find_package(hardware_interface REQUIRED)
 find_package(pluginlib REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_lifecycle REQUIRED)
 find_package(rcutils REQUIRED)
-find_package(trajectory_msgs REQUIRED)
 find_package(realtime_tools REQUIRED)
-find_package(control_msgs REQUIRED)
+find_package(trajectory_msgs REQUIRED)
 
 add_library(joint_trajectory_controller SHARED
   src/joint_trajectory_controller.cpp
@@ -29,14 +29,14 @@ target_include_directories(joint_trajectory_controller PRIVATE include)
 ament_target_dependencies(joint_trajectory_controller
   "builtin_interfaces"
   "controller_interface"
+  "control_msgs"
   "hardware_interface"
   "pluginlib"
   "rclcpp"
   "rclcpp_lifecycle"
   "rcutils"
-  "trajectory_msgs"
   "realtime_tools"
-  "control_msgs"
+  "trajectory_msgs"
 )
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
@@ -76,10 +76,12 @@ if(BUILD_TESTING)
   )
   ament_target_dependencies(test_trajectory_controller
     "controller_parameter_server"
+    "control_msgs"
     "hardware_interface"
     "rclcpp"
     "rclcpp_lifecycle"
     "rcutils"
+    "realtime_tools"
     "test_robot_hardware"
     "trajectory_msgs"
   )
@@ -97,6 +99,7 @@ endif()
 
 ament_export_dependencies(
   controller_interface
+  control_msgs
   hardware_interface
   rclcpp
   rclcpp_lifecycle
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -21,7 +21,7 @@
 
 #include "controller_interface/controller_interface.hpp"
 
-#include <control_msgs/msg/joint_trajectory_controller_state.hpp>
+#include "control_msgs/msg/joint_trajectory_controller_state.hpp"
 
 #include "hardware_interface/joint_command_handle.hpp"
 #include "hardware_interface/joint_state_handle.hpp"
@@ -34,8 +34,8 @@
 #include "rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp"
 #include "rclcpp_lifecycle/state.hpp"
 
-#include <realtime_tools/realtime_buffer.h>
-#include <realtime_tools/realtime_publisher.h>
+#include "realtime_tools/realtime_buffer.h"
+#include "realtime_tools/realtime_publisher.h"
 
 #include "trajectory_msgs/msg/joint_trajectory.hpp"
 #include "trajectory_msgs/msg/joint_trajectory_point.hpp"
@@ -109,10 +109,10 @@ private:
 
   bool is_halted = false;
 
-  typedef control_msgs::msg::JointTrajectoryControllerState JTrajControllerStateMsg;
-  typedef realtime_tools::RealtimePublisher<JTrajControllerStateMsg> StatePublisher;
-  typedef std::unique_ptr<StatePublisher>                            StatePublisherPtr;
-  rclcpp_lifecycle::LifecyclePublisher<JTrajControllerStateMsg>::SharedPtr publisher_;
+  using ControllerStateMsg = control_msgs::msg::JointTrajectoryControllerState;
+  using StatePublisher = realtime_tools::RealtimePublisher<ControllerStateMsg>;
+  using StatePublisherPtr = std::unique_ptr<StatePublisher>;
+  rclcpp_lifecycle::LifecyclePublisher<ControllerStateMsg>::SharedPtr publisher_;
   StatePublisherPtr state_publisher_;
 
   rclcpp::Duration state_publisher_period_ = rclcpp::Duration(1.0 / 50.0);
@@ -121,6 +121,8 @@ private:
   bool reset();
   void set_op_mode(const hardware_interface::OperationMode & mode);
   void halt();
+
+  void publish_state();
 };
 
 }  // namespace joint_trajectory_controller
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -21,6 +21,8 @@
 
 #include "controller_interface/controller_interface.hpp"
 
+#include <control_msgs/msg/joint_trajectory_controller_state.hpp>
+
 #include "hardware_interface/joint_command_handle.hpp"
 #include "hardware_interface/joint_state_handle.hpp"
 #include "hardware_interface/operation_mode_handle.hpp"
@@ -31,6 +33,9 @@
 
 #include "rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp"
 #include "rclcpp_lifecycle/state.hpp"
+
+#include <realtime_tools/realtime_buffer.h>
+#include <realtime_tools/realtime_publisher.h>
 
 #include "trajectory_msgs/msg/joint_trajectory.hpp"
 #include "trajectory_msgs/msg/joint_trajectory_point.hpp"
@@ -103,6 +108,15 @@ private:
   std::shared_ptr<trajectory_msgs::msg::JointTrajectory> traj_msg_home_ptr_ = nullptr;
 
   bool is_halted = false;
+
+  typedef control_msgs::msg::JointTrajectoryControllerState JTrajControllerStateMsg;
+  typedef realtime_tools::RealtimePublisher<JTrajControllerStateMsg> StatePublisher;
+  typedef std::unique_ptr<StatePublisher>                            StatePublisherPtr;
+  rclcpp_lifecycle::LifecyclePublisher<JTrajControllerStateMsg>::SharedPtr publisher_;
+  StatePublisherPtr state_publisher_;
+
+  rclcpp::Duration state_publisher_period_ = rclcpp::Duration(1.0 / 50.0);
+  rclcpp::Time last_state_publish_time_;
 
   bool reset();
   void set_op_mode(const hardware_interface::OperationMode & mode);
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -69,7 +69,7 @@ JointTrajectoryController::init(
   // with the lifecycle node being initialized, we can declare parameters
   lifecycle_node_->declare_parameter<std::vector<std::string>>("joints", joint_names_);
   lifecycle_node_->declare_parameter<std::vector<std::string>>("write_op_modes", write_op_names_);
-  lifecycle_node_->declare_parameter<int>("state_publish_rate", 50);
+  lifecycle_node_->declare_parameter<double>("state_publish_rate", 50.0);
 
   return CONTROLLER_INTERFACE_RET_SUCCESS;
 }
@@ -85,7 +85,7 @@ JointTrajectoryController::update()
     return CONTROLLER_INTERFACE_RET_SUCCESS;
   }
 
-  //when we receive a traj msg
+  // when we receive a traj msg
   if (traj_point_active_ptr_ && (*traj_point_active_ptr_)->is_empty() == false) {
     // find next new point for current timestamp
     auto traj_point_ptr = (*traj_point_active_ptr_)->sample(lifecycle_node_->now());
@@ -109,33 +109,7 @@ JointTrajectoryController::update()
     set_op_mode(hardware_interface::OperationMode::ACTIVE);
   }
 
-  // publish state
-  {
-    if (state_publisher_period_.seconds() == 0.0 &&
-        (last_state_publish_time_ + state_publisher_period_) < lifecycle_node_->now())
-    {
-      if (state_publisher_ && state_publisher_->trylock())
-      {
-        last_state_publish_time_ = lifecycle_node_->now();
-
-        state_publisher_->msg_.header.stamp = last_state_publish_time_;
-        // TODO(ddengster): The following code is ported from ros_controllers. 
-        // Port it when we're ready to put in the other trajectory code.
-
-        // state_publisher_->msg_.header.stamp          = time_data_.readFromRT()->time;
-        // state_publisher_->msg_.desired.positions     = desired_state_.position;
-        // state_publisher_->msg_.desired.velocities    = desired_state_.velocity;
-        // state_publisher_->msg_.desired.accelerations = desired_state_.acceleration;
-        // state_publisher_->msg_.actual.positions      = current_state_.position;
-        // state_publisher_->msg_.actual.velocities     = current_state_.velocity;
-        // state_publisher_->msg_.error.positions       = state_error_.position;
-        // state_publisher_->msg_.error.velocities      = state_error_.velocity;
-
-        state_publisher_->unlockAndPublish();
-      }
-    }
-  }
-
+  publish_state();
   return CONTROLLER_INTERFACE_RET_SUCCESS;
 }
 
@@ -248,33 +222,36 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous
   // joint_command_subscriber_->on_activate();
 
   // State publisher
-  {
-    int state_publish_rate = 50;
-    if (lifecycle_node_->has_parameter("state_publish_rate")) {
-      state_publish_rate = lifecycle_node_->get_parameter("state_publish_rate").get_value<int>();
-      RCLCPP_INFO_STREAM(logger, "Controller state will be published at "
-                            << state_publish_rate << "Hz.");
-    }
-    state_publisher_period_ = rclcpp::Duration(1.0 / (double)state_publish_rate);
-
-    publisher_ = lifecycle_node_->create_publisher<JTrajControllerStateMsg>("state", rclcpp::SystemDefaultsQoS());
-    state_publisher_.reset(new StatePublisher(publisher_));
-
-    int n_joints = joint_names_.size();
-
-    state_publisher_->lock();
-    state_publisher_->msg_.joint_names = joint_names_;
-    state_publisher_->msg_.desired.positions.resize(n_joints);
-    state_publisher_->msg_.desired.velocities.resize(n_joints);
-    state_publisher_->msg_.desired.accelerations.resize(n_joints);
-    state_publisher_->msg_.actual.positions.resize(n_joints);
-    state_publisher_->msg_.actual.velocities.resize(n_joints);
-    state_publisher_->msg_.error.positions.resize(n_joints);
-    state_publisher_->msg_.error.velocities.resize(n_joints);
-    state_publisher_->unlock();
-
-    last_state_publish_time_ = lifecycle_node_->now();
+  double state_publish_rate =
+    lifecycle_node_->get_parameter("state_publish_rate").get_value<double>();
+  RCLCPP_INFO_STREAM(
+    logger, "Controller state will be published at " <<
+      state_publish_rate << "Hz.");
+  if (state_publish_rate > 0.0) {
+    state_publisher_period_ =
+      rclcpp::Duration::from_seconds(1.0 / state_publish_rate);
+  } else {
+    state_publisher_period_ = rclcpp::Duration(0.0);
   }
+
+  publisher_ = lifecycle_node_->create_publisher<ControllerStateMsg>(
+    "state", rclcpp::SystemDefaultsQoS());
+  state_publisher_.reset(new StatePublisher(publisher_));
+
+  int n_joints = joint_names_.size();
+
+  state_publisher_->lock();
+  state_publisher_->msg_.joint_names = joint_names_;
+  state_publisher_->msg_.desired.positions.resize(n_joints);
+  state_publisher_->msg_.desired.velocities.resize(n_joints);
+  state_publisher_->msg_.desired.accelerations.resize(n_joints);
+  state_publisher_->msg_.actual.positions.resize(n_joints);
+  state_publisher_->msg_.actual.velocities.resize(n_joints);
+  state_publisher_->msg_.error.positions.resize(n_joints);
+  state_publisher_->msg_.error.velocities.resize(n_joints);
+  state_publisher_->unlock();
+
+  last_state_publish_time_ = lifecycle_node_->now();
 
   set_op_mode(hardware_interface::OperationMode::INACTIVE);
 
@@ -382,6 +359,28 @@ JointTrajectoryController::halt()
   }
   set_op_mode(hardware_interface::OperationMode::ACTIVE);
 }
+
+void JointTrajectoryController::publish_state()
+{
+  if (state_publisher_period_.seconds() <= 0.0) {
+    return;
+  }
+
+  if (lifecycle_node_->now() < (last_state_publish_time_ + state_publisher_period_)) {
+    return;
+  }
+
+  if (state_publisher_ && state_publisher_->trylock()) {
+    last_state_publish_time_ = lifecycle_node_->now();
+
+    state_publisher_->msg_.header.stamp = last_state_publish_time_;
+    // TODO(ddengster): Fill in the rest of the state data.
+    // Port it when we're ready to put in the other trajectory code (ros2_controllers PR #26).
+
+    state_publisher_->unlockAndPublish();
+  }
+}
+
 
 }  // namespace joint_trajectory_controller
 
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -69,6 +69,7 @@ JointTrajectoryController::init(
   // with the lifecycle node being initialized, we can declare parameters
   lifecycle_node_->declare_parameter<std::vector<std::string>>("joints", joint_names_);
   lifecycle_node_->declare_parameter<std::vector<std::string>>("write_op_modes", write_op_names_);
+  lifecycle_node_->declare_parameter<int>("state_publish_rate", 50);
 
   return CONTROLLER_INTERFACE_RET_SUCCESS;
 }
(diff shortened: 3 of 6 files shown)


ddengster:
@Karsten1987 Could use a review :)

bmagyar:
@ddengster could you please opt for rebasing on upstream master as opposed to merging? it keeps the review diff way cleaner

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -221,6 +246,37 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous
...
+  // State publisher
+  {
+    int state_publish_rate = 50;
+    if (lifecycle_node_->has_parameter("state_publish_rate")) {
+      state_publish_rate = lifecycle_node_->get_parameter("state_publish_rate").get_value<int>();
+      RCLCPP_INFO_STREAM(
+        logger, "Controller state will be published at " <<
+          state_publish_rate << "Hz.");
+    }
+    state_publisher_period_ = rclcpp::Duration(1.0 / static_cast<double>(state_publish_rate));
```
bmagyar: Can we fetch this as a double (who says Hz has to be integer?) to avoid that `static_cast` and let's add a non-zero positive check here, otherwise we may create a black hole

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -221,6 +246,37 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous
   // TODO(karsten1987): no lifecyle for subscriber yet
   // joint_command_subscriber_->on_activate();
 
+  // State publisher
+  {
+    int state_publish_rate = 50;
+    if (lifecycle_node_->has_parameter("state_publish_rate")) {
```
bmagyar: you set a default when declaring the param, there is no need to hardcode the default rate here for a second time nor to check if it exists.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -221,6 +246,37 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous
...
+      state_publish_rate = lifecycle_node_->get_parameter("state_publish_rate").get_value<int>();
+      RCLCPP_INFO_STREAM(
+        logger, "Controller state will be published at " <<
+          state_publish_rate << "Hz.");
+    }
+    state_publisher_period_ = rclcpp::Duration(1.0 / static_cast<double>(state_publish_rate));
+
+    publisher_ = lifecycle_node_->create_publisher<JTrajControllerStateMsg>(
+      "state", rclcpp::SystemDefaultsQoS());
+    state_publisher_.reset(new StatePublisher(publisher_));
```
bmagyar: Is there a reason for creating a copy of a local object only to shove it into a unique pointer?

How about this, instead of these two lines?

`state_publisher_ = std::make_unique(lifecycle_node_->create_publisher<ControllerStateMsg>("state", rclcpp::SystemDefaultsQoS()));`

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -104,6 +109,15 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
 
   bool is_halted = false;
 
+  typedef control_msgs::msg::JointTrajectoryControllerState JTrajControllerStateMsg;
+  typedef realtime_tools::RealtimePublisher<JTrajControllerStateMsg> StatePublisher;
+  typedef std::unique_ptr<StatePublisher> StatePublisherPtr;
+  rclcpp_lifecycle::LifecyclePublisher<JTrajControllerStateMsg>::SharedPtr publisher_;
+  StatePublisherPtr state_publisher_;
+
+  rclcpp::Duration state_publisher_period_ = rclcpp::Duration(1.0 / 50.0);
```
bmagyar: Please add some tests where this param is set to 10, you bring up the controller and validate you get state feedback ~10Hz and one where it is set to 0 and you validate that there is nothing published on the state topic.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: So there's a problem with casting to doubles with parameters, I reported it.
https://github.com/ros2/rclcpp/issues/1009

As a workaround, I used integers.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: There's some confusion here I think?; `publisher_` is of type LifecyclePublisher, and `StatePublisher` is of type RealtimePublisher, so there's no local copy.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: Ditto, found a better solution.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
ddengster: Added test in [1cb140c](https://github.com/ros-controls/ros2_controllers/pull/25/commits/1cb140c3501bbb0b7ecad761ae4e53579132ad16)

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: Fixed in [23e7b0d ](https://github.com/ros-controls/ros2_controllers/pull/25/commits/23e7b0de8c293346934a15ab088579e6a28beea1)

ddengster:
> @ddengster could you please opt for rebasing on upstream master as opposed to merging? it keeps the review diff way cleaner

My bad, I'm new to using git in this manner of collaboration. I'll be sure to use something like `git pull --rebase` next time. Should I expect to do a git rebase that squashes history before this goes into the main branch?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -84,31 +85,55 @@ JointTrajectoryController::update()
...
-  for (size_t index = 0; index < joint_num; ++index) {
-    registered_joint_cmd_handles_[index]->set_cmd(traj_point_ptr->positions[index]);
+    prev_traj_point_ptr_ = traj_point_ptr;
+    set_op_mode(hardware_interface::OperationMode::ACTIVE);
   }
 
-  prev_traj_point_ptr_ = traj_point_ptr;
-  set_op_mode(hardware_interface::OperationMode::ACTIVE);
+  // publish state
+  {
```
piraka9011: This block could be extracted into a separate function

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -84,31 +85,55 @@ JointTrajectoryController::update()
...
+        // state_publisher_->msg_.header.stamp          = time_data_.readFromRT()->time;
+        // state_publisher_->msg_.desired.positions     = desired_state_.position;
+        // state_publisher_->msg_.desired.velocities    = desired_state_.velocity;
+        // state_publisher_->msg_.desired.accelerations = desired_state_.acceleration;
+        // state_publisher_->msg_.actual.positions      = current_state_.position;
+        // state_publisher_->msg_.actual.velocities     = current_state_.velocity;
+        // state_publisher_->msg_.error.positions       = state_error_.position;
+        // state_publisher_->msg_.error.velocities      = state_error_.velocity;
+
+        state_publisher_->unlockAndPublish();
```
piraka9011: Do we still need this to be scoped if we're manually holding/releasing the mutex?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -221,6 +246,39 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous
...
+  {
+    int state_publish_rate = lifecycle_node_->get_parameter("state_publish_rate").get_value<int>();
+    RCLCPP_INFO_STREAM(
+      logger, "Controller state will be published at " <<
+        state_publish_rate << "Hz.");
+    if (state_publish_rate > 0) {
+      state_publisher_period_ =
+        rclcpp::Duration::from_seconds(1.0 / static_cast<double>(state_publish_rate));
+    } else {
+      state_publisher_period_ = rclcpp::Duration(0.0);
```
piraka9011: In `JointTrajectoryController::update()` we check `if (state_publisher_period_.seconds() > 0.0)`.
This line here means we won't perform any update when a negative rate is passed. Is that the intended behavior?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -441,3 +441,121 @@ TEST_F(TestTrajectoryController, correct_initialization_using_parameters) {
...
+    10,
+    [&](Msg::UniquePtr msg) {
+      (void)msg;
+      ++echo_received_counter;
+      // RCLCPP_INFO(rclcpp::get_logger("test_state_publisher"), "%d", echo_received_counter);
+    }
+  );
+
+  // wait for things to setup
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
piraka9011: Why wait for this amount? My concern is that `sleep_for`s will lead to flakey tests.
Is there a better way to check that the resources are ready?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -441,3 +441,121 @@ TEST_F(TestTrajectoryController, correct_initialization_using_parameters) {
...
+
+  // update for 1second
+  auto start_time = rclcpp::Clock().now();
+  rclcpp::Duration wait = rclcpp::Duration::from_seconds(1.0);
+  auto end_time = start_time + wait;
+  while (rclcpp::Clock().now() < end_time) {
+    traj_controller->update();
+  }
+
+  EXPECT_EQ(10, echo_received_counter);
```
piraka9011: Can we extract the `10` as a constant since it's used throughout the test.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -441,3 +441,121 @@ TEST_F(TestTrajectoryController, correct_initialization_using_parameters) {
...
+
+TEST_F(TestTrajectoryController, test_state_publish_rate) {
+  auto traj_controller = std::make_shared<joint_trajectory_controller::JointTrajectoryController>();
+  auto ret = traj_controller->init(test_robot, controller_name);
+  if (ret != controller_interface::CONTROLLER_INTERFACE_RET_SUCCESS) {
+    FAIL();
+  }
+
+  // fill in some data so we wont fail
+  auto traj_lifecycle_node = traj_controller->get_lifecycle_node();
```
piraka9011: Consider having you're own test fixture where you do all this setup so you don't have duplicate code.

travis/workspace.repos
```diff
@@ -3,3 +3,11 @@ repositories:
     type: git
     url: https://github.com/ros-controls/ros2_control.git
     version: master
+  control_msgs:
+    type: git
+    url: https://github.com/ros-controls/control_msgs.git
+    version: crystal-devel
```
piraka9011: Is this intended? @bmagyar can we get a `dashing` and `foxy` branch?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: Yes. That being said, we could improve it to throw an error, though inputting a negative rate is so bewildering that we can say whoever writes it probably wants to save keystrokes or something. I'd allow that flexibility.

travis/workspace.repos
ddengster: There are no dashing/foxy branches on https://github.com/ros-controls/control_msgs

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: I'm an advocate for not exposing extra functions to a higher/wider level namespace unless needed. That being said, I'll do so if many more insists on it.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: Yes, we don't need it. This scoping is a code style thing. See [reply](https://github.com/ros-controls/ros2_controllers/pull/25#discussion_r386767757) above.

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: Not that I know of.

Other tests are using it as well, see [this](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/test/test_trajectory_controller.cpp#L330). We should do the fix (if we find it) in a single pass.

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: Fixed in [505f8a2](https://github.com/ros-controls/ros2_controllers/pull/25/commits/505f8a2a6d1d985582501bc9492236ae8805e710)

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: Fixed in [505f8a2](https://github.com/ros-controls/ros2_controllers/pull/25/commits/505f8a2a6d1d985582501bc9492236ae8805e710)

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: Caught up with the issue and fixed in [0212a21](https://github.com/ros-controls/ros2_controllers/pull/25/commits/0212a2104627ec58052144bc839496bfdc14d81a)

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: There's nothing wrong with creating a private function within the class, that wouldn't expose it, only keep it a little more tidy.

joint_trajectory_controller/test/test_trajectory_controller.cpp
bmagyar: Indeed. We had a lot of issues in `ros_controllers` with flaky tests, best to root them out early on!

travis/workspace.repos
bmagyar: There were no changes made since the ROS2 port of `control_msgs`, hence why it is still on `crystal-devel`.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: The thing is I have to scroll/jump to the function via editor, which is likely a number of lines away (in larger codebases it gets annoying as the number of these 'only used in one place functions' increases), thus breaking concentration quite abit.

But anyway in the interests of keeping things moving along, fixed in [6e00b03](https://github.com/ros-controls/ros2_controllers/pull/25/commits/6e00b03b0e84207812733fcbdfee1f436c95cd60)

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -104,6 +109,15 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
 
   bool is_halted = false;
 
+  typedef control_msgs::msg::JointTrajectoryControllerState JTrajControllerStateMsg;
```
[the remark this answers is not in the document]
ddengster: [d3065b6](https://github.com/ros-controls/ros2_controllers/pull/25/commits/d3065b66f50879b91219ffc4cad4a71cbe2451ef)

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -104,9 +111,19 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
...
+  StatePublisherPtr state_publisher_;
+
+  rclcpp::Duration state_publisher_period_ = rclcpp::Duration(1.0 / 50.0);
+  rclcpp::Time last_state_publish_time_;
+
   bool reset();
   void set_op_mode(const hardware_interface::OperationMode & mode);
   void halt();
+
+  void publishState();
```
Karsten1987: please adhere to the code style in this class. `publish_state()`.

joint_trajectory_controller/CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ find_package(rclcpp REQUIRED)
 find_package(rclcpp_lifecycle REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(trajectory_msgs REQUIRED)
+find_package(realtime_tools REQUIRED)
+find_package(control_msgs REQUIRED)
```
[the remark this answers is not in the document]
Karsten1987: same below

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -69,6 +69,7 @@ JointTrajectoryController::init(
   // with the lifecycle node being initialized, we can declare parameters
   lifecycle_node_->declare_parameter<std::vector<std::string>>("joints", joint_names_);
   lifecycle_node_->declare_parameter<std::vector<std::string>>("write_op_modes", write_op_names_);
+  lifecycle_node_->declare_parameter<double>("state_publish_rate", 50.0);
```
Karsten1987: why do you think an update rate of `50hz` for the current state is needed?
Thinking forward, the preferred way of triggering a new trajectory is by calling an action service, which has in built feedback mechanism. Wouldn't that suffice?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -84,32 +85,31 @@ JointTrajectoryController::update()
...
+    size_t joint_num = registered_joint_cmd_handles_.size();
+    for (size_t index = 0; index < joint_num; ++index) {
+      registered_joint_cmd_handles_[index]->set_cmd(traj_point_ptr->positions[index]);
+    }
 
-  size_t joint_num = registered_joint_cmd_handles_.size();
-  for (size_t index = 0; index < joint_num; ++index) {
-    registered_joint_cmd_handles_[index]->set_cmd(traj_point_ptr->positions[index]);
+    prev_traj_point_ptr_ = traj_point_ptr;
+    set_op_mode(hardware_interface::OperationMode::ACTIVE);
```
Karsten1987: In fact, I generally disagree with this _change_. It's only really indentation and whitespaces. The only diff here should be the call to `publish_state()`.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
Karsten1987: I'd advocate for printing an error message and return an error here if negative. It's clearly negatively configured and the lifecycle should not continue in this case.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -325,6 +362,33 @@ JointTrajectoryController::halt()
...
+      // Port it when we're ready to put in the other trajectory code.
+
+      // state_publisher_->msg_.header.stamp          = time_data_.readFromRT()->time;
+      // state_publisher_->msg_.desired.positions     = desired_state_.position;
+      // state_publisher_->msg_.desired.velocities    = desired_state_.velocity;
+      // state_publisher_->msg_.desired.accelerations = desired_state_.acceleration;
+      // state_publisher_->msg_.actual.positions      = current_state_.position;
+      // state_publisher_->msg_.actual.velocities     = current_state_.velocity;
+      // state_publisher_->msg_.error.positions       = state_error_.position;
+      // state_publisher_->msg_.error.velocities      = state_error_.velocity;
```
Karsten1987: Do you want to open a follow-up ticket for this? I don't see a big point in having commented code checked in.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -441,3 +441,82 @@ TEST_F(TestTrajectoryController, correct_initialization_using_parameters) {
...
+
+  traj_controller->on_configure(traj_lifecycle_node->get_current_state());
+  traj_controller->on_activate(traj_lifecycle_node->get_current_state());
+
+  auto future_handle = std::async(
+    std::launch::async, [&executor]() -> void {
+      executor.spin();
+    });
+
+  typedef joint_trajectory_controller::JointTrajectoryController::ControllerStateMsg Msg;
```
Karsten1987: is that typedef really shorter than it's original?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -441,3 +441,82 @@ TEST_F(TestTrajectoryController, correct_initialization_using_parameters) {
...
+
+  static const int qos_level = 10;
```
Karsten1987: why is this `static`?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -441,3 +441,82 @@ TEST_F(TestTrajectoryController, correct_initialization_using_parameters) {
...
+  int echo_received_counter = 0;
+  rclcpp::Subscription<Msg>::SharedPtr subs = traj_lifecycle_node->create_subscription<Msg>(
+    "/state",
+    qos_level,
+    [&](Msg::UniquePtr msg) {
+      (void)msg;
+      ++echo_received_counter;
+      // RCLCPP_INFO(rclcpp::get_logger("test_state_publisher"), "%d", echo_received_counter);
```
Karsten1987: do you want this code to be part of the test or can this be deleted?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -83,6 +88,8 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
   on_shutdown(const rclcpp_lifecycle::State & previous_state) override;
 
+  typedef control_msgs::msg::JointTrajectoryControllerState ControllerStateMsg;
+
```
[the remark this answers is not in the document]
ddengster: See https://github.com/ros-controls/ros2_controllers/pull/25#discussion_r386113625.
One of the tests uses it to subscribe to messages of the same type.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -84,32 +85,31 @@ JointTrajectoryController::update()
     return CONTROLLER_INTERFACE_RET_SUCCESS;
   }
 
-  // when no traj msg has been received yet
-  if (!traj_point_active_ptr_ || (*traj_point_active_ptr_)->is_empty()) {
-    return CONTROLLER_INTERFACE_RET_SUCCESS;
-  }
```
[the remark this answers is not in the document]
ddengster: So am I on the early returns. But I needed publish the state despite there being no current goal, and so the changes turned out that way.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: feedback only occurs when there is a current action goal. state publishing publishes despite there being no action goal.

50hz was the previous setting in ros_controllers.

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: Initialized only once, though the compiler will likely optimize it out

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: I will likely put the code in with the action server; it's the part I need to get the numbers correct for my purposes. https://github.com/ros-controls/ros2_controllers/pull/26

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: It's test code and not widely used, so I gave a little leeway on this. The typedef in the header can stay that way.

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: Hm if someone else wants to debug this code they can uncomment it? No harm leaving it in imo.

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: initialize once. Though compiler optimizations will take care of that. It's a habit of mine though, I'll remove it.

joint_trajectory_controller/CMakeLists.txt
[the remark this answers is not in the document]
ddengster: Fixed (b49e51c)[https://github.com/ros-controls/ros2_controllers/pull/25/commits/b49e51cd429f46d59f584a283c183aeb4163ed5a]

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: see comment above

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -221,6 +221,40 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous
   // TODO(karsten1987): no lifecyle for subscriber yet
   // joint_command_subscriber_->on_activate();
 
+  // State publisher
+  {
```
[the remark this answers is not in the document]
ddengster: Style readability thing, discussed above. Fixed in [91e5549](https://github.com/ros-controls/ros2_controllers/pull/25/commits/91e554965480f76db58d12428d5abbd45c05ecb4)

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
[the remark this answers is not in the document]
bmagyar: It is within the class (therefore namespace) `JointTrajectoryController`. Let's not blindly follow the overly verbose practices of ROS2 codebases but apply some common sense. 

I agree with the comment on not putting this public though... We never intended this to be part of a public interface, please move it into the private section and use it to shorten variable declarations if you like but the test should define it's own `typedef` or `using`.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: Which is perfect, please remove the commented code from this PR.

joint_trajectory_controller/test/test_trajectory_controller.cpp
bmagyar: please either remove or change it to `RCLCPP_DEBUG` then

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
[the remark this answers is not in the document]
Karsten1987: +1 for `using` in the first place. Applies to all `typedef`s below.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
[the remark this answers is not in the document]
Karsten1987: why not calling `publish_state()` before that logic then?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -325,6 +360,33 @@ JointTrajectoryController::halt()
   set_op_mode(hardware_interface::OperationMode::ACTIVE);
 }
 
+void JointTrajectoryController::publish_state()
+{
+  if (state_publisher_period_.seconds() > 0.0 &&
```
Karsten1987: can this ever happen when being check during initialization?

Karsten1987: Also again only style, but I would also opt for early return here to avoid the nested `if`s

joint_trajectory_controller/test/test_trajectory_controller.cpp
Karsten1987: With the comment of not having the `typedef` in the public header, this should be addressed.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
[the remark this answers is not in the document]
ddengster: Because i need the results of the sampled trajectory. In PR #26 action feedback will also be given and I'll need to rearrange this update function anyway.

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: Fixed [b8e4091](https://github.com/ros-controls/ros2_controllers/pull/25/commits/b8e4091bebfdbbd91b297976afebea7328d0ae40)

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: Fixed [b8e4091](https://github.com/ros-controls/ros2_controllers/pull/25/commits/b8e4091bebfdbbd91b297976afebea7328d0ae40)

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: yes, giving negative/zero state_publish_rate resolves its results to state_publisher_period_ being 0.
f668d37

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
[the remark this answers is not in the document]
ddengster: fixed b8e4091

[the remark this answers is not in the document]
bmagyar: I think not. I think what Karsten meant was to replace all `typedef A::B XY` with `using XY = A::B`. On top of being more intuitive by following a format akin to variable assignment, this notation can also handle template parameters.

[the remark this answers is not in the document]
ddengster: my bad, was confusing it with the test typedef, fix in 03ff135

piraka9011:
Friendly reminder to rebase on master. Helps make it easier to review the diff.
Don't mind squashing extra commits (like those for styling).
```
git checkout master
git pull
git checkout trajectory_state_publish
git rebase master
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -325,6 +360,37 @@ JointTrajectoryController::halt()
...
+  if (lifecycle_node_->now() < (last_state_publish_time_ + state_publisher_period_)) {
+    return;
+  }
+
+  if (state_publisher_ && state_publisher_->trylock()) {
+    last_state_publish_time_ = lifecycle_node_->now();
+
+    state_publisher_->msg_.header.stamp = last_state_publish_time_;
+    // TODO(ddengster): The following code is ported from ros_controllers.
+    // Port it when we're ready to put in the other trajectory code.
```
bmagyar: don't push commented code blocks please

Karsten1987: I feel we had this comment here already. I'd recommend opening a follow up issue for it.

ddengster: removed in 99c036d

ddengster:
Alright, I've put in the fixes and I've cleaned up the commit log. let me know if there's something else I need to look out for or do.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -221,6 +221,38 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous
...
+  if (state_publish_rate > 0.0) {
+    state_publisher_period_ =
+      rclcpp::Duration::from_seconds(1.0 / state_publish_rate);
+  } else {
+    state_publisher_period_ = rclcpp::Duration(0.0);
+  }
+
+  publisher_ = lifecycle_node_->create_publisher<ControllerStateMsg>(
+    "state", rclcpp::SystemDefaultsQoS());
+  state_publisher_.reset(new StatePublisher(publisher_));
```
Karsten1987: this should be `std::make_unique`

bmagyar:
```suggestion
  state_publisher_ = std::make_unique<StatePublisher>(publisher_);
```

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9653/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5337/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7883/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9552/)

bmagyar:
All happy, merging!

Karsten1987:
@bmagyar how is this all happy? The windows build has quite a number of warnings introduced by this build. 

@ddengster do you mind addressing these issues please?

piraka9011:
[`Duration`](https://github.com/ros2/rclcpp/blob/7c1721a0b390be8242a6b824489d0bc861f6a0ad/rclcpp/src/rclcpp/duration.cpp#L34) takes seconds and nanoseconds so `1/50` should be converted into nanoseconds.

[`rcutils` has some macros](https://github.com/ros2/rcutils/blob/e7eb74c7a11cf207aa9330d77f44098339528cac/include/rcutils/time.h#L32) that can help with this.

bmagyar:
This PR has been dangling around for way too long with other work depending on it that can already be reviewed. I agree that we should fix all warnings from all builds but this can be done in a follow up PR.

Please don't panic merge these follow-up PRs either but let's give them proper reviews.

Karsten1987:
> This PR has been dangling around for way too long with other work depending on it that can already be reviewed. I agree that we should fix all warnings from all builds but this can be done in a follow up PR.

Let's just agree that we don't merge PRs if there are warnings/regression visible on CI. Then we also don't need any _panic_ follow-ups ;-)

bmagyar:
Please point me to any test regressions. 
If you want to treat warnings as errors, modern compilers have a setting for this (eg `-Werror`) which in turn should appropriately communicate how blocking they are.

PR #28, opened 2020-03-11
Add GitHub Action build workflow

piraka9011:
Add the build and test GitHub Action workflow.
This PR will be follow up by one for linting and one for codecov.
Added the `ros2.repos` list so that it includes the repos used by Travis as well.

Signed-off-by: Anas Abou Allaban <allabana@amazon.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,12 +15,13 @@ jobs:
       fail-fast: false
     steps:
       - uses: ros-tooling/setup-ros@0.0.15
-      - uses: ros-tooling/action-ros-ci@0.0.13
+      - uses: ros-tooling/action-ros-ci@master
         with:
           package-name: joint_state_controller joint_trajectory_controller
           colcon-mixin-name: coverage-gcc
           colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
-          vcs-repo-file-url: "${{ github.workspace }}/.github/workflows/ros2.repos"
+          vcs-repo-file-url: |
+            "${{ github.workspace }}/travis/workspace.repos"
       - uses: actions/upload-artifact@master
         with:
           name: colcon-logs-${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


Karsten1987:
I am not sure I follow why we need the `ros2.repos` file. I would prefer not hardcoding this list as it might change. Is there a way we can use the upstream one from `ros2/ros2`?

piraka9011:
> I would prefer not hardcoding this list as it might change. 

I agree, this is a WIP to verify that having this repo file does in fact result in a successful build. I will update this to use the workflow suggested [here](https://github.com/ros-tooling/action-ros-ci#build-with-a-custom-repos-or-rosinstall-file):

```yaml
# Run the generator and output the results to a file.
- run: |
    rosinstall_generator <package-name> --rosdistro <target-distro> \
    --deps-only --deps --upstream-development > /tmp/deps.repos
```

Karsten1987:
@piraka9011 any news on this? I would be happy to have at least the set of linters be part of the repo in order to iterate faster on existing PRs.

piraka9011:
Sorry about the delay, see #33.

bmagyar:
Is this ready for merging? The ros2 repos files is a little unfortunate but do you see how we could rid of it in the future?

piraka9011:
> Is this ready for merging? The ros2 repos files is a little unfortunate but do you see how we could rid of it in the future?

Let me try the rosinstall_generator workflow. Otherwise, I'll use the repos file.

bmagyar:
Any conclusion on this?

Karsten1987:
There's is this on the way: https://github.com/ros-tooling/action-ros-ci/pull/121
That would be a little bit more handy as we had to specify the ros2 upstream repos file plus an overlay one which only holds the ros-controls packages.

.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
...
+      fail-fast: false
+    steps:
+      - uses: ros-tooling/setup-ros@0.0.15
+      - uses: ros-tooling/action-ros-ci@master
+        with:
+          package-name: joint_state_controller joint_trajectory_controller
+          colcon-mixin-name: coverage-gcc
+          colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
+          vcs-repo-file-url: |
+            ${{ github.workspace }}/travis/workspace.repos
```
Karsten1987, line 24: I still think it makes more sense to specify the ros2 master upstream repos here, followed by a dedicated one for ros2_controllers. 
I would further make this travis independent, as it might get decommissioned at some point.

PR #33, opened 2020-03-23
Add linting workflows

piraka9011:
* Add GitHub Action linting workflows.

Signed-off-by: Anas Abou Allaban <aabouallaban@pm.me>


piraka9011:
> LGTM although it'd be nice if github already ran jobs before merging

It can, but not when it's coming from a fork.

PR #35, opened 2020-04-12
diff_drive_controller

ManuelMeraz:
I decided to write code for my rover in ros 2 and apparently a diff drive controller hadn't been implemented yet so I decided to make one. This one is very basic and only does open loop control, but wouldn't be very hard to extend. I decided this was a good enough pull request to have some base code for a diff drive controller. 

I'll be improving this controller as I get some odometry feedback and need to update odometry in my robot. 

I wrote a test as well, but the test is very basic and mainly tests the core functionally and not much logic behind the twist messsages being sent. 

I'm currently using ros dashing for my coding, so I couldn't use the code in the master branch.


bmagyar:
Hope you don't mind if we don't get to merge this. @piraka9011 has his port almost ready but we would be glad if you could give us a review on that upcoming PR!

ManuelMeraz:
Yeah that's not a problem! If someone is already further ahead, then that's better. Could you link me to his branch? I'd like to use his diff drive controller if it's better. I'd actually prefer not to implement this, but was planning because I had to.

piraka9011:
I think this is fine actually. I don't mind working on top of this/doing some refactoring to improve the code + tests.

.gitignore
```diff
@@ -0,0 +1,44 @@
...
+# Compiled Static libraries
+*.lai
+*.la
+*.a
+*.lib
+
+# Executables
+*.exe
+*.out
+*.app
```
piraka9011, line 32: Are all these necessary?

ManuelMeraz: Nope, not necessary. It was there simply for convenience. I use git status a lot and seeing a clean status makes me feel good :). I can remove and push if you'd like.

ManuelMeraz: Oh just caught where you put the message. And no they're not! I got this ignore from the standard cpp ignore file github generates and then added some myself. I'lll go ahead and remove these.

ManuelMeraz: Fixed.

ros_controllers/CMakeLists.txt
```diff
@@ -83,15 +84,36 @@ if(BUILD_TESTING)
...
+          test_diff_drive_controller test/test_diff_drive_controller.cpp
+          ENV config_file=${CMAKE_CURRENT_SOURCE_DIR}/test/config/test_diff_drive_controller.yaml)
+  target_include_directories(test_diff_drive_controller PRIVATE include)
+  target_link_libraries(
+          test_diff_drive_controller
+          default_controllers
+  )
+
+  ament_target_dependencies(
+          test_diff_drive_controller
```
bmagyar, line 108: Let's reserve "" for strings please

ros_controllers/CMakeLists.txt
```diff
@@ -83,15 +84,36 @@ if(BUILD_TESTING)
...
-    "controller_interface"
-    "controller_parameter_server"
-    "hardware_interface"
-    "lifecycle_msgs"
-    "rclcpp"
-    "rcutils"
-    "test_robot_hardware"
-    "trajectory_msgs"
+          test_trajectory_controller
+          "controller_interface"
```
bmagyar, line 88: let's reserve "" for strings please

bmagyar:
Ok @piraka9011 please give this a review and let me know when it is ready for a merge.

Let's keep it snappy, I'd like to merge it asap and then you should move this branch over to master where we have all CI set up. It should involve more refactoring required by all the linters but it hsould be fine. Also a new package will be required. I'm happy to give you a priority review on that one

ManuelMeraz:
I'm currently building ros2 foxy. Pretty new to ros 2, so I didn't even know how to build to the latest until just now. I'm currently in the process of porting it to be compatible with the latest build according to this documentation: https://index.ros.org/doc/ros2/Installation/Foxy/Linux-Development-Setup/#linux-latest

I'm hoping master will build off that.

ros_controllers/CMakeLists.txt
```diff
@@ -27,25 +27,26 @@ add_library(
...
-  "controller_manager"
-  "controller_parameter_server"
-  "hardware_interface"
-  "rclcpp"
-  "rclcpp_lifecycle"
-  "rcutils"
-  "sensor_msgs"
-  "trajectory_msgs"
+        default_controllers
+        builtin_interfaces
```
piraka9011, line 34: nit: 2 spaces

ros_controllers/include/ros_controllers/diff_drive_controller.hpp
```diff
@@ -15,69 +15,104 @@
...
+   std::vector<WheelHandle> registered_left_wheel_handles_{};
+   std::vector<WheelHandle> registered_right_wheel_handles_{};
 
-  controller_interface::controller_interface_ret_t init(
-    std::weak_ptr<hardware_interface::RobotHardware> robot_hardware,
-    const std::string & controller_name) override;
+   struct WheelParams
+   {
+      size_t wheels_per_side{1};
+      double separation{0.21}; // w.r.t. the midpoint of the wheel width
```
piraka9011, line 87:
```suggestion
      double wheel_separation{0.21}; // w.r.t. the midpoint of the wheel width
```

ros_controllers/include/ros_controllers/diff_drive_controller.hpp
```diff
@@ -15,69 +15,104 @@
...
-  controller_interface::controller_interface_ret_t update() override;
+   // Timeout to consider cmd_vel commands old
+   std::chrono::milliseconds cmd_vel_timeout_{500};
+   std::vector<double> left_previous_commands_{};
+   std::vector<double> right_previous_commands_{};
 
-private:
-  std::string name_{};
+   std::vector<std::string> write_op_names_{};
+   std::vector<hardware_interface::OperationModeHandle*> registered_operation_mode_handles_{};
```
piraka9011, line 100: nit: spacing here. consider running `ament_uncrustify --reformat <dir>`

ros_controllers/src/diff_drive_controller.cpp
```diff
@@ -14,12 +14,272 @@
 
 #include "ros_controllers/diff_drive_controller.hpp"
 
-namespace
+#include <lifecycle_msgs/msg/state.hpp>
+#include <utility>
+
+namespace {
+constexpr auto DEFAULT_COMMAND_TOPIC = "~/cmd_vel";
+}
```
piraka9011, line 22:
```suggestion
}  // namespace
```

ros_controllers/src/diff_drive_controller.cpp
```diff
@@ -14,12 +14,272 @@
 
 #include "ros_controllers/diff_drive_controller.hpp"
 
-namespace
+#include <lifecycle_msgs/msg/state.hpp>
+#include <utility>
```
piraka9011, line 18: Make sure to follow the [Google C++ style guide for include headers](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

These two should be above line#15 as:

```
#include <utility>
#include <lifecycle_msgs/msg/state.hpp>
```

ros_controllers/src/diff_drive_controller.cpp
```diff
@@ -14,12 +14,272 @@
...
+
+   return CallbackReturn::SUCCESS;
+}
+
+rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+DiffDriveController::on_activate(const rclcpp_lifecycle::State&)
+{
+   is_halted = false;
+   subscriber_is_active_ = true;
+   //   traj_point_active_ptr_ = &traj_external_point_ptr_;
```
piraka9011, line 213:
```suggestion

```

ros_controllers/src/diff_drive_controller.cpp
```diff
@@ -14,12 +14,272 @@
...
+
+   // Apply (possibly new) multipliers:
+   const auto wheels = wheel_params_;
+
+   const double wheel_separation = wheels.separation_multiplier * wheels.separation;
+   const double left_wheel_radius = wheels.left_radius_multiplier * wheels.radius;
+   const double right_wheel_radius = wheels.right_radius_multiplier * wheels.radius;
+
+   // Compute wheels velocities:
+   const auto& current_command = *velocity_msg_ptr_;
```
piraka9011, line 86:
```suggestion
   const auto& current_command = velocity_msg_ptr_.get();
```

Also verify the velocity is not null.

ros_controllers/src/diff_drive_controller.cpp
```diff
@@ -14,12 +14,272 @@
...
-  constexpr const char kDefaultCmdVelTopic[] = "cmd_vel";
+   subscriber_is_active_ = false;
+   return CallbackReturn::SUCCESS;
 }
 
-namespace ros_controllers
+rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+DiffDriveController::on_cleanup(const rclcpp_lifecycle::State&)
 {
+   velocity_msg_ptr_ = std::make_shared<Twist>();
```
piraka9011, line 227: use [`reset()`](https://en.cppreference.com/w/cpp/memory/shared_ptr/reset)

ros_controllers/test/test_diff_drive_controller.cpp
```diff
@@ -0,0 +1,270 @@
...
+#include "ros_controllers/diff_drive_controller.hpp"
+#include "std_msgs/msg/string.hpp"
+#include "test_robot_hardware/test_robot_hardware.hpp"
+
+#include "gtest/gtest.h"
+#include <array>
+#include <memory>
+#include <string>
+#include <thread>
+#include <vector>
```
piraka9011, line 29:
```suggestion
#include "gtest/gtest.h"

#include <array>
#include <memory>
#include <string>
#include <thread>
#include <vector>

#include "controller_parameter_server/parameter_server.hpp"
#include "hardware_interface/robot_hardware.hpp"
#include "lifecycle_msgs/msg/state.hpp"
#include "rclcpp/rclcpp.hpp"
#include "rcutils/get_env.h"
#include "ros_controllers/diff_drive_controller.hpp"
#include "std_msgs/msg/string.hpp"
#include "test_robot_hardware/test_robot_hardware.hpp"
```

ros_controllers/test/test_diff_drive_controller.cpp
```diff
@@ -0,0 +1,270 @@
...
+   static void TearDownTestCase()
+   {
+      rclcpp::shutdown();
+   }
+
+   /// Publish velocity msgs
+   /**
+    *  time_from_start - delay between each points
+    *  points - vector of trajectories. Each velocity consists of 3 joints
+    */
```
piraka9011, line 68: Update docs

ros_controllers/test/test_diff_drive_controller.cpp
```diff
@@ -0,0 +1,270 @@
...
+      int wait_count = 0;
+      auto topic = velocity_publisher->get_topic_name();
+      while (pub_node->count_subscribers(topic) == 0) {
+         if (wait_count >= 5) {
+            auto error_msg = std::string("publishing to ") + topic + " but no node subscribes to it";
+            throw std::runtime_error(error_msg);
+         }
+         std::this_thread::sleep_for(std::chrono::milliseconds(100));
+         ++wait_count;
+      }
```
piraka9011, line 80: This should only publish, why do we need this while loop?

ros_controllers/test/test_diff_drive_controller.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+TEST_F(TestDiffDriveController, wrong_initialization)
+{
+   auto uninitialized_robot = std::make_shared<test_robot_hardware::TestRobotHardware>();
+   auto diff_drive_controller =
+      std::make_shared<ros_controllers::DiffDriveController>(left_wheel_names, right_wheel_names, op_mode);
+   auto ret = diff_drive_controller->init(uninitialized_robot, controller_name);
+   if (ret != controller_interface::CONTROLLER_INTERFACE_RET_SUCCESS) {
+      FAIL();
+   }
```
piraka9011, line 107:
```suggestion
   ASSERT_EQ(ret, controller_interface::CONTROLLER_INTERFACE_RET_SUCCESS);
```

ros_controllers/test/test_diff_drive_controller.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+TEST_F(TestDiffDriveController, correct_initialization)
+{
+   auto initialized_robot = std::make_shared<test_robot_hardware::TestRobotHardware>();
+   initialized_robot->init();
+   auto diff_drive_controller =
+      std::make_shared<ros_controllers::DiffDriveController>(left_wheel_names, right_wheel_names, op_mode);
+   auto ret = diff_drive_controller->init(initialized_robot, controller_name);
+   if (ret != controller_interface::CONTROLLER_INTERFACE_RET_SUCCESS) {
```
piraka9011, line 120: ditto here

ros_controllers/test/test_diff_drive_controller.cpp
```diff
@@ -0,0 +1,270 @@
...
+   executor.add_node(diff_drive_lifecycle_node->get_node_base_interface());
+
+   auto state = diff_drive_lifecycle_node->configure();
+   ASSERT_EQ(State::PRIMARY_STATE_INACTIVE, state.id());
+
+   state = diff_drive_lifecycle_node->activate();
+   ASSERT_EQ(State::PRIMARY_STATE_ACTIVE, state.id());
+
+   // wait for the subscriber and publisher to completely setup
+   std::this_thread::sleep_for(std::chrono::seconds(2));
```
piraka9011, line 172:
```suggestion
auto start = clock::now();
auto timedout = True;
  while (!publisher.get_subscription_count > 0) {
    if ((clock::now() - start) > timeout) {
      timedout =  False;
    }
    rclcpp::spin_some(node);
  }
  ASSERT_TRUE(timedout);
```

ros_controllers/test/test_diff_drive_controller.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+   // send msg
+   const double linear = 1.0;
+   const double angular = 1.0;
+   publish(linear, angular);
+
+   // wait for msg is be published to the system
+   std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
piraka9011, line 180: Same as above.

ros_controllers/src/diff_drive_controller.cpp
ManuelMeraz: My bad. I thought I put it through clang format. I can format it with uncrustify like you ssaid.

ros_controllers/src/diff_drive_controller.cpp
ManuelMeraz: Yeah good call. 

I was trying to dereference the pointer right there instead of working with the raw pointer, so I can use the dot operator and call a clean variable rather than the straight pointer.

ros_controllers/src/diff_drive_controller.cpp
```diff
@@ -14,12 +14,272 @@
...
+   wheel_params_.wheels_per_side = registered_left_wheel_handles_.size();
+
+   left_previous_commands_ = std::vector<double>(0, left_wheel_names_.size());
+   right_previous_commands_ = std::vector<double>(0, right_wheel_names_.size());
+
+   velocity_msg_ptr_ = std::make_shared<Twist>();
+
+   auto callback = [this](const std::shared_ptr<Twist> msg) -> void {
+      if (subscriber_is_active_) {
+         *velocity_msg_ptr_ = *msg;
```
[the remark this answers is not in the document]
ManuelMeraz: Will moving a shared_ptr leave it in a valid state? Not sure if the pointer is used by other infrastructure code. I think moving is the right thing here by the way, I'm just not used to moving a shared_ptrs.

ros_controllers/src/diff_drive_controller.cpp
```diff
@@ -14,12 +14,272 @@
...
+   }
+
+   if (registered_left_wheel_handles_.empty() or registered_right_wheel_handles_.empty() or
+       registered_operation_mode_handles_.empty()) {
+      return CallbackReturn::ERROR;
+   }
+
+   wheel_params_.wheels_per_side = registered_left_wheel_handles_.size();
+
+   left_previous_commands_ = std::vector<double>(0, left_wheel_names_.size());
```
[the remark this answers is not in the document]
ManuelMeraz: initializing the previous commands that have been sent to 0's. Currently not using it, but I noticed it in the original header file. Might be used. And yeah you're right! They are flipped. good catch.

ros_controllers/src/diff_drive_controller.cpp
piraka9011: Yes please, otherwise CI will complain about it.

ros_controllers/src/diff_drive_controller.cpp
ManuelMeraz: Reset would leave the pointer in a nullptr state. I think this call is meant to set the values to their initial states, so I thought this was a convenient (albeit inefficient) way to do that. I could reassign a a freshly constructed vector3 to both angular and linear, or just manually assign 0's to all of the members. What do you think?

ros_controllers/src/diff_drive_controller.cpp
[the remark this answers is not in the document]
piraka9011: Yep, check out [this SA answer](https://stackoverflow.com/a/41874953) and the [cppreference](https://en.cppreference.com/w/cpp/memory/shared_ptr).

> Constructing a new shared_ptr using the raw underlying pointer owned by another shared_ptr leads to undefined behavior. 

Which is why you should use `std::move`

ros_controllers/test/test_diff_drive_controller.cpp
ManuelMeraz: Nothing yet. Left it there since it came from the trajectory controller and this controller will likely publish odometry.

ros_controllers/src/diff_drive_controller.cpp
piraka9011: I see, if that's the case, this is fine

ManuelMeraz:
Just merged master into my diff drive branch and put the diff_drive_controller into its own package. 

Need to do the following:
1. Apply @piraka9011 's change requests
2. Add a test to load the diff drive controller

piraka9011:
Build failure looks like a downstream dependency issue. Taking a look...

ManuelMeraz:
Hey guys, just an update. I'd say the diff drive controller is 80% complete (building with foxy now):

1. Most of the original logic is implemented from the original diff drive controller 
2. This needs some rigorous testing for bugs
3. Needs more unit and integration testing (odometry and speed limiter are not tested as of now)

It looks like the build is failing due to dependency on boost accumulators. It looks like ros2 is moving away from boost if possible? I can easily implement a rolling mean class, or find one. 

Spend the last week or so implementing a sensor driver for an IMU I just got because a driver wasn't available outside of arduino usage, so didn't make any progress on this until recently. 

I'll be testing this on my rover and seeing how it goes. I'll be adding in the ros2 navigation stack since I have my imu working. I just recently got a realsense t265 and d435 and the ros2 nodes kept crashing, so I might be focusing my contributions on that after I integrate navigation stack and finish up with this.

bmagyar:
Let's not reimplement boost if possible. A lot of people (especially in corporate environments) weren't happy that ROS1 was exporting boost everywhere, ROS2 cut back on this. The only thing that changed is that now we have to properly include boost, it won't fall on our laps automagically.

ManuelMeraz:
> Let's not reimplement boost if possible. A lot of people (especially in corporate environments) weren't happy that ROS1 was exporting boost everywhere, ROS2 cut back on this. The only thing that changed is that now we have to properly include boost, it won't fall on our laps automagically.

That's weird because it automatically linked for me on my machine. Not sure why. Although I am using many other packages rather than only building the ros2_controls repo

diff_drive_controller/package.xml
```diff
@@ -0,0 +1,51 @@
...
+  <build_depend>hardware_interface</build_depend>
+  <build_depend>pluginlib</build_depend>
+  <build_depend>rclcpp</build_depend>
+  <build_depend>rclcpp_lifecycle</build_depend>
+  <build_depend>rcutils</build_depend>
+  <build_depend>realtime_tools</build_depend>
+  <build_depend>geometry_msgs</build_depend>
+  <build_depend>nav_msgs</build_depend>
+  <build_depend>tf2_msgs</build_depend>
+  <build_depend>boost</build_depend>
```
gavanderhoorn, line 22: Could this be made more precise?

The `boost` key will drag in *all* of Boost ([here](https://github.com/ros/rosdistro/blob/fa9b8015feb7acebea9667f2dae443bb50da994b/rosdep/base.yaml#L275-L301)).

If only specific parts of Boost are needed, we should use the appropriate keys from [here](https://github.com/ros/rosdistro/blob/fa9b8015feb7acebea9667f2dae443bb50da994b/rosdep/base.yaml#L1624-L1826), which will significantly reduce the size of the Boost dependency.

diff_drive_controller/package.xml
```diff
@@ -0,0 +1,51 @@
...
+  <exec_depend>nav_msgs</exec_depend>
+  <exec_depend>tf2_msgs</exec_depend>
+  <exec_depend>controller_interface</exec_depend>
+  <exec_depend>hardware_interface</exec_depend>
+  <exec_depend>pluginlib</exec_depend>
+  <exec_depend>rclcpp</exec_depend>
+  <exec_depend>rclcpp_lifecycle</exec_depend>
+  <exec_depend>rcutils</exec_depend>
+  <exec_depend>geometry_msgs</exec_depend>
+  <exec_depend>boost</exec_depend>
```
gavanderhoorn, line 39: And here as well.

And if the runtime dependencies are different from the build dependencies (header-only libraries fi), it would be good to differentiate those as well.

diff_drive_controller/package.xml
ManuelMeraz: Thanks for pointing that out to me, I had no idea how to specify which boost library and couldn't find documentation for it

ManuelMeraz: Also it seems like `ament_target_dependencies` doesn't like cmake aliases, so I can't just pass in `Boost::boost` to get the header only stuff

diff_drive_controller/package.xml
ManuelMeraz: I couldn't find exact documentation for which library of boost accumulators is in, so i assume it's the math one which would be part of `Boost::boost`. We'll see.

diff_drive_controller/package.xml
```diff
@@ -0,0 +1,51 @@
...
+  <build_depend>hardware_interface</build_depend>
+  <build_depend>pluginlib</build_depend>
+  <build_depend>rclcpp</build_depend>
+  <build_depend>rclcpp_lifecycle</build_depend>
+  <build_depend>rcutils</build_depend>
+  <build_depend>realtime_tools</build_depend>
+  <build_depend>geometry_msgs</build_depend>
+  <build_depend>nav_msgs</build_depend>
+  <build_depend>tf2_msgs</build_depend>
+  <build_depend>libboost-dev</build_depend>
```
gavanderhoorn, line 22: This will still drag in just about everything of Boost, as this resolves to `libboost1.65-dev` on (fi) Ubuntu Bionic, which depends on [these additional packages](https://packages.ubuntu.com/bionic/libboost1.65-dev).

ManuelMeraz: okay, so how do I specify to only pull in the math library? I don't think a deb package exists to only pull the math library in without pulling everything else in.

diff_drive_controller/package.xml
gavanderhoorn: My "header only" reference was to the libraries themselves. The manifest is separate from the build script (ie: `CMakeLists.txt`).

Header-only libraries which are not included in any public headers don't need to be listed as `exec_depend`, nor as `build_export_depend`. So they would only appear as `build_depend`.

diff_drive_controller/package.xml
gavanderhoorn: Ah, I believe I see now where this comes from:

```
$ dpkg -S boost/accumulators/accumulators.hpp
libboost1.65-dev:amd64: /usr/include/boost/accumulators/accumulators.hpp
```

gavanderhoorn: I believe `libboost-math` is only a library which ships the runtime library.

The headers are still part of `libboost-dev`. So this key would seem to be correct.

As a run dependency it would have been nice to be able to spec [libboost-math](https://packages.ubuntu.com/bionic/libboost-math1.65.1), but that key doesn't exist, as you note.

ManuelMeraz: I see so the runtime components of math are in libboost-math but all header only math components live with the core library.

bmagyar:
Thanks for the work on this, we'll now work on bringing this to master.

PR #43, opened 2020-05-01
Add on_activate to JointStateController

youtalk:
`JointTrajectoryController::on_activate` calls `LifecyclePublisher::on_activate`. https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L411
But `JointStateController` calls it `on_configure` instead. This PR simply fixed it.

### DIFF
--- a/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
+++ b/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
@@ -44,6 +44,10 @@ public:
   rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
   on_configure(const rclcpp_lifecycle::State & previous_state) override;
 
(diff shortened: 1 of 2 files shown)


youtalk:
Do you mean this line? I think so but https://github.com/ros-controls/ros2_controllers/pull/43/files#diff-f2699e17e97eb97098cc1a30b7bfb8baR66

rclcpp also uses the `(void)` hack so I use it. Do you know alternative?
https://github.com/ros2/rclcpp/blob/b25213a186add0ee93e270d08c9cb5601ae88066/rclcpp/test/test_parameter_client.cpp#L29

Karsten1987:
I think if we add an `on_activate()` callback here, we should enhance this PR as well with a `on_deactivate()` callback.

youtalk:
@Karsten1987 Thank you for your feedback. I've added it.

PR #44, opened 2020-05-01
Add init method to call ControllerInterface::init

youtalk:
`JointTrajectoryController::init` calls `ControllerInterface::init`. https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L64-L68

But `JointStateController` has no initialization. This PR simply fixed it.


Karsten1987:
Why do we need this?

The default implementation [here](https://github.com/ros-controls/ros2_control/blob/master/controller_interface/src/controller_interface.cpp#L23-L50) should just work. I can see that we'd like to override the `init` function if we have a specific set of instructions to do for the joint state controller, but in this case, I think we don't need it.

youtalk:
@Karsten1987 Thank you for your feedback. I see and close this PR.

PR #36, opened 2020-04-22
[Reference Only/Draft] joint_trajectory_controller convert trajectories to effort values

ddengster:
Used in conjunction with https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1083

The code changes in this PR was intended to work with a 'coffeebot' simulation in gazebo. It has lots of hacks, and is only meant to be an incomplete guide to porting to ROS2.

Things of note will be put in reviews.

Update: Also used with moveit2 training https://github.com/ros-planning/moveit2/pull/187 with change 9a493a5

Issue #27, opened 2020-03-05
Tests should no longer use sleep_for


ddengster:
See https://github.com/ros-controls/ros2_controllers/pull/25#discussion_r386492622
- Find a way to wait for services
- Apply the change over all tests

bmagyar:
@v-lopez do you have an idea for this?

v-lopez:
I am still at the beginning of the learning process with all this. So not at the moment.
But I can take a look in a couple of weeks when I'm more familiar with ROS2 and ros2_control.

But I totally agree that random sleeps in tests are recipe for flakiness.

bmagyar:
@jaron-l this may be relevant for your tickets

christophfroehlich:
This got fixed already, there is not a single `sleep_for` anymore in this repository.

PR #53, opened 2020-06-03
Reorder incoming out of order joint_names in trajectory messages

ddengster:
As per title. For a use case, moveit sends joint messages with names out of order, so this would be nice to have.

siphoned from PR #36

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -143,7 +143,8 @@ private:
     const std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
   void feedback_setup_callback(
     std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
-  void remap_msg_trajectories(std::shared_ptr<trajectory_msgs::msg::JointTrajectory> trajectory_msg);
+  void remap_msg_trajectories(
+    std::shared_ptr<trajectory_msgs::msg::JointTrajectory> trajectory_msg);
 
   SegmentTolerances default_tolerances_;
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -144,7 +144,7 @@ private:
   void feedback_setup_callback(
     std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
 
-  // sorts the trajectories of the incoming message to our local order
+  // sorts the joints of the incoming message to our local order
   void sort_to_local_joint_order(
     std::shared_ptr<trajectory_msgs::msg::JointTrajectory> trajectory_msg);
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -143,7 +143,9 @@ private:
     const std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
   void feedback_setup_callback(
     std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
-  void remap_msg_trajectories(
+
+  // sorts the trajectories of the incoming message to our local order
+  void sort_to_local_joint_order(
     std::shared_ptr<trajectory_msgs::msg::JointTrajectory> trajectory_msg);
 
   SegmentTolerances default_tolerances_;
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -143,6 +143,7 @@ private:
     const std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
   void feedback_setup_callback(
     std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
+  void remap_msg_trajectories(std::shared_ptr<trajectory_msgs::msg::JointTrajectory> trajectory_msg);
 
   SegmentTolerances default_tolerances_;
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
@@ -135,7 +135,9 @@ inline std::vector<size_t> mapping(const T & t1, const T & t2)
   std::vector<size_t> mapping_vector(t1.size());  // Return value
   for (auto t1_it = t1.begin(); t1_it != t1.end(); ++t1_it) {
     auto t2_it = std::find(t2.begin(), t2.end(), *t1_it);
-    if (t2.end() == t2_it) {return std::vector<size_t>();} else {
+    if (t2.end() == t2_it) {
+      return std::vector<size_t>();
+    } else {
       const size_t t1_dist = std::distance(t1.begin(), t1_it);
       const size_t t2_dist = std::distance(t2.begin(), t2_it);
       mapping_vector[t1_dist] = t2_dist;
--- a/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
@@ -126,21 +126,18 @@ private:
  * If \p t1 is <tt>"{C, B}"</tt> and \p t2 is <tt>"{A, B, C, D}"</tt>, the associated mapping vector is
  * <tt>"{2, 1}"</tt>.
  */
-template <class T>
-inline std::vector<unsigned int> mapping(const T& t1, const T& t2)
+template<class T>
+inline std::vector<unsigned int> mapping(const T & t1, const T & t2)
 {
   typedef unsigned int SizeType;
 
   // t1 must be a subset of t2
   if (t1.size() > t2.size()) {return std::vector<SizeType>();}
 
-  std::vector<SizeType> mapping_vector(t1.size()); // Return value
-  for (typename T::const_iterator t1_it = t1.begin(); t1_it != t1.end(); ++t1_it)
-  {
+  std::vector<SizeType> mapping_vector(t1.size());  // Return value
+  for (typename T::const_iterator t1_it = t1.begin(); t1_it != t1.end(); ++t1_it) {
     typename T::const_iterator t2_it = std::find(t2.begin(), t2.end(), *t1_it);
-    if (t2.end() == t2_it) {return std::vector<SizeType>();}
-    else
-    {
+    if (t2.end() == t2_it) {return std::vector<SizeType>();} else {
       const SizeType t1_dist = std::distance(t1.begin(), t1_it);
       const SizeType t2_dist = std::distance(t2.begin(), t2_it);
       mapping_vector[t1_dist] = t2_dist;
--- a/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
@@ -121,6 +121,34 @@ private:
   bool sampled_already_ = false;
 };
 
+/**
+ * \return The map between \p t1 indices (implicitly encoded in return vector indices) to \t2 indices.
+ * If \p t1 is <tt>"{C, B}"</tt> and \p t2 is <tt>"{A, B, C, D}"</tt>, the associated mapping vector is
+ * <tt>"{2, 1}"</tt>.
+ */
+template <class T>
+inline std::vector<unsigned int> mapping(const T& t1, const T& t2)
+{
+  typedef unsigned int SizeType;
+
+  // t1 must be a subset of t2
+  if (t1.size() > t2.size()) {return std::vector<SizeType>();}
+
+  std::vector<SizeType> mapping_vector(t1.size()); // Return value
+  for (typename T::const_iterator t1_it = t1.begin(); t1_it != t1.end(); ++t1_it)
+  {
+    typename T::const_iterator t2_it = std::find(t2.begin(), t2.end(), *t1_it);
+    if (t2.end() == t2_it) {return std::vector<SizeType>();}
+    else
+    {
+      const SizeType t1_dist = std::distance(t1.begin(), t1_it);
+      const SizeType t2_dist = std::distance(t2.begin(), t2_it);
+      mapping_vector[t1_dist] = t2_dist;
+    }
+  }
+  return mapping_vector;
+}
+
 }  // namespace joint_trajectory_controller
 
 #endif  // JOINT_TRAJECTORY_CONTROLLER__TRAJECTORY_HPP_
--- a/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
@@ -122,7 +122,7 @@ private:
 };
 
 /**
- * \return The map between \p t1 indices (implicitly encoded in return vector indices) to \t2 indices.
+ * \return The map between \p t1 indices (implicitly encoded in return vector indices) to \p t2 indices.
  * If \p t1 is <tt>"{C, B}"</tt> and \p t2 is <tt>"{A, B, C, D}"</tt>, the associated mapping vector is
  * <tt>"{2, 1}"</tt>.
  */
--- a/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
@@ -127,19 +127,17 @@ private:
  * <tt>"{2, 1}"</tt>.
  */
 template<class T>
-inline std::vector<unsigned int> mapping(const T & t1, const T & t2)
+inline std::vector<size_t> mapping(const T & t1, const T & t2)
 {
-  typedef unsigned int SizeType;
-
   // t1 must be a subset of t2
-  if (t1.size() > t2.size()) {return std::vector<SizeType>();}
+  if (t1.size() > t2.size()) {return std::vector<size_t>();}
 
-  std::vector<SizeType> mapping_vector(t1.size());  // Return value
+  std::vector<size_t> mapping_vector(t1.size());  // Return value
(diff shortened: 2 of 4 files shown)


gavanderhoorn:
> For a use case, moveit sends joint messages with names out of order, so this would be nice to have.

It's actually a requirement.

There is no default order, so it's also impossible to be "out of order".

Consumers must make sure to re-order these messages such that they align with whatever assumptions subsequent code has.

ddengster:
> It's actually a requirement.

Oh? To clarify, I'm talking about trajectory messages having their joint_names jumbled up and the corresponding trajectories moved according to joint_names. This may not align with the initialization you gave to joint_trajectory_controller.

There was some joint remapping functionality in ros1 joint_trajectory_controller; it was also able to handle partial joints. So in some sense this is a port.

gavanderhoorn:
> Oh? 

Yes.

> To clarify, I'm talking about trajectory messages having their joint_names jumbled up and the corresponding trajectories moved according to joint_names.

Yes, I understand what you are referring to.

Consumers / Subscribers / Action servers cannot assume a particular order will be used for the joint names in incoming goals / messages. There has always been the requirement to check whether whatever comes in corresponds to what the server needs.

> This may not align with the initialization parameters you gave to joint_trajectory_controller.

Of course.

Again: there has never been a requirement to maintain that order. So publishers/action clients should be able to use whatever order they want/need. Subscribers/servers have the responsibility to be able to process all incoming orders.

ddengster:
oh ok, sounds great to me! seems like i misunderstood what you meant there.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -526,3 +526,69 @@ TEST_F(TestTrajectoryController, zero_state_publish_rate) {
...
+  // wait for things to setup
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+  {
+    trajectory_msgs::msg::JointTrajectory traj_msg;
+    std::vector<std::string> jumbled_joint_names {
+      test_robot->joint_name2, test_robot->joint_name3, test_robot->joint_name1
+    };
+    traj_msg.joint_names = jumbled_joint_names;
+    traj_msg.header.stamp.sec = 0;
```
bmagyar, line 565:
```suggestion
    traj_msg.header.stamp.sec = rclcpp::Time(0.0);
```
should do instead of these 2 lines

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -526,3 +526,69 @@ TEST_F(TestTrajectoryController, zero_state_publish_rate) {
...
+
+  traj_controller->on_configure(traj_lifecycle_node->get_current_state());
+  traj_controller->on_activate(traj_lifecycle_node->get_current_state());
+
+  auto future_handle = std::async(
+    std::launch::async, [&executor]() -> void {
+      executor.spin();
+    });
+  // wait for things to setup
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
bmagyar, line 557: can't we check for something specific being there from the node isntead? these wait-based setup steps typically go flaky first

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -526,3 +526,69 @@ TEST_F(TestTrajectoryController, zero_state_publish_rate) {
...
+    traj_msg.points[0].time_from_start = rclcpp::Duration::from_seconds(0.25);
+    traj_msg.points[0].positions.resize(3);
+    traj_msg.points[0].positions[0] = 2.0;
+    traj_msg.points[0].positions[1] = 3.0;
+    traj_msg.points[0].positions[2] = 1.0;
+
+    trajectory_publisher->publish(traj_msg);
+  }
+
+  // update for 0.5 seconds
```
bmagyar, line 578: would it be possible to factor this section out into a function (there are many of the exact same above in thsi file), something like "run_mini_simulation(test_robot, time_to_run_for)" ? This is a function we'd want to use a lot in future tests and would also make this one much more readable

we can also open a follow-up issue if you think it's unreasonable for this PR

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -607,6 +617,41 @@ void JointTrajectoryController::feedback_setup_callback(
...
+      std::vector<double> output;
+      output.resize(mapping.size(), 0.0);
+      for (auto index = 0ul; index < mapping.size(); ++index) {
+        unsigned int map_index = mapping[index];
+        output[map_index] = to_remap[index];
+      }
+      return output;
+    };
+
+  for (auto index = 0ul; index < trajectory_msg->points.size(); ++index) {
```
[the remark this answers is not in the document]
ddengster: just following the standards @Karsten1987 set in the earlier JTC PR. The codebase already uses a ton of `auto index = 0ul` so it'll be better to do it in a single pass.

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: replaced with `traj_msg.header.stamp = rclcpp::Time(0.0);` 3e77574

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -143,6 +143,8 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
     const std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
   void feedback_setup_callback(
     std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
+  void remap_msg_trajectories(
```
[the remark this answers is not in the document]
ddengster: added docstring 3e77574

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -607,6 +617,41 @@ void JointTrajectoryController::feedback_setup_callback(
     std::bind(&RealtimeGoalHandle::runNonRealtime, rt_active_goal_));
 }
 
+void JointTrajectoryController::remap_msg_trajectories(
```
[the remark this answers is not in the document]
ddengster: switched from remap_msg_trajectories to sort_to_local_joint_order 3e77574

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -121,6 +121,31 @@ class Trajectory
   bool sampled_already_ = false;
 };
 
+/**
+ * \return The map between \p t1 indices (implicitly encoded in return vector indices) to \t2 indices.
+ * If \p t1 is <tt>"{C, B}"</tt> and \p t2 is <tt>"{A, B, C, D}"</tt>, the associated mapping vector is
+ * <tt>"{2, 1}"</tt>.
+ */
+template<class T>
+inline std::vector<unsigned int> mapping(const T & t1, const T & t2)
```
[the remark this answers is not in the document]
ddengster: switched to size_t 3e77574

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -552,6 +553,14 @@ rclcpp_action::GoalResponse JointTrajectoryController::goal_callback(
         "Joints on incoming goal don't match the controller joints.");
       return rclcpp_action::GoalResponse::REJECT;
     }
+
+    std::vector<unsigned int> mapping_vector = mapping(goal->trajectory.joint_names, joint_names_);
```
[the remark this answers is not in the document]
ddengster: this function `goal_callback` is a callback for the action server to determine whether the message should be accepted. Saving values into its internal structures should be out of scope; it's why imho there is a third callback function `feedback_setup_callback` in which you do additional setup that you could do in `goal_callback` but don't. The action server examples display this trait and I'm inclined to follow it.

In addition, I think saving those values would expose the mapping vector to the class's internal scope and makes things more complex to people extending the code.

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: old issue, i havent found a good solution yet
https://github.com/ros-controls/ros2_controllers/issues/27

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -552,6 +553,14 @@ rclcpp_action::GoalResponse JointTrajectoryController::goal_callback(
...
+    if (mapping_vector.empty()) {
+      RCLCPP_ERROR(
+        lifecycle_node_->get_logger(),
+        "Joints on incoming goal don't match the controller joints.");
```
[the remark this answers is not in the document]
ddengster: changed error checking and error msg 0baec1c

joint_trajectory_controller/src/joint_trajectory_controller.cpp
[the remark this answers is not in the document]
ddengster: also 0baec1c got rid of this mapping call

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: yes let's put it in another issue

joint_trajectory_controller/src/joint_trajectory_controller.cpp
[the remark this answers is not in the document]
bmagyar: Thanks, I've created an issue.
https://github.com/ros-controls/ros2_controllers/issues/58

joint_trajectory_controller/test/test_trajectory_controller.cpp
bmagyar: Awesome, thanks

joint_trajectory_controller/test/test_trajectory_controller.cpp
bmagyar: https://github.com/ros-controls/ros2_controllers/issues/59

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -121,6 +121,29 @@ class Trajectory
...
+ * <tt>"{2, 1}"</tt>.
+ */
+template<class T>
+inline std::vector<size_t> mapping(const T & t1, const T & t2)
+{
+  // t1 must be a subset of t2
+  if (t1.size() > t2.size()) {return std::vector<size_t>();}
+
+  std::vector<size_t> mapping_vector(t1.size());  // Return value
+  for (typename T::const_iterator t1_it = t1.begin(); t1_it != t1.end(); ++t1_it) {
```
Karsten1987, line 136:
```suggestion
  for (auto t1_it = t1.begin(); t1_it != t1.end(); ++t1_it) {
```

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -121,6 +121,29 @@ class Trajectory
...
+    typename T::const_iterator t2_it = std::find(t2.begin(), t2.end(), *t1_it);
```
Karsten1987, line 137:
```suggestion
    auto t2_it = std::find(t2.begin(), t2.end(), *t1_it);
```

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -121,6 +121,29 @@ class Trajectory
...
+    if (t2.end() == t2_it) {return std::vector<size_t>();} else {
```
Karsten1987, line 138: that format looks weird. Line break after the `{`

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -554,6 +555,24 @@ rclcpp_action::GoalResponse JointTrajectoryController::goal_callback(
...
 
+  for (auto i = 0ul; i < goal->trajectory.joint_names.size(); ++i) {
+    bool found_matching = false;
+    std::string incoming_joint_name = goal->trajectory.joint_names[i];
+    for (auto j = 0ul; j < joint_names_.size(); ++j) {
+      if (incoming_joint_name == joint_names_[j]) {
+        found_matching = true;
+        break;
+      }
+    }
```
Karsten1987, line 566: `std::find_if`

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -554,6 +555,24 @@ rclcpp_action::GoalResponse JointTrajectoryController::goal_callback(
     }
   }
 
+  for (auto i = 0ul; i < goal->trajectory.joint_names.size(); ++i) {
```
bmagyar, line 558: Sorry I think I've missed this part. 
You are basically checking if every joint in the goal is supported by the controller.

I see the point of the first loop but not so much with the second one. Couldn't a simple `std::find` do that for loop managing a boolean flag?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -607,6 +627,41 @@ void JointTrajectoryController::feedback_setup_callback(
...
+{
+  // rearrange all points in the trajectory message based on mapping
+  std::vector<size_t> mapping_vector = mapping(trajectory_msg->joint_names, joint_names_);
+  auto remap = [](const std::vector<double> & to_remap, const std::vector<size_t> & mapping)
+    -> std::vector<double>
+    {
+      if (to_remap.size() != mapping.size()) {
+        return to_remap;
+      }
+      std::vector<double> output;
```
Karsten1987, line 641:
```suggestion
      std::vector<double> output{};
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -607,6 +627,41 @@ void JointTrajectoryController::feedback_setup_callback(
...
 
+void JointTrajectoryController::sort_to_local_joint_order(
+  std::shared_ptr<trajectory_msgs::msg::JointTrajectory> trajectory_msg)
+{
+  // rearrange all points in the trajectory message based on mapping
+  std::vector<size_t> mapping_vector = mapping(trajectory_msg->joint_names, joint_names_);
+  auto remap = [](const std::vector<double> & to_remap, const std::vector<size_t> & mapping)
+    -> std::vector<double>
+    {
+      if (to_remap.size() != mapping.size()) {
```
Karsten1987, line 638: is that normal behavior or should there be a warning to this?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: replace with std::find d93864f

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: use std::find d93864f

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: imo it looks cleaner without the empty initializer list

joint_trajectory_controller/src/joint_trajectory_controller.cpp
ddengster: added warning 3255499

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
ddengster: fix formatting  8628144

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: thanks!

joint_trajectory_controller/test/test_trajectory_controller.cpp
ddengster: removed 0.0 double to avoid misconception that rclcpp::Time takes in double 4adabfa

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: I agree

Karsten1987: why would it be clearer?

bmagyar: Why write more to denote *nothing*?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -144,6 +144,10 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   void feedback_setup_callback(
     std::shared_ptr<rclcpp_action::ServerGoalHandle<FollowJTrajAction>> goal_handle);
 
+  // sorts the trajectories of the incoming message to our local order
```
bmagyar, line 147:
```suggestion
  // sorts the joints of the incoming message to our local order
```

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -121,6 +121,31 @@ class Trajectory
   bool sampled_already_ = false;
 };
 
+/**
+ * \return The map between \p t1 indices (implicitly encoded in return vector indices) to \t2 indices.
```
bmagyar, line 125:
```suggestion
 * \return The map between \p t1 indices (implicitly encoded in return vector indices) to \p t2 indices.
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -554,6 +555,18 @@ rclcpp_action::GoalResponse JointTrajectoryController::goal_callback(
     }
   }
 
+  for (auto i = 0ul; i < goal->trajectory.joint_names.size(); ++i) {
+    std::string incoming_joint_name = goal->trajectory.joint_names[i];
```
bmagyar, line 559:
```suggestion
    const std::string& incoming_joint_name = goal->trajectory.joint_names[i];
```
to avoid creating a copy

bmagyar: but we can put this on a follow-up ticket if you want

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: thanks!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: This is clearly a nitpick issue we can follow up at a different time

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: I've fixed it up on your branch, waiting for CI to pass

PR #64, opened 2020-06-11
Refactor test cases

### LINKED ISSUE #39: [joint_trajectory_controller] Refactor test cases, smaller scope
Refactor the test cases such that they clearly convey what the code is testing and what the expectation should be. Always avoid if possible, reusing the context of a previous test for simplicity, this is what a fixture class should be for. Example vaguely resembling what's in the test ATM:

```
TEST_F(EmptyFixtureWeCouldHaveFilled, test_sampling)
{
  // various lines setting up the context, should be part of the fixture class

  {
  result = sample(parameters1);
  ASSERT_THAT(result, ...);
  }

  {
  result = sample(parameters2);
  ASSERT_THAT(result, ...);
  }

  {
  result = sample(parameters3);
  ASSERT_THAT(result, ...);
  }

}
```

The problem with this is that the tests can only report "something is wrong with `sample`", we'd still need to dig and look which inner case and why. To aide in this, the pattern above is sometimes augmented with comments for each context block. I've done a lot of this and found it quite bad.

Instead, consider something like this:
```
TEST_F(TestSampling, works_with_empty)
{
  result = sample(parameters1);
  ASSERT_THAT(result, ...);
}

TEST_F(TestSampling, fails_with_time_from_start_undefined)
{
  result = sample(parameters2);
  ASSERT_THAT(result, ...);
}

TEST_F(TestSampling, success_when_all_params_defined)
{
  result = sample(parameters3);
  ASSERT_THAT(result, ...);
}
```
The result when running the tests will show a lot more cases and be a lot more specific with failures. Also, note that in this case I used a specific class fixture for `TestSampling` only, it is good practice to use the fixture name to convey the general area we are testing so we don't have to duplicate it in the name of the case, e.g `TEST_F(TestJTC, sample_success_when_all_params_defined)`

Also, if you find that there's no need for a test fixture class, don't create an empty one.
The code below will still work:
```
TEST(TestSampling, success_when_all_params_defined)
{
  result = sample(parameters3);
  ASSERT_THAT(result, ...);
}
```

jordan-palacios:
Refactoring joint_trajectory_controller tests to fix #39.

I'm doing some cleanup since I'm at it.

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_actions.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_actions.cpp
@@ -45,38 +45,38 @@ protected:
 
   void SetUp()
   {
-    test_robot = std::make_shared<test_robot_hardware::TestRobotHardware>();
-    test_robot->init();
-    joint_names = {{test_robot->joint_name1, test_robot->joint_name2, test_robot->joint_name3}};
-    op_mode = {{test_robot->write_op_handle_name1}};
+    test_robot_ = std::make_shared<test_robot_hardware::TestRobotHardware>();
+    test_robot_->init();
+    joint_names_ = {{test_robot_->joint_name1, test_robot_->joint_name2, test_robot_->joint_name3}};
+    op_mode_ = {{test_robot_->write_op_handle_name1}};
 
-    node = std::make_shared<rclcpp::Node>("trajectory_test_node");
+    node_ = std::make_shared<rclcpp::Node>("trajectory_test_node");
 
-    traj_controller = std::make_shared<joint_trajectory_controller::JointTrajectoryController>();
-    auto ret = traj_controller->init(test_robot, controller_name);
+    traj_controller_ = std::make_shared<joint_trajectory_controller::JointTrajectoryController>();
+    auto ret = traj_controller_->init(test_robot_, controller_name_);
     if (ret != controller_interface::CONTROLLER_INTERFACE_RET_SUCCESS) {
       FAIL();
     }
 
-    traj_lifecycle_node = traj_controller->get_lifecycle_node();
-    rclcpp::Parameter joint_parameters("joints", joint_names);
-    traj_lifecycle_node->set_parameter(joint_parameters);
+    traj_lifecycle_node_ = traj_controller_->get_lifecycle_node();
+    rclcpp::Parameter joint_parameters("joints", joint_names_);
+    traj_lifecycle_node_->set_parameter(joint_parameters);
 
-    rclcpp::Parameter operation_mode_parameters("write_op_modes", op_mode);
-    traj_lifecycle_node->set_parameter(operation_mode_parameters);
+    rclcpp::Parameter operation_mode_parameters("write_op_modes", op_mode_);
+    traj_lifecycle_node_->set_parameter(operation_mode_parameters);
   }
 
   void SetUpActionClient()
   {
-    action_client = rclcpp_action::create_client<control_msgs::action::FollowJointTrajectory>(
-      node->get_node_base_interface(),
-      node->get_node_graph_interface(),
-      node->get_node_logging_interface(),
-      node->get_node_waitables_interface(),
-      controller_name + "/follow_joint_trajectory");
+    action_client_ = rclcpp_action::create_client<control_msgs::action::FollowJointTrajectory>(
+      node_->get_node_base_interface(),
+      node_->get_node_graph_interface(),
+      node_->get_node_logging_interface(),
+      node_->get_node_waitables_interface(),
+      controller_name_ + "/follow_joint_trajectory");
 
     bool response =
-      action_client->wait_for_action_server(std::chrono::seconds(1));
+      action_client_->wait_for_action_server(std::chrono::seconds(1));
     if (!response) {
       throw std::runtime_error("could not get action server");
     }
@@ -86,6 +86,7 @@ protected:
   {
     rclcpp::shutdown();
   }
+
   using FollowJointTrajectoryMsg = control_msgs::action::FollowJointTrajectory;
   using GoalHandle = rclcpp_action::ClientGoalHandle<FollowJointTrajectoryMsg>;
   using GoalOptions = rclcpp_action::Client<FollowJointTrajectoryMsg>::SendGoalOptions;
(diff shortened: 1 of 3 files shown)


joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -373,145 +396,84 @@ TEST_F(TestTrajectoryActions, test_goal_tolerances_success) {
...
-  GoalOptions opt;
-  opt.goal_response_callback =
-    std::bind(&TestTrajectoryActions::common_goal_response, this, _1);
-  opt.result_callback =
-    std::bind(&TestTrajectoryActions::common_result_response, this, _1);
-  opt.feedback_callback = nullptr;
+  traj_lifecycle_node_->declare_parameter("constraints.joint1.trajectory", 0.0);
+  traj_lifecycle_node_->declare_parameter("constraints.joint2.trajectory", 0.0);
+  traj_lifecycle_node_->declare_parameter("constraints.joint3.trajectory", 0.0);
+  traj_lifecycle_node_->set_parameter(
```
bmagyar, line 428: dear lord this 4-line break-up is ugly

let's add the value setting to the `declare_parameter` call, then please try to see how we can avoid uncrustify ruining the format...

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -250,108 +301,80 @@ TEST_F(TestTrajectoryActions, test_success_multi_point_sendgoal) {
...
+  EXPECT_NEAR(3.0, test_robot_->pos3, COMMON_THRESHOLD);
+}
 
-  // start again
-  common_goal_accepted = false;
-  common_resultcode = rclcpp_action::ResultCode::UNKNOWN;
-  controller_hw_thread = std::thread(thread_func);
+TEST_F(TestTrajectoryActions, test_goal_tolerances_multi_point_success) {
+  // set tolerance parameters
+  traj_lifecycle_node_->declare_parameter("constraints.joint1.goal", 0.0);
```
bmagyar, line 365: `declare_parameter` and then `set_parameter` are both called but we could directly set the values in `declare_parameter`!
http://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Node.html#a095ea977b26e7464d9371efea5f36c42

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -373,145 +396,84 @@ TEST_F(TestTrajectoryActions, test_goal_tolerances_success) {
...
-
-  executor.cancel();
+  EXPECT_NEAR(7.0, test_robot_->pos1, COMMON_THRESHOLD);
+  EXPECT_NEAR(8.0, test_robot_->pos2, COMMON_THRESHOLD);
+  EXPECT_NEAR(9.0, test_robot_->pos3, COMMON_THRESHOLD);
 }
 
 TEST_F(TestTrajectoryActions, test_state_tolerances_fail) {
   // set joint tolerance parameters
   double state_tol = 0.0001;
```
bmagyar, line 424: need more const

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -334,39 +340,35 @@ TEST_F(TestTrajectoryController, cleanup) {
...
 
   state = traj_lifecycle_node->cleanup();
   ASSERT_EQ(State::PRIMARY_STATE_UNCONFIGURED, state.id());
-  traj_controller->update();
-  test_robot->write();
+  traj_controller_->update();
+  test_robot_->write();
 
   // shouild be home pose again
   double threshold = 0.001;
```
bmagyar, line 357: common threshold?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -334,39 +340,35 @@ TEST_F(TestTrajectoryController, cleanup) {
...
+  test_robot_->write();
 
   state = traj_lifecycle_node->cleanup();
   ASSERT_EQ(State::PRIMARY_STATE_UNCONFIGURED, state.id());
-  traj_controller->update();
-  test_robot->write();
+  traj_controller_->update();
+  test_robot_->write();
 
   // shouild be home pose again
```
bmagyar, line 356:
```suggestion
  // should be home pose again
```

joint_trajectory_controller/test/test_trajectory_actions.cpp
bmagyar: also, can this be also a common threshold thing? I think it's fine to test with `0.001`  rad threshold instead of `0.0001` rad

bmagyar:
@jordan-palacios I've rebased this PR after merging #53 to ensure we don't break anything

bmagyar:
@jordan-palacios I've fixed the compilation on this one but haven't done the refactoring you made to the rest of the tests.

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -35,6 +35,8 @@ using trajectory_msgs::msg::JointTrajectoryPoint;
 using std::placeholders::_1;
 using std::placeholders::_2;
 
+static const double COMMON_THRESHOLD = 0.001;
```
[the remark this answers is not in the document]
jordan-palacios: Agreed.

joint_trajectory_controller/test/test_trajectory_actions.cpp
jordan-palacios: Done.

joint_trajectory_controller/test/test_trajectory_actions.cpp
jordan-palacios: Done.

joint_trajectory_controller/test/test_trajectory_actions.cpp
jordan-palacios: Done and done.

joint_trajectory_controller/test/test_trajectory_controller.cpp
jordan-palacios: Yes.

joint_trajectory_controller/test/test_trajectory_controller.cpp
jordan-palacios: Fixed.

jordan-palacios:
@bmagyar Ready for merge.

PR #65, opened 2020-06-12
Joint states are now published using the DynamicJointState message

### LINKED ISSUE #45: [joint_state_controller] Add publisher with "flexible joint state" message

jordan-palacios:
Fixes #45.

Going to clean up the code and add some tests too.

@bmagyar In the mean time, please let me know what you think.

### DIFF
--- a/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
+++ b/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
@@ -20,13 +20,9 @@
 #include <vector>
 
 #include "control_msgs/msg/dynamic_joint_state.hpp"
-
 #include "controller_interface/controller_interface.hpp"
-
 #include "joint_state_controller/visibility_control.h"
-
 #include "rclcpp_lifecycle/state.hpp"
-
 #include "sensor_msgs/msg/joint_state.hpp"
 
 namespace joint_state_controller
--- a/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
+++ b/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
@@ -29,16 +29,11 @@
 
 #include "sensor_msgs/msg/joint_state.hpp"
 
-class JointStateControllerTest;
-
 namespace joint_state_controller
 {
 
 class JointStateController : public controller_interface::ControllerInterface
 {
-private:
-  friend class ::JointStateControllerTest;
-
 public:
   JOINT_STATE_CONTROLLER_PUBLIC
   JointStateController();
@@ -59,7 +54,7 @@ public:
   rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
   on_deactivate(const rclcpp_lifecycle::State & previous_state) override;
 
-private:
+protected:
   std::vector<const hardware_interface::JointStateHandle *> registered_joint_handles_;
 
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<sensor_msgs::msg::JointState>>
--- a/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
+++ b/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
@@ -19,6 +19,8 @@
 #include <string>
 #include <vector>
 
+#include "control_msgs/msg/dynamic_joint_state.hpp"
+
 #include "controller_interface/controller_interface.hpp"
 
 #include "joint_state_controller/visibility_control.h"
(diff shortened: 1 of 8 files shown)


jordan-palacios:
@bmagyar CI failed to wrong version of `control-msgs` being used. https://github.com/ros-controls/ros2_controllers/pull/66 should fix it.

bmagyar:
Thanks @jordan-palacios for this however I'd ask you to not remove the publisher of JointStates, we should keep it for legacy reasons as there's a lot of the ecosystem that depend son it.

jordan-palacios:
Understood. Will restore.

jordan-palacios:
@bmagyar @Karsten1987 I'd like your opinion on d33969f. I think we could use this approach to address #27.

joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
```diff
@@ -27,11 +29,16 @@
...
 
+class JointStateControllerTest;
+
 namespace joint_state_controller
 {
 
 class JointStateController : public controller_interface::ControllerInterface
 {
+private:
+  friend class ::JointStateControllerTest;
```
bmagyar, line 40: could we please follow the inheritance-based setup for this? would really like to avoid friends

joint_state_controller/src/joint_state_controller.cpp
```diff
@@ -29,10 +32,8 @@ JointStateController::JointStateController()
 {}
 
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-JointStateController::on_configure(const rclcpp_lifecycle::State & previous_state)
+JointStateController::on_configure(const rclcpp_lifecycle::State & /*previous_state*/)
```
bmagyar, line 35: love this, thank you

joint_state_controller/test/test_joint_state_controller.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+#include <functional>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "gtest/gtest.h"
+
+#include "hardware_interface/types/hardware_interface_return_values.hpp"
+
```
bmagyar, line 25: I know I'm now *that* guy commenting on your whitespaces but I really don't think it's necessary to bloat the include section with this many newlines. I don't think it's evil to bundle ROS2 includes into one group

joint_state_controller/test/test_joint_state_controller.cpp
```diff
@@ -0,0 +1,289 @@
...
+#include "test_joint_state_controller.hpp"
+
+#include "test_robot_hardware/test_robot_hardware.hpp"
+
+using std::placeholders::_1;
+
+namespace
+{
+constexpr rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn NODE_SUCCESS =
```
bmagyar, line 42:
```suggestion
constexpr auto NODE_SUCCESS =
```
does this break it all?

joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
jordan-palacios: We can do inheritance. However, since I'm testing member variables this will require switching to `protected` instead of `private` in the `JointStateController` class. Is this acceptable?

joint_state_controller/test/test_joint_state_controller.cpp
jordan-palacios: I don't care either way. But having whitespaces between different package includes is how it's done pretty much everywhere. A few examples:

[joint_state_controller.hpp](https://github.com/ros-controls/ros2_controllers/blob/master/joint_state_controller/include/joint_state_controller/joint_state_controller.hpp)
[joint_trajectory_controller.hpp](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp)
[controller_interface.hpp](https://github.com/ros-controls/ros2_control/blob/master/controller_interface/include/controller_interface/controller_interface.hpp)
[controller_manager.hpp](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/include/controller_manager/controller_manager.hpp)

joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
jordan-palacios: We can. But then we need the member variables to be `protected` instead of `private` in order to test them (and we do a lot of that in these tests!). Acceptable?

bmagyar: absolutely

joint_state_controller/test/test_joint_state_controller.cpp
bmagyar: Yes I know and it also looks ugly and impractical. The fact that we have bad style in some files shouldn't force us to introduce new code following it. The next time we get to those we can fix this.

jordan-palacios: Understood. Will fix.

jordan-palacios:
> this looks great _however_ super bloated especially if we want to use this for all tests in this file. How about wrapping all this boilerplate into a conveniently named function `wait_for_message()`? we could redefine the same thing over at the diff drive controller tests

Done.

joint_state_controller/test/test_joint_state_controller.cpp
jordan-palacios: Done.

joint_state_controller/include/joint_state_controller/joint_state_controller.hpp
jordan-palacios: Done.

bmagyar: thank you!

Issue #37, opened 2020-04-26
[joint_trajectory_controller] Implement partial joint list handling


bmagyar:
The `joint_trajectory_controller` currently only accepts a joint trajectory goal if all joints have a defined goal. This is an unnecessary constraint which was already lifted in ROS Kinetic.

v-lopez:
@destogl I was going to start working on this, but @bmagyar mentioned that you may have something for this? Is it something you'd accept contributions to? If not I'll get on another issue

v-lopez:
@ddengster I was going to start working on this, but @bmagyar mentioned that you may have something for this? Is it something you'd accept contributions to? If not I'll get on another issue

ddengster:
thanks for the initiative! I don't have any plans for this, but you may want to wait for the joint sorting PR https://github.com/ros-controls/ros2_controllers/pull/53 to get in first.

v-lopez:
I will wait, thanks @ddengster.
I'm taking the time to think about how it can be done. Could any of you give me your opinion on the outline of the changes to be made below?

A partial joint trajectory implies that the "missing" joints hold their state until the end of the trajectory. (at the moment we're not bridging an existing trajectory into the new one).

I see two options:

1. We don't modify the `Trajectory``class and modify the partial message and fill the missing joints with the current state. 

2. We modify the `Trajectory` class so it can accept partial joints. I started with this idea in mind and is detailed at the end, but probably complicates it too much. 

I don't want any changes to be restrictive in the future when we provide a [current trajectory](https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h#L149) with the goal.

When we implement current_trajectory bridging, what would be the approach? Combine both trajectories outside the `Trajectory` class and provide the resulting msg to `Trajectory`?

I would like to know the scope of the `Trajectory` class.

Right now it bridges between current state and starting trajectory, so it's not too far fetched that it  would do the bridging between two trajectories, or fill a partial trajectory.



**Annex Trajectory class modifications for option 2**

`Trajectory` class keeps the current state of the joints in `state_before_traj_msg_`, but we will need to map from partial to global joints. We may need to pass the full joint_names in the constructor, and compute and store the remapping inside the `Trajectory` class. (we may move @ddengster code inside the

The `sample()` method provides a `JointTrajectoryPoint` with all the joints.

[...]

Issue #31, opened 2020-03-19
Port over partial joints functionality for joint_trajectory_controller


ddengster:
As per title.

PR #68, opened 2020-06-17
Partial joints

v-lopez:
Fixes https://github.com/ros-controls/ros2_controllers/issues/31
Fixes https://github.com/ros-controls/ros2_controllers/issues/37

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -121,6 +121,15 @@ JointTrajectoryController::update()
       error.accelerations[index] = 0.0;
     };
 
+  // Check if a new external message has been received from nonRT threads
+  auto current_external_msg = traj_external_point_ptr_->get_trajectory_msg();
+  auto new_external_msg = traj_msg_external_point_ptr_.readFromRT();
+  if (current_external_msg != *new_external_msg) {
+    fill_partial_goal(*new_external_msg);
+    sort_to_local_joint_order(*new_external_msg);
+    traj_external_point_ptr_->update(*new_external_msg);
+  }
+
   JointTrajectoryPoint state_current, state_desired, state_error;
   const auto joint_num = registered_joint_state_handles_.size();
   resize_joint_trajectory_point(state_current, joint_num);
@@ -133,8 +142,6 @@ JointTrajectoryController::update()
     state_current.accelerations[index] = 0.0;
   }
   state_current.time_from_start.set__sec(0);
-
-  std::lock_guard<std::mutex> guard(trajectory_mtx_);
 
   // currently carrying out a trajectory
   if (traj_point_active_ptr_ && (*traj_point_active_ptr_)->has_trajectory_msg() == false) {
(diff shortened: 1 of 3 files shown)


v-lopez:
@ddengster Could I get your take on these changes?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -630,6 +609,47 @@ void JointTrajectoryController::feedback_setup_callback(
...
+        it.positions.push_back(joint_state->get_position());
+        if (!it.velocities.empty()) {
+          it.velocities.push_back(0.0);
+        }
+        if (!it.accelerations.empty()) {
+          it.accelerations.push_back(0.0);
+        }
+        // TODO(v-lopez) is effort 0.0 valid here? Is it worse than joint_state->get_effort()?
+        if (!it.effort.empty()) {
+          it.effort.push_back(0.0);
```
ddengster, line 646: effort implies torque/force, see this [post](https://answers.ros.org/question/331421/ros-control-effort-controllers-explained/?answer=331449#post-id-331449). So setting 0 to torque/force sounds good to me

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -630,6 +609,47 @@ void JointTrajectoryController::feedback_setup_callback(
...
+      {
+        // joint found on msg
+        continue;
+      }
+      trajectory_msg->joint_names.push_back(joint_names_[index]);
+
+      const auto & joint_state = registered_joint_state_handles_[index];
+      for (auto & it : trajectory_msg->points) {
+        // Assume hold position with 0 velocity and acceleration for missing joints
+        it.positions.push_back(joint_state->get_position());
```
ddengster, line 637: This function is called on a separate thread from the main JTC updating loop, you'll get some kind of race condition if you call `joint_state->get_position()` here. 

`traj_external_point_ptr_->update(msg);` is where we sync up, we could put some kind of coordination or delay it until the first parsing of the message (this is how I did immediate paybacks of trajectory messages where messages are sent with time 0). Perhaps we could use a boolean per joint to indicate whether to retain the last known joint position.

v-lopez: Thanks for noticing this, it is indeed a problem.

Furthermore, On the [topic callback](https://github.com/ros-controls/ros2_controllers/pull/68/files/c910ecc50cb3c53bd6a6c88690f701660de85b43#diff-fbf197065df5790eaa93ff95fc223e8bR332) we update the trajectory without acquiring the `trajectory_mtx_` which is another race condition.

Probably when we receive a trajectory (whether from topic or action) on an "auxiliary thread" we should do some basic checks to ensure that we can accept it, and store it somewhere protected by a mutex.

Then on the next `update` call on the "main thread", we should process it, fill the partial goals and do whatever needs to be done.

Something along these lines I believe would fix these issues.

v-lopez: I've pushed a fix for these issues on this same branch. https://github.com/ros-controls/ros2_controllers/pull/68/commits/188c0e405d8d0db0cdd6c873e41bf9a04d4ffb06

PR #69, opened 2020-06-17
Update PAL Robotics's headers license to Apache 2

v-lopez:


Karsten1987:
@v-lopez can you sign your commit here though? Especially in cases like that it make sense to cope with the DCO.

v-lopez:
Done!

PR #70, opened 2020-06-18
Remove boost and add custom rolling mean accumulator

v-lopez:


diff_drive_controller/include/diff_drive_controller/rolling_mean_accumulator.hpp
```diff
@@ -0,0 +1,75 @@
...
+      sum_ -= *next_insert_;
+      *next_insert_ = val;
+    } else {
+      next_insert_ = buffer_.insert(next_insert_, val);
+    }
+
+    next_insert_++;
+    if (next_insert_ == buffer_.end()) {
+      next_insert_ = buffer_.begin();
+    }
```
Karsten1987, line 58: can the index of the buffer not simply be determined by using modulo (`%`) and a counter?

diff_drive_controller/include/diff_drive_controller/rolling_mean_accumulator.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * Computes the mean of the last accumulated elements
+ */
+template<typename T>
+class RollingMeanAccumulator
+{
+public:
+  explicit RollingMeanAccumulator(size_t rolling_window_size)
+  : rolling_window_size_(rolling_window_size)
+  {
+    buffer_.reserve(rolling_window_size);
```
Karsten1987, line 41: would it make sense to use `resize` or the initializer constructor to default initialize the values to 0?
That would remove the need for branching later on as the buffer is always completely filled.

diff_drive_controller/src/odometry.cpp
```diff
@@ -33,8 +33,8 @@ Odometry::Odometry(size_t velocity_rolling_window_size)
   left_wheel_old_pos_(0.0),
   right_wheel_old_pos_(0.0),
   velocity_rolling_window_size_(velocity_rolling_window_size),
-  linear_accumulator_(RollingWindow::window_size = velocity_rolling_window_size),
-  angular_accumulator_(RollingWindow::window_size = velocity_rolling_window_size)
+  linear_accumulator_(RollingMeanAccumulator(velocity_rolling_window_size)),
+  angular_accumulator_(RollingMeanAccumulator(velocity_rolling_window_size))
```
Karsten1987, line 37:
```suggestion
  linear_accumulator_(velocity_rolling_window_size),
  angular_accumulator_(velocity_rolling_window_size)
```

diff_drive_controller/test/test_accumulator.cpp
```diff
@@ -0,0 +1,43 @@
...
+  accum.accumulate(1.);
+  EXPECT_NEAR(1., accum.getRollingMean(), THRESHOLD);
+
+  accum.accumulate(5.);
+  EXPECT_NEAR(7. / 3., accum.getRollingMean(), THRESHOLD);
+
+  accum.accumulate(5.);
+  EXPECT_NEAR(12. / 4., accum.getRollingMean(), THRESHOLD);
+
+  // Start removing old values
```
Karsten1987, line 37: I guess with default initializing the values to `T(0)`, this separation of "old" vs. "new" values can be neglected.

diff_drive_controller/include/diff_drive_controller/rolling_mean_accumulator.hpp
v-lopez: Yeah, totally.

diff_drive_controller/include/diff_drive_controller/rolling_mean_accumulator.hpp
v-lopez: Yeah, that's a great idea.

I do need to keep one branching in there to prevent the counter from overflowing and to know the actual window size for the rolling mean computation.

The size_t overflow is a far feched possibility, but I'd rather take it into account. Even though if this accumulator is going to only be used here for odometry, the chance of it overflowing in our lifetime is essentially 0.

diff_drive_controller/test/test_accumulator.cpp
v-lopez: Yeah, well there's just a slight change where you stop dividing by the items inserted so far and start dividing by your window size.

diff_drive_controller/include/diff_drive_controller/rolling_mean_accumulator.hpp
v-lopez: I made some changes that avoid using branching as far as I could see on godbolt.org with optimizations. I'd like your feedback.

diff_drive_controller/test/test_accumulator.cpp
```diff
@@ -0,0 +1,43 @@
...
+  accum.accumulate(5.);
+  EXPECT_NEAR(16. / 4., accum.getRollingMean(), THRESHOLD);
+
+  accum.accumulate(5.);
+  EXPECT_NEAR(20. / 4., accum.getRollingMean(), THRESHOLD);
+}
```
bmagyar, line 43:
```suggestion
}

#include <cmath>
TEST(TestAccumulator, spam_accumulator)
{
  constexpr double THRESHOLD = 1e-12;
  diff_drive_controller::RollingMeanAccumulator<double> accum(10);

  for(int i=0; i<10000; ++i)
  {
    accum.accumulate(M_PI);
    EXPECT_NEAR(M_PI, accum.getRollingMean(), THRESHOLD);
  }
}
```

bmagyar: something along these lines may be a useful addition to these tests

v-lopez: Great suggestion, done.

PR #72, opened 2020-06-26
Remove controller parameter server dep

bmagyar:


bmagyar:
I've rebased on top of master which should resolve the build failure on the gh action

Issue #49, opened 2020-06-01
Port over tests for joint_trajectory_controller from ROS1


bmagyar:
https://github.com/ros-controls/ros_controllers/tree/melodic-devel/position_controllers/include/position_controllers

v-lopez:
Gonna be updating missing tests here.

### init_joint_trajectory_test missing tests

- [ ] `InitLogicCombine` Combination not implemented yet.
- [ ] `InitValuesCombine` Combination not implemented yet.
- [ ]  `WrappingSpec` Wrapping not implemented yet.
- [ ]  `IgnoreWrappingSpec` Wrapping not implemented yet.
- [ ] `GoalHandleTest` Goal Handle update not implemented yet
- [ ] `OtherTimeBase` Othe time base not implemented yet

### hold_trajectory_builder_test missing tests

All tests missing. HoldTrajectoryBuilder not implemented yet.

### joint_partial_trajectory_controller_test missing tests
- [ ] `invalidMessages` 
- [ ] `topicReplacesTopicTraj` `actionReplacesActionTraj` `actionReplacesTopicTraj` `topicReplacesActionTraj` Will only do one test for this
- [ ] `ignoreOldTopicTraj` `ignoreOldActionTraj` 
- [ ] `ignorePartiallyOldActionTraj`
- [ ] `executeParitalActionTrajInFuture`

v-lopez:
https://github.com/ros-controls/ros2_controllers/pull/75 Contains the invalidMessages and topicReplaces tests

Waiting on #68 to go through

v-lopez:
@bmagyar I have a question regarding: https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/test/joint_partial_trajectory_controller_test.cpp#L542

If I understand correctly that test, it does the following. Given 3 joints.

Send trajectory with several points: [(0, 0, 0), (1, 2, 3), (2, 4, 6)]
Wait until first waypoint is done, robot joints are at (1, 2, 3)

Send a partial trajectory only for the first joint [(-1), (-2)]. Where only the second point is in the future.

Expects that when the second trajectory time has passed, the joints are at: (-2, 4, 6). Since the first joint follows the partial trajectory, and the other joints follow the original trajectory.

Is this behavior correct? Shouldn't the second trajectory fully stop the execution of the first one, and then move only the partial joints, so the final state is (-2, 2, 3)?

[...]

PR #75, opened 2020-07-01
More jtc tests

v-lopez:
These changes originate from the partial-joint branch in https://github.com/ros-controls/ros2_controllers/pull/68

Let's wait until that is merged before accepting this one.

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -712,6 +712,8 @@ bool JointTrajectoryController::validate_trajectory_msg(
   }
 
   const auto trajectory_start_time = static_cast<rclcpp::Time>(trajectory.header.stamp);
+  // If the starting time it set to 0.0, it means the controller should start it now. 
+  // Otherwise we check if the trajectory ends before the current time, in which case it can be ignored.
   if (trajectory_start_time.seconds() != 0.0) {
     auto trajectory_end_time = trajectory_start_time;
     for (const auto & p : trajectory.points) {
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -525,7 +525,6 @@ void JointTrajectoryController::publish_state(
 
   if (state_publisher_ && state_publisher_->trylock()) {
     last_state_publish_time_ = lifecycle_node_->now();
-
     state_publisher_->msg_.header.stamp = last_state_publish_time_;
     state_publisher_->msg_.desired.positions = desired_state.positions;
     state_publisher_->msg_.desired.velocities = desired_state.velocities;
@@ -712,6 +711,21 @@ bool JointTrajectoryController::validate_trajectory_msg(
     return false;
   }
 
+  const auto trajectory_start_time = static_cast<rclcpp::Time>(trajectory.header.stamp);
+  if (trajectory_start_time.seconds() != 0.0) {
+    auto trajectory_end_time = trajectory_start_time;
+    for (const auto & p : trajectory.points) {
+      trajectory_end_time += p.time_from_start;
+    }
+    if (trajectory_end_time < lifecycle_node_->now()) {
+      RCLCPP_ERROR(
+        lifecycle_node_->get_logger(),
+        "Received trajectory with non zero time start time (%f) that ends on the past (%f)",
+        trajectory_start_time.seconds(), trajectory_end_time.seconds());
+      return false;
+    }
+  }
+
   for (auto i = 0ul; i < trajectory.joint_names.size(); ++i) {
     const std::string & incoming_joint_name = trajectory.joint_names[i];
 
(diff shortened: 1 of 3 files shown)


bmagyar:
I've squashed & rebased this to the latest master, with the recent merges to master the merge algorithm was confused, there were no conflicts really.

bmagyar:
@v-lopez I've created a backup of the old branch at pal/more-jtc-tests-orig

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -712,6 +711,21 @@ bool JointTrajectoryController::validate_trajectory_msg(
     return false;
   }
 
+  const auto trajectory_start_time = static_cast<rclcpp::Time>(trajectory.header.stamp);
+  if (trajectory_start_time.seconds() != 0.0) {
```
bmagyar, line 715:
```suggestion
  // If the starting time it set to 0.0, it means the controller should start it now. 
  // Otherwise we check if the trajectory ends before the current time, in which case it can be ignored.
  if (trajectory_start_time.seconds() != 0.0) {
```

bmagyar: I think a bit of a reminder is good to have here

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -724,6 +738,57 @@ bool JointTrajectoryController::validate_trajectory_msg(
...
+      RCLCPP_ERROR(
+        lifecycle_node_->get_logger(),
+        "Time between points %u and %u is not strictly increasing, it is %f and %f respectively",
+        i - 1, i, previous_traj_time.seconds(),
+        rclcpp::Duration(trajectory.points[i].time_from_start).seconds());
+      return false;
+    }
+    previous_traj_time = trajectory.points[i].time_from_start;
+
+    if (trajectory.joint_names.size() != trajectory.points[i].positions.size()) {
```
bmagyar, line 754: Nitpick: `trajectory.joint_names.size()` is written 8 times in this section. Perhaps a slightly shorter alternative could be created.

Also, I think a clever function could remove the repetitive parts of the "check this vector is empty then check for size, then get_logger(), general bla % parameters". 

```
bool validate_trajectory_point_vector_field(joint_names_size, vector_field, string_for_vector_field, i)
{
    if (!vector_field.empty() &&
      joint_names_size !=vector_field.size())
    {
      RCLCPP_ERROR(
        lifecycle_node_->get_logger(),
        "Mismatch between joint_names (%u) and %s (%u) at point #%u.",
        joint_names_size, string_for_vector_field, vector_field.size(), i);
      return false;
    }
}
```

and `&&` the results together and check whether it's false to early exit this loop?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -369,15 +467,20 @@ TEST_F(TestTrajectoryController, cleanup) {
 
   state = traj_lifecycle_node->cleanup();
   ASSERT_EQ(State::PRIMARY_STATE_UNCONFIGURED, state.id());
-  traj_controller_->update();
-  test_robot_->write();
+  // update for 0.25 seconds
+  const auto start_time = rclcpp::Clock().now();
```
bmagyar, line 471: could use `updateController()` here?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -139,16 +207,49 @@ class TestTrajectoryController : public ::testing::Test
...
+      }
       duration_total = duration_total + duration;
     }
 
     trajectory_publisher_->publish(traj_msg);
   }
 
+  void updateController(rclcpp::Duration wait_time = rclcpp::Duration::from_seconds(0.2))
+  {
+    auto start_time = rclcpp::Clock().now();
```
bmagyar, line 222:
```suggestion
    const auto start_time = rclcpp::Clock().now();
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -139,16 +207,49 @@ class TestTrajectoryController : public ::testing::Test
...
+    auto end_time = start_time + wait_time;
```
bmagyar, line 223:
```suggestion
    const auto end_time = start_time + wait_time;
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -678,3 +724,169 @@ TEST_F(TestTrajectoryController, test_partial_joint_list) {
 
   executor.cancel();
 }
+
+/**
+ * @brief invalid_message Test missmatched joint and reference vector sizes
```
bmagyar, line 729:
```suggestion
 * @brief invalid_message Test mismatched joint and reference vector sizes
```

let's save this extra `s` for a rainy day

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -581,54 +656,31 @@ TEST_F(TestTrajectoryController, test_jumbled_joint_order) {
...
   EXPECT_NEAR(2.0, test_robot_->pos2, COMMON_THRESHOLD);
   EXPECT_NEAR(3.0, test_robot_->pos3, COMMON_THRESHOLD);
-
-  executor.cancel();
 }
 
+/**
+ * @brief test_partial_joint_list Test sending trajectories with a subset of the controlled joints
+ */
 TEST_F(TestTrajectoryController, test_partial_joint_list) {
```
bmagyar, line 671: Could you please add a pretty much identical test where we check the behaviour of the controller to a partial list with the feature turned off?

v-lopez:
Thanks for the review.

I cannot address thisuuntil the week after the next one.

In https://github.com/ros-controls/ros2_controllers/issues/49 I am writing down the ROS1 tests in each file, I believe these cover just 3 files, there's a bunch more. I tried to keep below the 500 line limit suggestion

joint_trajectory_controller/src/joint_trajectory_controller.cpp
v-lopez: Done

joint_trajectory_controller/test/test_trajectory_controller.cpp
v-lopez: Yes, done.

v-lopez:
Should I keep adding tests to this PR or continue on a separate branch?

bmagyar:
@v-lopez I think this is a good amount of new stuff to be merged, please open a new PR with additional tests.

v-lopez:
Will do!

v-lopez:
@bmagyar I've rebased this and linters and tests are passing on my machine. Should be good to go when the actions finish.

bmagyar (approved):
Thanks very much for this, this is a huge improvement!

PR #92, opened 2020-08-31
position_controllers

### LINKED ISSUE #50: Implement position_controllers from ROS1
https://github.com/ros-controls/ros_controllers/tree/melodic-devel/position_controllers/include/position_controllers

jordan-palacios:
Port of position_controllers from ROS1.

Closes #50 .

### DIFF
--- a/position_controllers/src/joint_group_position_controller.cpp
+++ b/position_controllers/src/joint_group_position_controller.cpp
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "position_controllers/joint_position_controller.hpp"
+#include "position_controllers/joint_group_position_controller.hpp"
 #include "rclcpp/logging.hpp"
 #include "rclcpp/parameter.hpp"
 
@@ -23,15 +23,16 @@ constexpr auto kJPCLoggerName = "joint position controller";
 
 namespace position_controllers
 {
-using CallbackReturn = JointPositionController::CallbackReturn;
+using CallbackReturn = JointGroupPositionController::CallbackReturn;
 
-JointPositionController::JointPositionController()
+JointGroupPositionController::JointGroupPositionController()
 : forward_command_controller::ForwardCommandController()
 {
   logger_name_ = kJPCLoggerName;
 }
 
-CallbackReturn JointPositionController::on_configure(const rclcpp_lifecycle::State & previous_state)
+CallbackReturn JointGroupPositionController::on_configure(
+  const rclcpp_lifecycle::State & previous_state)
 {
   rclcpp::Parameter interface_param;
   if (!lifecycle_node_->get_parameter("interface_name", interface_param)) {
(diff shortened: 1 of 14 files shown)


position_controllers/test/test_load_joint_group_position_controller.cpp
```diff
@@ -0,0 +1,39 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+#include <memory>
+
+#include "controller_manager/controller_manager.hpp"
+#include "test_robot_hardware/test_robot_hardware.hpp"
+
```
bmagyar, line 20: From a pure unit testing perspective, having a test on forwarding commands may not be sensible as this would be duplicating forward command group controller tests *but* I think there's value in it from an integration test perspective. Would it be too much trouble to add one following the load test? We could slowly turn these into some higher level integration tests.

position_controllers/src/joint_group_position_controller.cpp
```diff
@@ -0,0 +1,51 @@
...
+  const rclcpp_lifecycle::State & previous_state)
+{
+  rclcpp::Parameter interface_param;
+  if (!lifecycle_node_->get_parameter("interface_name", interface_param)) {
+    lifecycle_node_->declare_parameter("interface_name", "position_command");
+  } else {
+    if (interface_param.as_string() != "position_command") {
+      RCLCPP_ERROR_STREAM(
+        rclcpp::get_logger(
+          logger_name_), "'interface_name' already set with an invalid value");
```
bmagyar, line 39:
```suggestion
          logger_name_), "'interface_name' already set with an invalid value");
```
Maybe worth also printing what thi sinvalid value is

bmagyar (commented):
Looks good in general, minor comments
The test comment could be addressed in a follow-up

position_controllers/src/joint_group_position_controller.cpp
jordan-palacios: Done.

position_controllers/test/test_load_joint_group_position_controller.cpp
jordan-palacios: I'm not sure I follow.

You'd like to have tests on the actual forwarded command for the `JointGroupPositionController`? Just like in [test_forward_command_controller.cpp](https://github.com/ros-controls/ros2_controllers/blob/master/forward_command_controller/test/test_forward_command_controller.cpp)?

bmagyar: Yes

bmagyar: https://github.com/ros-controls/ros2_controllers/issues/100

bmagyar (approved):

PR #94, opened 2020-09-09
WIP: Unstamped cmd_vel subscriber

### LINKED ISSUE #82: Implement Twist controllers (rather than TwistStamped) for mobile robot uses
Hi,

I see this remap from Twist -> TwistStamped in the header files for using the Stamped version of messages. While I support using Stamped messages long term, right now no other item in the ecosystem will be compatible with this (gazebo, nav2, etc) so we should be supporting Twist right now in Foxy and older. For Galactic and newer, we can target TwistStamped but it would require a coordinated effort to change that API for a newer distribution.

piraka9011:
Signed-off-by: Anas Abou Allaban <aabouallaban@pm.me>

Closes #82 

First take on creating an unstamped `Twist` subscriber.

I basically just take the unstamped Twist msg and put the velocity data in the `TwistStamped` message, setting the time in the header to the current node's ROS time.
We could also just do nothing with the header or leave it at 0.

If the high level approach is good, I'll start writing some tests.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -29,6 +29,7 @@
 namespace
 {
 constexpr auto DEFAULT_COMMAND_TOPIC = "~/cmd_vel";
+constexpr auto DEFUALT_COMMAND_UNSTAMPED_TOPIC = "~/cmd_vel_unstamped";
 constexpr auto DEFAULT_COMMAND_OUT_TOPIC = "~/cmd_vel_out";
 constexpr auto DEFAULT_ODOMETRY_TOPIC = "/odom";
 constexpr auto DEFAULT_TRANSFORM_TOPIC = "/tf";
@@ -96,6 +97,7 @@ DiffDriveController::init(
   lifecycle_node_->declare_parameter<int>("cmd_vel_timeout", cmd_vel_timeout_.count());
   lifecycle_node_->declare_parameter<bool>("publish_limited_velocity", publish_limited_velocity_);
   lifecycle_node_->declare_parameter<int>("velocity_rolling_window_size", 10);
+  lifecycle_node_->declare_parameter<bool>("use_stamped_vel", use_stamped_vel_);
 
   lifecycle_node_->declare_parameter<bool>("linear.x.has_velocity_limits", false);
   lifecycle_node_->declare_parameter<bool>("linear.x.has_acceleration_limits", false);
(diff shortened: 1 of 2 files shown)


v-lopez (approved):
LGTM. I would just suggest changing https://github.com/ros-controls/ros2_controllers/pull/94/files#diff-30490990ab677715a8a78a31a16c9605R51 to `using TwistStamped ....` otherwise it becomes a bit confusing.

piraka9011:
> LGTM. I would just suggest changing https://github.com/ros-controls/ros2_controllers/pull/94/files#diff-30490990ab677715a8a78a31a16c9605R51 to `using TwistStamped ....` otherwise it becomes a bit confusing.

I avoided doing that to minimize the diff in the first iteration.
Will add this to the refactor!

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -29,6 +29,7 @@
 namespace
 {
 constexpr auto DEFAULT_COMMAND_TOPIC = "~/cmd_vel";
+constexpr auto DEFUALT_COMMAND_UNSTAMPED_TOPIC = "~/cmd_vel_unstamped";
```
jordan-palacios, line 32: Typo:
```suggestion
constexpr auto DEFAULT_COMMAND_UNSTAMPED_TOPIC = "~/cmd_vel_unstamped";
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -392,18 +395,36 @@ CallbackReturn DiffDriveController::on_configure(const rclcpp_lifecycle::State &
...
+            "Can't accept new commands. subscriber is inactive");
+          return;
+        }
+
+        received_velocity_msg_ptr_ = std::move(msg);
+      });
+  } else {
+    velocity_command_unstamped_subscriber_ =
+      lifecycle_node_->create_subscription<geometry_msgs::msg::Twist>(
+      DEFAULT_COMMAND_TOPIC, rclcpp::SystemDefaultsQoS(), [this](
```
jordan-palacios, line 414: It is not clear to me if you want to use the same topic regardless of the message type or this is a typo and should be DEFAULT_COMMAND_UNSTAMPED_TOPIC` instead.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -96,6 +97,7 @@ DiffDriveController::init(
   lifecycle_node_->declare_parameter<int>("cmd_vel_timeout", cmd_vel_timeout_.count());
   lifecycle_node_->declare_parameter<bool>("publish_limited_velocity", publish_limited_velocity_);
   lifecycle_node_->declare_parameter<int>("velocity_rolling_window_size", 10);
+  lifecycle_node_->declare_parameter<bool>("use_stamped_vel", use_stamped_vel_);
```
jordan-palacios, line 100: Maybe default to `false` since it is the more "compatible" behaviour.

bmagyar:
Thanks guys, closing in favour of #143

PR #103, opened 2020-10-12
fix brace initializers

Karsten1987:
something went wrong with my last change and I brought back the clang warning.

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_controller.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller.cpp
@@ -916,7 +916,7 @@ TEST_F(TestTrajectoryController, test_trajectory_replace) {
   subscribeToState();
 
   std::vector<std::vector<double>> points_old {{{2., 3., 4.}}};
-  std::vector<std::vector<double>> points_partial_new {{{1.5}}};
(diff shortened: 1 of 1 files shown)


joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -916,7 +916,7 @@ TEST_F(TestTrajectoryController, test_trajectory_replace) {
   subscribeToState();
 
   std::vector<std::vector<double>> points_old {{{2., 3., 4.}}};
```
bmagyar, line 918: how come it doesn't complain about this line for the same reason? can you try changing this too for consistency pelase?

Karsten1987: it does not. I think there's two different constructors being called.

bmagyar (approved):

joint_trajectory_controller/test/test_trajectory_controller.cpp
bmagyar: It probably is the case. Let's keep an eye out if something else puts it back again!

3hpxy5

PR #104, opened 2020-10-15
Update ci

bmagyar:


.github/workflows/ci.yml
```diff
@@ -10,21 +10,21 @@ on:
 
 jobs:
   # Re-enable once ros2_controls is released completely
```
Karsten1987, line 12: can this comment be removed?

Karsten1987:
Looking at the CI, it looks like the binary job is failing due to pluginlib using boost instead of std.

destogl:
> Looking at the CI, it looks like the binary job is failing due to pluginlib using boost instead of std.

@bmagyar: you should add this line: [comment](https://github.com/ros-controls/ros2_control/pull/147#discussion_r499124758)

.github/workflows/ci.yml
bmagyar: gone

bmagyar:
Merging this as it resolves at least part of the CI failures.

PR #98, opened 2020-09-22
Adapt GH actions to target Foxy

Karsten1987:
~Turns out this was not a CI misconfiguration.~

~https://github.com/ros2/rclcpp/pull/1311 introduced an API break how callbacks for action goals are handled. Given that this (so far) only affects a single test, I don't see an immediate need for branching off and fixed the tests with a simple conditional given the version of `rclcpp`.~

Adapt GH actions:
* temporary disable ROS industrial CI until either ros2-controls is fully released or a way is found to pass in a custom overlay repos file
* disable codecov
* use Foxy docker image for source overlay builds

PR #111, opened 2020-11-17
Change test to work with Foxy and posterior action API

v-lopez:
This PR https://github.com/ros2/rclcpp/pull/1311 fixed a bug in rclcpp_action but had to change API for that, the Foxy API remains the same.

I haven't found a way to have specify a `goal_response_callback` compatible with both API versions, so instead I use the future obtained after sending the goal, so we don't have to branch the code for a test.

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_actions.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_actions.cpp
@@ -103,8 +103,6 @@ protected:
     rclcpp::Parameter stopped_velocity_parameters("constraints.stopped_velocity_tolerance", 0.0);
     traj_lifecycle_node_->set_parameter(stopped_velocity_parameters);
 
-    goal_options_.goal_response_callback =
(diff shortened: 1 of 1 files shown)


Karsten1987:
I've tackled it with the following change: https://github.com/ros-controls/ros2_controllers/pull/98/commits/63de3d9ff2c73c904ae6e93c6d867cd49c6eea4d

However, as discussed, we are exclusively targeting Foxy, so I don't see a need to deviate from the Foxy API until the Galactic release.

bmagyar:
@v-lopez shall we follow up on this?

v-lopez:
I'm swamped for the next months, but yeah we should eventually check this.

tylerjw:
I successfully rebased this PR here: https://github.com/ros-controls/ros2_controllers/pull/237.  Lmk if there are any changes I should make to it.  I think the clang-format may have changed and that resulted in the merge conflict when I cherry-picked it.  I think I got that fixed.

bmagyar:
Thanks for taking care of this @tylerjw

PR #125, opened 2020-12-11
Remove usage of (void) to disable linter warnings

bmagyar:


v-lopez (approved):

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -248,10 +248,8 @@ JointTrajectoryController::update()
 }
 
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-JointTrajectoryController::on_configure(const rclcpp_lifecycle::State & previous_state)
+JointTrajectoryController::on_configure(const rclcpp_lifecycle::State &)
```
destogl, line 251: @bmagyar  Didn't we decided to go with commented argument? e.g. `/ *previous_state */`

bmagyar: They are functionally equivalent but found removing it better as it's less stuff to read...

PR #127, opened 2020-12-13
Add lifecycle state checks

bmagyar:
On top of #124, diffed against master. This PR can contain all additional checks we add on lifecycle states or whether we should have them at all.
The only real diff ATM is reverting the removal of checks


v-lopez:
Since the lifecycle changes are enforced by the controller manager, I don't think we need to check whether we're being used properly. 
If we do need, we need to add on all controllers, and therefore will be copied by all future controller developers.

bmagyar:
I wouldn't want to spend time on this given the understanding that we want to go back to lifecycle nodes at some point.

bmagyar:
I've adjusted tests in #124 to make it pass without this

PR #130, opened 2020-12-19
Print exception msg

bmagyar:
making life easier


destogl:
Why fprint?

bmagyar:
Can't use loggers without a node context in a simple way. I'm trying to use `get_node()` everywhere which was made to throw is `node_ == nullptr` that helps detect if this happens. I added a test that provokes this failure (a configure call without init) but it failed with printing since it calls `get_node()->get_logger()`. That being said, any print is better than no print, so I went for `fprinf` as per the earlier recommendation of printf-derivatives that are thread-safe.

PR #131, opened 2020-12-22
ForwardCommandController declares parameters

bmagyar:
Tested against demo repo and also in conjunction with joint trajectory controller as well as diff drive controller (I hacked the demo system locally a bit ;))

### DIFF
--- a/forward_command_controller/src/forward_command_controller.cpp
+++ b/forward_command_controller/src/forward_command_controller.cpp
@@ -34,31 +34,46 @@ ForwardCommandController::ForwardCommandController()
   joints_command_subscriber_(nullptr)
 {}
 
+controller_interface::return_type
+ForwardCommandController::init(
+  const std::string & controller_name)
+{
+  auto ret = ControllerInterface::init(controller_name);
+  if (ret != controller_interface::return_type::SUCCESS) {
+    return ret;
+  }
+
+  try {
+    auto node = get_node();
+    node->declare_parameter<std::vector<std::string>>("joints", {});
+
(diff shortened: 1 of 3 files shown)


forward_command_controller/src/forward_command_controller.cpp
```diff
@@ -34,31 +34,46 @@ ForwardCommandController::ForwardCommandController()
   joints_command_subscriber_(nullptr)
 {}
 
+controller_interface::return_type
+ForwardCommandController::init(
+  const std::string & controller_name)
```
destogl, line 39: One line please

```suggestion
ForwardCommandController::init(const std::string & controller_name)
```

forward_command_controller/src/forward_command_controller.cpp
```diff
@@ -34,31 +34,46 @@ ForwardCommandController::ForwardCommandController()
...
+{
+  auto ret = ControllerInterface::init(controller_name);
+  if (ret != controller_interface::return_type::SUCCESS) {
+    return ret;
+  }
+
+  try {
```
destogl, line 46: Do we have to declare Paramus in the Init step, or can this be done in the configure step? (If in configure, maybe it doesn’t work for PAL use-case)

destogl (approved):
Just check if configure can be used...

forward_command_controller/src/forward_command_controller.cpp
bmagyar: All PAL ported controllers declared them in init and I'd like to think I'm familiar with their use case. It should be fine as-is

PR #143, opened 2021-01-20
Unstamped cmd_vel subscriber rebased

v-lopez:
Original work here: https://github.com/ros-controls/ros2_controllers/pull/94

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
@@ -141,6 +141,9 @@ protected:
 
   bool subscriber_is_active_ = false;
   rclcpp::Subscription<Twist>::SharedPtr velocity_command_subscriber_ = nullptr;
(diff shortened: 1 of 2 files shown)


v-lopez:
I've got some comments before this is merged.

Is this the behavior that we want? That you have to choose between enabling one topic or the other? The topic changes names and type.

I don't like too much that the interface of the controllers changes so much due to one parameter.

Since navigation is going to unstamped twist,  most of controller instances will use unstamped twist, which will add confusion if the default behavior is the other.

Are the navigation stack developers adamant on not using stamped twist?

bmagyar:
There are strong reasons for using the stamped message but the top-down approach would hardly work. This way we are bottom-up, exposing the new stamped interface such that in Galactic, the nav stack can and other tooling can use stamped everywhere.

I don't think of this feature as anything other than compatibility mode, we should aim to remove it at some point.

v-lopez:
Ok then!

bmagyar (approved):

PR #145, opened 2021-02-02
Diff drive parameter fixes

v-lopez:
-  Recover old speed limiter behavior, if unspecified min defaults to -max
- Change cmd_vel_timeout to seconds (double) as ROS1 instead of ms(int)


bmagyar (approved):
Thanks. We should strive for parameter parity as much as possible with ROS1.

PR #146, opened 2021-02-02
Fix diff drive twist concurrency issues

v-lopez:
**Concurrency issues**
Before this fix, a twist message could be received and stored one
thread, in the middle of the update() of the controller.

This would be fixed by making a copy of the shared pointer at the
beginning of the update() function, added realtime box to ensure safe
concurrent access to the pointer.

**Don't store limited command as last command**
Before these changes, the limited command overwrote the original
command, which mean that it too much more time to reach the commanded
speed.
We only want this behavior when the command is too old and we replace it
with 0 speed.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -156,8 +156,23 @@ controller_interface::return_type DiffDriveController::update()
 
   std::shared_ptr<Twist> last_msg;
   received_velocity_msg_ptr_.get(last_msg);
-  double & linear_command = last_msg->twist.linear.x;
-  double & angular_command = last_msg->twist.angular.z;
+
+  if (last_msg == nullptr) {
+    RCLCPP_WARN(logger, "Velocity message received was a nullptr.");
+    return controller_interface::return_type::ERROR;
+  }
+
+  const auto dt = current_time - last_msg->header.stamp;
+  // Brake if cmd_vel has timeout, override the stored command
+  if (dt > cmd_vel_timeout_) {
+    last_msg->twist.linear.x = 0.0;
+    last_msg->twist.angular.z = 0.0;
+  }
+
+  // linear_command and angular_command may be limited further by SpeedLimit,
+  // without affecting the stored twist command
+  double linear_command = last_msg->twist.linear.x;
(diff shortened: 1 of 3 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -373,11 +379,12 @@ CallbackReturn DiffDriveController::on_configure(const rclcpp_lifecycle::State &
       std::make_shared<realtime_tools::RealtimePublisher<Twist>>(limited_velocity_publisher_);
   }
 
-  received_velocity_msg_ptr_ = std::make_shared<Twist>();
+  const Twist empty_twist;
+  received_velocity_msg_ptr_.set(std::make_shared<Twist>(empty_twist));
```
bmagyar, line 383: does a `shared_ptr` and the `emplace` below play well together in this case?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -520,7 +536,7 @@ bool DiffDriveController::reset()
   velocity_command_subscriber_.reset();
   velocity_command_unstamped_subscriber_.reset();
 
-  received_velocity_msg_ptr_.reset();
+  received_velocity_msg_ptr_.set(nullptr);
```
bmagyar, line 539: why?

diff_drive_controller/src/diff_drive_controller.cpp
v-lopez: The `previous_commands_` stores Twist objects, not poitners. I think it should be fine.

diff_drive_controller/src/diff_drive_controller.cpp
v-lopez: Afaik this is the only way to replicate with the RealtimeBox the nullptr reset of the shared_ptr.

bmagyar: :+1:

diff_drive_controller/src/diff_drive_controller.cpp
bmagyar: Ah! Right. `make_shared` creates a copy that it manages

bmagyar (approved):

PR #154, opened 2021-03-09
Multi interface Forward Controller

destogl:
A proposal for forward controller enabling multiple interfaces defined by a parameter. 

Difference to classical `ForwardCommandController` is in input parameters, i.e., joint-interface combinations, for example:

`ForwardCommandController` parameters:
``` yaml
joints:
  - joint1
  - ...
  - joint7
interface: position
```

`MultiInterfaceForwardController` parameters:
``` yaml
joint: joint1
interfaces:
  - position
  - velocity
  - emergency_stop
```

### DIFF
--- a/forward_command_controller/forward_command_plugin.xml
+++ b/forward_command_controller/forward_command_plugin.xml
@@ -4,4 +4,10 @@
     The forward command controller commands a group of joints in a given interface
   </description>
   </class>
+  <class name="forward_command_controller/MultiInterfaceForwardCommandController"
+         type="forward_command_controller::MultiInterfaceForwardCommandController" base_class_type="controller_interface::ControllerInterface">
+    <description>
+      MultiInterfaceForwardController ros2_control controller.
+    </description>
+  </class>
 </library>
--- a/forward_command_controller/include/forward_command_controller/forward_command_controller.hpp
+++ b/forward_command_controller/include/forward_command_controller/forward_command_controller.hpp
@@ -23,8 +23,6 @@
 
 namespace forward_command_controller
 {
-using CallbackReturn = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn;
-
 /**
  * \brief Forward command controller for a set of joints.
  *
@@ -44,7 +42,7 @@ public:
 
 protected:
   void declare_parameters() override;
-  CallbackReturn read_parameters() override;
+  controller_interface::CallbackReturn read_parameters() override;
 
   std::vector<std::string> joint_names_;
   std::string interface_name_;
--- a/forward_command_controller/include/forward_command_controller/forward_command_controller.hpp
+++ b/forward_command_controller/include/forward_command_controller/forward_command_controller.hpp
@@ -78,6 +78,8 @@ protected:
   std::vector<std::string> joint_names_;
   std::string interface_name_;
 
+  std::vector<std::string> command_interface_types_;
+
   realtime_tools::RealtimeBuffer<std::shared_ptr<CmdType>> rt_command_ptr_;
   rclcpp::Subscription<CmdType>::SharedPtr joints_command_subscriber_;
 
--- a/effort_controllers/src/joint_group_effort_controller.cpp
+++ b/effort_controllers/src/joint_group_effort_controller.cpp
@@ -25,7 +25,6 @@ namespace effort_controllers
 JointGroupEffortController::JointGroupEffortController()
 : forward_command_controller::ForwardCommandController()
 {
-  logger_name_ = "joint effort controller";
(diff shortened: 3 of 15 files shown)


forward_command_controller/include/forward_command_controller/forward_command_controller.hpp
```diff
@@ -72,6 +72,8 @@ class ForwardCommandController : public controller_interface::ControllerInterfac
   controller_interface::return_type update() override;
 
 protected:
+  virtual CallbackReturn read_parameters();
```
AndyZe, line 75: Not related to this PR, but I just noticed: the parent class is spelled wrong. Should be `controller_interface::ControllerInterface`

forward_command_controller/src/multi_interface_forward_controller.cpp
```diff
@@ -30,15 +30,16 @@ MultiInterfaceForwardController::MultiInterfaceForwardController()
...
   }
 
   try {
-    get_node()->undeclare_parameter("joints");
-    get_node()->undeclare_parameter("interface_name");
-    get_node()->declare_parameter<std::vector<std::string>>("joints_interfaces", {});
+    node_->undeclare_parameter("joints");
+    node_->undeclare_parameter("interface_name");
+    node_->declare_parameter<std::vector<std::string>>(
+      "joints_interfaces", joints_interfaces_);
```
AndyZe, line 42: A comment about what's happening here would be great. I don't understand without really digging into it

forward_command_controller/include/forward_command_controller/multi_interface_forward_controller.hpp
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2021, PickNik, Inc.
+// Copyright (c) 2021, Stogl Robotics Consulting UG (haftungsbeschränkt) (template)
```
bmagyar, line 2: just one please

forward_command_controller/src/multi_interface_forward_controller.cpp
```diff
@@ -0,0 +1,71 @@
...
+controller_interface::return_type
+MultiInterfaceForwardController::init(const std::string & controller_name)
+{
+  auto ret = ForwardCommandController::init(controller_name);
+  if (ret != controller_interface::return_type::SUCCESS) {
+    return ret;
+  }
+
+  try {
+    node_->undeclare_parameter("joints");
```
bmagyar, line 39: Seeing this I'm tempted to say we shouldn't even inherit from ForwardCommandController. It is not a specialization of a controller if you rip all parameters out and replace them with new ones.

forward_command_controller/forward_command_plugin.xml
```diff
@@ -4,4 +4,10 @@
     The forward command controller commands a group of joints in a given interface
   </description>
   </class>
+  <class name="forward_command_controller/MultiInterfaceForwardController"
+         type="forward_command_controller::MultiInterfaceForwardController" base_class_type="controller_interface::ControllerInterface">
+    <description>
+      MultiInterfaceForwardController ros2_control controller.
```
bmagyar, line 10: A better description would be good here. We should make it clear that this controller is for some very specific use-cases and in general recommend ppl to use the single interface version to avoid mixing different units in the (essentially) untyped input ROS message.

forward_command_controller/include/forward_command_controller/forward_command_controller.hpp
AndyZe: It looks like this has been fixed in the meantime.  :+1:  (Which probably broke API, but oh well)

forward_command_controller/include/forward_command_controller/multi_interface_forward_controller.hpp
destogl: of course. This is the reason for draft PR :) Not everything is cleared, but rather auto-generated :)

forward_command_controller/src/multi_interface_forward_controller.cpp
destogl: I partially agree, but there is a big overlap on the other side. What do you think to add a virtual ForwardController without parameters and then inheriting from it for both, `ForwardCommandController` and `MultiInterfaceForwardController`?

forward_command_controller/forward_command_plugin.xml
destogl: Yeah... This controller could become a basis for GPIO controller actually. We could discuss this in the WG-meeting.

anfemosa (approved):
lgtm

bmagyar (approved):
please fix the conflict again, otherwise :+1:

PR #156, opened 2021-03-14
Add basic user docs pages for each package

bmagyar:
Content used by https://github.com/ros-controls/control.ros.org/pull/9


doc/controllers_index.rst
```diff
@@ -0,0 +1,16 @@
+.. _controllers:
+
+Controllers user documentation
+==============================
+
+.. include:: ../diff_drive_controller/doc/userdoc.rst
```
destogl, line 6: I would not include them all into one file, but rather reference them. Why not simply using `index.rst` as the main file in the each directory and than simply auto-generate all the links to them:

See here:
https://github.com/StoglRobotics/ros_team_workspace/blob/rolling/docs/use-cases/index.rst

This generate nice menu structure in the docs (check [here](https://stoglrobotics.github.io/ros_team_workspace/use-cases/index.html))

diff_drive_controller/doc/userdoc.rst
```diff
@@ -0,0 +1,24 @@
+.. _diff_drive_controller_userdoc:
+
+diff_drive_controller
+---------------------
+
+Controller for differential drive wheel systems. Control is in the form of a velocity command, that is split then sent on the two wheels of a differential drive wheel base. Odometry is computed from the feedback from the hardware, and published.
```
destogl, line 6:
```suggestion
Controller for mobile systems with differential drive.
Input for control are velocity commands splitter for two wheels of the differential drive base.
Odometry is computed from the hardware's feedback and published.
```

destogl (commented):
I have to continue the review of the files, but for starters...

doc/controllers_index.rst
bmagyar: It looks pretty much the same to me:

Screenshot from 2021-03-17 07-18-04

Happy to have an alternative approach but I don't wish to spend more time on this now.

bmagyar:
@destogl  ping?

destogl (approved):
@bmagyar please merge.

Issue #158, opened 2021-03-25
Create joint_state_broadcaster to replace joint_state_controller


bmagyar:
We've recently decided to suffix all not-really-controllers (joint_state_controller, force_torque_controller, etc from ROS1) with *broadcaster* to follow the nomenclature of tf and to avoid name clashes with publishers (ROS) and observers (control theory). 

Task is considered complete when
* there's a new package `joint_state_broadcaster`, with all references to `joint_state_controller` gone
* the `joint_state_controller` package still exists but includes a very minimal amount of code to make the old name still available -> you can probably pull this off by tweaking the `joint_state_plugin.xml` file and without having to have additional C++ code
* add a deprecation note to `joint_state_plugin.xml` and other visible places (if any) to `joint_state_controller`
* one can load either `joint_state_broadcaster` or `joint_state_controller` and both produce the same result

Supersedes #110

matthew-reynolds:
I don't have permissions in this repo to self-assign, but I will take this on.

matthew-reynolds:
I think this can be closed

PR #160, opened 2021-03-28
Rename joint_state_controller -> joint_state_broadcaster

matthew-reynolds:
Addresses #158.
 - Renamed controller to broadcaster
 - Updated documentation to use new terminology
 - Added shim controller with old name, and appropriate deprecation notices

---

Unfortunately, we require a little bit of C++ in the deprecated `joint_state_controller` in order for `pluginlib` to lookup the correct plugin, since `pluginlib` won't let us look up `.so`s generated from other ROS packages.

I also retained the `test_load_joint_state_controller` test, so that we can make sure the old name is still loadable even with the redirection.

In addition to the unit test, I ran the `ros2_control_demos` demo and ensured the correct data was published on `/joint_states` when using both the old and new `joint_state_controller/broadcaster` names.


joint_state_controller/doc/userdoc.rst
```diff
@@ -3,15 +3,10 @@
...
-Commands
-^^^^^^^^
-
-This is not a real controller, therefore is takes no commands.
-
-Hardware interface type
-^^^^^^^^^^^^^^^^^^^^^^^
-
-All available *joint state interfaces* are used by this controller.
+This package will be removed in ROS H-Turtle. Instead, use package
```
matthew-reynolds, line 9: Not sure if this ("H-Turle") is correct - Let me know and I can adjust the deprecation notice if required

destogl: I thinks this is OK. H-Turtle is next LTS and there we want to get rid of this :)

joint_state_broadcaster/joint_state_plugin.xml
```diff
@@ -0,0 +1,7 @@
+<library path="joint_state_broadcaster">
+  <class name="joint_state_broadcaster/JointStateBroadcaster" type="joint_state_broadcaster::JointStateBroadcaster" base_class_type="controller_interface::ControllerInterface">
+  <description>
+    The joint state broadcaster publishes the current joint state
```
destogl, line 4:
```suggestion
    The joint state broadcaster publishes the current states from ros2_control system.
```

I propose something more general. Especially because we can publish now all kind of states in the /dynamic_joint_state.

joint_state_broadcaster/package.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>joint_state_broadcaster</name>
+  <version>0.2.0</version>
+  <description>Broadcaster to publish joint state</description>
+  <maintainer email="bence.magyar.robotics@gmail.com">Bence Magyar</maintainer>
```
destogl, line 6:
```suggestion
  <maintainer email="bence.magyar.robotics@gmail.com">Bence Magyar</maintainer>
  <maintainer email="denis@stogl.de">Denis Stogl</maintainer>
```

I would add also myself. @bmagyar is this OK? If you think that the list is overblown, you can leave me out :)

joint_state_broadcaster/src/joint_state_broadcaster.cpp
```diff
@@ -0,0 +1,237 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
```
destogl, line 1:
```suggestion
// Copyright 2021 ros2_control development team
```

I don't think OSRF contributed much for this. At least after the most recent changes where the framework is completely restructured.

@bmagyar what do you think?

destogl (commented):
Just a few comments since we are doing a major change. The comments are mostly for @bmagyar.

@matthew-reynolds thanks for a great work!

joint_state_broadcaster/package.xml
bmagyar: if you want to get emails, so be it! :D

bmagyar (approved):

PR #161, opened 2021-03-29
[Joint Trajectory Controller] Add parameters for special cases

destogl:
Implements the following special cases:

- JTC can be used for open-loop control. States from the hardware are read when starting the controller. This enables its use for robots/manipulators which have offset between command and states, e.g., often the case with hydraulics manipulators.
- JTC can deduce position and velocity when positions or accelerations are as input in JointTrajectory.


bmagyar:
Please split this into 2 PRs
* open-loop mode
* deduce position and velocity

destogl:
> Please split this into 2 PRs
> 
>     * open-loop mode
> 
>     * deduce position and velocity

Done in #189 and #190.

Issue #132, opened 2020-12-23
Flaky joint_trajectory_controller test


bmagyar:
It causes sporadic failures on the CI

```
Function 'run_target_test' returned with code '0' after 0 min 27 sec
build/joint_trajectory_controller/Testing/20201223-1353/Test.xml: 10 tests, 0 errors, 1 failure, 0 skipped
- test_trajectory_actions
  <<< failure message
    -- run_test.py: invoking following command in '/root/target_ws/build/joint_trajectory_controller':
     - /root/target_ws/build/joint_trajectory_controller/test_trajectory_actions --gtest_output=xml:/root/target_ws/build/joint_trajectory_controller/test_results/joint_trajectory_controller/test_trajectory_actions.gtest.xml
    Running main() from /opt/ros/foxy/src/gtest_vendor/src/gtest_main.cc
    [==========] Running 6 tests from 1 test case.
    [----------] Global test environment set-up.
    [----------] 6 tests from TestTrajectoryActions
    [ RUN      ] TestTrajectoryActions.test_success_single_point_sendgoal
    [INFO] [1608731594.223464761] [test_joint_trajectory_controller]: Controller state will be published at 50Hz.
    [INFO] [1608731594.224467156] [test_joint_trajectory_controller]: Action status changes will be monitored at 20Hz.
    [INFO] [1608731594.528069895] [test_joint_trajectory_controller]: Received new action goal
    [INFO] [1608731594.528172294] [test_joint_trajectory_controller]: Accepted new action goal
    [INFO] [1608731595.028759186] [test_joint_trajectory_controller]: Goal reached, success!
    [       OK ] TestTrajectoryActions.test_success_single_point_sendgoal (2025 ms)
    [ RUN      ] TestTrajectoryActions.test_success_multi_point_sendgoal
    [INFO] [1608731596.241568953] [test_joint_trajectory_controller]: Controller state will be published at 50Hz.
    [INFO] [1608731596.241845852] [test_joint_trajectory_controller]: Action status changes will be monitored at 20Hz.
    [INFO] [1608731596.544673796] [test_joint_trajectory_controller]: Received new action goal
    [INFO] [1608731596.545481792] [test_joint_trajectory_controller]: Accepted new action goal
    [INFO] [1608731596.846115746] [test_joint_trajectory_controller]: Goal reached, success!
    [       OK ] TestTrajectoryActions.test_success_multi_point_sendgoal (2015 ms)
[...]

PR #164, opened 2021-04-05
Fix JTC segfault

matthew-reynolds:
# Purpose
Addresses #132. Fixes a segfault that could occur in regular use of the joint_trajectory_controller, as exposed by the unit tests.

# Summary
There was a race condition in the JTC, where `rt_active_goal_` could be `reset()` in one thread (`joint_trajectory_controller.cpp:568`) but then dereferenced elsewhere. This would cause a segfault.

For example, in `JointTrajectoryController::update()`, we check that `rt_active_goal_` is non-null, and then dereference it a couple lines later. But if we get unlucky with timing, we can reset the shared_ptr in another thread in-between the non-null check and the dereference. We don't currently have the appropriate thread safety mecanisms in place.

By taking a copy of the `rt_active_goal_` shared ptr before checking and using it, we ensure the local copy will never becoming invalid while we're holding it. The `RealtimeBuffer` is required since we need to read and write to the shared ptr concurrently from multiple threads.

# Testing done
:heavy_check_mark: `colcon test --packages-select joint_trajectory_controller --retest-until-fail 100`

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -186,7 +186,8 @@ JointTrajectoryController::update()
         }
       }
 
-      if (rt_active_goal_) {
+      const auto active_goal = get_active_goal();
+      if (active_goal) {
         // send feedback
         auto feedback = std::make_shared<FollowJTrajAction::Feedback>();
         feedback->header.stamp = node_->now();
@@ -195,7 +196,7 @@ JointTrajectoryController::update()
         feedback->actual = state_current;
         feedback->desired = state_desired;
         feedback->error = state_error;
-        rt_active_goal_->setFeedback(feedback);
+        active_goal->setFeedback(feedback);
 
         // check abort
         if (abort || outside_goal_tolerance) {
@@ -208,17 +209,16 @@ JointTrajectoryController::update()
             RCLCPP_WARN(node_->get_logger(), "Aborted due to goal tolerance violation");
             result->set__error_code(FollowJTrajAction::Result::GOAL_TOLERANCE_VIOLATED);
           }
-          rt_active_goal_->setAborted(result);
-          rt_active_goal_.reset();
+          active_goal->setAborted(result);
+          clear_active_goal();
         }
-
         // check goal tolerance
-        if (!before_last_point) {
+        else if (!before_last_point) {
           if (!outside_goal_tolerance) {
             auto res = std::make_shared<FollowJTrajAction::Result>();
             res->set__error_code(FollowJTrajAction::Result::SUCCESSFUL);
-            rt_active_goal_->setSucceeded(res);
-            rt_active_goal_.reset();
+            active_goal->setSucceeded(res);
+            clear_active_goal();
 
             RCLCPP_INFO(node_->get_logger(), "Goal reached, success!");
           } else if (default_tolerances_.goal_time_tolerance != 0.0) {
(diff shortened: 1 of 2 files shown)


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -583,17 +585,18 @@ void JointTrajectoryController::feedback_setup_callback(
...
+  rt_goal->preallocated_feedback_->joint_names = joint_names_;
+  rt_goal->execute();
+  rt_active_goal_.writeFromNonRT(rt_goal);
+
   // Setup goal status checking timer
   goal_handle_timer_ = node_->create_wall_timer(
     action_monitor_period_.to_chrono<std::chrono::seconds>(),
-    std::bind(&RealtimeGoalHandle::runNonRealtime, rt_active_goal_));
+    std::bind(&RealtimeGoalHandle::runNonRealtime, rt_goal));
 }
```
matthew-reynolds, line 600: I think this needs further work, probably in a follow-up PR.

In both the old and new code, I worry that we might end up in a condition where we're overwriting the `goal_handle_timer_` while the previous `rt_active_goal_` has pending operations, and thus dropping those operations. I think we should call `goal_handle_timer_::execute_callback()` or `rt_active_goal_.readFromNonRT()->runNonRealtime()` before creating the new timer.

Similarly, I think we could probably end up in a race condition where something is resetting the `rt_active_goal_` right after this function writes the new `rt_goal`, and we could end up losing the goal handle. This would involve a little more work to solve, for example only resetting the `rt_active_goal_` if `new_data_available_ == false`. That variable is private, but something along those lines.

Would appreciate your thoughts so we can consider appropriate follow-up PRs.

matthew-reynolds:
The CI failure looks unrelated: `colcon: error: Mixin 'coverage-gcc' is not available for 'test'`.

ros-tooling/setup-ros v0.1.3 was just released a few days ago, maybe something changed?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -208,17 +209,18 @@ JointTrajectoryController::update()
             RCLCPP_WARN(node_->get_logger(), "Aborted due to goal tolerance violation");
             result->set__error_code(FollowJTrajAction::Result::GOAL_TOLERANCE_VIOLATED);
           }
-          rt_active_goal_->setAborted(result);
-          rt_active_goal_.reset();
-        }
+          active_goal->setAborted(result);
+          // TODO(matthew-reynolds): Need a lock-free write here
```
bmagyar, line 213: Please create a follow-up issue with a short write-up and insert the link of that here (and other instances) instead. Much easier to follow up on it that way.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: This write-up above is a perfect start for the description of a follow-up issue ;)

bmagyar (commented):
looks good to me, just 2 notes about creating follow-up issues

bmagyar:
Thanks a bunch for this fix, a big step toward a green pipeline! :D

joint_trajectory_controller/src/joint_trajectory_controller.cpp
matthew-reynolds: @destogl beat me to it! See #166

joint_trajectory_controller/src/joint_trajectory_controller.cpp
matthew-reynolds: Done, see #168 and #169

PR #173, opened 2021-04-21
Add admittance controller

destogl:
Add stock admittance controller.

Description - TBD﻿


AndyZe:
I think you should squash my commits down to 1   :+1:

admittance_controller/src/incremental_ik_calculator.cpp
```diff
@@ -74,8 +73,17 @@ bool IncrementalIKCalculator::convertCartesianDeltasToJointDeltas(Eigen::VectorX
   svd_ = Eigen::JacobiSVD<Eigen::MatrixXd>(jacobian_, Eigen::ComputeThinU | Eigen::ComputeThinV);
   matrix_s_ = svd_.singularValues().asDiagonal();
   pseudo_inverse_ = svd_.matrixV() * matrix_s_.inverse() * svd_.matrixU().transpose();
```
olivier-stasse, line 75: My two cents on this part of the code. If you are going through a singularity, basically if the "geometrical/configuration Jacobian" is near losing a rank,  the singular values will be going near zero, and the inverse will generate extremely high speed. The usual way is to fix this problem is to set to zero the speed of the joint which is going through the singularity. If you have a powerful robot, clamping the speed will not be enough.

admittance_controller/include/admittance_controller/admittance_rule.hpp
```diff
@@ -0,0 +1,162 @@
...
+  std::string ik_group_name_;
+
+  // Controller frames
+  std::string control_frame_;
+  std::string endeffector_frame_;
+  std::string fixed_world_frame_;
+  std::string sensor_frame_;
+
+  // Admittance parameters
+  //   bool unified_mode_ = false;
```
AndyZe, line 81: No idea what unified mode means. Needs a comment, eventually

admittance_controller/include/admittance_controller/admittance_rule.hpp
```diff
@@ -0,0 +1,162 @@
...
+  Eigen::VectorXd delta_theta_;
+
+  // Transformation variables
+  std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
+  std::shared_ptr<tf2_ros::TransformListener> tf_listener_;
+
+  geometry_msgs::msg::WrenchStamped measured_force_;
+  geometry_msgs::msg::WrenchStamped measured_force_control_frame_;
+
+  geometry_msgs::msg::PoseStamped current_origin_;
```
AndyZe, line 100: Current origin of what? Needs a comment

admittance_controller/src/admittance_controller.cpp
```diff
@@ -94,83 +309,119 @@ const
...
-
-  // Calculate desired Cartesian displacement of the robot
-  //delta_x_ = ...
+  // get input commands
+  // TODO(destogl): Enable this when unified mode is used
+  auto input_force_cmd = input_force_command_.readFromRT();
+  auto input_pose_cmd = input_pose_command_.readFromRT();
 
-  // Get current robot joint angles
+  // Position has to always there
```
AndyZe, line 397: ^ sentence doesn't make sense

doc/admittance_controller.rst
```diff
@@ -0,0 +1,58 @@
+admittance_controller
+======================
+
+The controller uses admittance dynamics to calculate robot movements caused by an external force.
+The standard admittance control law: F = M\*a + D\*v + K\*(x_d - x) is used.
+
+The paramters are adjustable using ``dynamic_reconfigure`` functionality and rqt-plugin.
+
+The controller has following parameters:
```
AndyZe, line 9: Really appreciate this documentation!

admittance_controller/src/admittance_rule.cpp
```diff
@@ -240,22 +249,25 @@ controller_interface::return_type AdmittanceRule::update(
...
-  // Get current robot joint angles
-
-  // Convert Cartesian deltas to joint angle deltas via Jacobian
-  //   if (!ik_->convertCartesianDeltasToJointDeltas(delta_x_, sensor_frame_, delta_theta_))
-  //   {
-  //     RCLCPP_ERROR(get_node()->get_logger(), "Conversion of Cartesian deltas to joint deltas failed.");
-  //     return controller_interface::return_type::ERROR;
-  //   }
+  transform = tf_buffer_->lookupTransform(ik_base_frame_, ik_tip_frame_, tf2::TimePointZero);
+  std::vector<double> relative_desired_pose_v_(relative_desired_pose_vec_.begin(), relative_desired_pose_vec_.end());
```
AndyZe, line 261: No underscore:  `relative_desired_pose_v`

admittance_controller/include/admittance_controller/incremental_ik_calculator.hpp
```diff
@@ -31,16 +32,27 @@ class IncrementalIKCalculator
...
-  IncrementalIKCalculator(std::shared_ptr<rclcpp::Node> & node);
+  IncrementalIKCalculator(const std::shared_ptr<rclcpp::Node> & node, const std::string & group_name);
 
   /**
    * \brief Convert Cartesian delta-x to joint delta-theta, using the Jacobian.
-   * \param delta_x input Cartesian deltas
-   * \param delta_x_frame input name of the delta_x tf frame
-   * \param delta_theta output
+   * \param delta_x_vec input Cartesian deltas (x, y, z, rx, ry, rz)
+   * \param ik_base_to_tip_trafo transformation between ik base and ik tip
```
AndyZe, line 40: I hate this `_trafo` suffix. Can we call it:  `ik_base_to_tip_tf`

admittance_controller/include/admittance_controller/admittance_rule.hpp
```diff
@@ -34,21 +34,22 @@ class AdmittanceRule
 public:
   AdmittanceRule() = default;
 
-  controller_interface::return_type configure(rclcpp::Clock::SharedPtr clock);
+  controller_interface::return_type configure(rclcpp::Node::SharedPtr node);
 
   controller_interface::return_type update(
```
AndyZe, line 39: Worth a comment here:

// This is where the admittance calculation is done

admittance_controller/src/admittance_rule.cpp
```diff
@@ -240,22 +249,25 @@ controller_interface::return_type AdmittanceRule::update(
     }
   }
 
+  // TODO: Add here desired vector at the IK tip and not tool!!!!
```
AndyZe, line 252: I'm going to add notes for several lines that aren't reachable from this PR.

Line 213 (above) seems unnecessary. Please double-check

AndyZe: L227:

TODO: this formula could cause a "gimbal lock" if the angular difference is large

AndyZe: Above L239, add comment:

// Acceleration is regarded as the control input. It is integrated to calculate velocity and acceleration.

admittance_controller/src/admittance_rule.cpp
```diff
@@ -240,22 +249,25 @@ controller_interface::return_type AdmittanceRule::update(
...
+
   // Do clean conversion to the goal pose using transform and not messing with Euler angles
   convert_array_to_message(relative_desired_pose_vec_, relative_desired_pose_);
   tf2::doTransform(current_pose_control_frame_, desired_pose_, relative_desired_pose_);
 
   // Use Jacobian-based IK
   // Calculate desired Cartesian displacement of the robot
```
AndyZe, line 259: Update the comment (prob my mistake):

// Convert desired position deltas to joint deltas

admittance_controller/src/incremental_ik_calculator.cpp
AndyZe: What about clamping acceleration? We have the capacity to do that fairly easily with Reflexxes.

olivier-stasse: A simple implementation will be to iterate over the diagonal element of matrix_s_ and compute the following:
`pinv_matrix_s_(i,i) = matrix_s_(i,i) /( matrix_s_(i,i) * matrix_s_(i,i) + rho )`
with rho a regularisation parameter to be defined by the user.

You can find a tutorial on damped least square here:
https://link.springer.com/content/pdf/10.1007/BF01254007.pdf

Two notes on this: 
1/ This implies that you will not be following exactly the admittance behavior near a singularity. 
2/ many consider that formulating the pseudo inverse as an optimization problem is making its incremental extension easier.

For instance if you want to take into account the robot limits it is much more easier when written as an optimization problem.

@AndyZe this is a good idea. As far as I understood the Reflexx library it allows to define the trajectory dynamics in the task space, but I am not sure that the robot dynamics is taken into account. If not this is fine as long as you do not go very fast. The faster you will go, the more important is to take into account the robot dynamics. Clamping the acceleration may not be always the most appropriate, it can be too late and too constraining depending if you are near a limit or not. Reflexx was started/written by Kroger Tarsten who was the head of IAR-PR. @destogl may have more precise comments than me on this point as he recently graduated from this team. (Congrats by the way @destogl , even if this is quite late).

nbbrooks: Fyi I am working on pulling in the [servo approach](https://github.com/ros-planning/moveit2/blob/7064da6e4b714481047fe55803e4bca0bbf32ada/moveit_ros/moveit_servo/src/servo_calcs.cpp#L696) for this into the admittance controller.

admittance_controller/src/admittance_controller.cpp
```diff
@@ -0,0 +1,640 @@
...
+    // TODO(destogl): add error handling
+    if ((*input_joint_cmd)->points[0].positions.empty()) {
+      for (auto index = 0u; index < num_joints; ++index) {
+        joint_deltas[index] = (*input_joint_cmd)->points[0].velocities[index] * period.seconds();
+      }
+    } else {
+      for (auto index = 0u; index < num_joints; ++index) {
+        // TODO(destogl): ATTENTION: This does not work properly, deltas are getting neutralized and robot is not moving on external forces
+        // TODO(anyone): Is here OK to use shortest_angular_distance?
+        joint_deltas[index] = angles::shortest_angular_distance(current_joint_states.positions[index], (*input_joint_cmd)->points[0].positions[index]);
```
AndyZe, line 489: First time looking at this in awhile!

I just wanted to note, this should probably use shortest_angular_distance_with_large_limits().

https://github.com/ros/angles/blob/811da0230f13348fd40337f63219c3b1520088f3/angles/include/angles/angles.h#L222

PR #175, opened 2021-04-23
Avoid std::stringstream

matthew-reynolds:
Similar to https://github.com/ros-controls/ros2_control/pull/391

We rarely use the `_STREAM` variants of RCLCPP logging, so we might as well just avoid them altogether to avoid the `std::stringstream`.

While I was in there, I also corrected the printf format for `size_t`. We previously had a mix of `%d` and `%u`. For correctness and portability, it should be `%zu`, which is specifically for `size_t`s.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -575,7 +575,7 @@ CallbackReturn DiffDriveController::configure_side(
   auto logger = node_->get_logger();
 
   if (wheel_names.empty()) {
-    RCLCPP_ERROR(logger, "No %s wheel names specified", side.c_str());
+    RCLCPP_ERROR(logger, "No '%s' wheel names specified", side.c_str());
     return CallbackReturn::ERROR;
   }
 
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -192,7 +192,7 @@ controller_interface::return_type DiffDriveController::update()
 
       if (std::isnan(left_position) || std::isnan(right_position)) {
         RCLCPP_ERROR(
-          logger, "Either the left or right wheel position is invalid for index [%d]",
+          logger, "Either the left or right wheel position is invalid for index [%zu]",
           index);
         return controller_interface::return_type::ERROR;
       }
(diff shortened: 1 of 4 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -575,9 +575,7 @@ CallbackReturn DiffDriveController::configure_side(
   auto logger = node_->get_logger();
 
   if (wheel_names.empty()) {
-    std::stringstream ss;
-    ss << "No " << side << " wheel names specified.";
-    RCLCPP_ERROR(logger, ss.str().c_str());
+    RCLCPP_ERROR(logger, "No %s wheel names specified", side.c_str());
```
destogl, line 578:
```suggestion
    RCLCPP_ERROR(logger, "No '%s' wheel names specified", side.c_str());
```

forward_command_controller/src/forward_command_controller.cpp
```diff
@@ -82,7 +82,7 @@ CallbackReturn ForwardCommandController::on_configure(
       rt_command_ptr_.writeFromNonRT(msg);
     });
 
-  RCLCPP_INFO_STREAM(get_node()->get_logger(), "configure successful");
+  RCLCPP_INFO(get_node()->get_logger(), "configure successful");
```
destogl, line 85: When you are already editing this, can you take a look at the following use?

```suggestion
  RCLCPP_INFO(node_->get_logger(), "configure successful");
```

forward_command_controller/src/forward_command_controller.cpp
```diff
@@ -162,12 +162,11 @@ controller_interface::return_type ForwardCommandController::update()
   }
 
   if ((*joint_commands)->data.size() != command_interfaces_.size()) {
-    RCLCPP_ERROR_STREAM_THROTTLE(
+    RCLCPP_ERROR_THROTTLE(
       get_node()->get_logger(),
```
destogl, line 166:
```suggestion
      node_->get_logger(),
```

destogl (approved):
Just a few small suggestions. Let me know what do you think about them. Thanks!

bmagyar:
I am wondering if there's a way to use logging properly without the bloat of `get_logger()` and other ugly stuff like that. `RCUTILS_LOG_ERROR_NAMED` seems to have a similar effect... any opinions/experience with this?

bmagyar (approved):

matthew-reynolds:
As I recall, the `RCLCPP` logging macros just take a string name from the `Node`'s logger which is then passed down to the `RCUTILS` macros. The `RCUTILS` macros are basically the same as the ROS1 `LOG_NAMED` macros.

I think the cleanest and most standard is to stick with the `RCLCPP` macros, and to use `node_` everywhere. We mostly use `node_` internally and `get_node()` externally, which I think is the right move - It's not too verbose and it matches the standard ROS C++ approach. I've got a PR ready that I'll put up once I get approval from work...

PR #178, opened 2021-04-30
Set C++ version to C++14 to fix macOS build

nkalupahana:
On macOS, when building `joint_state_controller`, the system defaults to C++11 and the build fails, because C++14 features are used. This fix forces the C++ standard version to C++14, thus fixing the issue. (This C++14 defaulting code is copied from the CMakeLists in `joint_trajectory_controller`, so I'm guessing that the fact that it wasn't in here was just an oversight.)


Karsten1987 (approved):
thank you!

PR #185, opened 2021-05-19
Reset external trajectory message upon activation

nbbrooks:
Reset external trajectory message to prevent preserving the old goal on systems with hardware offsets

- Fix has_trajectory_msg() function: two wrongs were making a right so functionally things were fine

### DIFF
--- a/joint_trajectory_controller/src/trajectory.cpp
+++ b/joint_trajectory_controller/src/trajectory.cpp
@@ -296,7 +296,7 @@ Trajectory::time_from_start() const
 bool
 Trajectory::has_trajectory_msg() const
 {
-  return trajectory_msg_.get() != 0;
+  return trajectory_msg_.get() != nullptr;
 }
 
 }  // namespace joint_trajectory_controller
--- a/joint_trajectory_controller/src/trajectory.cpp
+++ b/joint_trajectory_controller/src/trajectory.cpp
@@ -296,7 +296,7 @@ Trajectory::time_from_start() const
 bool
 Trajectory::has_trajectory_msg() const
 {
-  return !trajectory_msg_;
+  return trajectory_msg_.get() != 0;
 }
 
 }  // namespace joint_trajectory_controller
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -148,7 +148,7 @@ JointTrajectoryController::update()
   state_current.time_from_start.set__sec(0);
 
   // currently carrying out a trajectory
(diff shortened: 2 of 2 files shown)


matthew-reynolds (commented):
> to prevent preserving the old goal on systems with hardware offsets

What do you mean by "hardware offsets"? This seems like a good change regardless, I'm just not following this point.

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -296,7 +296,7 @@ Trajectory::time_from_start() const
 bool
 Trajectory::has_trajectory_msg() const
 {
-  return !trajectory_msg_;
+  return trajectory_msg_.get() != 0;
```
matthew-reynolds, line 299: Nitpick: 
```suggestion
  return trajectory_msg_.get() != nullptr;
```

bmagyar:
Could be related to https://github.com/ros-controls/ros2_controllers/pull/161 ?

joint_trajectory_controller/src/trajectory.cpp
nbbrooks: Good point, thanks

nbbrooks:
Yes, by hardware offsets I am referring to #161. I was in a rush getting this pushed up and did not write a good description. Essentially, the bug this address is if you are using `hardware_state_has_offset` to account for arms which may move while the controller is deactivated. Consider this workflow:

Command movement through JTC topic
// movement begins
Deactivate JTC
// movement stops
An arm offset is introduced (see motivations in 161)
Reactivate JTC
// movement resumes, trying to jump to previous point in pre-offset trajectory

Currently, the above workflow would see the previous trajectory being preserved and resumed when the JTC was reactivated. I do not think this is intended behavior, or at the very least, should not be if operating with `hardware_state_has_offset`.

destogl:
> Consider this workflow:
> 
>     * Command movement through JTC topic
> 
>     * // movement begins
> 
>     * Deactivate JTC
> 
>     * // movement stops
> 
>     * An arm offset is introduced (see motivations in 161)
> 
>     * Reactivate JTC
> 
>     * // movement resumes, trying to jump to previous point in pre-offset trajectory

This is actually also needed always when JTC is used along other controllers. So the issues exist always when robot if moved with another controller and switched back to JTC.

destogl (approved):

bmagyar:
@nbbrooks I have a fix ready for you but can't push as you disabled `Allow edits from maintainers.`

Please run `ament_uncrustify --reformat` and push or enable me pushing

bmagyar (approved):
I'd even consider this a bugfix, thanks!

PR #189, opened 2021-05-28
Enable JTC for hardware having offset from state measurements

destogl:
This PR replaces the first part of #161.

The goal is short:
- support "open-loop" control when hardware state is not exactly following the robot commands, i.e., there is systematic offset in the state measurements
- add approach to change controller without "jumps" from the command to state values:
  - when started HW interface has to set command_interfaces to N aN
  - controller checks if command interfaces are NaN --> then is a controller started for the first time, so read states
  - if command interface have a value --> so a controller was running already (supports switching between controllers)


joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -124,6 +124,12 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
     hardware_interface::HW_IF_EFFORT,
   };
 
+  // Parameters for some special cases, e.g. hydraulics powered robots
+  /// Read hardware states only when starting controller. This is useful when robot is not exactly
+  /// following the commanded trajectory.
+  bool hardware_state_has_offset_ = false;
+  trajectory_msgs::msg::JointTrajectoryPoint current_state_when_offset_;
+
```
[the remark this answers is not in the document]
destogl: Thanks for the great comment! This is an excellent idea. I was also not happy with the naming!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -307,6 +281,92 @@ JointTrajectoryController::update()
...
+      assign_point_from_interface(state.velocities, joint_command_interface_[1]);
+    } else {
+      state.velocities.clear();
+      has_values = false;
+    }
+  } else {
+    state.velocities.clear();
+  }
+  // Acceleration is used only in combination with velocity
+  if (has_acceleration_state_interface_) {
```
destogl, line 356: Use here the same logic as in line L302

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -165,32 +149,19 @@ JointTrajectoryController::update()
...
 
   // currently carrying out a trajectory
   if (traj_point_active_ptr_ && (*traj_point_active_ptr_)->has_trajectory_msg()) {
     // if sampling the first time, set the point before you sample
     if (!(*traj_point_active_ptr_)->is_sampled_already()) {
-      (*traj_point_active_ptr_)->set_point_before_trajectory_msg(
-        node_->now(), state_current);
+      if (open_loop_control_) {
+        (*traj_point_active_ptr_)->set_point_before_trajectory_msg(
+          node_->now(), last_commanded_state_);
```
destogl, line 160: Should `last_commanded_state_` here be reduced to existing state interface from the hardware? Or is it OK to internally manage more control-states then can be received from hardware.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -67,6 +67,8 @@ JointTrajectoryController::init(const std::string & controller_name)
   node_->declare_parameter<double>("state_publish_rate", 50.0);
   node_->declare_parameter<double>("action_monitor_rate", 20.0);
   node_->declare_parameter<bool>("allow_partial_joints_goal", allow_partial_joints_goal_);
+  node_->declare_parameter<bool>(
+    "open_loop_control", open_loop_control_);
```
bmagyar, line 71:
```suggestion
  node_->declare_parameter<bool>("open_loop_control", open_loop_control_);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: I think this should only set the ones you'd see from commands too, otherwise may result in strange discrepancies. Happy to have a follow-up issue for it

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -630,6 +694,17 @@ JointTrajectoryController::on_activate(const rclcpp_lifecycle::State &)
   traj_point_active_ptr_ = &traj_external_point_ptr_;
   last_state_publish_time_ = node_->now();
 
+  // Initialize current state storage if hardware state has tracking offset
```
bmagyar, line 697: should all the last command management happen *only* when in open loop?

bmagyar (approved):
Mostly nitpick, otherwise looks good to me!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
bmagyar: Understood, thanks!

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -231,13 +240,29 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
...
+    trajectory_msgs::msg::JointTrajectoryPoint & point, size_t size)
+  {
+    point.positions.resize(size);
+    if (has_velocity_state_interface_) {
+      point.velocities.resize(size);
+    }
+    if (has_acceleration_state_interface_) {
+      point.accelerations.resize(size);
+    }
+  }
```
v-lopez, line 265: nitpick, is there any reason why this is on the header?

v-lopez (approved):
LGTM, just minor comment that can be ignored.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
destogl: Not really. I simply thought it's a basic utility function, so I put it there. Should I move this to cpp file? Or should I create a utility namespace outside of the class for this?

v-lopez: Move to the cpp file for sure, otherwise if this gets included twice it will produce a linker double definition error.

PR #192, opened 2021-06-02
disable griper on OSX

Karsten1987:
`experimental/optional` is not supported on OSX. I believe the proper fix would be to `ifdef` the logic around optional, but for now I am simply disabling the compilation on OSX.

### DIFF
--- a/gripper_controllers/CMakeLists.txt
+++ b/gripper_controllers/CMakeLists.txt
@@ -1,6 +1,11 @@
 cmake_minimum_required(VERSION 3.5)
 project(gripper_controllers)
 
+if(APPLE)
+  message(WARNING "gripper controllers are not available on OSX")
+  return()
+endif()
+
 # Default to C++14
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)


destogl (commented):
Is this really needed to be merged to master? Do we actually habe OSX users? With all the fixes, it sounds like ros2_control is not hardly working on it.

Karsten1987:
I don't see why this PR disturbs any non-OSX users. ROS2 (and ros2_control) works pretty well on OSX so far. I've developed it from day one on OSX ;-)

But it's fine for me if we don't want to merge this. I'll continue working on an internal fork then.

nkalupahana:
@Karsten1987 What's the status on a fix for this? For now, on my own fork, I'm just switching to C++17, where <optional> is a header on macOS, but that can't be merged in for foxy. All I can think of for a proper fix for this is to use `boost::optional`, but that's not super ideal. Do you have any ideas? I'm free to take this on.

bmagyar:
Merging hack, hopefully we can remove it in galactic

Karsten1987:
@nkalupahana thanks for the interest. Foxy does not support c++17, we're bound to c++14. I believe that's where the experimental part came in. If we ever decide to branch out for galactic, we can leverage all c++17 features including `<optional>` (amongst others). 
For Foxy though, I don't see a really nice workaround at the moment. Introducing boost is just as brutal and at least for the rest of ros2, we've mainly worked around having it. Disabling the gripper controller for now at least gets me going on OSX with foxy.

PR #204, opened 2021-06-17
Fix deprecation warnings on Rolling, remove rcutils dependency

bmagyar:
FYI @Timple

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -42,7 +42,7 @@
 #include "trajectory_msgs/msg/joint_trajectory.hpp"
 #include "trajectory_msgs/msg/joint_trajectory_point.hpp"
 
-using namespace std::chrono_literals;
+using namespace std::chrono_literals; // NOLINT
 
 namespace rclcpp_action
 {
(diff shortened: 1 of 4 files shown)


Timple (approved):

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -35,13 +36,14 @@
...
 #include "rclcpp_lifecycle/lifecycle_publisher.hpp"
 #include "rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp"
-#include "rcutils/time.h"
 #include "realtime_tools/realtime_buffer.h"
 #include "realtime_tools/realtime_publisher.h"
 #include "realtime_tools/realtime_server_goal_handle.h"
 #include "trajectory_msgs/msg/joint_trajectory.hpp"
 #include "trajectory_msgs/msg/joint_trajectory_point.hpp"
 
+using namespace std::chrono_literals; // NOLINT
```
matthew-reynolds, line 45: Suggest to avoid a `using namespace` in a header. `std::chrono_literals` is admittedly a fairly safe namespace to pull in, but it's also easy and much cleaner to narrow the scope and just stick it in the class or, better yet, right in the few functions where it's needed in the .hpp and .cpp

bmagyar: That was my first attempt but `using namespace` is not really something the compiler is happy about when it comes to that. I'd also add that this is a header of a controller, the only place where this is re-used is if someone inherits from it.

v-lopez (approved):

destogl (approved):
Thanks!

PR #205, opened 2021-06-23
Fixes for Windows

Ace314159:
I changed a couple things to get this working on Windows.
1. For the same reason as #192, the gripper controller doesn't work on Windows as well, so I disabled it.
2. `M_PI` is not defined unless `_USE_MATH_DEFINES` is also defined
3. I had to export some protected functions of the joint trajectory controller to fix missing symbol errors.
4. Lastly, `cxxabi.h` doesn't exist on MSVC, so I didn't include it when MSVC was detected. It still still seems to compile fine without it.

### DIFF
--- a/diff_drive_controller/test/test_accumulator.cpp
+++ b/diff_drive_controller/test/test_accumulator.cpp
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#ifdef _MSC_VER
 #define _USE_MATH_DEFINES
+#endif
 #include <diff_drive_controller/rolling_mean_accumulator.hpp>
 
 #include <gmock/gmock.h>
--- a/diff_drive_controller/test/test_accumulator.cpp
+++ b/diff_drive_controller/test/test_accumulator.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#define _USE_MATH_DEFINES
 #include <diff_drive_controller/rolling_mean_accumulator.hpp>
 
 #include <gmock/gmock.h>
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -194,7 +194,8 @@ protected:
   // sorts the joints of the incoming message to our local order
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC void sort_to_local_joint_order(
     std::shared_ptr<trajectory_msgs::msg::JointTrajectory> trajectory_msg);
-  JOINT_TRAJECTORY_CONTROLLER_PUBLIC bool validate_trajectory_msg(const trajectory_msgs::msg::JointTrajectory & trajectory) const;
+  JOINT_TRAJECTORY_CONTROLLER_PUBLIC bool validate_trajectory_msg(
+    const trajectory_msgs::msg::JointTrajectory & trajectory) const;
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC void add_new_trajectory_msg(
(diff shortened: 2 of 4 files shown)


bmagyar:
Could we have `_USE_MATH_DEFINES` added in the cmake as a flag by any chance? It would move the file-based hack to a more hidden place and may avoid having to add it to more files in the future where we have a need for M_PI.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -179,39 +179,39 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   rclcpp::Duration action_monitor_period_ = rclcpp::Duration(RCUTILS_MS_TO_NS(50));
 
   // callbacks for action_server_
-  rclcpp_action::GoalResponse goal_callback(
+  JOINT_TRAJECTORY_CONTROLLER_PUBLIC rclcpp_action::GoalResponse goal_callback(
```
bmagyar, line 182: Please try to make it look nice at least, the visibility macros should be out of the way of the actual code.
```suggestion
  JOINT_TRAJECTORY_CONTROLLER_PUBLIC
  rclcpp_action::GoalResponse goal_callback(
```

bmagyar:
Also please fix uncrustify and cpplint

Ace314159:
I could add the define using `add_definitions` and that would fix this issue as well as future issues for only the `diff_drive_controller`. This would be fine in this case because it is used in a cpp file. However, if `M_PI` were to be used in a header file, then having it in the CMake would require anyone using the header file to also have that define, which I think is not ideal.

Karsten1987:
thanks @Ace314159 for picking this up. Let's see what the official ROS2 buildfarm is saying:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14806/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9555/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12472/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14959/)

Ace314159:
They all seem to fail at a similar point involving logging. I don't think I modified that code, and I don't get that error. Do you have any idea what might be causing that?

Karsten1987:
I think I misconfigured the job, let's try again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14813/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9562/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12479/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14968/)

Ace314159:
Looks like the only failing test is a linter one in ros2_control. I just put a fix for that.

Is the expected way to fix these linter errors to run `ament_uncrustify`. Is there a way to tell it to do it for all the packages in a workspace without having to manually specify all the packages.

Karsten1987:
>  This would be fine in this case because it is used in a cpp file. However, if M_PI were to be used in a header file, then having it in the CMake would require anyone using the header file to also have that define, which I think is not ideal.

I agree with this logic. I think the current solution pretty neat.

Karsten1987 (approved):

diff_drive_controller/test/test_accumulator.cpp
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#define _USE_MATH_DEFINES
```
destogl, line 15:
```suggestion
#ifndef _MSC_VER
#define _USE_MATH_DEFINES
#endif
```

destogl (approved):
It looks good now.

@Karsten1987 would it be possible to get Windows CI integrated?

PR #208, opened 2021-06-27
Add dependency to force_torque_sensor_broadcaster

nkalupahana:
`force_torque_sensor_broadcaster` uses `realtime_tools`, but it isn't marked as a dependency in its `package.xml`. This can cause builds to fail if colcon decides to build `realtime_tools` after `force_torque_sensor_broadcaster` (see CI: https://github.com/nkalupahana/ros2-foxy-macos/runs/2924666462?check_suite_focus=true)

This PR adds `realtime_tools` as a dependency, thus fixing this build issue. (CI: https://github.com/nkalupahana/ros2-foxy-macos/runs/2926064882?check_suite_focus=true)

### DIFF
--- a/force_torque_sensor_broadcaster/package.xml
+++ b/force_torque_sensor_broadcaster/package.xml
@@ -17,6 +17,7 @@
   <depend>pluginlib</depend>
   <depend>rclcpp</depend>
   <depend>rclcpp_lifecycle</depend>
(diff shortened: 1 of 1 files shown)


bmagyar:
Great find, thanks!

bmagyar (approved):

PR #211, opened 2021-07-06
use rolling mean from rcppmath

Karsten1987:
as per title.

With https://github.com/ros2/rcpputils/pull/133 being merged, we really should use rcppmath for the rolling mean.


destogl:
Where is this actually used?

will this available for foxy, galactic, and rolling, or do we have to branch out.

Karsten1987:
you need to branch out, I don't think it's getting backported to any existing distros. And this goes back to the discussion we had in the last WG meeting, that I'd really appreciate having master being the development branch.

destogl (approved):
It looks good.

bmagyar:
Thanks, love to see PRs with a negative balance ;) 
Not long until branching, this goes on the wait list until then.

bmagyar (approved):

bmagyar:
waiting a little more on https://github.com/ros2/rcpputils/issues/157

bmagyar:
Finally merging!

PR #218, opened 2021-07-16
[Joint State Broadcaster] Add option to publish joint states to local topics

destogl:
This PR is related/inspired by #217 and #218.

Changes:
- Use local topics for publishing joint states
- Parameterize local namespace topics.
- Unify test structure.
- Use get_node() instead of node_ everywhere.

What do you think?

### DIFF
--- a/joint_state_broadcaster/doc/userdoc.rst
+++ b/joint_state_broadcaster/doc/userdoc.rst
@@ -19,10 +19,7 @@ Parameters
 ----------
 
 ``use_local_topics``
-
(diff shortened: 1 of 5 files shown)


Timple:
Can't this be handled with a simple remap in the launchfile? Is just as much configuration for a user as modifying a parameter.

destogl:
> Can't this be handled with a simple remap in the launchfile? Is just as much configuration for a user as modifying a parameter.

Not really. What is you have multiple JointStateBroadcasters running? Then, which one would be remapped?

bmagyar (approved):

bmagyar:
@destogl this looks good to me, feel free to merge after resolving the conflicts

PR #222, opened 2021-08-12
Delete failing parameter undeclare in JointGroupPositionController

### LINKED ISSUE #221: JointGroupPositionController fails to initialize on Rolling due to rclcpp "undeclare_parameter" API change
There have already been [some fixes to other controllers](https://github.com/ros-controls/ros2_controllers/pull/199) for the rclcpp parameter API changes. Currently (as of 0.4.1) the `JointGroupPositionController` fails to initialize because it tries to undeclare a statically-typed parameter at [these lines](https://github.com/ros-controls/ros2_controllers/blob/6a53981982eeac1822bfca8e9d5e40156870c0b1/position_controllers/src/joint_group_position_controller.cpp#L41-L47), which produces an error:

```
[ros2_control_node-1] [INFO] [1628693935.184140389] [controller_manager]: Loading controller 'position_controller'
[ros2_control_node-1] Exception thrown during init stage with message: parameter 'interface_name' has invalid type: cannot undeclare an statically typed parameter 
[ros2_control_node-1] [ERROR] [1628693935.194137855] [controller_manager]: Could not initialize the controller named 'position_controller'
```

I wanted to get some additional context and input before I try to fix this myself: why is it necessary to undeclare the `interface_name` parameter declared by `ForwardCommandController`, and does anyone else already have a fix in mind?

schornakj:
This is one way to fix #221 and allow loading the JointGroupPositionController plugin on Rolling.

If I understand correctly, the JointGroupPositionController just sets the value of `interface_name_` to `hardware_interface::HW_IF_POSITION` instead of reading it from a parameter, so it shouldn't make a functional difference whether the parameter is explicitly undeclared or simply not read.

### DIFF
--- a/effort_controllers/src/joint_group_effort_controller.cpp
+++ b/effort_controllers/src/joint_group_effort_controller.cpp
@@ -40,8 +40,10 @@ JointGroupEffortController::init(
   }
 
   try {
-    // undeclare interface parameter used in the general forward_command_controller
-    get_node()->undeclare_parameter("interface_name");
+    // Explicitly set the interface parameter declared by the forward_command_controller
+    // to match the value set in the JointGroupEffortController constructor.
+    get_node()->set_parameter(rclcpp::Parameter("interface_name", hardware_interface::HW_IF_EFFORT));
+
   } catch (const std::exception & e) {
     fprintf(stderr, "Exception thrown during init stage with message: %s \n", e.what());
     return controller_interface::return_type::ERROR;
--- a/effort_controllers/src/joint_group_effort_controller.cpp
+++ b/effort_controllers/src/joint_group_effort_controller.cpp
@@ -43,7 +43,7 @@ JointGroupEffortController::init(
     // Explicitly set the interface parameter declared by the forward_command_controller
     // to match the value set in the JointGroupEffortController constructor.
     get_node()->set_parameter(
-          rclcpp::Parameter("interface_name", hardware_interface::HW_IF_EFFORT));
+      rclcpp::Parameter("interface_name", hardware_interface::HW_IF_EFFORT));
   } catch (const std::exception & e) {
     fprintf(stderr, "Exception thrown during init stage with message: %s \n", e.what());
     return controller_interface::return_type::ERROR;
--- a/effort_controllers/src/joint_group_effort_controller.cpp
+++ b/effort_controllers/src/joint_group_effort_controller.cpp
@@ -42,8 +42,8 @@ JointGroupEffortController::init(
   try {
     // Explicitly set the interface parameter declared by the forward_command_controller
     // to match the value set in the JointGroupEffortController constructor.
-    get_node()->set_parameter(rclcpp::Parameter("interface_name", hardware_interface::HW_IF_EFFORT));
-
+    get_node()->set_parameter(
+          rclcpp::Parameter("interface_name", hardware_interface::HW_IF_EFFORT));
   } catch (const std::exception & e) {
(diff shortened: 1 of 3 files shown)


destogl:
I think that solution goes into right direction. Maybe would be even better to set the parameter to `hardware_interface::HW_IF_POSITION` there. This would make it more explicit for someone reading it. And since we don't have dynamic reconfigure here its change during runtime would not change anything.

@schornakj What do you think?

position_controllers/src/joint_group_position_controller.cpp
```diff
@@ -38,14 +38,6 @@ JointGroupPositionController::init(const std::string & controller_name)
     return ret;
   }
 
-  try {
-    // undeclare interface parameter used in the general forward_command_controller
-    get_node()->undeclare_parameter("interface_name");
```
destogl, line 43: Add something like:

```
get_node()->set_parameter("interface_name", hardware_interface::HW_IF_POSITION);
```

destogl (commented):
I think that solution goes into the right direction. Maybe would be even better to set the parameter to `hardware_interface::HW_IF_POSITION` there. This would make it more explicit for someone reading it. And since we don't have dynamic reconfigure, its change during runtime would not change anything.

@schornakj What do you think?

position_controllers/src/joint_group_position_controller.cpp
bmagyar: setting is a very good alternative solution to this, I was a little uneasy about undeclaring anyways...

bmagyar:
also, in general, this should then apply to all forward command controller derivatives, right?

position_controllers/src/joint_group_position_controller.cpp
schornakj: I agree that setting it is a good idea. I've updated the branch to implement this change.

schornakj:
> also, in general, this should then apply to all forward command controller derivatives, right?

Good point, there are other plugins with the same issue. I can fix those here as well.

effort_controllers/src/joint_group_effort_controller.cpp
```diff
@@ -40,8 +40,10 @@ JointGroupEffortController::init(
   }
 
   try {
-    // undeclare interface parameter used in the general forward_command_controller
-    get_node()->undeclare_parameter("interface_name");
+    // Explicitly set the interface parameter declared by the forward_command_controller
+    // to match the value set in the JointGroupEffortController constructor.
+    get_node()->set_parameter(rclcpp::Parameter("interface_name", hardware_interface::HW_IF_EFFORT));
+
   } catch (const std::exception & e) {
```
destogl, line 47: I would use this opportunity also to make this catch more specific. According to docstring `set_parameter` method throws `rclcpp::exceptions::ParameterNotDeclaredException`. Should be make this more explicit? @schornakj, @bmagyar what do you think?

destogl (approved):
One small comment since we are already editing this part of the code.

@schornakj thanks for your PR and for addressing our comments very fast.

effort_controllers/src/joint_group_effort_controller.cpp
schornakj: I think I'd actually continue catching `std::exception`. If a different exception is thrown other than `ParameterNotDeclaredException` (which is not likely) we still want to catch it and cleanly exit the initialization function by returning an error code.

schornakj:
@destogl @bmagyar I applied the formatting changes required by `ament_cpplint`, so this is ready for CI again.

bmagyar (approved):
thanks a lot for this!

Issue #149, opened 2021-02-10
Joint trajectory controller detailed parameters don't exist on param server


livanov93:
As stated [here](https://github.com/PickNikRobotics/Universal_Robots_ROS2_Driver/issues/36) and [here](https://github.com/PickNikRobotics/Universal_Robots_ROS2_Driver/issues/36) parameters for checking the error through trajectory and on the goal position are not propagated on the parameter server.
We also had problems with checking the errors on the last trajectory point - the problem occurs [here](https://github.com/ros-controls/ros2_controllers/blob/a80ce8d2b9a22fa83978976bc4257ede70dbc810/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp#L147)

destogl:
@livanov93 Can you explain a bit more about your test-setup and what is expected behavior and what is the issue?

livanov93:
While testing joint_trajectory_controller on  the real hardware I was getting the `Aborted due to goal tolerance violation` error. Afterwards, decided to add parameter per joint to increase goal tolerance:

The paramaters were defined in the following manner:
```

controller_manager:
  ros__parameters:
    update_rate: 600  # Hz

    forward_command_controller_position:
      type: forward_command_controller/ForwardCommandController

    joint_state_controller:
      type: joint_state_controller/JointStateController

    ur_joint_trajectory_controller:
      type: joint_trajectory_controller/JointTrajectoryController

ur_joint_trajectory_controller:
  ros__parameters:
    joints:
      - shoulder_pan_joint
      - shoulder_lift_joint
      - elbow_joint
      - wrist_1_joint
      - wrist_2_joint
      - wrist_3_joint
    state_publish_rate: 100.0
    action_monitor_rate: 20.0
    allow_partial_joints_goal: false
    constraints:
      stopped_velocity_tolerance: 0.01
      goal_time: 0.0
      shoulder_pan_joint:
        trajectory: 0.05
        goal:       0.03
      shoulder_lift_joint:
        trajectory: 0.05
        goal:       0.03
      elbow_joint:
        trajectory: 0.05
        goal:       0.03
      wrist_1_joint:
        trajectory: 0.05
        goal:       0.03
      wrist_2_joint:
        trajectory: 0.05
        goal:       0.03
      wrist_3_joint:
        trajectory: 0.05
        goal:       0.03

forward_command_controller_position:
  ros__parameters:
    joints:
      - shoulder_pan_joint
[...]

PR #224, opened 2021-08-16
Avoid redeclaration of parameters (automatically declared parameters)

livanov93:
PR addresses #149. Parameters are needed for checks within jtc. 
As mentioned [here](https://github.com/ros-controls/ros2_controllers/issues/149#issuecomment-780488925) in the future adjusting detailed per joint parameters could be done dynamically.

EDIT:
PR enables existence of parameters uploaded via overrides (e.g. yaml files). It is blocked by [#ros2_control/PR504](https://github.com/ros-controls/ros2_control/pull/504)

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -59,55 +59,48 @@ controller_interface::return_type DiffDriveController::init(const std::string & 
 
   try
   {
-    auto node = get_node();
     // with the lifecycle node being initialized, we can declare parameters
-    node->declare_parameter<std::vector<std::string>>(
-      "left_wheel_names", std::vector<std::string>());
-    node->declare_parameter<std::vector<std::string>>(
-      "right_wheel_names", std::vector<std::string>());
+    auto_declare<std::vector<std::string>>("left_wheel_names", std::vector<std::string>());
+    auto_declare<std::vector<std::string>>("right_wheel_names", std::vector<std::string>());
 
-    node->declare_parameter<double>("wheel_separation", wheel_params_.separation);
-    node->declare_parameter<int>("wheels_per_side", wheel_params_.wheels_per_side);
-    node->declare_parameter<double>("wheel_radius", wheel_params_.radius);
-    node->declare_parameter<double>(
-      "wheel_separation_multiplier", wheel_params_.separation_multiplier);
-    node->declare_parameter<double>(
-      "left_wheel_radius_multiplier", wheel_params_.left_radius_multiplier);
-    node->declare_parameter<double>(
-      "right_wheel_radius_multiplier", wheel_params_.right_radius_multiplier);
+    auto_declare<double>("wheel_separation", wheel_params_.separation);
+    auto_declare<int>("wheels_per_side", wheel_params_.wheels_per_side);
+    auto_declare<double>("wheel_radius", wheel_params_.radius);
+    auto_declare<double>("wheel_separation_multiplier", wheel_params_.separation_multiplier);
+    auto_declare<double>("left_wheel_radius_multiplier", wheel_params_.left_radius_multiplier);
+    auto_declare<double>("right_wheel_radius_multiplier", wheel_params_.right_radius_multiplier);
 
-    node->declare_parameter<std::string>("odom_frame_id", odom_params_.odom_frame_id);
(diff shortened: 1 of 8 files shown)


destogl:
Are you sure this didn't work before? Controllers should have auto-declaration of parameters enabled...

Would be nice to get some test for this, showing what exactly wasn't working.

livanov93:
Pretty sure, I will try to make tests to show the problem.

destogl:
[This](https://github.com/ros-controls/ros2_control/blob/108671dd07fd520bf00ab25ff45b594d55060a3f/controller_interface/src/controller_interface.cpp#L33) line enables auto-declaration of parameters. That's the reason I am asking...

Ace314159:
I can confirm that this is needed. If you change the log level to debug, it will log the parameters. Regardless of the parameters specified, it is always 0.

bmagyar (approved):
Ok. We may need to revisit this in Galactic but if it's blocking on Foxy, I'm tempted to merge it now

bmagyar:
Speaking too soon perhaps :D 

@livanov93 @Ace314159 , what ROS distro are you guys running?
The tests came back with this:
```
 - joint_trajectory_controller.PositionVelocityAccelerationTrajectoryControllers/TrajectoryControllerTestParameterized correct_initialization_using_parameters/1
  <<< failure message
    unknown file
    C++ exception with description "parameter 'constraints.joint1.trajectory' has already been declared" thrown in the test body.
  >>>
- joint_trajectory_controller.PositionVelocityAccelerationTrajectoryControllers/TrajectoryControllerTestParameterized correct_initialization_using_parameters/2
  <<< failure message
    unknown file
    C++ exception with description "parameter 'constraints.joint1.trajectory' has already been declared" thrown in the test body.
  >>>
```

Is something wrong with how we test this perhaps? Or some bug in rclcpp that makes it work with the test setup but not otherwise?

livanov93:
@bmagyar 
I am on foxy, however [f996475 ](https://github.com/ros-controls/ros2_controllers/pull/224/commits/f9964752fb2b6bad95732dc8acddbc0bb8cb51c0) will fix the test passing where configure is called multiple times (like [here](https://github.com/ros-controls/ros2_controllers/blob/d34800cf2a8621c84b4e2add5154d507cfd1f875/joint_trajectory_controller/test/test_trajectory_controller.cpp#L277) and [here](https://github.com/ros-controls/ros2_controllers/blob/d34800cf2a8621c84b4e2add5154d507cfd1f875/joint_trajectory_controller/test/test_trajectory_controller.cpp#L277)) within the same test. The controller should be configurable multiple times definitely, so this commit tackles that problem.

Still I think that there is a misunderstanding of `allow_undeclared_parameters` in [controller_interface](https://github.com/ros-controls/ros2_control/blob/79ce11f06903e3dc1958c5957be4dff6a9cc9bef/controller_interface/src/controller_interface.cpp#L30).
In [documentation](https://github.com/ros2/rclcpp/blob/d0cd6bb0a46f5b2841e9f282c9eecfa8cffe1ffa/rclcpp/include/rclcpp/node_options.hpp#L337) we can find that this is enough for putting the parameters into node's namespace without declaration.
If you take a closer look few lines on a flag `automatically_declare_parameters_from_overrides` and its [documentation](https://github.com/ros2/rclcpp/blob/d0cd6bb0a46f5b2841e9f282c9eecfa8cffe1ffa/rclcpp/include/rclcpp/node_options.hpp#L357) it says if it is `false` the parameters from overrides (yaml files) won't appear even though `allow_undeclared_parameters` is `true`.

There is a small discussion on the documentation improvements [here](https://github.com/ros2/rclcpp/pull/1751) and [here](https://github.com/ros2/rclcpp/issues/1749).

So in my understanding controller_interface should have `allow_undeclared_parameters` and `automatically_declare_parameters_from_overrides` to `true` and then in each instance of controller (e.g. joint_trajectory_controller, JointGroupVelocityController...) we should check if node hase some parameter or not, and then act accordingly.

bmagyar:
Using `automatically_declare_parameters_from_overrides` sounds like a better fix than this.
I vaguely remember something about removing that for some reason... let's check

bmagyar:
https://github.com/ros-controls/ros2_control/pull/276

It seems this was removed due to some error report from you earlier ;) could you test locally that fixing in controller_interface.cpp actually works?

bmagyar (approved):

destogl (approved):
Looks good and sensible chnage! Thanks!

Issue #171, opened 2021-04-18
[JTC] Implement `velocity` and `effort` versions of JointTrajectoryController


destogl:
#140 implements a generic JointTrajectoryController supporting joint, velocity, and acceleration interfaces and their combinations.

For the brevity of PR, pure `velocity` or `effort` controllers are not implemented. Their commands are generated using the desired state and state error using a velocity feedforward term plus a corrective PID term (see ROS1 implementation).
Usage of this term can be controlled in the code by using `using_closed_loop_hw_interface_adapter` variable as mentioned [here](https://github.com/ros-controls/ros2_controllers/blob/74f7f52fc5f94013c16f497e863ebffd4c080bfe/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L248). This is just a proposal. Please feel free to change this at your convenience.

christophfroehlich:
I'll close this as this is implemented already

Issue #135, opened 2020-12-31
[joint_trajectory_controller] Implement [vel, acc, effort] controller variants


Jarijaas:
Tl;dr:  Port Joint Trajectory Controller variants from the ROS1: [joint_trajectory_controller.cpp](https://github.com/ros-controls/ros_controllers/blob/1a085c37d212d1b52f44b8b1ebb6167cef90cc65/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L36-L89)

---

Hey,

So I have been testing this controller with my robot. When I set `time_from_start: { sec: 5, nanosec: 0 } }`, traversing the trajectory should take 5 seconds, which is true, but the velocity is incorrect. The controller correctly set's the positions, but because the controller does not write new velocity values to the  command interface, the traversed trajectory is unsmooth; the joint moves too fast, so the joint movement stops before the new intermediate position is written.

Shouldn't the velocity be set using the command interface or did I misunderstood something?

I believe, that the velocity value should be set around [here](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L169-L187)

destogl:
Hi @Jarijaas,

this depends on the used hardware (robot). probably your robot is trying to drive maximal velocity (shortest time) between two positions. The current implementation of `joint_trajectory_controller` enables only position interface toward hardware ([see line#83](https://github.com/ros-controls/ros2_controllers/blob/b0e3d198667d885f8a0fbd46c02a247ffdeb85fb/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L83)). I believe this is the same functionality as in ROS1.

How do you expect that the controller behaves:
* set position and velocity for hardware, or, 
* set only velocity in the hardware.

Also, which hardware, i.e. interface implementation are you using?

Jarijaas:
Hey @destogl,

My hardware interface is based on the [demo](https://github.com/ros-controls/ros2_control_demos/tree/master/ros2_control_demo_hardware), I added velocity interface to it, and I use it to control dynamixel servos.

I would expect that the `joint_trajectory_controller` should support setting the desired velocity value for the joints using the hardware interface.  I'm probably going to fork this controller and add velocity hardware command interface support to it, as this would suit my needs.

gavanderhoorn:
@destogl wrote:
[...]

PR #225, opened 2021-08-23
[JTC] Implement effort-only command interface

Ace314159:
This addresses the effort part of #135 and #171. I've tested this on our team's robot and it works identically to the ROS1 version from our testing.

I just realized that the trajectory tolerance parameters fix (the second commit) is also addressed by #224. I can remove the commit if you'd like to merge that first instead.


Ace314159:
The test failures seem to be in `force_torque_sensor_broadcaster`, but I didn't modify that part of the code. Any ideas for what might be happening?

destogl:
> The test failures seem to be in `force_torque_sensor_broadcaster`, but I didn't modify that part of the code. Any ideas for what might be happening?

Don't worry about that. The tests are flaky.

Could you rebase the code since we changed the formatting?

Ace314159:
@destogl Does it look good now?

bmagyar (commented):
The changes look good in general, thanks for keeping it up to date!
Do you think this should be called effort though? I think that the values you are using are basically acceleration or strongly derived from it. Does it relate to forces in any way? Effort in ros_control was used somewhat as a catch-all for "the third value" next to Position and Velocity

Ace314159:
@bmagyar I'm not too sure either, but I just tried to mimic the behavior of ros_control.

However, I don't think this is acceleration. My team is using the effort value as the power output for the motor, so perhaps a better name might be power?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -525,6 +541,23 @@ JointTrajectoryController::on_configure(const rclcpp_lifecycle::State &)
...
+    for (const auto & joint_name : joint_names_)
+    {
+      // Init PID gains from ROS parameter server
+      const std::string prefix = "gains." + joint_name;
+      const auto k_p = node_->declare_parameter<double>(prefix + ".p", 0.0);
+      const auto k_i = node_->declare_parameter<double>(prefix + ".i", 0.0);
+      const auto k_d = node_->declare_parameter<double>(prefix + ".d", 0.0);
+      const auto i_clamp = node_->declare_parameter<double>(prefix + ".i_clamp", 0.0);
+      const auto velocity_ff = node_->declare_parameter<double>("velocity_ff." + joint_name, 0.0);
+      // Initialize PID
```
livanov93, line 555: use `auto_declare` from [here](https://github.com/ros-controls/ros2_control/blob/7a0d5169c6a0f4aa8381b0e57de22ee28e6172bf/controller_interface/include/controller_interface/controller_interface.hpp#L110)

livanov93 (changes requested):

Ace314159:
@livanov93 Does this look good?

Do you know if the test failures are just caused by flaky tests or if my changes caused it? There seem to be test failures on master as well.

livanov93:
> @livanov93 Does this look good?
> 
> Do you know if the test failures are just caused by flaky tests or if my changes caused it? There seem to be test failures on master as well.

I think it is flaky test. The rest looks fine.

destogl:
> However, I don't think this is acceleration. My team is using the effort value as the power output for the motor, so perhaps a better name might be power?

My two cents about this. I think that you are actually using acceleration, which results in moment/power of the motor. The only thing what is intransparently happening in PID-part is conversion between acceleration and power by gain of the PID.

IMHO, the proper way for this would be to use acceleration interface and convert in robot driver acceleration to power. then if you are using other controllers you don't have to adjust gain of it for your specific robot.

(I am delighted if someone gives me a reason that this "proper" was is not so "proper") :)

Ace314159:
@destogl Perhaps power is not the right term. In the motor code, the type is called `percent output`. The difference between what we supply to the motor and acceleration is that when we supply a value of 0, the motor completely stops, but an acceleration of 0 would still preserve the velocity. However, I don't think this is velocity either since the value we output doesn't map linearly to position

How would you convert acceleration to power? I was able to find some formulas online, but they required knowing the mass and a time for the system.

destogl:
@Ace314159 what I can remember circular acceleration of a motor is proportional to its moment. But this discussion is probably not sensible because what you are saying you have some kind of abstract value. So we can go with the term "EFFORT" in this case. What do you think?

P.S. For personal interest, I am wondering how are you tuning your PID then if you don't have any idea about the physics of the value the motor is accepting.

Ace314159:
@destogl Yeah, I think that sounds good.

We don't really know the physics of what the motor is accepting, but we are assuming/hoping it is consistent (which is probably not true). We can read values like the position and velocity of the motor, and we know that a positive value turns the motor in direction and a negative value turns it in the other direction. Then, we just tuned our PID following the guidelines we found from [StackExchange](https://robotics.stackexchange.com/questions/167/what-are-good-strategies-for-tuning-pid-loops). I'm not sure if a PID is the best way to do what we want, but from our cursory testing it seems to work well enough. Does this answer your question?

theunkn0wn1:
Is there anything blocking this merge?

Ace314159:
This is good on my end. @bmagyar @destogl Is there anything else you'd like me to do?

theunkn0wn1:
@Ace314159 Would it be possible to get a backport of this for the Foxy branch?

Ace314159:
@theunkn0wn1 Yes, I'll try to make a PR for that once this gets in since my team is using Foxy as well.

Ace314159:
@bmagyar @destogl Pinging again, but is there anything blocking this merge?

destogl:
Hi @Ace314159, we are waiting for #239 to get merged, and then we have to rebase this PR to it. This would reduce some merge collisions.

It would help us very much if you could rebase your changes to #239. Would you try this?

Ace314159:
Thanks for the response! It looks like there are still some comments on #239, so I'll wait for those to be resolved before rebasing.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -191,39 +197,41 @@ controller_interface::return_type JointTrajectoryController::update(
...
         assign_interface_from_point(joint_command_interface_[2], state_desired.accelerations);
       }
-      // TODO(anyone): Add here "if using_closed_loop_hw_interface_adapter" (see ROS1) - #171
-      //       if (check_if_interface_type_exist(
-      //           command_interface_types_, hardware_interface::HW_IF_EFFORT)) {
-      //         assign_interface_from_point(joint_command_interface_[3], state_desired.effort);
-      //       }
+      if (has_effort_command_interface_ && use_closed_loop_pid_adapter)
+      {
+        assign_interface_from_point(joint_command_interface_[3], tmp_command_);
```
destogl, line 233: Should we use here the same logic as for velocity? To assign efforts directly if no pids are used?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -533,7 +537,7 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
   }
 
   // TODO(livanov93): change when other option is implemented
```
destogl, line 535: This is not needed anymore.

destogl (commented):
Just a few comments.

I am a bit concerned about missing tests.

Ace314159:
I agree. I'll try to add tests following the structure of the ones added for the velocity-only interface.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
Ace314159: Actually, would it make sense to directly set the effort or use it without PIDs? Velocity is a physical quantity, so I can see why you can set it directly. However, as we discussed and concluded before, effort is an abstract quantity, which I think means that it's not possible to directly set it. It's only used indirectly to modify the physical quantities position, velocity, and acceleration via the PID.

Ace314159:
Do these tests look good?

I don't think the tests that seem to be failing are due to my changes since they seem to be failing on main as well.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
destogl: I don't know about that. Exactly because it is an abstract quantity, it could be “anything”. On the other hand, you are right because we would need another interpolator to see it directly. So until we have one… we can keep it also as it is.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -505,6 +514,8 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
     contains_interface_type(command_interface_types_, hardware_interface::HW_IF_VELOCITY);
   has_acceleration_command_interface_ =
     contains_interface_type(command_interface_types_, hardware_interface::HW_IF_ACCELERATION);
+  has_effort_command_interface_ =
+    contains_interface_type(command_interface_types_, hardware_interface::HW_IF_EFFORT);
 
   if (has_velocity_command_interface_)
```
destogl, line 520: It would be good to have also checks if command interfaces are set properly. I guess that we are currently supporting "effort" as use-only command interface.

Can you add those after line 544, similarly as done here, but without “else”. Important is to give error output if “effort” and one other command interface are defined.

Also, according to tests, it appears that the effort interface has as a requirement `"position"` and  `"velocity"` state interfaces. Can you check this also? This would help users to find faster issues with illegal configurations.

(It seems that this last check is missing for the velocity-only interface too...)

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1199,10 +1243,6 @@ TEST_F(TrajectoryControllerTest, incorrect_initialization_using_interface_parame
   command_interface_types_ = {"bad_name"};
   set_parameter_and_check_result();
 
-  // command interfaces: effort not yet implemented
-  command_interface_types_ = {"effort"};
```
destogl, line 1203: It would be then great to add here checks for illegal configurations, as mentioned in the comment above.

destogl: The same for velocity-only mode would be nice to have.

destogl (changes requested):
A few more changes! They are just minor, but let's make the code "finished" and helpful for users.

Thanks for the very fast follow-up!!!

destogl:
> Do these tests look good?
> 
> I don't think the tests that seem to be failing are due to my changes since they seem to be failing on main as well.

This is all fine.

Ace314159:
Thanks for the quick reviews!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -622,6 +616,22 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
       "interfaces are not present.");
     return CallbackReturn::FAILURE;
   }
+  else if (has_velocity_command_interface_ && command_interface_types_.size() == 1)
```
destogl, line 619: This checks here are a bit hard to understand. And it could happen that they pass if “acceleration state interface” is present. (This is also not explicitly tested.) It seems to me that there are too many checks with “else if”.

What do you think? Would it be better to separate checks? Then the code would be cleaner and functionality should be preserved, e.g.:
```suggestion
  if (has_velocity_command_interface_ && command_interface_types_.size() == 1)
```

destogl (approved):
Thanks for all the followups! I agree with this changes. I have only one minor comment that is potentially not covered with tests. It seems to me that there can happen error output if "acceleration" state interface is used.

Otherwise, very nice work!!!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
Ace314159: Just to clarify, should we allow or not allow the acceleration state interface when using effort-only or velocity-only command interfaces?

destogl: We can allow it, but I think that in this case if acceleration interface is there, other combination will not be checked. Maybe I am misunderstanding something (?).

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -608,29 +608,33 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
...
-  }
-  else if (has_effort_command_interface_)
+  else
   {
-    RCLCPP_ERROR(
-      logger,
-      "'effort' command interface can only be used alone if 'velocity' and "
-      "'position' state interfaces are present");
-    return CallbackReturn::FAILURE;
+    if (has_acceleration_state_interface_)
```
destogl, line 613: I think that changing this line should result in issues… Because there is no actual check for "position" and "velocity" interfaces. In the previous code, this was checked implicitly. So here should stay `else if`.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -608,29 +608,33 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
...
+    {
+      RCLCPP_ERROR(
+        logger,
+        "'acceleration' state interface cannot be used if 'position' and 'velocity' "
+        "interfaces are not present.");
+      return CallbackReturn::FAILURE;
+    }
+    if (has_velocity_command_interface_ && command_interface_types_.size() == 1)
```
destogl, line 621: This is good now.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -608,29 +608,33 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
...
+    {
+      RCLCPP_ERROR(
+        logger,
+        "'velocity' command interface can only be used alone if 'velocity' and "
+        "'position' state interfaces are present");
+      return CallbackReturn::FAILURE;
+    }
+    // effort is always used alone so no need for size check
+    if (has_effort_command_interface_)
```
destogl, line 630: This should have failed on tests… because always when you are having “effort” command interface this will fail. We should at least check if there is only one interface presents, as done for the velocity. But I would be more happy if we make this checks more explicit.

They are correct, but unclear/confusing.

destogl (changes requested):

destogl (approved):
Sorry for the last review, I was reading the code wrongly!

Everything is fine! Thanks!

Ace314159:
@destogl Is there anything else you'd like me to do? Are we just waiting for the review from @bmagyar

destogl:
> @destogl Is there anything else you'd like me to do? Are we just waiting for the review from @bmagyar

I am waiting for another review. I will also ping other people in today's WG meeting.

lorepieri8:
Any news on this?

aprotyas (approved):

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -191,39 +191,48 @@ controller_interface::return_type JointTrajectoryController::update(
...
+        }
+      }
       if (has_position_command_interface_)
       {
         assign_interface_from_point(joint_command_interface_[0], state_desired.positions);
       }
       if (has_velocity_command_interface_)
       {
-        if (!use_closed_loop_pid_adapter)
+        if (use_closed_loop_pid_adapter)
```
bmagyar, line 212: This set of cascading `if`-s should be refactored into something more readable.
I suggest 2 things:
* make the selection of choosing the command somehow more explicit with different variable names, etc: `tmp_command_` is not very intuitive
* refactor all the `assign_interface_from_point` uses into a function that gets a parameter, the command (whether it's desired state or the pid command)

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -595,7 +599,7 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
 
   if (has_velocity_state_interface_)
   {
-    if (!contains_interface_type(state_interface_types_, hardware_interface::HW_IF_POSITION))
+    if (!has_position_state_interface_)
```
bmagyar, line 602: Investigate how to refactor these cascading `if`s so that it is easier to read which conditions together lead to a failure.
This section of  the code looks like a table of booleans where in some cases an error is flagged up.

bmagyar (approved):
The PR is good to go!

Two follow-up issues I created for refactoring:
https://github.com/ros-controls/ros2_controllers/issues/325
https://github.com/ros-controls/ros2_controllers/issues/326

Ace314159:
@bmagyar  Would it be possible to backport this to foxy? Someone else also requested this and this would be useful for my team as well.

theunkn0wn1:
Yeah, a foxy backport would be extremely useful for my team.

destogl:
Sorry guys but that is not possible because of too many changes in the mean time. you can try to resolve conflicting PRs and then we are happy to merge. We don't have resources to do this.

GeorgeVe:
Hee, I quikcly created a WIP backport to foxy for the joint_trajectory_controller with an effort-only command interface. It works for my team in simulation with Gazebo. It is on our forked GitLab repository: (https://gitlab.com/project-march/libraries/control/ros2_controllers/-/tree/mergify/bp/foxy/pr-225/joint_trajectory_controller).  Feel free to check it out, but **I can't guarantee that it is 100% correct and finished**. For example, the feature of 'allow_integration_in_goal_trajectories_' is also added but only partially implemented.

rohitmenon86:
Hi, I have created a backport for galactic on my fork here https://github.com/rohitmenon86/ros2_controllers.git  for the joint_trajectory_controller ONLY with an effort-only command interface. I am still testing it

bmagyar:
PRs are welcome

PR #231, opened 2021-08-30
Update CI to Galactic

bmagyar:


.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           required-ros-distributions: foxy
```
v-lopez, line 35: Not sure about what this does, but should it be galactic?

bmagyar:
```suggestion
          required-ros-distributions: galactic
```

PR #232, opened 2021-08-30
refactor get_current_state to get_state

VX792:
Based on this issue:
https://github.com/ros-controls/ros2_control/issues/505


bmagyar (approved):
This plus #234 should fix the CI on master, thanks!

PR #235, opened 2021-09-07
Removing explicit Boost disable for controllers

bailaC:
Further changes for Controllers. This will remove the restriction for Boost from Controllers.

ros2_control PR : https://github.com/ros-controls/ros2_control/pull/514


bailaC:
@destogl , @bmagyar , Please take a look.

bmagyar:
I'd ask to update the title. We are not enabling boost dependencies since we do not depend on them. What changed is that we do not need to explicitly disable boost with pluginlib anymore as that is the default. It will remain disabled.

bailaC:
Title updated.

destogl (approved):
Looks good thanks!

bmagyar (approved):

PR #239, opened 2021-09-13
[JointTrajectoryController] Add velocity-only command option for JTC with closed loop controller

livanov93:
Introduce velocity-only command interface for `JointTrajectoryController` and extend tests for its usage.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -150,6 +150,7 @@ protected:
   std::vector<PidPtr> pids_;
   std::chrono::steady_clock::time_point last_update_time_;
   std::vector<double> velocity_ff_;
+  std::vector<double> command_;
 
   // TODO(karsten1987): eventually activate and deactivate subscriber directly when its supported
   bool subscriber_is_active_ = false;
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -135,6 +135,7 @@ protected:
   InterfaceReferences<hardware_interface::LoanedCommandInterface> joint_command_interface_;
   InterfaceReferences<hardware_interface::LoanedStateInterface> joint_state_interface_;
 
+  bool has_position_state_interface_ = false;
   bool has_velocity_state_interface_ = false;
   bool has_acceleration_state_interface_ = false;
   bool has_position_command_interface_ = false;
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -148,9 +148,10 @@ protected:
   bool use_closed_loop_pid_adapter = false;
   using PidPtr = std::shared_ptr<control_toolbox::Pid>;
   std::vector<PidPtr> pids_;
-  std::chrono::steady_clock::time_point last_update_time_;
-  std::vector<double> velocity_ff_;
-  std::vector<double> command_;
+    /// Feed-forward velocity weight factor when calculating closed loop pid adapter's command
+  std::vector<double> ff_velocity_scale_;
+    /// reserved storage for result of the command when closed loop pid adapter is used
+  std::vector<double> tmp_command_;
 
   // TODO(karsten1987): eventually activate and deactivate subscriber directly when its supported
   bool subscriber_is_active_ = false;
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -24,6 +24,7 @@
 
 #include "control_msgs/action/follow_joint_trajectory.hpp"
 #include "control_msgs/msg/joint_trajectory_controller_state.hpp"
+#include "control_toolbox/pid.hpp"
 #include "controller_interface/controller_interface.hpp"
 #include "hardware_interface/types/hardware_interface_type_values.hpp"
 #include "joint_trajectory_controller/tolerances.hpp"
@@ -145,6 +146,10 @@ protected:
   // There should be PID-approach used as in ROS1:
   // https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h#L283
   bool use_closed_loop_pid_adapter = false;
+  using PidPtr = std::shared_ptr<control_toolbox::Pid>;
+  std::vector<PidPtr> pids_;
+  std::chrono::steady_clock::time_point last_update_time_;
+  std::vector<double> velocity_ff_;
 
   // TODO(karsten1987): eventually activate and deactivate subscriber directly when its supported
   bool subscriber_is_active_ = false;
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -148,9 +148,9 @@ protected:
   bool use_closed_loop_pid_adapter = false;
   using PidPtr = std::shared_ptr<control_toolbox::Pid>;
   std::vector<PidPtr> pids_;
-    /// Feed-forward velocity weight factor when calculating closed loop pid adapter's command
+  /// Feed-forward velocity weight factor when calculating closed loop pid adapter's command
   std::vector<double> ff_velocity_scale_;
-    /// reserved storage for result of the command when closed loop pid adapter is used
+  /// reserved storage for result of the command when closed loop pid adapter is used
   std::vector<double> tmp_command_;
 
   // TODO(karsten1987): eventually activate and deactivate subscriber directly when its supported
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -205,11 +205,11 @@ controller_interface::return_type JointTrajectoryController::update(
           // Update PIDs
           for (auto i = 0ul; i < joint_num; ++i)
           {
-            tmp_command_[i] =
-              (state_desired.velocities[i] * ff_velocity_scale_[i]) +
-              pids_[i]->computeCommand(
-                state_desired.positions[i] - state_current.positions[i],
-                state_desired.velocities[i] - state_current.velocities[i], period.seconds());
+            tmp_command_[i] = (state_desired.velocities[i] * ff_velocity_scale_[i]) +
+                              pids_[i]->computeCommand(
+                                state_desired.positions[i] - state_current.positions[i],
+                                state_desired.velocities[i] - state_current.velocities[i],
+                                (uint64_t)period.nanoseconds());
           }
 
           assign_interface_from_point(joint_command_interface_[1], tmp_command_);
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -558,7 +558,7 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
       // Initialize PID
       pids_[i] = std::make_shared<control_toolbox::Pid>(k_p, k_i, k_d, i_clamp, -i_clamp);
     }
-    //TODO(livanov93) add other option when implemented
+    // TODO(livanov93): add other option when implemented
   }
 
   // Read always state interfaces from the parameter because they can be used
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -211,7 +211,6 @@ controller_interface::return_type JointTrajectoryController::update(
                                 state_desired.velocities[i] - state_current.velocities[i],
                                 (uint64_t)period.nanoseconds());
           }
-
           assign_interface_from_point(joint_command_interface_[1], tmp_command_);
         }
       }
@@ -592,6 +591,10 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
     }
   }
 
+  if (contains_interface_type(state_interface_types_, hardware_interface::HW_IF_POSITION))
+  {
+    has_position_state_interface_ = true;
+  }
   if (contains_interface_type(state_interface_types_, hardware_interface::HW_IF_VELOCITY))
   {
     has_velocity_state_interface_ = true;
@@ -830,8 +833,17 @@ CallbackReturn JointTrajectoryController::on_deactivate(const rclcpp_lifecycle::
   // TODO(anyone): How to halt when using effort commands?
   for (size_t index = 0; index < joint_names_.size(); ++index)
   {
-    joint_command_interface_[0][index].get().set_value(
-      joint_command_interface_[0][index].get().get_value());
+    if (has_position_command_interface_)
+    {
+      joint_command_interface_[0][index].get().set_value(
+        joint_command_interface_[0][index].get().get_value());
+    }
+
+    if (has_velocity_command_interface_)
+    {
+      joint_command_interface_[1][index].get().set_value(
+        joint_command_interface_[1][index].get().get_value());
+    }
   }
 
   for (size_t index = 0; index < allowed_interface_types_.size(); ++index)
@@ -1027,7 +1039,19 @@ void JointTrajectoryController::fill_partial_goal(
       for (auto & it : trajectory_msg->points)
       {
         // Assume hold position with 0 velocity and acceleration for missing joints
-        it.positions.push_back(joint_command_interface_[0][index].get().get_value());
+        if (!it.positions.empty())
+        {
+          if (has_position_command_interface_)
+          {
+            // copy last command if cmd interface exists
+            it.positions.push_back(joint_command_interface_[0][index].get().get_value());
+          }
+          else if (has_position_state_interface_)
+          {
+            // copy current state if state interface exists
+            it.positions.push_back(joint_state_interface_[0][index].get().get_value());
+          }
+        }
         if (!it.velocities.empty())
         {
           it.velocities.push_back(0.0);
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -591,10 +591,7 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
     }
   }
 
-  if (contains_interface_type(state_interface_types_, hardware_interface::HW_IF_POSITION))
-  {
-    has_position_state_interface_ = true;
-  }
+  has_position_state_interface_ = contains_interface_type(state_interface_types_, hardware_interface::HW_IF_POSITION;
   if (contains_interface_type(state_interface_types_, hardware_interface::HW_IF_VELOCITY))
   {
     has_velocity_state_interface_ = true;
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -196,7 +196,27 @@ controller_interface::return_type JointTrajectoryController::update(
       }
       if (has_velocity_command_interface_)
       {
-        assign_interface_from_point(joint_command_interface_[1], state_desired.velocities);
+        if (!use_closed_loop_pid_adapter)
+        {
+          assign_interface_from_point(joint_command_interface_[1], state_desired.velocities);
+        }
+        else
+        {
+          std::vector<double> command(pids_.size());
+          const auto period = std::chrono::steady_clock::now() - last_update_time_;
+
+          // Update PIDs
+          for (auto i = 0ul; i < joint_num; ++i)
+          {
+            command[i] =
+              (state_desired.velocities[i] * velocity_ff_[i]) +
+              pids_[i]->computeCommand(
+                state_desired.positions[i] - state_current.positions[i],
+                state_desired.velocities[i] - state_current.velocities[i], period.count());
+          }
+
+          assign_interface_from_point(joint_command_interface_[1], command);
+        }
       }
       if (has_acceleration_command_interface_)
       {
@@ -501,10 +521,6 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
     if (command_interface_types_.size() == 1)
     {
       use_closed_loop_pid_adapter = true;
-      // TODO(anyone): remove this when implemented
-      RCLCPP_ERROR(logger, "using 'velocity' command interface alone is not yet implemented yet.");
-      return CallbackReturn::FAILURE;
-      // if velocity interface can be used without position if multiple defined
     }
     else if (!has_position_command_interface_)
     {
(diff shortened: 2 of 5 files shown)


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -195,7 +195,27 @@ controller_interface::return_type JointTrajectoryController::update()
...
       if (has_velocity_command_interface_)
       {
-        assign_interface_from_point(joint_command_interface_[1], state_desired.velocities);
+        if (!use_closed_loop_pid_adapter)
+        {
+          assign_interface_from_point(joint_command_interface_[1], state_desired.velocities);
+        }
+        else
+        {
+          std::vector<double> command(pids_.size());
```
destogl, line 204: Please initialize this variable in the configure method and not on each update

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -524,6 +540,26 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
     return CallbackReturn::FAILURE;
   }
 
+  if (has_velocity_command_interface_ && use_closed_loop_pid_adapter)
+  {
+    pids_.resize(joint_names_.size());
+    velocity_ff_.resize(joint_names_.size());
```
destogl, line 546: There should be `num_joints` define (or similar variable).

```suggestion
    pids_.resize(num_joints);
    velocity_ff_.resize(num_joints);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -524,6 +540,26 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
...
+      const std::string prefix = "gains." + joint_names_[i];
+      const auto k_p = auto_declare<double>(prefix + ".p", 0.0);
+      const auto k_i = auto_declare<double>(prefix + ".i", 0.0);
+      const auto k_d = auto_declare<double>(prefix + ".d", 0.0);
+      const auto i_clamp = auto_declare<double>(prefix + ".i_clamp", 0.0);
+      velocity_ff_[i] = auto_declare<double>("velocity_ff/" + joint_names_[i], 0.0);
+      // Initialize PID
+      pids_[i] = std::make_shared<control_toolbox::Pid>(k_p, k_i, k_d, i_clamp, -i_clamp);
+    }
+    // TODO(livanov93): add other option when implemented
```
destogl, line 560: what other option(s)?

destogl (commented):
This looks good. Can you add some tests?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
livanov93: @destogl  Was alluding on effort interface, [here](https://github.com/livanov93/ros2_controllers/blob/2d2970d01cdaee11387b92bee090a668c084f489/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L543)  we will have to leave only `use_closed_loop_pid_adapter` variable in the if condition. 
Wrong choice of word - change when implemented would be more appropriate.

livanov93: ros-controls/ros2_controllers@68d838d

joint_trajectory_controller/src/joint_trajectory_controller.cpp
livanov93: ros-controls/ros2_controllers@68d838d

joint_trajectory_controller/src/joint_trajectory_controller.cpp
livanov93: ros-controls/ros2_controllers@68d838d

livanov93:
> This looks good. Can you add some tests?

I can. I will follow the currently existing tests in the jtc package and add one to test velocity interface.

destogl:
> I can. I will follow the currently existing tests in the jtc package and add one to test velocity interface.

Maybe you are lucky and can add simply a new parameterized test. Check the end of the file (before the last test). Last test should also be changed because it will fail if VELOCITY interface is enabled.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -145,6 +146,11 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   // There should be PID-approach used as in ROS1:
   // https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h#L283
   bool use_closed_loop_pid_adapter = false;
+  using PidPtr = std::shared_ptr<control_toolbox::Pid>;
+  std::vector<PidPtr> pids_;
+  std::chrono::steady_clock::time_point last_update_time_;
+  std::vector<double> velocity_ff_;
```
destogl, line 152:
```suggestion
  /// Feed-forward velocity weight factor when calculating closed loop pid adapter's command
  std::vector<double> ff_velocity_scale_;
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -145,6 +146,11 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
...
+  std::vector<double> command_;
```
destogl, line 153: Add doc string comment and rename to avoid confusion in the future.

```suggestion
  /// reserved storage for result of the command when closed loop pid adapter is used
  std::vector<double> tmp_command_;
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -196,7 +196,26 @@ controller_interface::return_type JointTrajectoryController::update(
...
+          assign_interface_from_point(joint_command_interface_[1], state_desired.velocities);
+        }
+        else
+        {
+          const auto period = std::chrono::steady_clock::now() - last_update_time_;
+
+          // Update PIDs
+          for (auto i = 0ul; i < joint_num; ++i)
+          {
+            command_[i] =
```
destogl, line 210:
```suggestion
            tmp_command_[i] =
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -196,7 +196,26 @@ controller_interface::return_type JointTrajectoryController::update(
...
+              (state_desired.velocities[i] * velocity_ff_[i]) +
+              pids_[i]->computeCommand(
+                state_desired.positions[i] - state_current.positions[i],
+                state_desired.velocities[i] - state_current.velocities[i], period.count());
+          }
+
+          assign_interface_from_point(joint_command_interface_[1], command_);
```
destogl, line 217:
```suggestion
          assign_interface_from_point(joint_command_interface_[1], tmp_command_);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -196,7 +196,26 @@ controller_interface::return_type JointTrajectoryController::update(
...
+        }
+        else
+        {
+          const auto period = std::chrono::steady_clock::now() - last_update_time_;
+
+          // Update PIDs
+          for (auto i = 0ul; i < joint_num; ++i)
+          {
+            command_[i] =
+              (state_desired.velocities[i] * velocity_ff_[i]) +
```
destogl, line 211:
```suggestion
              (state_desired.velocities[i] * ff_velocity_scale_[i]) +
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -525,6 +540,28 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
     return CallbackReturn::FAILURE;
   }
 
+  // TODO(livanov93): change when other option is implemented
+  if (has_velocity_command_interface_ && use_closed_loop_pid_adapter)
+  {
+    size_t num_joints = joint_names_.size();
+    pids_.resize(num_joints);
+    velocity_ff_.resize(num_joints);
+    command_.resize(num_joints, 0.0);
```
destogl, line 549:
```suggestion
    ff_velocity_scale_.resize(num_joints);
    tmp_command_.resize(num_joints, 0.0);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -525,6 +540,28 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
...
+
+    // Init PID gains from ROS parameter server
+    for (size_t i = 0; i < pids_.size(); ++i)
+    {
+      const std::string prefix = "gains." + joint_names_[i];
+      const auto k_p = auto_declare<double>(prefix + ".p", 0.0);
+      const auto k_i = auto_declare<double>(prefix + ".i", 0.0);
+      const auto k_d = auto_declare<double>(prefix + ".d", 0.0);
+      const auto i_clamp = auto_declare<double>(prefix + ".i_clamp", 0.0);
+      velocity_ff_[i] = auto_declare<double>("velocity_ff/" + joint_names_[i], 0.0);
```
destogl, line 559: Propose also to rename the parameter

```suggestion
      ff_velocity_scale_[i] = auto_declare<double>("ff_velocity_scale/" + joint_names_[i], 0.0);
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1104,6 +1104,17 @@ INSTANTIATE_TEST_CASE_P(
       std::vector<std::string>({"position", "velocity", "acceleration"}),
       std::vector<std::string>({"position", "velocity", "acceleration"}))));
 
+// only velocity controller
```
destogl, line 1107: It would also be cool to add a dedicated test for parameters. It can be parameterized and then depending on interfaces checked. See `test_joint_trajectory_controller_utils.hpp` file (or similar name).

destogl (changes requested):
It looks good, just some minor comment to polish things

joint_trajectory_controller/src/joint_trajectory_controller.cpp
livanov93: ros-controls/ros2_controllers@09e6954

joint_trajectory_controller/src/joint_trajectory_controller.cpp
livanov93: ros-controls/ros2_controllers@09e6954

joint_trajectory_controller/src/joint_trajectory_controller.cpp
livanov93: ros-controls/ros2_controllers@09e6954

joint_trajectory_controller/src/joint_trajectory_controller.cpp
livanov93: ros-controls/ros2_controllers@09e6954

joint_trajectory_controller/src/joint_trajectory_controller.cpp
livanov93: ros-controls/ros2_controllers@09e6954

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
livanov93: ros-controls/ros2_controllers@09e6954

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
livanov93: ros-controls/ros2_controllers@09e6954
ros-controls/ros2_controllers@deb098e

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -145,6 +146,11 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   // There should be PID-approach used as in ROS1:
   // https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h#L283
   bool use_closed_loop_pid_adapter = false;
+  using PidPtr = std::shared_ptr<control_toolbox::Pid>;
+  std::vector<PidPtr> pids_;
+  std::chrono::steady_clock::time_point last_update_time_;
```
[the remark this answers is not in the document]
livanov93: ros-controls/ros2_controllers@09e6954
ros-controls/ros2_controllers@deb098e

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -196,7 +196,26 @@ controller_interface::return_type JointTrajectoryController::update(
...
       {
-        assign_interface_from_point(joint_command_interface_[1], state_desired.velocities);
+        if (!use_closed_loop_pid_adapter)
+        {
+          assign_interface_from_point(joint_command_interface_[1], state_desired.velocities);
+        }
+        else
+        {
+          const auto period = std::chrono::steady_clock::now() - last_update_time_;
+
```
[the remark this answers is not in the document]
livanov93: ros-controls/ros2_controllers@09e6954
ros-controls/ros2_controllers@deb098e

joint_trajectory_controller/test/test_trajectory_controller.cpp
livanov93: What do we want to test here? Parameter propagation from overrides (e.g. yaml files) or we want to add parametrized test that will receive joint_names, interfaces, etc (this one already exists, only the joint names are hardcoded).

destogl: We want to run all existing test on this. Not only propagation, but calculations and setup in general.

livanov93:
@destogl what is the decision with this?

destogl (approved):
This looks fine! thanks!

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -496,18 +530,36 @@ TEST_P(TrajectoryControllerTestParameterized, test_partial_joint_list)
...
 
   if (
     std::find(command_interface_types_.begin(), command_interface_types_.end(), "velocity") !=
     command_interface_types_.end())
   {
-    // TODO(anyone): need help here - we should at least estimate the sign of the velocity
-    //     EXPECT_NEAR(traj_msg.points[0].velocities[1], joint_vel_[0], threshold);
-    //     EXPECT_NEAR(traj_msg.points[0].velocities[0], joint_vel_[1], threshold);
+    // estimate the sign of the velocity
+    if (0 > copysign(2.0, traj_msg.points[0].positions[0] - initial_joint2_cmd))
```
bmagyar, line 547:
```suggestion
    // joint rotates forward
    if (traj_msg.points[0].positions[0] - initial_joint2_cmd < 0.0)
```

and please apply a similar change to the other 3 ifs/else ifs below

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
```diff
@@ -94,6 +94,10 @@ class TestableJointTrajectoryController
     return last_commanded_state_;
   }
 
+  bool get_has_position_command_interface() { return has_position_command_interface_; }
```
bmagyar, line 97:
```suggestion
  bool has_position_command_interface() { return has_position_command_interface_; }
```
let's improve readability here

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
```diff
@@ -94,6 +94,10 @@ class TestableJointTrajectoryController
...
+
+  bool get_has_velocity_command_interface() { return has_velocity_command_interface_; }
```
bmagyar, line 99:
```suggestion
  bool has_velocity_command_interface() { return has_velocity_command_interface_; }
```
and here

bmagyar (changes requested):

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -267,6 +267,7 @@ TEST_P(TrajectoryControllerTestParameterized, correct_initialization_using_param
 
   // This call is replacing the way parameters are set via launch
   SetParameters();
+  SetPidParameters();
```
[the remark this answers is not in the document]
bmagyar: probably best to keep it here as `   // This call is replacing the way parameters are set via launch ` suggests parameters may change and `SetPidParameters()` reads them and sets 'em so it makes sense on first read

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -525,6 +537,28 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
...
+    for (size_t i = 0; i < pids_.size(); ++i)
+    {
+      const std::string prefix = "gains." + joint_names_[i];
+      const auto k_p = auto_declare<double>(prefix + ".p", 0.0);
+      const auto k_i = auto_declare<double>(prefix + ".i", 0.0);
+      const auto k_d = auto_declare<double>(prefix + ".d", 0.0);
+      const auto i_clamp = auto_declare<double>(prefix + ".i_clamp", 0.0);
+      ff_velocity_scale_[i] = auto_declare<double>("ff_velocity_scale/" + joint_names_[i], 0.0);
+      // Initialize PID
+      pids_[i] = std::make_shared<control_toolbox::Pid>(k_p, k_i, k_d, i_clamp, -i_clamp);
```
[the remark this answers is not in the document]
bmagyar: actually... this is fine

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -525,6 +537,28 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
...
+    tmp_command_.resize(num_joints, 0.0);
+
+    // Init PID gains from ROS parameter server
+    for (size_t i = 0; i < pids_.size(); ++i)
+    {
+      const std::string prefix = "gains." + joint_names_[i];
+      const auto k_p = auto_declare<double>(prefix + ".p", 0.0);
+      const auto k_i = auto_declare<double>(prefix + ".i", 0.0);
+      const auto k_d = auto_declare<double>(prefix + ".d", 0.0);
+      const auto i_clamp = auto_declare<double>(prefix + ".i_clamp", 0.0);
```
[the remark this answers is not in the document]
bmagyar: `auto_declare` will read if it's defined and declare & set & return default when not defined.

I'd really prefer not to add another round of reading parameters, that's a sign that either ROS2 params are broken or we are doing something wrong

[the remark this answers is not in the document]
destogl: What do you mean with another round? I find this approach OK.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -795,8 +833,17 @@ CallbackReturn JointTrajectoryController::on_deactivate(const rclcpp_lifecycle::
...
+    if (has_position_command_interface_)
+    {
+      joint_command_interface_[0][index].get().set_value(
+        joint_command_interface_[0][index].get().get_value());
+    }
+
+    if (has_velocity_command_interface_)
+    {
+      joint_command_interface_[1][index].get().set_value(
+        joint_command_interface_[1][index].get().get_value());
```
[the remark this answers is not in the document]
destogl: This is a good catch, although it could be tricky if we want to do "on-the fly controller switch", i.e., during movement.

joint_trajectory_controller/test/test_trajectory_controller.cpp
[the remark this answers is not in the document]
destogl: I check this now deeper and it seems good as it is. This is used in another metho in .hpp: `SetUpAndActivateTrajectoryController`

destogl (approved):
I still approve this after @bmagyar's changes. We should only clarify if we should reset initial velocity to 0.0 or to current velocity of the hardware.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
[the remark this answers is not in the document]
destogl: Wouldn't be better to write: (additional check is probably needed). If no state interface is existing, then we should reset to 0 (probably).
```suggestion
        joint_state_interface_[1][index].get().get_value());
```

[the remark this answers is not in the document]
bmagyar: well... let's go with 0 following the slack discussion and discuss `on-the-fly` controller switching which has not yet been in any scope ;)

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -795,8 +821,17 @@ CallbackReturn JointTrajectoryController::on_deactivate(const rclcpp_lifecycle::
...
```
bmagyar, line 833:
```suggestion
      joint_command_interface_[1][index].get().set_value(0.0);
```

bmagyar (approved):

destogl (approved):
@bmagyar you can go with merge

joint_trajectory_controller/src/joint_trajectory_controller.cpp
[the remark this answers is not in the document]
livanov93: `auto_declare` does exactly what @bmagyar said, so I don't think another parameter reading is needed here.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -456,9 +476,19 @@ TEST_P(TrajectoryControllerTestParameterized, test_jumbled_joint_order)
...
+  if (traj_controller_->get_has_position_command_interface())
+  {
+    EXPECT_NEAR(1.0, joint_pos_[0], COMMON_THRESHOLD);
+    EXPECT_NEAR(2.0, joint_pos_[1], COMMON_THRESHOLD);
+    EXPECT_NEAR(3.0, joint_pos_[2], COMMON_THRESHOLD);
+  }
+
+  if (traj_controller_->get_has_velocity_command_interface())
+  {
+    EXPECT_GT(0.0, joint_vel_[0]);
```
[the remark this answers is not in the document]
livanov93: @destogl The idea was to set jumbled order of joint names and to set velocity with a wrong sign. The test is initialized, and the commands and states are set to zero. Here I want joints to go from {0.0, 0.0, 0.0} to {1.0, 2.0, 3.0} with velocity {-0.1, -0.1, -0.1} and if the velocity command interface is active, I know that PID should revert these velocity values into positive ones.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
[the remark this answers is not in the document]
livanov93: @bmagyar sure? what else could be done here?  Maybe create PIDs in the constructor, read the parameters in `on_init` and then reset the already existing PIDs here in `on_configure`?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -557,6 +591,10 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
     }
   }
 
+  if (contains_interface_type(state_interface_types_, hardware_interface::HW_IF_POSITION))
+  {
+    has_position_state_interface_ = true;
+  }
```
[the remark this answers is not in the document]
livanov93: You did it already :)

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -304,10 +309,19 @@ TEST_P(TrajectoryControllerTestParameterized, correct_initialization_using_param
...
+  }
 
-  EXPECT_NEAR(3.3, joint_pos_[0], allowed_delta);
-  EXPECT_NEAR(4.4, joint_pos_[1], allowed_delta);
-  EXPECT_NEAR(5.5, joint_pos_[2], allowed_delta);
+  if (traj_controller_->has_velocity_command_interface())
+  {
+    EXPECT_LT(0.0, joint_vel_[0]);
+    EXPECT_LT(0.0, joint_vel_[1]);
+    EXPECT_LT(0.0, joint_vel_[2]);
```
bmagyar, line 323:
```suggestion
    EXPECT_LE(0.0, joint_vel_[0]);
    EXPECT_LE(0.0, joint_vel_[1]);
    EXPECT_LE(0.0, joint_vel_[2]);
```

PR #241, opened 2021-09-18
Add time and period to update function

bmagyar:
Following up https://github.com/ros-controls/ros2_control/pull/520

PR #244, opened 2021-09-21
Use lifecycle node as base for controllers

vatanaksoytezer:
~~Adds the lifecycle nodes back, that are removed in #124. Depends on ros-controls/ros2_control#538 and #241, since ros-controls/ros2_control#520 is merged but #241 hasn't.~~

Depends on
- ros-controls/ros2_control#538


vatanaksoytezer:
@destogl @bmagyar this should be pretty close to complete though, controller manager tests are failing due to controller state not being correctly changed (for eg. starts in inactive instead of unconfigured). Appreciate if you guys have any ideas why this might be the case.

vatanaksoytezer:
So again most of this should be done, but auto declare doesn't seem to work as intended wit lifecycle_nodes as it is with rclcpp::Node.

bmagyar (approved):

bmagyar:
depends on https://github.com/ros-controls/ros2_control/pull/538 but otherwise LGTM

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -92,7 +92,7 @@ struct SegmentTolerances
  * \return Trajectory segment tolerances.
  */
 SegmentTolerances get_segment_tolerances(
-  const rclcpp::Node & node, const std::vector<std::string> & joint_names)
+  const rclcpp_lifecycle::LifecycleNode & node, const std::vector<std::string> & joint_names)
```
destogl, line 95: We should probably here use SharedPtr to the node and not the object itself. This is usually done in the rest of the code-base.

destogl (approved):

Briancbn (approved):

diff_drive_controller/test/test_diff_drive_controller.cpp
```diff
@@ -389,14 +389,14 @@ TEST_F(TestDiffDriveController, correct_initialization_using_parameters)
   rclcpp::executors::SingleThreadedExecutor executor;
   executor.add_node(controller_->get_node()->get_node_base_interface());
 
-  auto state = controller_->configure();
-  assignResourcesPosFeedback();
+  auto state = controller_->get_node()->configure();
+  assignResources();
```
destogl, line 393:
```suggestion
  assignResourcesPosFeedback();
```

harderthan (approved):
Looks good to me

bijoua29 (approved):

gripper_controllers/include/gripper_controllers/hardware_interface_adapter.hpp
```diff
@@ -46,7 +46,7 @@ class HardwareInterfaceAdapter
   bool init(
     std::experimental::optional<
       std::reference_wrapper<hardware_interface::LoanedCommandInterface>> /* joint_handle */,
-    const rclcpp::Node::SharedPtr & /* node */)
+    std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & /* node */)
```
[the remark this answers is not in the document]
destogl: Would it be better to pass a reference instead of shard_ptr here too?

bmagyar (approved):

bmagyar (approved):

aprotyas (approved):

bmagyar:
Several CI failures are caused by the lack of release of ros2_control which'd cause similar build issues on the buildfarm. After merging this I'll push a release of both to restore the CI results

Issue #229, opened 2021-08-26
diff_drive_controller shall accept velocity feedback from motors


roncapat:
At a certain point in ROS1 implementation, `estimate_velocity_from_position` parameter switch was inserted, and odometry could also be computed by integration of velocity values coming directly from motors.

In ROS2, this in not an option. Although somewhere it is written that the controller works "through a velocity interface" this is true only for the input chain, not the feedback one.

I propose a port of the feature also in ROS2 , and at least specify clearly what are the required Hardware Interfaces (in/out) for this controller in the documentation.

bmagyar:
Yes! I'm quite sad that we never got to port more of the diff drive functionality over to ROS2 as we are still fairly busy with the framework itself. I'd be very happy to review some PRs for this though!

roncapat:
I searched again the sources of the implementation I mentioned in the opening of this issue. From mobile, didn't found yet the code I swear to have seen at some time in some branch of ros_controllers official repo, but I remember it was something very similar to what can be seen here:
https://github.com/clearpathrobotics/ros_controllers/pull/10

Starting from this code, I think that in the next week I can open a draft PR, or maybe even already test it in my company on our retrofitted Seekur Jr Rover.

roncapat:
Closing since #260 got merged.

Issue #48, opened 2020-06-01
Port over tests for diff_drive_controller from ROS1


bmagyar:
So many and so lovely: https://github.com/ros-controls/ros_controllers/tree/melodic-devel/diff_drive_controller/test

christophfroehlich:
stale PR: #108

PR #260, opened 2021-10-22
Add velocity feedback option for diff_drive_controller

roncapat:
Implements #229.
Builds on a fresh Ubuntu 20.04 ROS2 Dashing install alongside the master branch of ros2-controls.


roncapat:
I hope today to already fix the linter errors :)

roncapat:
Only sync to master for now, no need to re-run workflows till I fix the other errors :)

roncapat:
I enabled workflows on my fork. Actually, the only one failing is [this one](https://github.com/roncapat/ros2_controllers/runs/4090576281). Can someone help me with the log?

roncapat:
The issue on my fork workflows is related with `test_force_torque_sensor_broadcaster`. All other tests pass. Is there a fix?

bmagyar:
Sorry for the delay on this: can you please add a test for this?

roncapat:
I may try... Never done. What behaviour I should test?

roncapat:
In the meantime I try to learn about writing test, I paste here the log of the failing job:
```
   2: [ RUN      ] ForceTorqueSensorBroadcasterTest.SensorName_Publish_Success
  2: 1636098896.664650 [0] test_force: using network interface eth0 (udp/10.1.1.107) selected arbitrarily from: eth0, docker0
  2: /home/runner/work/ros2_controllers/ros2_controllers/ros_ws/src/cav7w0ygslo/ros2_controllers/force_torque_sensor_broadcaster/test/test_force_torque_sensor_broadcaster.cpp:96: Failure
  2: Expected equality of these values:
  2:   wait_for(subscription)
  2:     Which is: 1
  2:   rclcpp::WaitResultKind::Ready
  2:     Which is: 0
  2: /home/runner/work/ros2_controllers/ros2_controllers/ros_ws/src/cav7w0ygslo/ros2_controllers/force_torque_sensor_broadcaster/test/test_force_torque_sensor_broadcaster.cpp:260: Failure
  2: Expected equality of these values:
  2:   wrench_msg.header.frame_id
  2:     Which is: ""
  2:   frame_id_
  2:     Which is: "fts_sensor_frame"
  2: [  FAILED  ] ForceTorqueSensorBroadcasterTest.SensorName_Publish_Success (20168 ms)
  2: [ RUN      ] ForceTorqueSensorBroadcasterTest.InterfaceNames_Publish_Success
  2: 1636098916.833366 [0] test_force: using network interface eth0 (udp/10.1.1.107) selected arbitrarily from: eth0, docker0
  2: /home/runner/work/ros2_controllers/ros2_controllers/ros_ws/src/cav7w0ygslo/ros2_controllers/force_torque_sensor_broadcaster/test/test_force_torque_sensor_broadcaster.cpp:96: Failure
  2: Expected equality of these values:
  2:   wait_for(subscription)
  2:     Which is: 1
  2:   rclcpp::WaitResultKind::Ready
  2:     Which is: 0
  2: /home/runner/work/ros2_controllers/ros2_controllers/ros_ws/src/cav7w0ygslo/ros2_controllers/force_torque_sensor_broadcaster/test/test_force_torque_sensor_broadcaster.cpp:284: Failure
  2: Expected equality of these values:
  2:   wrench_msg.header.frame_id
  2:     Which is: ""
  2:   frame_id_
  2:     Which is: "fts_sensor_frame"
  2: [  FAILED  ] ForceTorqueSensorBroadcasterTest.InterfaceNames_Publish_Success (20211 ms)
```

From the source code, seems that the subscription times out. FYI

roncapat:
@bmagyar I ran locally the existing tests, but noticed that the test YAML file is not taken into account! (I vary values, recompile, run test and print some parameters and nothing changes). 

In CMake:
```
ament_add_gmock(test_diff_drive_controller
    test/test_diff_drive_controller.cpp 
    ENV config_file=${CMAKE_CURRENT_SOURCE_DIR}/test/config/test_diff_drive_controller.yaml)
```

but parameters passing in this way is documented nowhere and seems to me not to work correctly. Are we sure that up to now it has always worked well?

roncapat:
I learned to write basic tests. In particular, now there are test that validate the effectiveness of the new parameter I introduced on the loading of other position or velocity feedback interfaces.

diff_drive_controller/src/odometry.cpp
```diff
@@ -86,6 +86,30 @@ bool Odometry::update(double left_pos, double right_pos, const rclcpp::Time & ti
   return true;
 }
 
+bool Odometry::updateFromVelocity(double left_vel, double right_vel, const rclcpp::Time & time)
```
bmagyar, line 89: line #70-86 are basically the same as this function. I think that for now we could call `updateFromVelocity` in `update()` until we discover we need some extra special implementation in either position or velocity feedback mode.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -586,10 +624,12 @@ CallbackReturn DiffDriveController::configure_side(
   registered_handles.reserve(wheel_names.size());
   for (const auto & wheel_name : wheel_names)
   {
+    auto interface_name = feedback_type();
```
bmagyar, line 627: can this also be `const`?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -183,28 +189,59 @@ controller_interface::return_type DiffDriveController::update(
...
-
-      if (std::isnan(left_position) || std::isnan(right_position))
+      double left_position_mean = 0.0;
+      double right_position_mean = 0.0;
+      for (size_t index = 0; index < wheels.wheels_per_side; ++index)
       {
-        RCLCPP_ERROR(
-          logger, "Either the left or right wheel position is invalid for index [%zu]", index);
-        return controller_interface::return_type::ERROR;
+        const double left_position =
```
bmagyar, line 198: The code in these two branches of the `if` are essentially the same. Could you please try to do code validation, etc before the `if`?

You could even generalize the print by using the `feedback_type()` function or something alike

bmagyar:
Thanks for the work on this @roncapat  and indeed probably that way of configuring tests may not work anymore as we moved from one distro to another. I'll create a separate issue to look into that.

roncapat:
I addressed all comments. Now I run the format workflow on my branch and push format fixes if needed.

roncapat:
@bmagyar done :)

diff_drive_controller/src/odometry.cpp
bmagyar: thanks!

bmagyar (approved):
Thanks very much for this!

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
```diff
@@ -88,10 +88,11 @@ class DiffDriveController : public controller_interface::ControllerInterface
 protected:
   struct WheelHandle
   {
-    std::reference_wrapper<const hardware_interface::LoanedStateInterface> position;
+    std::reference_wrapper<const hardware_interface::LoanedStateInterface> feedback;
     std::reference_wrapper<hardware_interface::LoanedCommandInterface> velocity;
```
destogl, line 92: We should probably also rename this in the follow-up PR to make the variable name clearer.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -183,28 +189,36 @@ controller_interface::return_type DiffDriveController::update(
...
+    right_feedback_mean /= wheels.wheels_per_side;
 
-    odometry_.update(left_position_mean, right_position_mean, current_time);
+    if (odom_params_.position_feedback)
+    {
+      odometry_.update(left_feedback_mean, right_feedback_mean, current_time);
+    }
+    else
+    {
+      odometry_.updateFromVelocity(left_feedback_mean, right_feedback_mean, current_time);
```
destogl, line 220: Why not use `period` instead of calculating it manually?

```suggestion
      odometry_.updateFromVelocity(left_feedback_mean, right_feedback_mean, period);
```

diff_drive_controller/test/config/test_diff_drive_controller.yaml
```diff
@@ -17,6 +17,7 @@ diff_drive_controller:
     pose_covariance_diagonal: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
     twist_covariance_diagonal: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
 
+    position_feedback: false
```
destogl, line 20: is there other option missing: `position_feedback: true`

destogl (commented):
@roncapat, @bmagyar  please check/address my comments

diff_drive_controller/test/config/test_diff_drive_controller.yaml
roncapat: No, in my previous comments I also observed that this file is not being loaded correctly by the tests. So I just added the line as an example of configuration and reference, but it is not being used by tests.

diff_drive_controller/src/diff_drive_controller.cpp
roncapat: You mean calculating 'const double dt = time.seconds() - timestamp_.seconds();' outside updateFromVelocity() and update() instead of inside? This will remove redundancy but still allow `(dt < 0.0001)` check for update() call

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
roncapat: You mean choosing a different name than "feedback"? Suggestions?

bmagyar: I think it's ok for now

roncapat:
@destogl any feedbacks on the review answers? I may have some time this week to fix/implement/polish so that we can eventually merge.

diff_drive_controller/test/config/test_diff_drive_controller.yaml
bmagyar: Long story short, we need to address this ticket: https://github.com/ros-controls/ros2_controllers/issues/48

roncapat: @bmagyar is your reply to me or @destogl? To me it seems #48 is not a blocking PR for this one.

bmagyar: I was not meaning to block this PR with that comment, just cross referencing

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
destogl: I am actually proposing to rename the "velocity" to "command" or something like this.

diff_drive_controller/src/diff_drive_controller.cpp
destogl: we should have `period` available in the `update` method

destogl (approved):
OK, let's move with it and address my other comments in the follow-up PRs.

diff_drive_controller/src/diff_drive_controller.cpp
wmmc88, line 220: Shouldnt this be `left_feedback_mean * dt` and `right_feedback_mean * dt` since `left_vel` and `right_vel` used in `integrateExact` are supposed to be the position deltas and not the instantaneous velocities? In the `position_feedback == true` path, `left_vel` and `right_vel` [are passed as position deltas](https://github.com/ros-controls/ros2_controllers/blob/dedb120993c6a95e240c2cf7d7730767ab701296/diff_drive_controller/src/odometry.cpp#L62). The fork that the feedback_param was based off of also [multiplies by dt](https://github.com/clearpathrobotics/ros_controllers/blob/22e9c2cf64b029abd88eafedae6b988a3845585f/diff_drive_controller/src/odometry.cpp#L128)

roncapat: Sorry, I'm from mobile now, but your line of thought seems reasonable at a first look. In these days I'm on vacation so I can't evaluate and fix, maybe it is worth pushing a new PR with the fixes you have in mind explaining there what's wrong with current code.

wmmc88: It also doesn't really make sense to me why both pose and twist get computed from vel when this param is enabled. Doesn't it make more sense to compute odom pose using position deltas calculated from the position interface and then use instantaneous vel from the velocity interface to calculate the odom twist? This is the approach that the clearpath fork ended up using, allowing for two params to disable this behavior (originally implemented in https://github.com/clearpathrobotics/ros_controllers/pull/18, but available in their latest indigo branch). Thoughts on this?

roncapat: @wmmc88 do you opened any issue or PR related to this?

For your last message, I undestand that you are talking about using simultaneously the feedback position and velocity interfaces. Is is correct? In my undestanding, for now there is only support here for one of the two IF/s at time, but using both if available could be interesting (not to be seen as a bug, but as a feature enhancement).

wmmc88: > @wmmc88 do you opened any issue or PR related to this?

nope. havent opened an issue and haven't had any time to address it myself in a pr.

> For your last message, I undestand that you are talking about using simultaneously the feedback position and velocity interfaces. Is is correct? In my undestanding, for now there is only support here for one of the two IF/s at time, but using both if available could be interesting (not to be seen as a bug, but as a feature enhancement).

yeah this wasn't meant to be an issue with this pr. I'm more just suggesting as an enhancement that the pose be calculated by the position interface data and the twist be calculated from the velocity interface data. I believe that is a more optimal solution and that is what the linked clearpath fork did(they reverted a change similar to this pr and went with the pose from position/twist from velocity approach).

roncapat: I reviewed now, and you are true. Since the code calls "vel" the position delta (misleadingly), I made confusion. A multiplication with `dt` will convert back instantaneous velocity to deltas. Might now also consider to do the multiplication inside updateFromVelocity() to hide this detail at interface level. Will open PR now.

PR #267, opened 2021-11-12
Starting to debug diff drive controller in sim.

destogl:
ros-simulation/gazebo_ros2_control#97 - we have to sync with this

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -133,9 +133,10 @@ InterfaceConfiguration DiffDriveController::state_interface_configuration() cons
 }
 
 controller_interface::return_type DiffDriveController::update(
-  const rclcpp::Time & time, const rclcpp::Duration & /*period*/)
+  const rclcpp::Time & current_time, const rclcpp::Duration & period)
 {
   auto logger = node_->get_logger();
+  // TODO(destogl): remove this checks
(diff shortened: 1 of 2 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -157,13 +156,20 @@ controller_interface::return_type DiffDriveController::update(
...
   }
 
-  const auto age_of_last_command = current_time - last_command_msg->header.stamp;
-  // Brake if cmd_vel has timeout, override the stored command
-  if (age_of_last_command > cmd_vel_timeout_)
-  {
-    last_command_msg->twist.linear.x = 0.0;
-    last_command_msg->twist.angular.z = 0.0;
-  }
+//   if (last_command_msg->header.stamp.sec != 0 and last_command_msg->header.stamp.nanosec != 0)
```
bmagyar, line 159: I think we discussed this enough but basically there should be a common policy on this and we should offer only one solution. An optional addition can be to opt-out of this feature altogether, in that case one could publish commands with timestamps from different time sources, we wouldn't know.

destogl:
Obsolete.

Raamkishore:
The changes are not reflected in the main repository and its branches.

Raamkishore:
I guess the changes are not updated in all the branches. I cloned the galactic branch and it took me a good amount of time to realise the problem with just the diff drive controller.

bmagyar:
Hence why this is still an open PR

PR #268, opened 2021-11-16
[FTS Broadcaster] Add support for publishing forces in other frames on additional topics

destogl:
This has to be cleaned a bit to split “additional topics” and filtering functionalities into separate PRs.

I would be glad to get some help with tests.

Documentation on new parameter is also missing.

### DIFF
--- a/force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
+++ b/force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
@@ -18,8 +18,12 @@
 
 #include "force_torque_sensor_broadcaster/force_torque_sensor_broadcaster.hpp"
 
+#include <limits>
 #include <memory>
 #include <string>
+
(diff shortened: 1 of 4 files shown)


force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
```diff
@@ -18,9 +18,13 @@
 
 #include "force_torque_sensor_broadcaster/force_torque_sensor_broadcaster.hpp"
 
+#include <limits>
 #include <memory>
 #include <string>
```
bmagyar, line 23:
```suggestion
#include <string>
#include <vector>
```

bmagyar (changes requested):
Looks good to me but please please add 2 new tests:
* checking the wrench
* checking additional frames

bmagyar:
@destogl are you ready to move this ahead?

christophfroehlich:
included in #559

PR #269, opened 2021-11-17
Add support for filtering of wrenches in the broadcaster.

destogl:
- use filters as plugins for filtering data
- add publisher for filtered data

depends on #268﻿ (containes also changes from it)


christophfroehlich:
included in #559

PR #281, opened 2022-01-05
Fix angular velocity direction of diff_drive_controller odometry

### LINKED ISSUE #280: Diff_drive_controller's odometry rotates in wrong direction
The angular velocity direction of the `diff_drive_controller`'s odometry is the wrong way around. This causes the odometry topic and the transform to rotate in the inverse direction.

This bug is introduced by #260.

PaulVerhoeckx:
Fixes #280


bmagyar (approved):
Ouch, very nice catch, thanks!
I wonder why no unit test has picked this up...

bmagyar:
@mergifyio backport to galactic

Timple:
Would it make sense to do a new bloom-release?
Before the next galactic sync picks up this bug, but not the fix?

PR #297, opened 2022-02-12
Remove version from clang-format

### LINKED ISSUE #295: Clang Format in pre-commit file could be broken on newer systems
## Context

Hey folks, I'm having problems doing a simple commit because of our pre-commit configuration, there is one step in which we have defined `clang-format-10`, and by default at least on my machine, brew is pulling version 13 of that program.

So, my question is, do you think it is worth just removing that kind of restriction from our pre-commit file? or are we using some specific settings from that version?
Btw, I already tried removing that specification from pre-commit, and now I can run the process without problems.

I can do this in a separate PR if you want, or if you think this would be good as a "first issue", feel free to edit the Issue and keep it open for someone else.

## Environment

- OS: macOS
- Version: 11.6.3
- Architecture: arm64

erickisos:
Closes #295

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,6 @@
 ---
-Language:        Cpp
-BasedOnStyle:  Google
+Language: Cpp
+BasedOnStyle: Google
 
 ColumnLimit: 100
 AccessModifierOffset: -2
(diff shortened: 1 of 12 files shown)


destogl (approved):
Seems OK to me and CI likes it.

erickisos:
It seems like at least, I can't merge it because the errors on some checks

bmagyar (approved):

erickisos:
Thank you @bmagyar and @destogl !

PR #283, opened 2022-01-05
Add reset() for the controllers

bailaC:
This PR is the follow up of the https://github.com/ros-controls/ros2_controllers/issues/247. Now the usage of reset() function of RealtimeBuffer is added in controllers (wherever possible).

PR #142, opened 2021-01-19
Update linter workflow.

Xi-HHHM:
Hello, this is a PR according to the issue #141.

PR #97, opened 2020-09-21
fix brace initializers on clang

Karsten1987:
Signed-off-by: Karsten Knese <Karsten1987@users.noreply.github.com>

PR #301, opened 2022-02-23
scaled jtc

fmauch:
- Initial copy of the ROS1 driver
- Change pkg versions to 0.0.0
- Update CMakeLists and package.xml for:
- Clean out ur_controllers, it needs a complete rewrite
- Update package.xml files so `ros2 pkg list` shows all pkgs
- Silence `ament_lint_cmake` errors
- Add XML schema to all `package.xml` files
- Add force torque controller.
- Claim individual resources.
- Add force torque controller.
- Claim individual resources.
- Remove lifecycle node (update with ros2_control changes).
- Update ft state controller with ros2_control changes.
- Prepare for testing.
- Update force torque state controller.
- Fix clang tidy in multiple pkgs.
- Update licence.
- Initial version of the speed_scaling_state_controller
- Added publisher for speed scaling factor
- Code formating and cleanup
- Added publisher rate
- Removed last remnants of joint_state_controller
- Removed controller from config file to realign with current branch status
- Only load speed scaling interface
- Code cleanup
- Scaled Joint Trajectory Controller (#43)
- Use GitHub Actions, use pre-commit formatting (#56)
- Fix warning about deprecated controller_interface::return_type::SUCCESS (#68)
- Add support for gpios, update MoveIt and ros2_control launching (#66)
- Review CI by correcting the configurations (#71)
- Last fix-ups
- Some intermediate commit
- Last fix-ups
- Fix move to home bug (#92)
- Reintegrating missing ur_client_library dependency since the break the building process (#97)
- Update for changes to ros2_control and ros2_controllers
- Correct check for fixed has_trajectory_msg()
- Update ur_controllers/src/scaled_joint_trajectory_controller.cpp
- Update ur_controllers/src/scaled_joint_trajectory_controller.cpp
- Update ur_controllers/src/scaled_joint_trajectory_controller.cpp
- Update ur_controllers/src/scaled_joint_trajectory_controller.cpp
- Update ur_controllers/src/scaled_joint_trajectory_controller.cpp
- Correct formating, include std::vector and update ros2_controller to master branch in repo file.
- Modifiy parameter declaration - approach equalization with ros-controls dependencies (#152)
- Update CI configration to supprot galactic and rolling (#142)
- Update main branch with ros-controls changes (#160)
- Update controllers adding dt in to update as in ros2_control (#171)
- Add missing dependency on angles and update formatting for linters. (#283)
- Stripped references to ur_controllers
- Formatting as requested in this repo


destogl (changes requested):
Let's keep only  `scaled_joint_trajectory_controller` in this PR and move it to `joint_trajectory_controller` package.

Also, adding a test or two for it would be good. Probably some existing JTC-tests can be copied and adjusted to take into account scaling factor.

scaled_controllers/include/scaled_controllers/scaled_joint_trajectory_controller.hpp
```diff
@@ -0,0 +1,66 @@
...
+//----------------------------------------------------------------------
+/*!\file
+ *
+ * \author  Marvin Große Besselmann grosse@fzi.de
+ * \date    2021-02-18
+ *
+ */
+//----------------------------------------------------------------------
+#ifndef SCALED_CONTROLLERS__SCALED_JOINT_TRAJECTORY_CONTROLLER_HPP_
+#define SCALED_CONTROLLERS__SCALED_JOINT_TRAJECTORY_CONTROLLER_HPP_
```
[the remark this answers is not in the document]
bijoua29: I agree. I would propose adding the scaled controller to the joint_trajectory_controller package. In fact, as suggested in the description, I would rather add the speed scaling as a field in the JTC instead of creating a new controller. There is some duplication of code and it seems brittle since changes in JTC may have to be reflected in this controller.

scaled_controllers/include/scaled_controllers/speed_scaling_state_broadcaster.hpp
```diff
@@ -0,0 +1,70 @@
...
+/*!\file
+ *
+ * \author  Marvin Große Besselmann grosse@fzi.de
+ * \date    2021-02-10
+ *
+ */
+//----------------------------------------------------------------------
+
+#ifndef SCALED_CONTROLLERS__SPEED_SCALING_STATE_BROADCASTER_HPP_
+#define SCALED_CONTROLLERS__SPEED_SCALING_STATE_BROADCASTER_HPP_
```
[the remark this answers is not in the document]
bijoua29: I also agree removing the broadcaster. It seems a little overkill and it should rather be published as status of the controller.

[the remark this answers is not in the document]
fmauch: You mean as a status of the modified JTC? Sure, that sounds reasonable.

destogl (commented):
@fmauch just pinging this again. What's your plan about this PR?

I see two main things before do proper merging:

- [ ] move code to `joint_trajectory_controller`
- [ ] rather extend `joint_trajectory_controller` with a parameter then duplicate its code (what do you think about this?

bmagyar:
@fmauch @destogl

can we agree we want to move the scaled jtc into this repo?

christophfroehlich:
> @fmauch @destogl
> 
> can we agree we want to move the scaled jtc into this repo?

As far as I can remember we decided to add the scaling-feature to the existing JTC?

fmauch:
Yes, that was the plan. @destogl told me a couple of times that he planned to do it at some point, so I kept my feed still until now.

destogl:
Yes, I wanted to do this. It will be done in the next 2 weeks, since we have a release of application with this feature. So I will clean this up to be ready for merge.

fmauch:
I've updated the PR with the current state from ur_controllers' scaled JTC. As discussed above, I modified the JTC to accept a scaling interface. There is still some work to be planned, hence converting to draft:

- [ ] Make the speed scaling interface optional (If not specified, speed scaling of 1 should be assumed)
- [ ] Make the speed scaling configurable by a service call. (This doesn't necessarily have to be done in this PR, though)
- [ ] Add documentation about speed scaling
- [ ] Add tests for speed scaling

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -189,25 +202,35 @@ controller_interface::return_type JointTrajectoryController::update(
...
   {
+    // Adjust time with scaling factor
+    TimeData time_data;
+    time_data.time = time;
+    rcl_duration_value_t t_period = (time_data.time - time_data_.readFromRT()->time).nanoseconds();
+    time_data.period = rclcpp::Duration::from_nanoseconds(scaling_factor_ * t_period);
+    time_data.uptime = time_data_.readFromRT()->uptime + time_data.period;
+    rclcpp::Time traj_time =
+      time_data_.readFromRT()->uptime + rclcpp::Duration::from_nanoseconds(t_period);
+    time_data_.writeFromNonRT(time_data);
```
christophfroehlich, line 213: This should be a writeFromRT, but this doesn't exist see #168 
I'm not sure what is the better way to perform this, but with the set_hold_position I did it that way:
```suggestion    
    time_data_.reset();
    time_data_.initRT(time_data);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -934,6 +957,13 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
 controller_interface::CallbackReturn JointTrajectoryController::on_activate(
   const rclcpp_lifecycle::State &)
 {
+  // Setup time_data buffer used for scaling
+  TimeData time_data;
+  time_data.time = get_node()->now();
+  time_data.period = rclcpp::Duration::from_nanoseconds(0);
+  time_data.uptime = get_node()->now();
+  time_data_.initRT(time_data);
```
christophfroehlich, line 965: or
```suggestion
  time_data_.writeFromNonRT(time_data);
```
I guess? If this is wrong, I have to change the other buffers in on_activate

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -114,12 +115,24 @@ JointTrajectoryController::state_interface_configuration() const
...
 controller_interface::return_type JointTrajectoryController::update(
   const rclcpp::Time & time, const rclcpp::Duration & period)
 {
+  if (state_interfaces_.back().get_name() == params_.speed_scaling_interface_name)
+  {
+    scaling_factor_ = state_interfaces_.back().get_value();
+  }
+  else
+  {
+    RCLCPP_ERROR(
```
christophfroehlich, line 131: Using a throttled logger or
```suggestion
    RCLCPP_ERROR_ONCE(
```

christophfroehlich (changes requested):
I'm late on this topic, and don't have experience with UR.
Where does the info of the scaling come from? Why is this implemented as a variable coming from the hardware?
But you mentioned a service call tbd, so that users can probably set the speed from a GUI and we could add it to rqt_joint_trajectory_controller.

If there isn't just position interface: Shouldn't it scale the velocity and acceleration according to the scaling factor? This would be a more general solution, though not perfect because we cannot scale effort values.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: I'll leave this unresolved for now. Once the scaling interface is optional it would probably make sense to return ERROR at this point (If scaling is configured to be used, but no interface is found).

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: The [docs](https://control.ros.org/master/doc/ros2_controllers/doc/writing_new_controller.html) state

>  This method is part of the real-time loop, therefore avoid any reservation of memory and, in general, keep it as short as possible.

so, probably  `initRT` is the correct approach?

fmauch:
> Why is this implemented as a variable coming from the hardware? But you mentioned a service call tbd, so that users can probably set the speed from a GUI and we could add it to rqt_joint_trajectory_controller.

On URs this is indeed coming from the hardware. The value used is actually a combination between the "Speed slider" and any safety-related scaling done by the controller. Imagine a straight horizontal arm rotating around its base with a restricted Cartesian velocity setup in the safety configuration. the UR will actually slow down the motion if rotating at full speed would cause a velocity above the limit.

Hence, we read the actual value from the hardware in that case. For robots that don't have this kind of mechanism it might make sense to add this using a service call, though.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: possibly? but is a `reset()` before necessary then?

christophfroehlich:
> On URs this is indeed coming from the hardware. The value used is actually a combination between the "Speed slider" and any safety-related scaling done by the controller. Imagine a straight horizontal arm rotating around its base with a restricted Cartesian velocity setup in the safety configuration. the UR will actually slow down the motion if rotating at full speed would cause a velocity above the limit.

Got it. Do you use a position interface only, or does the UR scale the velocity values internally and you only stretch the trajectory in time?

fmauch:
> Got it. Do you use a position interface only, or does the UR scale the velocity values internally and you only stretch the trajectory in time?

Indeed, up until now we only used it with a position interface. With velocities (or even efforts) things get a bit more complicated... I guess I'll have to think about that conceptually a bit more...

christophfroehlich:
> Indeed, up until now we only used it with a position interface. With velocities (or even efforts) things get a bit more complicated... I guess I'll have to think about that conceptually a bit more...

I think there is no way to scale efforts properly. And velocity+acceleration would be trivial only assuming the scaling being constant.

fmauch:
@bmagyar is there a reason this got closed? I was hoping to finish this until next week.

firesurfer:
@fmauch Whats the state? You wrote in april you are hoping to finish it within the week.
@bmagyar Same question as in april. Is there a reason this PR was closed?


Btw. this PR might be the right one to follow up on the discussion we had in: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/883

Let me know @fmauch if I can help you with upstreaming the scaled JTC. ( I have a certain interest because of #1182 :D )

PR #276, opened 2021-12-13
joint_state_broadcaster to use realtime tools

bmagyar:

PR #308, opened 2022-03-02
Foxy:  `joint_state_broadcaster` to use realtime tools

zultron:
This is a backport to Foxy of #276, "`joint_state_broadcaster` to use realtime tools".

In some tests in our Foxy-based project, before this patch, the controller manager `update()` function could take up to 33ms to return.

After backporting this patch, I see about 40µs max during normal operation and 80µs max when activating the controller.  Three orders of magnitude less!  Thanks for the original fix, @bmagyar .


destogl (approved):
Thanks for this!

I know this is ABI braking change, but it is also critical for performance.

bmagyar:
Wow, great feedback, thanks!

bmagyar (approved):

PR #309, opened 2022-03-07
Fix linking of target `test_load_joint_trajectory_controller`

kb0n:
Fixes "No such file or directory" error:
`joint_trajectory_controller.hpp:27:10: fatal error: control_toolbox/pid.hpp: No such file or directory`

It needs `control_toolbox`, but depends on `joint_trajectory_controller` anyway. Hence the single-target linking, just as it is done for all other test-targets here as well.

fixup for 1fd8bc37f284098aed39d4e6bd324a8528a72ad4

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -87,6 +87,7 @@ if(BUILD_TESTING)
     test/test_load_joint_trajectory_controller.cpp
   )
   target_include_directories(test_load_joint_trajectory_controller PRIVATE include)
+  target_link_libraries(test_load_joint_trajectory_controller joint_trajectory_controller)
   ament_target_dependencies(test_load_joint_trajectory_controller
     controller_manager
     realtime_tools


bmagyar:
hi,

the change seems sensible but I wonder how come we did not see this issue come up on the CI. What OS are you running / what ROS distro?

kb0n:
> hi,
> 
> the change seems sensible but I wonder how come we did not see this issue come up on the CI. What OS are you running / what ROS distro?

I'm on Debian 11 (bullseye), using a rather recent ros2-rolling installation according to https://docs.ros.org/en/rolling/Installation/Ubuntu-Development-Setup.html

JafarAbdi (approved):
We run into the same issue today in our CI

destogl:
No, this is not an issue! The point of the test is to load JTC using pluginlib, as controller manger would do. And having controller manager to be linked to all and every controller is not only impossible, but also not intended.

bmagyar:
I agree that this only happens to accidentally fix it but it is quite a nuclear solution.
How about adding `control_toolbox` to `ament_target_dependencies` for this test?

Note that it'd also be nice to reproduce this with a CI somehow.

bmagyar (changes requested):
I agree that this only happens to accidentally fix it but it is quite a nuclear solution.
How about adding control_toolbox to ament_target_dependencies for this test?

Note that it'd also be nice to reproduce this with a CI somehow.

bmagyar:
Closing in favour of #319 , please reopen if issue still persists

PR #311, opened 2022-03-09
gripper_action_controller: Disable writing until action is active

SyllogismRXS:
When the gripper_action_controller starts it writes zero's to the gripper desired position, which causes the gripper to start moving on startup, which is not the behavior of the joint_trajectory_controller. I modified the gripper_action_controller to disable writing to the hardware unless an action goal is active.

I also had to link some of the tests to control_toolbox in order to build correctly.

### DIFF
--- a/gripper_controllers/include/gripper_controllers/gripper_action_controller_impl.hpp
+++ b/gripper_controllers/include/gripper_controllers/gripper_action_controller_impl.hpp
@@ -64,6 +64,11 @@ template <const char * HardwareInterface>
 controller_interface::return_type GripperActionController<HardwareInterface>::update(
   const rclcpp::Time & /*time*/, const rclcpp::Duration & /*period*/)
 {
+  if (!rt_active_goal_) {
+    set_hold_position();
+    return controller_interface::return_type::OK;
+  }
+
   command_struct_rt_ = *(command_.readFromRT());
 
   const double current_position = joint_position_state_interface_->get().get_value();
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -80,6 +80,7 @@ if(BUILD_TESTING)
     realtime_tools
     ros2_control_test_assets
     trajectory_msgs
+    control_toolbox
   )
 
   ament_add_gtest(
@@ -91,6 +92,7 @@ if(BUILD_TESTING)
     controller_manager
     realtime_tools
     ros2_control_test_assets
+    control_toolbox
   )
 
   ament_add_gtest(


joint_trajectory_controller/CMakeLists.txt
```diff
@@ -91,6 +92,7 @@ if(BUILD_TESTING)
     controller_manager
     realtime_tools
     ros2_control_test_assets
+    control_toolbox
```
destogl, line 95: I am very sure that this is not needed. Is `control_toolbox` used explicitly in the test?

gripper_controllers/include/gripper_controllers/gripper_action_controller_impl.hpp
```diff
@@ -64,6 +64,11 @@ template <const char * HardwareInterface>
 controller_interface::return_type GripperActionController<HardwareInterface>::update(
   const rclcpp::Time & /*time*/, const rclcpp::Duration & /*period*/)
 {
+  if (!rt_active_goal_) {
+    set_hold_position();
+    return controller_interface::return_type::OK;
+  }
```
destogl, line 70: This solution will not work if there is given command from topic and not action. Why not put proper state interface value when starting controller?

SyllogismRXS: Do you mean setting a default initial value for the controller? It's potentially dangerous to have the controller start moving servos on startup with a real robot system without an explicit request. When developing on a physical system, you can't assume that the robot will be in the same physical state on startup every time. The joint_trajectory_controller operates in the same way. When you start the ros2 control system, it doesn't start moving the servos to an initial value until a plan is requested.

SyllogismRXS: I only tested this with the action interface, not the topic interface. I'll look into how to achieve the same effect for both interfaces.

joint_trajectory_controller/CMakeLists.txt
SyllogismRXS: Both tests gave me compile errors without including control_toolbox because the tests include test_trajectory_controller_utils.hpp, which includes joint_trajectory_controller.hpp, which includes control_toolbox/pid.hpp. Is there a way to have the includes from joint_trajectory_controller.hpp be included in the tests automatically with CMake?

Serafadam: I guess this is related to https://github.com/ros-controls/ros2_controllers/pull/309#issuecomment-1065910124 ?

gripper_controllers/include/gripper_controllers/gripper_action_controller_impl.hpp
destogl: @SyllogismRXS did you have time to make this also working on topic interface?

We should probably get state interface from the gripper on the beginning and set commands to states in `on_activate`.

destogl: We should for sure have the following:

- in `on_activate` commands should be assigned to states; something like here storing into last_commanded_state_; maybe some variables like this if there is no other ([example](https://github.com/ros-controls/ros2_controllers/blob/afe216fa016150a493a29ae7e452fd5e8eef81ec/admittance_controller/src/admittance_controller.cpp#L372)
- check in update is new goal/command has arrived. The best would be to store the goal/command in a realtime buffer es done in JTC [example](https://github.com/ros-controls/ros2_controllers/blob/76bf6ae70c432565c8767fba14c9adc75028292e/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1218) and then using this in the update loop [like here](https://github.com/ros-controls/ros2_controllers/blob/76bf6ae70c432565c8767fba14c9adc75028292e/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L136)
This could be done exactly like this using pointers. If a command is sent, then one could reset it to `nullptr`.

Does this help a bit?

SyllogismRXS:
This issue of the gripper changing position to the default zero position doesn't occur in ROS Humble now. This PR can be closed without merging it in.

PR #315, opened 2022-03-15
CMakeLists.txt: added missing control_toolbox dependencies

DasRoteSkelett:
The tests of joint_trajectory_controller need to depend on
control_toolbox, oherwise, the
(which should imho be in <> instead of "") will report
No such file or directory when building against ros2_rolling
source compile.

Signed-off-by: Matthias Schoepfer <m.schoepfer@rethinkrobotics.com>

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -77,6 +77,7 @@ if(BUILD_TESTING)
     hardware_interface
     rclcpp
     rclcpp_lifecycle
+    control_toolbox
     realtime_tools
     ros2_control_test_assets
     trajectory_msgs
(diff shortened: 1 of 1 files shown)


bmagyar (approved):

destogl (approved):

destogl:
> (which should imho be in <> instead of "") 

Just out of curiosity, why do you think that?

We are following “the rule” that inside `<>` are system/std-library dependencies, other deps are in `""`)

DasRoteSkelett:
@destogl : 
From reading https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html, it would suggest to me to use "" when including header from inside the package, and <> else.

PR #318, opened 2022-03-23
Chained Controllers Example: Forwarding-Chained controllers.

destogl:
this PR is only for exemplary purposes to enable demonstration of chained controllers.


christophfroehlich:
We have a demo now in the demos, see https://github.com/ros-controls/ros2_control_demos/pull/318
Is there anything more inside this PR, or can we close it?

christophfroehlich:
I think we can close it, as we have now two demos with forward-chaining in the demos.

PR #320, opened 2022-03-31
Minor cleanup & add TODOs for more

bmagyar:


jaron-l:
How are TODOs used in this organization? Are they supposed to act as flags for minor things that anyone can pick up? I'm not sure how most people use them, but I've only used them as flags for work that needs to be done before the request is merged.

destogl:
@jaron-l you are right, todos should be resolved before merge or is bigger changes are needed we should open an follow up issue describing it and referencing it in the code.

Other than that, this PR should be 'draft' because we will open a lot of issues from the todos.

(You can tackle some of them if you like)

christophfroehlich:
I'll close this PR in favor of #944, let's discuss things there.

PR #321, opened 2022-04-01
Lifecycle updates

jackcenter:
I was unable to build on the Galactic branch without updating the tests and nodes to lifecycle nodes. This PR should allow the ros2_controllers galactic branch to build in Galactic with ros2_control on the master branch. 

I also reverted #283 because the reset function was not being found. I'm sure there is a way to use that function still, but I haven't looked into it.


destogl:
Hi,

We are actually targeting master-master and galactic-galactic branches to be compatible. This PR obviously breaks this.

PR #323, opened 2022-04-05
Variable renaming in the joint traj controller

AndyZe:
The output variable for `sample()` was named wrong.

This work is sponsored by RE2 Robotics.


destogl (approved):
Thanks! This makes it clearer!

arne48 (approved):

PR #324, opened 2022-04-05
Add Ruckig jerk-limited smoothing to the joint trajectory controller

AndyZe:
Optionally smooth one outgoing command at a time. This is nice because the trajectories will now obey velocity, acceleration, and jerk limits.

Depends on the `joint_limits` PR in the ros2_control repo:  https://github.com/ros-controls/ros2_control/pull/462

To use this optional feature, you just need to add a parameter to the yaml file for the controller, e.g. `ros2_controllers.yaml`:

```
panda_arm_controller:
  ros__parameters:
    command_interfaces:
      - position
    state_interfaces:
      - position
      - velocity
    joints:
      - panda_joint1
...
      - panda_joint7

    ruckig_jerk_smoothing: true
```

TODO:
- [x] Make this optional
- [ ] Read velocity/accel/jerk limits, likely from `joint_limits.yaml`  (I'm aware of https://github.com/ros-controls/ros2_control/pull/462)

This work is sponsored by RE2 Robotics.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
@@ -29,13 +29,14 @@ namespace interpolation_methods
 enum class InterpolationMethod
 {
   NONE,
-  VARIABLE_DEGREE_SPLINE
+  VARIABLE_DEGREE_SPLINE,
+  RUCKIG
 };
 
 const InterpolationMethod DEFAULT_INTERPOLATION = InterpolationMethod::VARIABLE_DEGREE_SPLINE;
 
 const std::unordered_map<InterpolationMethod, std::string> InterpolationMethodMap(
-  {{InterpolationMethod::NONE, "none"}, {InterpolationMethod::VARIABLE_DEGREE_SPLINE, "splines"}});
+  {{InterpolationMethod::NONE, "none"}, {InterpolationMethod::VARIABLE_DEGREE_SPLINE, "splines"}, {InterpolationMethod::RUCKIG, "ruckig"}});
 
 [[nodiscard]] inline InterpolationMethod from_string(const std::string & interpolation_method)
 {
@@ -49,13 +50,19 @@ const std::unordered_map<InterpolationMethod, std::string> InterpolationMethodMa
   {
     return InterpolationMethod::VARIABLE_DEGREE_SPLINE;
   }
+  else if (
+    !interpolation_method.compare(
+      InterpolationMethodMap.at(InterpolationMethod::RUCKIG)) == 0)
+  {
+    return InterpolationMethod::RUCKIG;
+  }
   // Default
   else
   {
     RCLCPP_INFO_STREAM(
       LOGGER,
       "No interpolation method parameter was given. Using the default, VARIABLE_DEGREE_SPLINE.");
-    return InterpolationMethod::VARIABLE_DEGREE_SPLINE;
+    return DEFAULT_INTERPOLATION;
   }
 }
 }  // namespace interpolation_methods
--- a/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
@@ -36,7 +36,8 @@ enum class InterpolationMethod
 const InterpolationMethod DEFAULT_INTERPOLATION = InterpolationMethod::VARIABLE_DEGREE_SPLINE;
 
 const std::unordered_map<InterpolationMethod, std::string> InterpolationMethodMap(
-  {{InterpolationMethod::NONE, "none"}, {InterpolationMethod::VARIABLE_DEGREE_SPLINE, "splines"}, {InterpolationMethod::RUCKIG, "ruckig"}});
+  {{InterpolationMethod::NONE, "none"}, {InterpolationMethod::VARIABLE_DEGREE_SPLINE, "splines"},
+  {InterpolationMethod::RUCKIG, "ruckig"}});
 
 [[nodiscard]] inline InterpolationMethod from_string(const std::string & interpolation_method)
 {
@@ -45,13 +46,13 @@ const std::unordered_map<InterpolationMethod, std::string> InterpolationMethodMa
     return InterpolationMethod::NONE;
   }
   else if (
-    !interpolation_method.compare(
+    interpolation_method.compare(
       InterpolationMethodMap.at(InterpolationMethod::VARIABLE_DEGREE_SPLINE)) == 0)
   {
     return InterpolationMethod::VARIABLE_DEGREE_SPLINE;
   }
   else if (
-    !interpolation_method.compare(
+    interpolation_method.compare(
       InterpolationMethodMap.at(InterpolationMethod::RUCKIG)) == 0)
   {
     return InterpolationMethod::RUCKIG;
--- a/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
@@ -16,6 +16,7 @@
 #define JOINT_TRAJECTORY_CONTROLLER__TRAJECTORY_HPP_
 
 #include <memory>
+#include <ruckig/ruckig.hpp>
 #include <vector>
 
 #include "joint_trajectory_controller/interpolation_methods.hpp"
@@ -76,7 +77,7 @@ public:
    *    return false
    *
    * \param[in] sample_time Time at which trajectory will be sampled.
-   * \param[in] interpolation_method Specify whether splines, another method, or no interpolation at all.
+   * \param[in] interpolation_method Specify whether splines, ruckig smoothing, another method, or no interpolation at all.
    * \param[out] expected_state Calculated new at \p sample_time.
    * \param[out] start_segment_itr Iterator to the start segment for given \p sample_time. See description above.
    * \param[out] end_segment_itr Iterator to the end segment for given \p sample_time. See description above.
@@ -105,13 +106,15 @@ public:
    * \param[in] time_b Time at which the segment state equals \p state_b.
    * \param[in] state_b State at time \p time_b.
    * \param[in] sample_time The time to sample, between time_a and time_b.
+   * \param[in] do_ruckig_smoothing Optionally apply Ruckig jerk-limited smoothing.
    * \param[out] output The state at \p sample_time.
    */
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC
-  void interpolate_between_points(
+  bool interpolate_between_points(
     const rclcpp::Time & time_a, const trajectory_msgs::msg::JointTrajectoryPoint & state_a,
     const rclcpp::Time & time_b, const trajectory_msgs::msg::JointTrajectoryPoint & state_b,
-    const rclcpp::Time & sample_time, trajectory_msgs::msg::JointTrajectoryPoint & output);
+    const rclcpp::Time & sample_time, const bool do_ruckig_smoothing,
+    trajectory_msgs::msg::JointTrajectoryPoint & output);
 
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC
   TrajectoryPointConstIter begin() const;
@@ -137,6 +140,11 @@ public:
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC
   bool is_sampled_already() const { return sampled_already_; }
 
+  void reset_ruckig_smoothing()
+  {
+    have_previous_ruckig_output_ = false;
+  }
+
 private:
   void deduce_from_derivatives(
     trajectory_msgs::msg::JointTrajectoryPoint & first_state,
@@ -149,7 +157,17 @@ private:
   rclcpp::Time time_before_traj_msg_;
   trajectory_msgs::msg::JointTrajectoryPoint state_before_traj_msg_;
 
-  bool sampled_already_ = false;
+  bool sampled_already_;
+
+  // For Ruckig jerk-limited smoothing
+  std::unique_ptr<ruckig::Ruckig<ruckig::DynamicDOFs>> smoother_;
+  ruckig::InputParameter<ruckig::DynamicDOFs> ruckig_input_{0};
+  ruckig::OutputParameter<ruckig::DynamicDOFs> ruckig_output_{0};
+  // To avoid instability, Ruckig runs in a closed-loop fashion:
+  // Ruckig output at cycle i is used as the initial state for cycle i+1.
+  // This flag determines whether we need to initialize the state or use the previous
+  // Ruckig output.
+  bool have_previous_ruckig_output_;
 };
 
 /**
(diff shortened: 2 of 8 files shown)


joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -270,6 +300,62 @@ void Trajectory::interpolate_between_points(
                                 t[2] * 12.0 * coefficients[4] + t[3] * 20.0 * coefficients[5];
     }
   }
+  // Apply velocity, acceleration, and jerk limits with Ruckig
+  if (period.seconds() > 0)
+  {
+    ruckig::Ruckig<ruckig::DynamicDOFs> smoother(dim, period.seconds());
```
AndyZe, line 306: It might seem strange to declare a new smoothing object with every cycle, but `period` could change. So I think it has to be done this way.

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -270,6 +293,66 @@ void Trajectory::interpolate_between_points(
                                 t[2] * 12.0 * coefficients[4] + t[3] * 20.0 * coefficients[5];
     }
   }
+  // Apply velocity, acceleration, and jerk limits with Ruckig
+  if ((period.seconds() > 0) && do_ruckig_smoothing)
+  {
+    ruckig::Ruckig<ruckig::DynamicDOFs> smoother(dim, period.seconds());
```
AndyZe, line 299: It might seem strange to declare a new smoothing object with every cycle, but dim and period could change. So I think it has to be done this way.

https://github.com/pantor/ruckig/blob/master/include/ruckig/ruckig.hpp

I filed an issue asking if a method can be added to update in a nicer way.

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -124,14 +131,28 @@ class Trajectory
...
   rclcpp::Time time_before_traj_msg_;
   trajectory_msgs::msg::JointTrajectoryPoint state_before_traj_msg_;
 
-  bool sampled_already_ = false;
+  bool sampled_already_;
+
+  // For Ruckig jerk-limited smoothing
+  ruckig::InputParameter<ruckig::DynamicDOFs> ruckig_input_{0};
+  ruckig::OutputParameter<ruckig::DynamicDOFs> ruckig_output_{0};
+  // To avoid instability, Ruckig runs in a closed-loop fashion.
```
destogl, line 151: What does this mean, “runs in the closed-loop fashion”? I thought that it is rather “open-loop” because input for the next calculation is the result of the previous one, or do we really get information from the hardware?

Is rucking stateless, or does it store it state in the library?

destogl: Ok I see in the rest of the comment. I think we should reformulate this a bit to avoid misunderstanding.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -485,7 +489,7 @@ CallbackReturn JointTrajectoryController::on_configure(const rclcpp_lifecycle::S
       // TODO(anyone): This flag is not used for now
       // There should be PID-approach used as in ROS1:
       // https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/include/joint_trajectory_controller/hardware_interface_adapter.h#L283
-      use_closed_loop_pid_adapter = true;
+      use_closed_loop_pid_adapter_ = true;
```
destogl, line 492: Can we rename a variable in a separate PR? It would be a bit cleaner and easier to back port.

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -23,20 +23,36 @@
...
+  time_before_traj_msg_(0),
+  sampled_already_(false),
+  have_previous_ruckig_output_(false) {}
+
+Trajectory::Trajectory(std::shared_ptr<trajectory_msgs::msg::JointTrajectory> joint_trajectory)
+: trajectory_msg_(joint_trajectory),
+  trajectory_start_time_(static_cast<rclcpp::Time>(joint_trajectory->header.stamp)),
+  sampled_already_(false),
+  have_previous_ruckig_output_(false) {}
+
```
destogl, line 47: Why those two additional constructors?

joint_trajectory_controller/src/trajectory.cpp
destogl: I would not say that `dim` can change, it is usually defined with the number of joint JTC is controlling. But yes, `update` in general may vary. Is there a possibility to update only the period?

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -270,6 +293,68 @@ void Trajectory::interpolate_between_points(
...
+      ruckig_input_.current_position = ruckig_output_.new_position;
+      ruckig_input_.current_velocity = ruckig_output_.new_velocity;
+      ruckig_input_.current_acceleration = ruckig_output_.new_acceleration;
+    }
+    // else, need to initialize the robot state and joint limits
+    else
+    {
+      // For velocity/acceleration/jerk-limited command smoothing
+      ruckig_input_ = ruckig::InputParameter<ruckig::DynamicDOFs>(dim);
+      ruckig_output_ = ruckig::OutputParameter<ruckig::DynamicDOFs>(dim);
```
destogl, line 315: This could also be defined “statically” in the constructor.

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -270,6 +293,68 @@ void Trajectory::interpolate_between_points(
...
+
+      // TODO(andyz): check if kinematic limits are defined in URDF or joint_limits.yaml
+      ruckig_input_.max_velocity = std::vector<double>(dim, DEFAULT_MAX_VELOCITY);
+      ruckig_input_.max_acceleration = std::vector<double>(dim, DEFAULT_MAX_ACCELERATION);
+      ruckig_input_.max_jerk = std::vector<double>(dim, DEFAULT_MAX_JERK);
```
destogl, line 320: Also define statically on trajectory level. I even think that we can define limits even for each trajectory separately (even point?)

joint_trajectory_controller/test/test_trajectory.cpp
```diff
@@ -98,78 +102,87 @@ TEST(TestTrajectory, sample_trajectory_positions)
...
+  rclcpp::Duration period = rclcpp::Duration::from_nanoseconds(1e7);
+  bool do_ruckig_smoothing = true;
 
   double duration_first_seg = 1.0;
   double velocity = (p1.positions[0] - point_before_msg.positions[0]) / duration_first_seg;
 
   // sample at trajectory starting time
   {
-    traj.sample(time_now, expected_state, start, end);
+    traj.sample(time_now, start, period, do_ruckig_smoothing, output_point, end);
```
destogl, line 115: Other than renaming/adjusting tests, is there a possibility to actually test for Jerk limits?

destogl (changes requested):
I like this very much. I have commented about small changes we should discuss before merging. I would even be ready to merge this with the current architecture after details are clarified. 


Nevertheless we should open two follow-up issues at least for:

1. separate “classical” trajectory interpolation and those using rucking into separate classes.
2. add test for jerk limits.
3. integrate joint limits from URDF/yaml. (should be anyway done in a follow-up)

AndyZe:
Closing this - I'm going to try another approach, which is to completely replace the quintic spline interpolation.

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
AndyZe: Added a colon to imply the user should keep reading --   :+1:

`// To avoid instability, Ruckig runs in a closed-loop fashion:`

joint_trajectory_controller/src/trajectory.cpp
AndyZe: OK, I was confused about `dim` changing. I see that it can't now. I'll refactor this.

joint_trajectory_controller/src/trajectory.cpp
AndyZe: Well, it's the same two constructors. I just initialized a couple more variables in the initializer list. `sampled_already_` and `have_previous_ruckig_output_` were initialized.

joint_trajectory_controller/src/trajectory.cpp
AndyZe: I don't think they can go in the constructor in case the input `std::shared_ptr<trajectory_msgs::msg::JointTrajectory>` is null. I'll look for a better place to initialize them, though.

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -149,7 +157,17 @@ class Trajectory
   rclcpp::Time time_before_traj_msg_;
   trajectory_msgs::msg::JointTrajectoryPoint state_before_traj_msg_;
 
-  bool sampled_already_ = false;
+  bool sampled_already_;
+
+  // For Ruckig jerk-limited smoothing
+  std::unique_ptr<ruckig::Ruckig<ruckig::DynamicDOFs>> smoother_;
+  ruckig::InputParameter<ruckig::DynamicDOFs> ruckig_input_{0};
```
AndyZe, line 164: This was probably my own code but I think the `0` argument is hard to understand. Let's add a constant at the top of the file and use it here:

```
// Allow Ruckig objects to be initialized with dynamic size
constexpr size_t RUCKIG_DYNAMIC_DOFS = 0;
```

```suggestion
  ruckig::InputParameter<ruckig::DynamicDOFs> ruckig_input_{RUCKIG_DYNAMIC_DOFS};
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -197,161 +205,162 @@ controller_interface::return_type JointTrajectoryController::update(
...
+    if (!valid_point)
+    {
+      return controller_interface::return_type::ERROR;
+    }
+
+    bool tolerance_violated_while_moving = false;
+    bool outside_goal_tolerance = false;
+    bool within_goal_time = true;
+    double time_difference = 0.0;
+    const bool before_last_point = end_segment_itr != (*traj_point_active_ptr_)->end();
```
AndyZe, line 217: Nitpick, for readability

```suggestion
    const bool before_last_point = (end_segment_itr != (*traj_point_active_ptr_)->end());
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -181,6 +186,9 @@ controller_interface::return_type JointTrajectoryController::update(
     if (!(*traj_point_active_ptr_)->is_sampled_already())
```
AndyZe, line 186: Since `*traj_point_active_ptr_` is used a lot and it's kind of annoying to read, you could do something like this:

`auto active_traj_point = *traj_point_active_ptr_;`

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -90,6 +117,8 @@ bool Trajectory::sample(
     return false;
   }
 
+  auto do_ruckig_smoothing =
+    interpolation_method ==interpolation_methods::InterpolationMethod::RUCKIG;
```
AndyZe, line 121: Readability nit

```suggestion
    interpolation_method == (interpolation_methods::InterpolationMethod::RUCKIG);
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -138,11 +167,15 @@ bool Trajectory::sample(
       // Do interpolation
       else
       {
-        // it changes points only if position and velocity do not exist, but their derivatives
+        // it changes points only if position and velocity are not exist, but their derivatives
```
AndyZe, line 170: Seems like the grammar was correct previously?

```suggestion
        // it changes points only if position and velocity do not exist, but their derivatives
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -64,6 +90,7 @@ bool Trajectory::sample(
   TrajectoryPointConstIter & start_segment_itr, TrajectoryPointConstIter & end_segment_itr)
 {
   THROW_ON_NULLPTR(trajectory_msg_)
+  // TODO(anyone): this shouldn't be initialized at runtime
   output_state = trajectory_msgs::msg::JointTrajectoryPoint();
```
AndyZe, line 94: I think this line is OK, actually, since the size of  `trajectory_msgs::msg::JointTrajectoryPoint();` is known. This line (probably) won't add any overhead. See https://stackoverflow.com/a/3835902

What will hurt runtime is a few lines down, when `output_state` is resized.

`output_state = state_before_traj_msg_;`

I'm glad you're thinking about this though!

The best improvement I can think of is:

```suggestion
  `trajectory_msgs::msg::JointTrajectoryPoint& output_state;`
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -166,14 +199,16 @@ bool Trajectory::sample(
...
   const rclcpp::Time & time_a, const trajectory_msgs::msg::JointTrajectoryPoint & state_a,
   const rclcpp::Time & time_b, const trajectory_msgs::msg::JointTrajectoryPoint & state_b,
-  const rclcpp::Time & sample_time, trajectory_msgs::msg::JointTrajectoryPoint & output)
+  const rclcpp::Time & sample_time, const bool do_ruckig_smoothing,
+  trajectory_msgs::msg::JointTrajectoryPoint & output)
 {
   rclcpp::Duration duration_so_far = sample_time - time_a;
   rclcpp::Duration duration_btwn_points = time_b - time_a;
 
+  // TODO(anyone): this shouldn't be resized at runtime
```
AndyZe, line 211: Same deal, per  https://stackoverflow.com/questions/3835857/in-c-does-using-static-variables-in-a-function-make-it-faster/3835902#3835902

```suggestion
```

joint_trajectory_controller/test/test_trajectory.cpp
```diff
@@ -62,10 +63,11 @@ TEST(TestTrajectory, initialize_trajectory)
     const auto now = rclcpp::Clock().now();
     auto traj = joint_trajectory_controller::Trajectory(empty_msg);
     const auto traj_starttime = traj.time_from_start();
+    rclcpp::Duration period = rclcpp::Duration::from_nanoseconds(1e7);
 
-    trajectory_msgs::msg::JointTrajectoryPoint expected_point;
+    trajectory_msgs::msg::JointTrajectoryPoint output_point;
```
AndyZe, line 68: I noticed as well that `expected_point` was named in a funny way. Glad to see it renamed  :+1:

joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
```diff
@@ -29,13 +29,15 @@ namespace interpolation_methods
 enum class InterpolationMethod
 {
   NONE,
-  VARIABLE_DEGREE_SPLINE
+  VARIABLE_DEGREE_SPLINE,
+  RUCKIG
```
AndyZe, line 33: It might be confusing that splines run, followed by Ruckig. Maybe this should be renamed, SPLINES_THEN_RUCKIG?

```suggestion
  SPLINES_THEN_RUCKIG
```

AndyZe:
Closing this for now because deciding when to reset Ruckig is not trivial.

For example, should it reset after every new trajectory message? That won't work if commands are being streamed to the robot, as in teleoperation.

gavanderhoorn:
Ruckig works rather well as a `joint_group_(position|velocity)_controller`.

PR #328, opened 2022-04-11
feat: :construction_worker: add gtests to ctest

jaron-l:
Adds support for `ctest` by adding `gtest_discover_tests` to all of the `CMakeLists.txt` files. This allows one to use `colcon test --ctest-args ...`. I have been using this a lot to filter tests so I thought I'd share it case others want to use it.

### DIFF
--- a/diff_drive_controller/CMakeLists.txt
+++ b/diff_drive_controller/CMakeLists.txt
@@ -63,6 +63,7 @@ if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
   find_package(controller_manager REQUIRED)
   find_package(ros2_control_test_assets REQUIRED)
+  include(GoogleTest)
 
   ament_add_gmock(test_diff_drive_controller
     test/test_diff_drive_controller.cpp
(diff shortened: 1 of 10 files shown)


jaron-l:
As an FYI, I discovered that this results in a slowdown of running the tests. Probably not a big deal when run tests once, but if using with an colcon retest flags or looping manually, you'll notice. I created an issue at colcon/colcon-cmake#120.

bmagyar:
I'm a little reluctant about adding this change. I think the majority of users won't use ctest to filter their tests, but rather call the test binary directly.

At the same time I'm worried about the extra amount of code in the cmake + the explicit gtest imports... I suggest we close this PR instead of merging

jaron-l:
> I'm a little reluctant about adding this change. I think the majority of users won't use ctest to filter their tests, but rather call the test binary directly.
> 
> At the same time I'm worried about the extra amount of code in the cmake + the explicit gtest imports... I suggest we close this PR instead of merging

That's fine. I just thought I'd offer it as I used it a lot in debugging #263. It's really mostly useful if you want the tests to show up independently with colcon test.

PR #334, opened 2022-04-20
Use system time in all tests to avoid error with different time sources.

destogl:
This PR solves (a new) issues caused by difference in time source (probably some changes in rclcpp cause that). 
An example of the issue can be found [here](https://github.com/ros-controls/ros2_controllers/runs/6030141354?check_suite_focus=true#step:6:6121).


bmagyar:
There still seems to be errors... shouldn't this PR fix them?

```
 - joint_trajectory_controller.OnlyEffortTrajectoryControllers/TrajectoryControllerTestParameterized test_ignore_partial_old_trajectory/1
  <<< failure message
    unknown file
    C++ exception with description "can't compare times with different time sources" thrown in the test body.
  >>>
```

bmagyar (approved):

bmagyar:
Only the joint_state_broadcaster tests fail on the CI ATM

PR #296, opened 2022-02-12
Use time argument on update function instead of node time

erickisos:
Closes #290

PR #340, opened 2022-04-26
Preallocate JTC variables to avoid resizing in realtime loops

AndyZe:
- Preallocate several variables
- `last_commanded_state_` was previously used for open-loop control. I think it's redundant with `state_desired_` now, so delete it


bmagyar (approved):

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -272,12 +267,9 @@ controller_interface::return_type JointTrajectoryController::update(
           }
           else
           {
-            assign_interface_from_point(joint_command_interface_[3], state_desired.effort);
+            assign_interface_from_point(joint_command_interface_[3], state_desired_.effort);
           }
         }
-
-        // store command as state when hardware state has tracking offset
-        last_commanded_state_ = state_desired;
```
mechwiz, line 280: Hmmm, isn't it possible that a scenario could occur where the last commanded state and state desired differ (such as when a path tolerance isn't met)?
That would probably affect the starting point used in open loop control since the state desired at that point would not be the last commanded state

AndyZe: Well, let's look at how `last_commanded_state_` was updated prior to this PR:

1. It gets set equal to state_desired at [L280](https://github.com/ros-controls/ros2_controllers/blob/7bd93cc9f38d9342b1e0ca3c63bcf1d25e7c6ee4/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L280)

2. It gets initialized at [L826](https://github.com/ros-controls/ros2_controllers/blob/7bd93cc9f38d9342b1e0ca3c63bcf1d25e7c6ee4/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L826)/[833](https://github.com/ros-controls/ros2_controllers/blob/7bd93cc9f38d9342b1e0ca3c63bcf1d25e7c6ee4/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L833)

This PR doesn't change 1 nor 2. So I think the functionality is equivalent. I guess the variable name was misleading.

AndyZe: I tested this with the `open_loop` flag just now. Seems to work OK in simulation. It did help me catch a bug with `dof_` initialization, though.

https://user-images.githubusercontent.com/11284393/166473164-96a38ab9-7db6-47fd-8657-7840755fd46e.mp4

mechwiz: While that is true, `state_desired` gets updated on every iteration when the call to sample is made [here](https://github.com/ros-controls/ros2_controllers/blob/7bd93cc9f38d9342b1e0ca3c63bcf1d25e7c6ee4/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L185). This is not the case of `last_commanded_state_` which only gets set to `state_desired` if no tolerance is violated. So in the iteration of when a tolerance is violated, `state_desired` will be the new command (that hasn't yet been commanded and will proceed to not be commanded) and `last_commanded_state` will be the last command (that has already been commanded). This means when a new trajectory comes in after the tolerance violation, the `last_commanded_state_` will be used to set the starting point in the trajectory when in `open_loop_control` which is correct.

I'm thinking that with your changes, even when the `state_desired_` command has not been commanded in a case of a tolerance violation, it'll still be used as the starting point when a new trajectory comes in which I don't think we want.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -272,14 +267,13 @@ controller_interface::return_type JointTrajectoryController::update(
...
       }
 
+      last_commanded_state_ = state_desired;
```
mechwiz, line 275:
```suggestion
      last_commanded_state_ = state_desired_;
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -822,14 +818,21 @@ controller_interface::CallbackReturn JointTrajectoryController::on_activate(
   last_state_publish_time_ = get_node()->now();
 
   // Initialize current state storage if hardware state has tracking offset
+  resize_joint_trajectory_point(state_current_, dof_);
+  resize_joint_trajectory_point(state_desired_, dof_);
+  resize_joint_trajectory_point(state_error_, dof_);
   resize_joint_trajectory_point(last_commanded_state_, joint_names_.size());
```
mechwiz, line 824:
```suggestion
  resize_joint_trajectory_point(last_commanded_state_, dof_);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
AndyZe: That makes sense. Done

AndyZe:
I don't understand why quite a few unit tests are failing. Marking as draft until that's fixed.

mechwiz:
I think #296 broke the JTC tests which is why many of them fail.
I think #334 is an attempt to fix the issue but it hasn't yet been resolved.

AndyZe:
OK, it does look like this PR does not make the test failures worse. Re-marking it as ready for review.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -822,14 +827,21 @@ controller_interface::CallbackReturn JointTrajectoryController::on_activate(
   last_state_publish_time_ = get_node()->now();
 
   // Initialize current state storage if hardware state has tracking offset
-  resize_joint_trajectory_point(last_commanded_state_, joint_names_.size());
+  resize_joint_trajectory_point(state_current_, dof_);
+  resize_joint_trajectory_point(state_desired_, dof_);
+  resize_joint_trajectory_point(state_error_, dof_);
+  resize_joint_trajectory_point(last_commanded_state_, dof_);
```
destogl, line 833: can we resize those in `on_configure`? Since this should be, if possible, real-time capable method.

destogl (approved):
I find this very good! Thanks! Just one small comment.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
AndyZe: That makes sense. Done and tested.

destogl:
The failing CI things are OK. Merging this.

PR #344, opened 2022-05-02
Tricycle Controller

tonynajjar:
Hey! We at [Pixel Robotics](https://www.pixel-robotics.eu/) recently started our migration to ros2_control. We are in the process of creating a tricycle controller and decided to contribute back.

This PR is not yet complete but we wanted to get a first feedback round already.

The main things still in progress are:

1- The process_twist_command still needs some logic to take care of corner cases to make the navigation smoother

2- We have a particular need (that might actually be generalized for other people) to apply some safety validation on the steering angle (alpha) and wheel speed (Ws) according to the robot state (e.g sensors triggered) before sending it to the HW interface (send_commands). This validation is done in a separate node that contains the state of the robot, this is why we publish the steering angle and wheel speed in update() and send the validated commands in send_command(). 
This surely looks like bad practice as it increases latency, how else would you do it?


bmagyar:
Thanks for the contribution! Could you please retarget the PR to the master branch? We don't add new features to already released distros, only backport them from Rolling

tonynajjar:
I was doubting so, will do, I think there are minor changes I will need to do

tonynajjar:
done https://github.com/ros-controls/ros2_controllers/pull/345

Issue #288, opened 2022-01-26
Ackermann Steering Controller


armuren:
## Background

Aside from issue #264 have there been any attempts at an ackermann controller for ros2-controllers?  From what I can tell of that issue, the port is no longer being worked on.

Here's what I currently have:

1. A project with a URDF description that is gazebo valid and spawns a simple 4 wheel robot
2. Integration with gazebo_ros2_control linking to the ros2-control Diff Drive controller config
3. the small city world in Gazebo with a few cars, streets, stop signs, traffic lights, etc

My plan is to follow the instructions in this [comment](https://github.com/ros-controls/ros2_controllers/issues/264#issuecomment-996646394) and begin attempting to implement an ackerman steering controller.  

If I am successful in this, I may then open a PR for the controller, plus a PR in ros-control-demos with the content I described above.

Any input, feedback, or direction is highly welcomed!

bmagyar:
This would be a very welcome change, thank you! Would you mind posting the same message in #264  that you are essentially planning to take over this effort to notify the people over there? And this issue could be closed as a duplicate of that.

aboarya:
Sure thing! Will do...

tonynajjar:
Hi @armuren, I'm sure many are impatiently waiting for your contributions! Could you share any updates on your progress?

tonynajjar:
We waited so impatiently that we decided to give it a try ourselves :smile: : https://github.com/ros-controls/ros2_controllers/pull/344/files
It's not exactly ackermann but I think with minor changes it could also work for ackermann as the kinematics are very similar as far as I know

armuren:
@tonynajjar my apologies! I got distracted and was not able to follow up on this.  I will close this PR

Issue #264, opened 2021-11-08
Mobile Robotics Controllers (Ackermann, four_wheels, swerve, ...)


tomlogan501:
Hi,
I create just this topic to keep track of the transfert of those controllers from ROS1 to ROS2.
We can discuss the main changes and the benefit to do so (new architecture, cyclic, ...)

harderthan:
Hello. 

I create [ackermann-steering-controller-ros2 repository](https://github.com/harderthan/ackermann-steering-controller-ros2) and implement a migration for ros1 [ackermann_steering_controller](https://github.com/ros-controls/ros_controllers/tree/noetic-devel/ackermann_steering_controller)  

I finished only to build the migrated code. I have a todo list below.

#### TODO 
- [x] Trouble shooting to build
- [ ] Add the Odometry/Velocity Calculation on ackermann_steering_controller.cpp
- [ ] Add test codes for ackermann_steering_controller.cpp
- [ ] Update Odometry.hpp and .cpp
- [ ] Update documents (remarks, descriptions, etc.)
- [ ] Add Test Code
- [ ] PR

Advice and questions are welcome. thank you.

Timple:
Be carefull with 'just' porting from ROS1.
For ROS1 ackermann, the implementation seemed [off](https://github.com/ros-controls/ros_controllers/issues/412).

Also support for front-wheel drive seems to be missing, but that might be a new feature request.

harderthan:
> For ROS1 ackermann, the implementation seemed off.  

@Timple  Reallly thank you for the feedback! I didn't know it.  

Do you know if there was any discussion about adding an Ackermann controller? I'm not sure how I should contribute. Does it make sense to read and implement all [this design_drafts](https://github.com/ros-controls/roadmap/tree/master/design_drafts)? 

And, I couldn't find the information about 'front-wheel drive'.
Do you mean `four_wheel_steering_controller`?

bmagyar:
@harderthan please don't create a separate repository with code copied. 
The way to contribute changes and get feedback is to first fork this repo and add a new package.

For porting of controllers you don't need a design draft. I agree with @Timple *however* I think the first step should be to do the porting of that controller as-is, then you can go and improve it. Doing too many things at once will only get you into a situation where it's impossible to trace your steps back and impossible to review as well.

Here's my suggestion for a chain of PRs once you have most things working (applicable to any of the controllers above):
[...]

PR #345, opened 2022-05-02
Tricycle controller

tonynajjar:
Hey! We at [Pixel Robotics](https://www.pixel-robotics.eu/) recently started our migration to ros2_control. We are in the process of creating a tricycle controller and decided to contribute back.


tricycle_controller/include/tricycle_controller/tricycle_controller.hpp
```diff
@@ -0,0 +1,189 @@
...
+#include "realtime_tools/realtime_buffer.h"
+#include "realtime_tools/realtime_publisher.h"
+#include "tf2_msgs/msg/tf_message.hpp"
+#include "tricycle_controller/odometry.hpp"
+#include "tricycle_controller/speed_limiter.hpp"
+#include "tricycle_controller/visibility_control.h"
+
+namespace tricycle_controller
+{
+using CallbackReturn = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn;
```
tonynajjar, line 48: needs to be changed to controller_interface::CallbackReturn but I'm getting this error:

`error: ‘CallbackReturn’ in namespace ‘controller_interface’ does not name a type.

`

tonynajjar: Actually I cloned ros2_control and ros2_controllers to master and when I try to build diff_drive_controller I get the same error. 
The rest of my system is on galactic, could it be related...

AndyZe:
Random question, out of curiosity:  does this work for (steered wheel in the front, 2 in the back) as well as (2 wheels in front, steered wheel in the back)?

tonynajjar:
Isn't what you described the same configuration turned around?

tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -0,0 +1,670 @@
...
+namespace
+{
+constexpr auto DEFAULT_COMMAND_TOPIC = "cmd_vel";
+constexpr auto DEFAULT_COMMAND_UNSTAMPED_TOPIC =
+  "cmd_vel";  // TODO: revert to cmd_vel_unstamped and remap topic
+constexpr auto DEFAULT_COMMAND_OUT_TOPIC = "cmd_vel_out";
+constexpr auto DEFAULT_ACKERMANN_OUT_TOPIC = "ackermann";
+constexpr auto DEFAULT_VALIDATED_ACKERMANN_TOPIC = "validated_ackermann";
+constexpr auto DEFAULT_ODOMETRY_TOPIC = "odom";
+constexpr auto DEFAULT_TRANSFORM_TOPIC = "tf";
```
tonynajjar, line 41: Is there a way to remap these topics from a launch file?

destogl: yes, check [here](https://github.com/ros-controls/ros2_control_demos/blob/c02942a892dde71f9e2df481b8007ef7f5e173fb/ros2_control_demo_bringup/launch/rrbot.launch.py#L57-L62) for the example

destogl: In a controller, I always recommend to use local topic , i.e. "~/<topic_name>" so you can remap them from "<controller_name>/<topic_name>" to anything you like

tonynajjar: Thanks fixed!

arne48 (commented):
Adding new controllers for new drive-train configurations should be always welcome.
For this repository there was a previous effort started to provide an Ackermann-Steering controller #288 #264.

And I am wondering because the "inverse" tricycle configuration with two driven wheels in the front would be a good candidate to use an ackermann-steering-controller.
So could the "tricycle" configuration (one driven wheel in the front) maybe just be considered as a specific wheel configuration of the same system at least from the kinematic point of view?

If my assumption is correct maybe a good next step could be to resurrect the efforts for the ackermann-steering-controller and giving it a wider choice of parameters so that your configuration would be covered as well.

My preference to make the tricycle controller a configuration for the general ackermann-steering-controller is based on the ackermann's versatility and that it is still missing for ROS2.

tonynajjar:
Hi @arne48. I do agree that the tricycle controller might be a special case of a more general ackermann controller that can be achieved with additional configuration. 
However, I think we could add a tricycle controller as a first step and see from there how to generalize it to ackermann. I am given the time to work on the tricycle configuration because this is what my company uses; I'm not sure I'll have time to work on an ackermann controller right now and I think it would be a shame to lose this initial effort, which can be easily extended.

arne48 (commented):
I agree with you that it would be a shame to let a good working controller go to waste.
Also it seems that the development/migration of the ackermann-steering-controller is currently stalled somehow.
And in this case I could see the merit of an easy to use specialized controller.
Especially for tricycles, which I think are likely to be a common drive train design for mobile robots besides differential and holonomic setups.

But this is finally a decision for the maintainers of the ros2_controllers package like @bmagyar  and if they want to add more specialized controllers and the thereby growing codebase for cases which could otherwise also be covered with more general controllers through configuration.

As you said during the absence of the ackermann-steering-controller this could be added as a temporary solution but if the ackermann-steering-controller should come, those temporary controllers and their replacement would cause problems in future releases where this controller then got replaced.
So once a controller got added - at least in my opinion - it should be a permanent part of the package.

So now before we would go into a more detailed code review I would like to wait for the opinion of the maintainers in regards to this topic.

Lets hope for a positive response :)

tonynajjar:
>  if the ackermann-steering-controller should come, those temporary controllers and there replacement would cause problems in future releases where this controller then got replaced.

I totally understand this point now. It's my first time contributing to a project where breaking changes affect many people so I didn't give it much thought :sweat_smile:.

tonynajjar:
While investigating on how to turn this into a more general-purpose ackermann controller I stumbled on http://wiki.ros.org/steer_bot_hardware_gazebo.

Their approach for navigating an ackermann vehicle was to generate tricycle commands with a tricycle controller (similar to the one in this PR) and transform these to ackermann commands in the hardware interface.

Do you consider the transformation tricycle->ackermann to be the responsibility of the controller or the hardware interface?

@bmagyar I'd appreciate guidance on how to continue with this PR.

tonynajjar:
Any chance of this PR going forward or should I simply close it? Frankly, I don't know if the maintainers are overwhelmed with work but I'd appreciate a simple answer if you want people to keep contributing.

arne48:
@bmagyar I think we should add this controller. As it seems that in the meatime the efforts for general Ackermann controller are currently are currently stagnant, so this controller could cover most use-cases and therefore has a stand-alone value.
Plus it is rather simple and shouldn't be too hard to maintain especially because I would guess @tonynajjar at Pixel Robotics will continue to use their controller and might have a look at  potential issues that arise.
So can we proceed to integrate this controller into the package?

tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -0,0 +1,629 @@
...
+
+bool TricycleController::reset()
+{
+  odometry_.resetOdometry();
+
+  // release the old queue
+  std::queue<TwistStamped> empty_twist;
+  std::swap(previous_commands_, empty_twist);
+
+  // TODO: clear handles
```
bmagyar, line 496: system alert: commented code found

tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -0,0 +1,629 @@
...
+  if (state_handle == state_interfaces_.cend())
+  {
+    RCLCPP_ERROR(
+      get_node()->get_logger(), "Unable to obtain joint state handle for %s",
+      steering_joint_name.c_str());
+    return CallbackReturn::ERROR;
+  }
+
+  // Lookup the velocity command interface
+  const auto command_handle = std::find_if(
```
bmagyar, line 580: this pattern repeats quite often. Could a helper function be of use here? Params would be the interfaces vector, name to look for, interface name (type) to look for -> returns std::optional< handle >

tricycle_controller/test/test_load_tricycle_controller.cpp
```diff
@@ -0,0 +1,44 @@
...
+  std::shared_ptr<rclcpp::Executor> executor =
+    std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+
+  controller_manager::ControllerManager cm(
+    std::make_unique<hardware_interface::ResourceManager>(
+      ros2_control_test_assets::minimal_robot_urdf),
+    executor, "test_controller_manager");
+
+  ASSERT_NO_THROW(cm.load_controller(
+    "test_tricycle_controller", "tricycle_controller/TricycleController"));
```
bmagyar, line 41: This is a good start but I think you forgot to add the file where you actually test controller behaviour

bmagyar (changes requested):

bmagyar:
Also please rebase this PR against the latest master branch

tonynajjar:
Thanks for the review @bmagyar. I synced to master and also added our latest changes which were on a different branch.

This PR still needs some work, not everything is well tested, now that I have your confirmation that this is the way forward, I can work on that. Will request another review once it's ready.

tricycle_controller/include/tricycle_controller/speed_limiter.hpp
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+#ifndef TRICYCLE_CONTROLLER__SPEED_LIMITER_HPP_
+#define TRICYCLE_CONTROLLER__SPEED_LIMITER_HPP_
+
+#include <cmath>
+
+namespace tricycle_controller
+{
+class SpeedLimiter
```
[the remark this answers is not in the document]
tonynajjar: I also felt it wrong to just copy it. I think it definitely makes sense to have some kind of generic rate limiter. This is similar: https://frankaemika.github.io/libfranka/rate__limiting_8h.html

tricycle_controller/src/tricycle_controller.cpp
tonynajjar: Is this important to you? I tried implementing the function you proposed, which is a good idea for reducing lines of codes, I just couldn't really make it compile, for reasons I don't understand. If it's important, I can spend some more time on it.

tricycle_controller/test/test_load_tricycle_controller.cpp
tonynajjar: Ah that was boilerplate, we forgot to write a test. Will do

tonynajjar:
You might have noticed that the rate limiters actually limit the steering angle and traction speed as opposed to the Twist (e.g in the diff drive controller). I think that Twist limiting should be done in the local planner/controller (e.g TEB does it). 
I implemented it like that mainly so that our gazebo simulation could be more realistic, e.g so that the steering angle change is slower. Rate limiting on our real motors is not needed as they have that feature built-in.
Happy to hear opinions.

tonynajjar:
Ready for review again, I made a demo also so you can try it out: https://github.com/ros-controls/gazebo_ros2_control/pull/145/files

tricycle_controller/test/test_load_tricycle_controller.cpp
tonynajjar: Added

tricycle_controller/src/tricycle_controller.cpp
bmagyar: It could be a follow-up issue to refactor this. I think other controllers could benefit from it too. The issue is probably the std::reference_wrapper not playing ball with other ref-style wrappers :) Please open an issue and reference it here, we can close the discussion then

tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -0,0 +1,629 @@
...
+  return state_interfaces_config;
+}
+
+controller_interface::return_type TricycleController::update(
+  const rclcpp::Time & /*time*/, const rclcpp::Duration & /*period*/)
+{
+  const auto current_time =
+    get_node()
+      ->get_clock()
+      ->now();  // original is current_time = time; time is always sim time when using gazebo_ros2_control. This is a hack to use system time instead.
```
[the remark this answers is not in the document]
bmagyar: Right. Please use ROS time (not sim ;)) in here for now

tricycle_controller/include/tricycle_controller/speed_limiter.hpp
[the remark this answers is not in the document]
bmagyar: Create an issue with a vague description of what we could move from this apcakge to control_toolbox and reference it here please

tricycle_controller/CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
...
+  rcpputils
+  realtime_tools
+  tf2
+  tf2_msgs
+)
+
+pluginlib_export_plugin_description_file(controller_interface tricycle_controller.xml)
+
+install(TARGETS ${PROJECT_NAME}
+
```
bmagyar, line 59:
```suggestion
```

tricycle_controller/src/tricycle_controller.cpp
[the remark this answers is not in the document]
tonynajjar: Ah right, this should already be done, you can take a look. I will mark this conversation as resolved, feel free to open it again if you have more comments there

tricycle_controller/include/tricycle_controller/speed_limiter.hpp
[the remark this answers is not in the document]
tonynajjar: https://github.com/ros-controls/ros2_controllers/issues/387

tricycle_controller/src/tricycle_controller.cpp
tonynajjar: I would very much like it in a follow-up issue, here it is: https://github.com/ros-controls/ros2_controllers/issues/388

tonynajjar:
@bmagyar is there anything in the tests I need to fix? I think the failures are not related to my changes

tonynajjar:
Hi, can I know what is still needed to move this forward? It's in the back of my head for a long time now.

bmagyar:
:ship: 

Thanks for sticking through the process!

bmagyar (approved):

tonynajjar:
@bmagyar thank you for the review and thank you @pixel-robotics (@jplapp) for financing the time put into this. Looking forward for community feedback and improving it

PR #346, opened 2022-05-02
Check for nans in JTC in command interface

mechwiz:
I noticed that there was an edge case in the JTC when partial goals are filled that nans weren't checked in the command interface (for those joints not specified in the goal) when assigning the last commanded joint value (assuming a command interface was specified). This actually occurred in a project I was working on and led to undefined behavior. This PR fixes this by implementing the check.

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -1064,7 +1064,8 @@ void JointTrajectoryController::fill_partial_goal(
         // Assume hold position with 0 velocity and acceleration for missing joints
         if (!it.positions.empty())
         {
-          if (has_position_command_interface_)
+          if (has_position_command_interface_ &&
+              !std::isnan(joint_command_interface_[0][index].get().get_value()))
           {
             // copy last command if cmd interface exists
             it.positions.push_back(joint_command_interface_[0][index].get().get_value());


bmagyar (approved):

AndyZe (approved):

AndyZe (commented):
Looking more closely at this, another part of the problem is that 
`last_commanded_state_` is never resized. Unlike the other variables that get used in JTC `update()` whose size gets updated for every iteration. [For example](https://github.com/ros-controls/ros2_controllers/blob/8f4dbaea1443499ebf5dc6a70eef5712f5db4b86/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L149).

`last_commanded_state_` is the only variable whose size doesn't get updated!

This actually relates to #340.

AndyZe:
@mechwiz can you test PR #340? I'll rebase it on this PR. You have a pretty unusual test case of using open-loop control mode plus partial goals.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1064,7 +1064,8 @@ void JointTrajectoryController::fill_partial_goal(
         // Assume hold position with 0 velocity and acceleration for missing joints
         if (!it.positions.empty())
         {
-          if (has_position_command_interface_)
+          if (has_position_command_interface_ &&
+              !std::isnan(joint_command_interface_[0][index].get().get_value()))
           {
             // copy last command if cmd interface exists
             it.positions.push_back(joint_command_interface_[0][index].get().get_value());
```
AndyZe, line 1071: Check 2 lines below this:

`else if (has_position_state_interface_)`

Should that be an `else if` or just an `if`? Seems like it shouldn't be dependent on L1067 at all.

AndyZe (changes requested):

joint_trajectory_controller/src/joint_trajectory_controller.cpp
AndyZe: Oh, I get it. L1073 is for the case where a joint is included but it doesn't have a position command interface.

AndyZe (approved):

joint_trajectory_controller/src/joint_trajectory_controller.cpp
mechwiz: I think it should still be an `else if` because I think the idea is that if there is a command interface for said joint and it's not a NAN, then assign that joint value to be the last commanded state. Otherwise if that's not the case then just use whatever the current value is as reported from the state interface. If it's an `if`, then if the joint value was already set via the command interface, it would get overwritten

mechwiz: lol, just refreshed after commenting and saw your second message. yep, you got it

PR #360, opened 2022-06-07
Fix exception about parameter already been declared & Change default c++ version to 17

JafarAbdi:
1- Make the default C++ version to 17 since rclcpp uses features that were introduced in that version https://github.com/ros2/rclcpp/search?q=variant
2- Use auto_declare for `multi_interface_forward_command_controller` and `forward_command_controller` to fix the following exception `Exception thrown during init stage with message: parameter 'joints' has already been declared `


mechwiz:
Same here, thanks for fixing this Jafar!

bmagyar (approved):
Great, thanks for the fix!

PR #365, opened 2022-06-16
Properly retrieve parameters in the Joint Trajectory Controller

### LINKED ISSUE #364: JointTrajectoryController does not load parameters correctly
The JTC loads parameters [like this](https://github.com/ros-controls/ros2_controllers/blob/801b708f58d0e63fab635d03f6bf4fc36e512a02/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L64):

`auto_declare<bool>("open_loop_control", open_loop_control_);`

The definition of auto_declare [is here](https://github.com/ros-controls/ros2_control/blob/643483355ba1c2c6c4d029eed7c7d3d6e3bcb116/controller_interface/include/controller_interface/controller_interface_base.hpp#L131):

```
  template <typename ParameterT>
  auto auto_declare(const std::string & name, const ParameterT & default_value)
  {
    if (!node_->has_parameter(name))
    {
      return node_->declare_parameter<ParameterT>(name, default_value);
    }
    else
    {
      return node_->get_parameter(name).get_value<ParameterT>();
    }
  }
```

I think the returned value is ignored.

AndyZe:
Fixes #364.

There are probably similar fixes that should be made for the other controller types.

I looked into adding the [[nodiscard]] attribute to `auto_declare` but it doesn't work with templates.  :(

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -55,17 +55,18 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
   try
   {
     // with the lifecycle node being initialized, we can declare parameters
-    auto_declare<std::vector<std::string>>("joints", joint_names_);
-    auto_declare<std::vector<std::string>>("command_interfaces", command_interface_types_);
-    auto_declare<std::vector<std::string>>("state_interfaces", state_interface_types_);
-    auto_declare<double>("state_publish_rate", 50.0);
-    auto_declare<double>("action_monitor_rate", 20.0);
-    auto_declare<bool>("allow_partial_joints_goal", allow_partial_joints_goal_);
-    auto_declare<bool>("open_loop_control", open_loop_control_);
-    auto_declare<bool>(
+    joint_names_ = auto_declare<std::vector<std::string>>("joints", joint_names_);
+    command_interface_types_ =
+      auto_declare<std::vector<std::string>>("command_interfaces", command_interface_types_);
+    state_interface_types_ =
+      auto_declare<std::vector<std::string>>("state_interfaces", state_interface_types_);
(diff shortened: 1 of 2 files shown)


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -55,17 +55,13 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
...
-    auto_declare<bool>(
+    joint_names_ = auto_declare<std::vector<std::string>>("joints", joint_names_);
+    command_interface_types_ = auto_declare<std::vector<std::string>>("command_interfaces", command_interface_types_);
+    state_interface_types_ = auto_declare<std::vector<std::string>>("state_interfaces", state_interface_types_);
+    allow_partial_joints_goal_ = auto_declare<bool>("allow_partial_joints_goal", allow_partial_joints_goal_);
+    open_loop_control_ = auto_declare<bool>("open_loop_control", open_loop_control_);
+    allow_integration_in_goal_trajectories_ = auto_declare<bool>(
       "allow_integration_in_goal_trajectories", allow_integration_in_goal_trajectories_);
-    auto_declare<double>("constraints.stopped_velocity_tolerance", 0.01);
-    auto_declare<double>("constraints.goal_time", 0.0);
```
AndyZe, line 68: As far as I can tell, constraints.goal_time isn't used anywhere. Same for `stopped_velocity_tolerance`, `state_publish_rate`, and `action_monitor_rate`.

mechwiz: yeah, these params I think are pretty much just used in the tolerances.hpp file https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp#L103

bmagyar (approved):

PR #367, opened 2022-06-17
Rename the "abort" variable in the joint traj controller

AndyZe:
I think this name was misleading. There could be many reasons to abort a trajectory, including an abort request from the action client.

Looking at Line 200, this variable specifically means that we should abort because a joint tolerance was violated while moving.


destogl (approved):
Definitely improvement of the code base! Thanks!

bmagyar (approved):

PR #368, opened 2022-06-21
[JTC] Hold position if tolerance is violated even during non-active goal

mechwiz:
This PR handles tolerance-checking logic for when non-active goals are running. This is specifically relevant if a trajectory is published to the joint_trajectory topic (instead of using the action server) as they are considered to be non-active goals.


destogl (commented):
Thanks for the contribution! This looks like a simple solution to this. Would it be possible to write a small test for this behavior? We would then feel better in the future to know that behavior is not broken (again).

erickisos (approved):

rosterloh (approved):

bmagyar (approved):

PR #374, opened 2022-07-01
Add an option to skip interpolation in the joint trajectory controller

AndyZe:
This should be useful if a user has a trajectory that is already interpolated to a fine degree, or if the low-level controllers just don't handle fine waypoint spacing well.

To use, just set the parameter in the yaml file. If the parameter is unset, it defaults to the previous method (which is spline-based). The default spline-based method doesn't obey joint limits (position/vel/accel/jerk), which is another reason to use this.

```
panda_arm_controller:
  ros__parameters:
    command_interfaces:
      - position
      - velocity
      - acceleration
    state_interfaces:
      - position
      - velocity
      - acceleration
    joints:
      - panda_joint1
      - panda_joint2
      - panda_joint3
      - panda_joint4
      - panda_joint5
      - panda_joint6
      - panda_joint7
    interpolation_method: none   # (or "splines")
    joint_limits:
      panda_joint1:
        has_velocity_limits: true
        max_velocity: 2.1750
        has_acceleration_limits: true
        max_acceleration: 3.75
      panda_joint2:
        ...
```

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
@@ -12,12 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef JOINT_TRAJECTORY_CONTROLLER_INTERPOLATION_METHODS_HPP_
-#define JOINT_TRAJECTORY_CONTROLLER_INTERPOLATION_METHODS_HPP_
+#ifndef JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
+#define JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
 
 namespace joint_trajectory_controller
 {
-enum class InterpolationMethod {NONE, SPLINE};
+enum class InterpolationMethod
+{
+  NONE,
+  SPLINE
+};
 }  // namespace joint_trajectory_controller
 
-#endif  // JOINT_TRAJECTORY_CONTROLLER_INTERPOLATION_METHODS_HPP_
+#endif  // JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
--- a/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
@@ -1,0 +1,23 @@
+// Copyright (c) 2022 ros2_control Development Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef JOINT_TRAJECTORY_CONTROLLER_INTERPOLATION_METHODS_HPP_
+#define JOINT_TRAJECTORY_CONTROLLER_INTERPOLATION_METHODS_HPP_
+
+namespace joint_trajectory_controller
+{
+enum class InterpolationMethod {NONE, SPLINE};
+}  // namespace joint_trajectory_controller
+
+#endif  // JOINT_TRAJECTORY_CONTROLLER_INTERPOLATION_METHODS_HPP_
--- a/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
@@ -15,9 +15,16 @@
 #ifndef JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
 #define JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
 
+#include <rclcpp/rclcpp.hpp>
 #include <string>
+#include <unordered_map>
 
 namespace joint_trajectory_controller
+{
+static const rclcpp::Logger LOGGER =
+  rclcpp::get_logger("joint_trajectory_controller.interpolation_methods");
+
+namespace interpolation_methods
 {
 enum class InterpolationMethod
 {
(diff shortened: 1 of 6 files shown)


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -67,6 +67,22 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
       "allow_integration_in_goal_trajectories", allow_integration_in_goal_trajectories_);
     state_publish_rate_ = auto_declare<double>("state_publish_rate", 50.0);
     action_monitor_rate_ = auto_declare<double>("action_monitor_rate", 20.0);
+
+    std::string interpolation_method = "";
+    interpolation_method = auto_declare<std::string>("interpolation_method", "splines");
+    if (interpolation_method == "none")
```
bmagyar, line 73: this if-else thing is basically a `from_string(... str)` method that could live alongside the class enum, helping with maintainability

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -121,12 +130,23 @@ bool Trajectory::sample(
...
-      // it changes points only if position and velocity are not exist, but their derivatives
-      deduce_from_derivatives(
-        point, next_point, state_before_traj_msg_.positions.size(), (t1 - t0).seconds());
-
-      interpolate_between_points(t0, point, t1, next_point, sample_time, output_state);
+      std::cout << "Somewhere prior to the first point" << std::endl;
+      // If interpolation is disabled, just forward the next waypoint
+      if (interpolation_method == joint_trajectory_controller::InterpolationMethod::NONE)
+      {
+        std::cout << "Traj pass-through" << std::endl;
```
bmagyar, line 137: debug print?

bmagyar (changes requested):

joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
+#define JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
+
+namespace joint_trajectory_controller
+{
+enum class InterpolationMethod
+{
+  NONE,
+  SPLINE
```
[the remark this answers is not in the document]
AndyZe: I guess I'll call what currently exists "VARIABLE_DEGREE_SPLINE"

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -88,22 +90,29 @@ bool Trajectory::sample(
...
-    // it changes points only if position and velocity are not exist, but their derivatives
-    deduce_from_derivatives(
-      state_before_traj_msg_, first_point_in_msg, state_before_traj_msg_.positions.size(),
-      (first_point_timestamp - time_before_traj_msg_).seconds());
-
-    interpolate_between_points(
-      time_before_traj_msg_, state_before_traj_msg_, first_point_timestamp, first_point_in_msg,
-      sample_time, output_state);
+    // If interpolation is disabled, just forward the next waypoint
+    if (interpolation_method == joint_trajectory_controller::InterpolationMethod::NONE)
```
destogl, line 101: Shouldn't we check this before calling sample in the `update` method?

What I understand, we will anyway have also ruckig based interpolation this means that we would need some base class for interpolations. If so we can then just decide at the begining in the configure which interpolation should be used and if non is used then we have default "passthrough" interpolation.

destogl: this should also reduce number of checks.

destogl (changes requested):

joint_trajectory_controller/src/trajectory.cpp
AndyZe: I don't think Ruckig is going to be a drop-in replacement for the spline interpolation that's currently there. It requires position/vel/accel inputs. What if there is only a `position` interface active? Then Ruckig will not have vel/accel input. So we will often still need to run some type of interpolation before Ruckig.

I'm trying to figure out the logic re. your comment on `update`...

bmagyar (approved):
Thanks!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -67,6 +67,9 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
       "allow_integration_in_goal_trajectories", allow_integration_in_goal_trajectories_);
     state_publish_rate_ = auto_declare<double>("state_publish_rate", 50.0);
     action_monitor_rate_ = auto_declare<double>("action_monitor_rate", 20.0);
+
+    std::string interpolation_string = auto_declare<std::string>("interpolation_method", "splines");
```
destogl, line 71: Since these strings are managed by the enum method. it would make sense to define those there so that everything is in one place. Now we have here string definition for parameters in one file and checking the same strings in other file. You could do 2 things, add const expr for each string and also add the full list of those so that all can be simple accessed.

joint_trajectory_controller/test/test_trajectory.cpp
```diff
@@ -31,8 +31,13 @@
 
 using namespace std::chrono_literals;
 
+namespace
+{
 // Floating-point value comparison threshold
 const double EPS = 1e-8;
+const joint_trajectory_controller::InterpolationMethod DEFAULT_INTERPOLATION =
```
destogl, line 38: This should be defined also in Enum class. So everything is in one place.

destogl (changes requested):
Few comments to increase maintainability and clarity in the code. Sorry, guys if I am starting to be too picky, and please tell me if I am drifting away in my requests/concerns.

joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
```diff
@@ -0,0 +1,46 @@
...
+    return InterpolationMethod::NONE;
+  }
+  else if (!interpolation_method.compare("splines") == 0)
+  {
+    return InterpolationMethod::VARIABLE_DEGREE_SPLINE;
+  }
+  // Default
+  else
+  {
+    return InterpolationMethod::VARIABLE_DEGREE_SPLINE;
```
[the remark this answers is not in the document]
bmagyar: Or just use a logger assigned to this file?

joint_trajectory_controller/include/joint_trajectory_controller/interpolation_methods.hpp
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+#ifndef JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
+#define JOINT_TRAJECTORY_CONTROLLER__INTERPOLATION_METHODS_HPP_
+
+#include <string>
+
+namespace joint_trajectory_controller
+{
+enum class InterpolationMethod
```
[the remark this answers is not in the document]
bmagyar: IMO it's better in a separate file ¯\_(ツ)_/¯

joint_trajectory_controller/src/joint_trajectory_controller.cpp
AndyZe: I don't completely understand. Shall I do something like this?

https://github.com/ros-planning/moveit2/blob/33b075d15f8d9ec64ca8fea98cf1573b7c581c47/moveit_ros/moveit_servo/include/moveit_servo/status_codes.h#L46

destogl: Yes, please. So, the strings should be somewhere defined as constants to avoid unintentional error is multiple instances of those get misaligned.

AndyZe: OK, done.

destogl (approved):
Thanks for all the follow-ups!

Issue #376, opened 2022-07-04
realtime_tools isn't an export dependency from joint_state_broadcaster


GeorgeVe:
**Describe the bug**
The joint_state_controller doesn't build when building the `joint_state_controller`, `joint_state_broadcaster` and realtime_tools from source code. (in foxy, not sure about other distributions)

**To Reproduce**
Clone `ros2_controllers` package
Clone `realtime_tools` package
Use `colcon build` 
-> It will fail with message: 

`In file included from .../ros2_controllers/joint_state_controller/src/joint_state_controller.cpp:15:
.../install/joint_state_broadcaster/include/joint_state_broadcaster/joint_state_broadcaster.hpp:28:10: fatal error: realtime_tools/realtime_publisher.h: No such file or directory
   28 | #include "realtime_tools/realtime_publisher.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.`

**Expected behavior**
It should build, and find the realtime_tools package. 

**Additional context**
Note that fi you do `colcon build --packages-skip joint_state_controller` it does build. Meaning that `joint_state_broadcaster` on itself does build (and thus can find the realtime_publisher.h file). Only when building the `joint_state_controller` that requires the `joint_state_broadcaster` it can't find the package `realtime_tools`.

PR #377, opened 2022-07-04
Add realtime_tools as export dependency

GeorgeVe:
Fix for #376 

**Describe the bug**
The joint_state_controller doesn't build when building the joint_state_controller, joint_state_broadcaster and realtime_tools from source code. (in foxy, not sure about other distributions)

**To Reproduce**
Clone ros2_controllers package
Clone realtime_tools package
Use colcon build
-> It will fail with message:

In file included from .../ros2_controllers/joint_state_controller/src/joint_state_controller.cpp:15: .../install/joint_state_broadcaster/include/joint_state_broadcaster/joint_state_broadcaster.hpp:28:10: fatal error: realtime_tools/realtime_publisher.h: No such file or directory 28 | #include "realtime_tools/realtime_publisher.h" | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ compilation terminated.

**Expected behavior**
It should build, and find the realtime_tools package.

**Additional context**
Note that fi you do colcon build --packages-skip joint_state_controller it does build. Meaning that joint_state_broadcaster on itself does build (and thus can find the realtime_publisher.h file). Only when building the joint_state_controller that requires the joint_state_broadcaster it can't find the package realtime_tools.

### DIFF
--- a/joint_state_broadcaster/CMakeLists.txt
+++ b/joint_state_broadcaster/CMakeLists.txt
@@ -88,6 +88,7 @@ ament_export_dependencies(
   controller_interface
   rclcpp_lifecycle
   sensor_msgs
+  realtime_tools
 )
 ament_export_include_directories(
   include


bmagyar:
The PR itself looks ok but what I don't get is:
* the buildfarm didn't complain, seems to build fine there
* the source-based CI fails on *something* related to the joint_state_broadcaster but not building it. It is a test that fails but unfortunately the output was not available, I'm re-running that stage in hopes of getting some feedback

Anything unusual you guys do on Foxy that this happens?

bmagyar:
Yeah probably the build order & build dependencies. Thanks for the PR, merging...

PR #382, opened 2022-07-11
Add admittance controller with full changes overview

destogl:
Replaces #173.

#370 should be merged into it. All changes to Joint Trajectory Controller should be moved from here.


AndyZe:
I wouldn't mind if you squash a bunch of my commits together. 99 seems excessive   ;)

PR #384, opened 2022-07-21
Generate Parameter Library for Joint Trajectory Controller

tylerjw:
Signed-off-by: Tyler Weaver <maybe@tylerjw.dev>

This is an PR to converting the parameter usage in JTC to use https://github.com/PickNikRobotics/generate_parameter_library


joint_trajectory_controller/package.xml
```diff
@@ -14,6 +14,7 @@
   <build_depend>angles</build_depend>
   <build_depend>pluginlib</build_depend>
   <build_depend>realtime_tools</build_depend>
+  <build_depend>generate_parameter_library</build_depend>
```
AndyZe, line 17: Alphabetize

joint_trajectory_controller/include/joint_trajectory_controller/validate_joint_trajectory_controller_parameters.hpp
```diff
@@ -0,0 +1,55 @@
+Result command_interface_type_combinations(rclcpp::Parameter const& parameter) {
```
AndyZe, line 1: Copyright needed

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -0,0 +1,73 @@
...
+    validation: {
+      lower_bounds: [0.1]
+    }
+  }
+  interpolation_method: {
+    type: string,
+    default_value: "splines",
+    description: "The type of interopoation to use if at all",
+    validation: {
+      one_of<>: [["splines", "none"]],
```
AndyZe, line 71: This is nice

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -0,0 +1,73 @@
+joint_trajectory_controller:
+  joints: {
+    type: string_array,
+    default_value: [],
+    description: "Names of joints used by the controller",
```
AndyZe, line 5: So is this going to replace the list of interfaces in the URDF? I'm not sure about that.

It's nice having it in the URDF so it applies to all controllers, not just the JTC. It looks like what you have here will need to be duplicated a lot.

tylerjw: No, this just replaces the YAML parameter declaration and getting. This does not change anything about how interfaces from the URDF work.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
tylerjw: With this validation, if you accidentially set this value to something else you'll get a helpful error message that says something like this:

```Parameter failed validation: 'bla' set for 'interpolation_method' is not in the set ['splines', 'none']```

Comment (AndyZe) [reply joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml:71]: OK. Well, that case is already covered and it even allows for a default.

```
[[nodiscard]] inline InterpolationMethod from_string(const std::string & interpolation_method)
{
  if (interpolation_method.compare(InterpolationMethodMap.at(InterpolationMethod::NONE)) == 0)
  {
    return InterpolationMethod::NONE;
  }
  else if (
    !interpolation_method.compare(
      InterpolationMethodMap.at(InterpolationMethod::VARIABLE_DEGREE_SPLINE)) == 0)
  {
    return InterpolationMethod::VARIABLE_DEGREE_SPLINE;
  }
  // Default
  else
  {
    RCLCPP_INFO_STREAM(
      LOGGER,
      "No interpolation method parameter was given. Using the default, VARIABLE_DEGREE_SPLINE.");
    return InterpolationMethod::VARIABLE_DEGREE_SPLINE;
  }
}
```

File: joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+    default_value: 20.0,
+    description: "Rate status changes will be monitored",
+    validation: {
+      lower_bounds: [0.1]
+    }
+  }
+  interpolation_method: {
+    type: string,
+    default_value: "splines",
+    description: "The type of interopoation to use if at all",
```
AndyZe, line 69:
```suggestion
    description: "The type of interpolation to use, if any",
```

AndyZe (commented):
Overall I think this looks like an improvement. There are more changes that will need to be made, though, like making `dof_` constant in `trajectory.cpp`.

joint_trajectory_controller/include/joint_trajectory_controller/validate_joint_trajectory_controller_parameters.hpp
tylerjw: Done

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
tylerjw: I think I should be able to remove that code from the `from_string` as with this you won't ever be able to get to the code path where it is invalid.

tylerjw:
> Overall I think this looks like an improvement. There are more changes that will need to be made, though, like making `dof_` constant in `trajectory.cpp`.

Can you help me understand what you mean by this by pointing to the code you want me to update?

joint_trajectory_controller/package.xml
tylerjw: Done

tylerjw:
> I think we should talk about in the ros2_control WG meeting. Basically `dof_` should become constant (I think). That's going to change a bunch of other stuff in good ways. Then this PR can be applied on top of it.

https://github.com/ros-controls/ros2_controllers/issues/394

tylerjw:
This passes all CI jobs but the ABI compatibility checks one which doesn't use a repos file so I wasn't able to add it as a source dependency. We'll bloom the package but it won't be ready for use without a repos file until the next sync.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -0,0 +1,108 @@
...
+    type: string_array,
+    default_value: [],
+    description: "Names of joints used by the controller",
+  }
+  command_interfaces: {
+    type: string_array,
+    default_value: [],
+    description: "Names of command interfaces to claim",
+    validation: {
+      unique<>: null,
```
destogl, line 12:
```suggestion
      unique<>: null,
      not_empty<>: null,
```

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -0,0 +1,109 @@
+joint_trajectory_controller:
+  joints: {
+    type: string_array,
+    default_value: [],
+    description: "Names of joints used by the controller",
```
destogl, line 5:
```suggestion
    description: "Names of joints used by the controller",
    validation: {
      unique<>: null,
      not_empty<>: null,
    } 
```

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -0,0 +1,113 @@
...
+    validation: {
+      unique<>: null,
+      not_empty<>: null,
```
destogl, line 8:
```suggestion
```

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -0,0 +1,113 @@
...
+    } 
+  }
+  command_interfaces: {
+    type: string_array,
+    default_value: [],
+    description: "Names of command interfaces to claim",
+    validation: {
+      unique<>: null,
+      not_empty<>: null,
```
destogl, line 17:
```suggestion
```

joint_trajectory_controller/CMakeLists.txt
```diff
@@ -29,12 +29,24 @@ foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
...
   src/trajectory.cpp
 )
 target_include_directories(${PROJECT_NAME} PRIVATE include)
-ament_target_dependencies(${PROJECT_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
+ament_target_dependencies(${PROJECT_NAME}
+  ${THIS_PACKAGE_INCLUDE_DEPENDS}
+)
+target_link_libraries(${PROJECT_NAME}
+  joint_trajectory_controller_parameters
+)
```
destogl, line 49:
```suggestion
target_link_libraries(${PROJECT_NAME} joint_trajectory_controller_parameters)
ament_target_dependencies(${PROJECT_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -46,32 +46,20 @@
...
-    std::string interpolation_string = auto_declare<std::string>(
-      "interpolation_method", interpolation_methods::InterpolationMethodMap.at(
-                                interpolation_methods::DEFAULT_INTERPOLATION));
-    interpolation_method_ = interpolation_methods::from_string(interpolation_string);
+    // Create the parameter listener and get the parameters
+    param_listener_ = std::make_shared<ParamListener>(get_node());
+    params_ = param_listener_->get_params();
+
+    // Set interpolation method from string parameter
+    interpolation_method_ = interpolation_methods::from_string(params_.interpolation_method);
```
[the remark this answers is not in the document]
destogl: Should be moved to the configure.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -98,10 +84,10 @@ JointTrajectoryController::command_interface_configuration() const
       dof_);
     std::exit(EXIT_FAILURE);
   }
-  conf.names.reserve(dof_ * command_interface_types_.size());
-  for (const auto & joint_name : command_joint_names_)
+  conf.names.reserve(dof_ * params_.command_interfaces.size());
+  for (const auto & joint_name : params_.joints)
```
destogl, line 88:
```suggestion
  for (const auto & joint_name : command_joint_name_)
```

bmagyar (approved):

PR #385, opened 2022-07-21
Parameter loading fixup in diff_drive and gripper controllers

AndyZe:
A follow-up from #365. I think this fixes it everywhere


bmagyar (approved):
thanks!

Issue #392, opened 2022-07-26
Dynamic memory allocation in the joint trajectory controller


AndyZe:
Here are some spots of dynamic memory allocation in the realtime `update` method of the JTC we should try to get rid of:

In `fill_partial_goal()`:
https://github.com/ros-controls/ros2_controllers/blob/d8d661607f770e704b774249a31534fa0eb0d46b/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1084

And several other places.

In `sort_to_local_joint_order()`:

`output.resize(mapping.size(), 0.0);`

Theonewhomadethings:
@christophfroehlich could I tackle this?

PR #393, opened 2022-07-27
Small improvement in remapping

AndyZe:
Resizing this output vector with every iteration isn't necessary.

The idea here is, make the variable `static` so it retains its value between iterations. Only remap if the dimension of the vector changes.

Partial fix to #392


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1136,8 +1136,12 @@ void JointTrajectoryController::sort_to_local_joint_order(
         get_node()->get_logger(), "Invalid input size (%zu) for sorting", to_remap.size());
       return to_remap;
     }
-    std::vector<double> output;
-    output.resize(mapping.size(), 0.0);
+    static std::vector<double> output(dof_, 0.0);
```
AndyZe, line 1139: If it's safe to assume `dof_` never changes, this can get simpler.

`dof_` changes here:

```
controller_interface::CallbackReturn JointTrajectoryController::on_configure(
{
  const auto logger = get_node()->get_logger();

  // update parameters
  joint_names_ = get_node()->get_parameter("joints").as_string_array();
  if ((dof_ > 0) && (joint_names_.size() != dof_))
  {
    RCLCPP_ERROR(
      logger,
      "The JointTrajectoryController does not support restarting with a different number of DOF");
    // TODO(andyz): update vector lengths if num. joints did change and re-initialize them so we
    // can continue
  }

  dof_ = joint_names_.size();
```

So it could hypothetically change but that would be really unusual

AndyZe: I think it doesn't make any sense to allow `dof_` to change since the yaml config file contains a list of joint names to control. Example config file here:  https://github.com/ros-planning/moveit_resources/blob/ros2/panda_moveit_config/config/ros2_controllers.yaml

Agree, @bmagyar or @bmagyar ?

destogl (approved):
I agree, although here it is not necessary to have great performance since this is (or at least should be) a non-realtime method.

AndyZe:
well, this method is actually called in the `update` loop. I'll probably end up closing this PR if `dof_` becomes constant because it won't be relevant anymore.

PR #414, opened 2022-08-11
Add parameter to disable closed loop PID adapter manually

destogl:
Enable disabling closed-loop PID in JTC when velocity-only interfaces are used.

This is useful in cases where only velocity output should be used but closed-loop PID should not be used. 
For example when there are no state interfaces available or cotroller is used in a chain and state interfaces can not be directly used for calculation.

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -60,6 +60,8 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
       auto_declare<std::vector<std::string>>("command_interfaces", command_interface_types_);
     state_interface_types_ =
       auto_declare<std::vector<std::string>>("state_interfaces", state_interface_types_);
+    disable_closed_loop_pid_adapter_ =
+      auto_declare<bool>("disable_closed_loop_pid_adapter", disable_closed_loop_pid_adapter_);
     allow_partial_joints_goal_ =
       auto_declare<bool>("allow_partial_joints_goal", allow_partial_joints_goal_);
     open_loop_control_ = auto_declare<bool>("open_loop_control", open_loop_control_);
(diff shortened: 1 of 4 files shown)


AndyZe:
I think it would help a lot if this stuff were documented somewhere. Got a link?

I'm a bit hazy on how it's all supposed to work, so this is hard to review.

Does the closed-loop PID adapter convert position commands to effort?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -285,14 +287,10 @@ controller_interface::return_type JointTrajectoryController::update(
         }
         if (has_effort_command_interface_)
         {
-          if (use_closed_loop_pid_adapter_)
+          if (use_closed_loop_pid_adapter_)  // this should be always set to true when using effort
```
AndyZe, line 290: I think a table of which combinations of these things are valid would be great. It's hard to follow with comments scattered around the code. Or I guess you could use a std::map of which combinations are valid.

- has_position_command_interface_
- has_velocity_command_interface_
- has_effort_command_interface_
- use_closed_loop_pid_adapter_
          
 Maybe you should add these checks:
 
```
 if (has_effort_command_interface_ && !use_closed_loop_pid_adapter_)
   // error
if (has_effort_command_interface_ && !has_position_command_interface)
  // error
if (has_effort_command_interface_ && has_velocity_interface_)
  // error
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -176,6 +176,58 @@ TEST_P(TrajectoryControllerTestParameterized, check_interface_names_with_command
...
+TEST_P(TrajectoryControllerTestParameterized, check_use_of_closed_loop_pid_adapter)
+{
+  SetUpTrajectoryController();
+
+  const auto state = traj_controller_->get_node()->configure();
+  ASSERT_EQ(state.id(), State::PRIMARY_STATE_INACTIVE);
+
+  EXPECT_FALSE(traj_controller_->get_disable_closed_loop_pid_adapter());  // default value
+
+  // closed loop PID adapter will be use when velocity- or effort-only interfaces are used
```
AndyZe, line 188:
```suggestion
  // closed loop PID adapter will be used when velocity- or effort-only interfaces are used
```

I think another way to say this is, "The joint trajectory controller always operates in the position domain. If velocity-only or effort-only interfaces are requested, a PID controller is used to convert position to velocity or effort."

PR #415, opened 2022-08-15
test: :white_check_mark: fix and add back joint_trajectory_controller state_topic_consistency

jaron-l:
- fixes and uncomments: TrajectoryControllerTestParameterized.state_topic_consistency

There are two main differences from how the test was before.

1. at line 378-379 in my proposed branch, I change the check to check that the desired position is at the initial position rather than at 0s
2. at line 405-408 in my proposed branch, I check that the error is 0 rather than being empty

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_controller.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller.cpp
@@ -375,10 +375,10 @@ TEST_P(TrajectoryControllerTestParameterized, state_topic_consistency)
   }
 
   // No trajectory by default, no desired state or error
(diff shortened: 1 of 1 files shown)


jaron-l:
This is part of fixing #132

bmagyar (approved):
Thanks!

bmagyar:
Unfortunately the CI doesn't seem to angry

jaron-l:
> Unfortunately the CI doesn't seem to agree

Yeah, I'll mark it draft until I figure it out.

bmagyar:
Awesome. Thanks!!

PR #420, opened 2022-08-17
build: :construction_worker: remove foxy rhel build

jaron-l:
- there is no foxy rhel docker image because it osrf doesn't publish rhel packages for foxy


jaron-l:
For documentation/proof, see [the ros release/target schedule](https://ros.org/reps/rep-2000.html#id40)

destogl (approved):

PR #422, opened 2022-08-18
fix: :bug: make bare exceptions more narrow

jaron-l:
- fixes pre-commit pipeline

For `joint_limits_urdf.py`, I was able to determine the exception they were looking for and put those but for `joint_trajectory_controller.py` it was unclear to me if they were just trying to do a general exception or were looking for something specific and just didn't put it. For the second case, the fix is still better because it will ignore exceptions not based on `Exception` (ex. SigInt) and fail those how one would expect. If anyone knows what exactly the excepts are looking for or if it was just a general catch, feel free to comment what your know, especially @Noel215 since you ported it (though I understand if you don't know either).


destogl (approved):
Thank you very much for this! :heart:

bmagyar (approved):
Looks good to me too. I previously tried to convert some of them into ifs but that still doesn't protect against the index error, etc really not happy with this exception-driven-development approach here but unfortunately it's fairly common w XML parsing :/

That being said, let's merge this for now

PR #425, opened 2022-08-25
Add an initialization of the gripper action command to load current joint position.

ShotaAk:
The gripper action controller is missing an initialization of the `command_` instance to load current joint position.
This causes that the controller moves focely a joint to 0 position at startup.

This PR adds an initialization to fix it.

This PR solves same problem of https://github.com/ros-controls/ros2_controllers/pull/311 with differing approach.

Reference: the init implementation of ROS 1 gripper action controller:
https://github.com/ros-controls/ros_controllers/blob/252d6584b17ae88e614a6c9cd30ba76ef258a193/gripper_action_controller/include/gripper_action_controller/gripper_action_controller_impl.h#L91-L97


bmagyar (approved):
Happy with this, waiting for CI

PR #426, opened 2022-08-26
improved mathematics of tricycle controller to avoid infinity or close to infinity state 

ros2torial:
the updated mathematics equation avoid denominator to be 0 or close to 0. Also it provides more accurate results.

### DIFF
--- a/tricycle_controller/include/tricycle_controller/tricycle_controller.hpp
+++ b/tricycle_controller/include/tricycle_controller/tricycle_controller.hpp
@@ -106,8 +106,8 @@ protected:
     const std::string & traction_joint_name, std::vector<TractionHandle> & joint);
   CallbackReturn get_steering(
     const std::string & steering_joint_name, std::vector<SteeringHandle> & joint);
-  double convert_trans_rot_vel_to_steering_angle(double v, double omega, double wheelbase);
-  std::tuple<double, double> twist_to_ackermann(double linear_command, double angular_command);
+  //double convert_trans_rot_vel_to_steering_angle(double v, double omega, double wheelbase);
+  std::tuple<double, double> twist_to_tricycle(double linear_command, double angular_command);
 
   std::string traction_joint_name_;
   std::string steering_joint_name_;
--- a/tricycle_controller/include/tricycle_controller/tricycle_controller.hpp
+++ b/tricycle_controller/include/tricycle_controller/tricycle_controller.hpp
@@ -106,7 +106,6 @@ protected:
     const std::string & traction_joint_name, std::vector<TractionHandle> & joint);
   CallbackReturn get_steering(
     const std::string & steering_joint_name, std::vector<SteeringHandle> & joint);
-  //double convert_trans_rot_vel_to_steering_angle(double v, double omega, double wheelbase);
   std::tuple<double, double> twist_to_tricycle(double linear_command, double angular_command);
 
   std::string traction_joint_name_;
--- a/tricycle_controller/src/tricycle_controller.cpp
+++ b/tricycle_controller/src/tricycle_controller.cpp
@@ -206,7 +206,7 @@ controller_interface::return_type TricycleController::update(
   }
 
   // Compute wheel velocity and angle
-  auto [alpha_write, Ws_write] = twist_to_ackermann(linear_command, angular_command);
+  auto [alpha_write, Ws_write] = twist_to_tricycle(linear_command, angular_command);
 
   // Reduce wheel speed until the target angle has been reached
   double alpha_delta = abs(alpha_write - alpha_read);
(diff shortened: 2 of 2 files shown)


bmagyar:
FYI @tonynajjar

tricycle_controller/include/tricycle_controller/tricycle_controller.hpp
```diff
@@ -106,8 +106,8 @@ class TricycleController : public controller_interface::ControllerInterface
     const std::string & traction_joint_name, std::vector<TractionHandle> & joint);
   CallbackReturn get_steering(
     const std::string & steering_joint_name, std::vector<SteeringHandle> & joint);
-  double convert_trans_rot_vel_to_steering_angle(double v, double omega, double wheelbase);
-  std::tuple<double, double> twist_to_ackermann(double linear_command, double angular_command);
+  //double convert_trans_rot_vel_to_steering_angle(double v, double omega, double wheelbase);
```
bmagyar, line 109: please don't leave commented code in PRs, we use version control software to be able to recover old code

tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -637,31 +637,29 @@ CallbackReturn TricycleController::get_steering(
...
   // using naming convention in http://users.isr.ist.utl.pt/~mir/cadeiras/robmovel/Kinematics.pdf
-  double alpha, Ws;
-
-  if (Vx == 0 && theta_dot != 0)
-  {  // is spin action
-    alpha = theta_dot > 0 ? M_PI_2 : -M_PI_2;
-    Ws = abs(theta_dot) * wheel_params_.wheelbase / wheel_params_.radius;
-    return std::make_tuple(alpha, Ws);
-  }
+  double traction_velocity = sqrt(pow((angular_command * wheel_params_.wheelbase),2) + pow((linear_command),2)) / wheel_params_.radius;
```
bmagyar, line 653: in comment, there's a reference to the nomenclature used in the "old math"

please do something similar for the "new math" so we don't need to reverse-compile code to math

tonynajjar: @ros2torial any reason why you want to change the nomenclature in the first place? Is it not clear enough with the documentation provided?

tonynajjar:
> FYI @tonynajjar

@bmagyar thanks, Is there a way I can get pinged automatically for PRs for the tricycle controller? (If you would like me to help out review them of course)

tonynajjar (commented):
> improved mathematics of tricycle controller to avoid infinity or close to infinity state

In which cases does an inifinity state happen?

tricycle_controller/src/tricycle_controller.cpp
ros2torial: @bmagyar @tonynajjar I have kept **linear_command** and **angular_command** name as it is easy to understand and same nomenclature is used in diff_drive_controller. Regarding alpha, I have changed its name to **steering_wheel_angle** in the latest commit as it is also easy to understand and this name is as per the nomenclature only (_If the **steering wheel** is set to an **angle** α(t) from the straight-line direction_).  
Let me know if the new variable names are ok else we can keep the previous names.

ros2torial:
@tonynajjar in the steering wheel angle calculation, atan can cause issue when the denominator term is close to 0 or 0. That's why I changed it to atan2 as it takes care of everything.  Regarding Ws calculation if we keep cos in denominator then it will also cause similar issue, that's why i have used another equation which also calculates Ws without having any term in denominator.

tonynajjar:
> atan can cause issue when the denominator term is close to 0 or 0

I agree that the denominator cannot be 0, that's why we check for it before. (in all cases except for `wheel_params_.radius` which we can reasonable assume is not 0).

On the other hand when the denominator is close to 0, then the ratio very big, which is fine because atan tends to PI/2. Can you be more specific on what issues atan causes?

Also, the output range of atan2 is [-180, 180] whereas the one from atan  is [-90, 90]. According to the conventions this code follows (referenced in the code), we want the latter range


> Regarding Ws calculation if we keep cos in denominator then it will also cause similar issue

The denominator in the Ws calculation cannot be exactly 0 (again, assuming the radius is not 0) because alpha cannot  be exactly PI/2 or -PI/2 because theta_dot and Vx are checked not to be 0.


> that's why i have used another equation which also calculates Ws without having any term in denominator.

You don't mention where that equation comes from and I can't really make any sense of it.


It would help if you can say what kind of issues on a higher level you are facing (e.g code crash, very high speeds, wrong angles, etc..) then maybe we can help better

ros2torial:
@tonynajjar 
> I agree that the denominator cannot be 0, that's why we check for it before. 

yes, if we use atan then we need to check for the denominator = 0 condition which you have done.

> On the other hand when the denominator is close to 0, then the ratio very big, which is fine because atan tends to PI/2. Can you be more specific on what issues atan causes?

I checked that atan and atan2 are giving same result at normal input. When denominator is close to 0 there is very slight difference from 14th digit after decimal which is negligible. Although atan is taking 3-10X more computation time than atan2.

```cpp
#include <iostream>
#include <cmath>
#include <chrono>
#include <iomanip>
using namespace std;

int main() {

  double a,b,x,y;
  x = 1.0;
  y = 000000000000000000.1;
  std::cout << std::fixed;
  std::cout << std::setprecision(20);
  auto start = std::chrono::high_resolution_clock::now();
  a = atan(x/y);
  auto finish = std::chrono::high_resolution_clock::now();
  std::cout << std::chrono::duration_cast<std::chrono::nanoseconds>(finish-start).count() << "ns\n";
  std::cout << a << "\n";
  
  auto start2 = std::chrono::high_resolution_clock::now();
  b = atan2(x,y);
  auto finish2 = std::chrono::high_resolution_clock::now();
  std::cout << std::chrono::duration_cast<std::chrono::nanoseconds>(finish2-start2).count() << "ns\n";
  std::cout << b << "\n";

  return 0;
}
```

> Also, the output range of atan2 is [-180, 180] whereas the one from atan is [-90, 90]. According to the conventions this code follows (referenced in the code), we want the latter range

to keep the atan2 in [-90, 90] range i have already included the below conditions

```cpp
if(steering_wheel_angle > M_PI_2)
  {steering_wheel_angle = steering_wheel_angle - M_PI;}
if(steering_wheel_angle < -M_PI_2)
  {steering_wheel_angle = steering_wheel_angle + M_PI;}
```

ros2torial:
@tonynajjar 

> You don't mention where that equation comes from and I can't really make any sense of it.

From section **Kinematic model in the robot frame** of http://users.isr.ist.utl.pt/~mir/cadeiras/robmovel/Kinematics.pdf 
if we simplify the equation by using **sin^2 + cos^2 = 1** then we will get the equation which i have used and this new equation is independent of steering angle.

> It would help if you can say what kind of issues on a higher level you are facing

there is no issue in using the equation which you have used as you have taken care of denominator = 0 condition. The use of atan2 will reduce computation time and as far as i know it is recommended.

tonynajjar:
I don't know what's wrong with your timing code but swapping the order showsatan2 as the faster option. See [here](https://onecompiler.com/cpp/3yek935f9)

tonynajjar:
It seems like your timing code is doing weird things, I just switched the order and now atan2 is slower: https://onecompiler.com/cpp/3yekem5gd

tonynajjar:
In general I think this PR doesn't bring much value, neither in performance nor in clarity/simplification (on the contrary in this regard). @bmagyar leaving it to you

ros2torial:
@tonynajjar I agree. there is some problem in the timing code. overall atan2 is slower than atan.

PR #427, opened 2022-08-29
Use a "steady clock" when measuring time differences

AndyZe:
I've found out recently the difference between the default rclcpp clock (a `system clock`) versus a `steady clock`. The steady clock is monotonically increasing, meaning it can't go backward in time. So it is more suitable for measuring time differences.

Making this type of change fixed the flakiness of some of our MoveIt CI tests, so I hope it helps here.

Reference reading:  https://design.ros2.org/articles/clock_and_time.html#time-abstractions


bmagyar (approved):
Very good catch, thanks!!

PR #428, opened 2022-09-01
Fix for high CPU usage by JTC in gzserver

arshadlab:
* Change type cast wall timer period from second to nanoseconds.
    create_wall_timer() expects delay in nanoseconds (duration object) however the type cast to seconds will result in 0 (if duration is less than 1s) and thus causing timer to be fired non stop resulting in very high CPU usage.

* Reset smartpointer so that create_wall_timer() call can destroy previous trajectory timer.
   node->create_wall_timer() first removes timers associated with expired smartpointers before servicing current request.  The JTC timer pointer gets overwrite only after the create_wall_timer() returns and thus not able to remove previous trajectory timer resulting in upto two timers running for JTC during trajectory execution.  Althougth the previous timer does nothing but still get fired.

Signed-off-by: Arshad Mehmood <arshadm78@yahoo.com>


destogl (approved):
Thanks for fixing this and and extensive analysis! Very good eye for details!

bmagyar (approved):

PR #435, opened 2022-09-19
Fix the version of generate_parameter_library for CI

tylerjw:
I am migrating the package tcb_span into its own repo and I don't want to break ros2_control's CI while I am doing that so this PR fixes the version used by CI to the hash just before this upcoming change:

- https://github.com/PickNikRobotics/generate_parameter_library/pull/69
- https://github.com/picknikrobotics/cpp_polyfills.git

### DIFF
--- a/ros2_controllers.rolling.repos
+++ b/ros2_controllers.rolling.repos
@@ -11,7 +11,3 @@ repositories:
     type: git
     url: https://github.com/ros-controls/realtime_tools.git
     version: master
-  generate_parameter_library:
-    type: git
-    url: https://github.com/picknikrobotics/generate_parameter_library.git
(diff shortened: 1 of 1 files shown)


destogl:
Thanks for this! But I am not even sure if we need Generate Parameter Library as source at all. What do you think?

I think the library is mostly stable and we should be OK working with the released versions.

tylerjw:
I think you should be but there were a few bugs since the last release. If you are ok accepting this I'm going to bloom a new release today.

bmagyar:
Go for it. If something breaks, we have this PR as backup!

destogl (approved):
If everything gets green then we can merge this if now then we can close this for now.

tylerjw:
Closing for now, we are good now on this.

Issue #421, opened 2022-08-18
Comment tests back in for joint_trajectory_controller


jaron-l:
**Describe the bug**
Some tests were commented out in joint_trajectory_controller/test/test_trajectory_controller.cpp during #391 to get tests to a working state. Comment these tests back in and fix why they broke.

PR #441, opened 2022-09-29
test: :white_check_mark: add activation_and_receive_command

jaron-l:
- also change updateController to use sleeps instead of CPU intense while loop

refs: #421

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_controller.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller.cpp
@@ -199,47 +199,27 @@ TEST_P(TrajectoryControllerTestParameterized, activate)
   executor.cancel();
 }
 
-// TEST_F(TestTrajectoryController, activation) {
-//   auto traj_controller = std::make_shared<ros_controllers::JointTrajectoryController>(
-//     joint_names_, op_mode_);
-//   auto ret = traj_controller->init(test_robot_, controller_name_);
-//   if (ret != controller_interface::return_type::OK) {
-//     FAIL();
-//   }
-//
-//   auto traj_node = traj_controller->get_node();
-//   rclcpp::executors::MultiThreadedExecutor executor;
-//   executor.add_node(traj_node->get_node_base_interface());
-//
-//   auto state = traj_controller_->configure();
-//   ASSERT_EQ(state.id(), State::PRIMARY_STATE_INACTIVE);
-//
-//   state = traj_node->activate();
-//   ASSERT_EQ(state.id(), State::PRIMARY_STATE_ACTIVE);
-//
-//   // wait for the subscriber and publisher to completely setup
-//   std::this_thread::sleep_for(std::chrono::seconds(2));
-//
-//   // send msg
-//   builtin_interfaces::msg::Duration time_from_start;
-//   time_from_start.sec = 1;
-//   time_from_start.nanosec = 0;
-//   std::vector<std::vector<double>> points {{{3.3, 4.4, 5.5}}};
-//   publish(time_from_start, points, rclcpp::Time());
-//   // wait for msg is be published to the system
-//   std::this_thread::sleep_for(std::chrono::milliseconds(1000));
-//   executor.spin_once();
-//
-//   traj_controller->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01));
-//   resource_manager_->write();
-//
-//   // change in hw position
-//   EXPECT_EQ(3.3, joint_pos_[0]);
-//   EXPECT_EQ(4.4, joint_pos_[1]);
-//   EXPECT_EQ(5.5, joint_pos_[2]);
-//
-//   executor.cancel();
-// }
+TEST_F(TrajectoryControllerTest, receive_command)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+  SetUpAndActivateTrajectoryController(executor);
+
+  // send msg
+  auto duration = rclcpp::Duration::from_nanoseconds(1);
+  builtin_interfaces::msg::Duration time_from_start{duration};
+  std::vector<std::vector<double>> points{{{3.3, 4.4, 5.5}}};
+  publish(time_from_start, points, rclcpp::Time());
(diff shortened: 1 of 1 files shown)


progtologist:
I am not sure about the activation of the unit test that was commented out, but I can comment on this:
> also change updateController to use sleeps instead of CPU intense while loop
The code that replaced it does not do the same thing. The test is supposed to run update as fast as possible for `wait_time` duration, while the code that replaces it only runs update twice. I don't think the test was intended to do that.

jaron-l:
> I am not sure about the activation of the unit test that was commented out, but I can comment on this:
> > also change updateController to use sleeps instead of CPU intense while loop
> The code that replaced it does not do the same thing. The test is supposed to run update as fast as possible for `wait_time` duration, while the code that replaces it only runs update twice. I don't think the test was intended to do that.

Are you suggesting that I should separate the PR? The reason I included it was because it was related to the failure mode that the test would have if simply commented in. Though, yes, I could technically separate the fixes, I thought it trivial to add in as the PR is small.

If you're suggesting that the change to `updateController` was unwarranted, then I'm interested in hearing why. My experience with the function is that it unnecessarily made tests take longer and would make my CPU ramp up unnecessarily because of the run-as-fast-as-possible loop, which I think is generally avoided as it will restrict the ability of other threads to run. Also, I don't get why running the update function would be needed more than twice anyways. Perhaps you could explain that?

progtologist:
> Are you suggesting that I should separate the PR? The reason I included it was because it was related to the failure mode that the test would have if simply commented in. Though, yes, I could technically separate the fixes, I thought it trivial to add in as the PR is small.

No, I agree with you that the change is quite small so I think that it could be included in this PR.

> If you're suggesting that the change to `updateController` was unwarranted, then I'm interested in hearing why. My experience with the function is that it unnecessarily made tests take longer and would make my CPU ramp up unnecessarily because of the run-as-fast-as-possible loop, which I think is generally avoided as it will restrict the ability of other threads to run. Also, I don't get why running the update function would be needed more than twice anyways. Perhaps you could explain that?

I am not sure how the change you suggested would be shorter, as it should still take the same amount of time, the only difference being that on the one unit test it runs in full tilt for `wait_time`, while in your suggestion it sleeps for `wait_time`. 
It should also not restrict the ability of other threads to run, as it runs on a single thread, all the other threads of your system should be available to perform other tasks on your computer.

Now regarding why running the update function more than twice is needed, I find it hard to believe that the original author who wrote the unit test wrote a while loop like that instead of a simple `call_function(); sleep(); call_function();` without good reason. 

My comment in your code is just stating the obvious, that the code you replaced is not 1-1 equivalent. It doesn't run the exact same thing but in a more efficient way, it modifies what the unit test does.

Lastly, this function is called a few lines below, so it seems required to call update a few times for [these](https://github.com/ros-controls/ros2_controllers/blob/77a2c0658d89267eb7ee39a8aeb70d9bc39d89c6/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp#L382) checks to pass, is it not?

jaron-l:
> I am not sure how the change you suggested would be shorter, as it should still take the same amount of time, the only difference being that on the one unit test it runs in full tilt for `wait_time`, while in your suggestion it sleeps for `wait_time`. It should also not restrict the ability of other threads to run, as it runs on a single thread, all the other threads of your system should be available to perform other tasks on your computer.
> 
> Now regarding why running the update function more than twice is needed, I find it hard to believe that the original author who wrote the unit test wrote a while loop like that instead of a simple `call_function(); sleep(); call_function();` without good reason.
> 
> My comment in your code is just stating the obvious, that the code you replaced is not 1-1 equivalent. It doesn't run the exact same thing but in a more efficient way, it modifies what the function does.

I see your point. The change in the function is not all that different is that it just replaces a while-loop running for a period of time with some sleeps for roughly that same period of time. But I would argue that this methodology is not good practice for a unit test. I would think we would want our tests to run as fast as possible and not depend on a certain period of time passing. I tried implementing this without the sleeps but one of the tests crashes when accessing uninitialized memory and would require some bug fixing. I would guess that the while loop was originally added for the same reason I added it, because they couldn't get it to work otherwise and don't quite get the complexity of the tests. Perhaps I should remove the change to `updateController` that I've made in this PR and set it aside for another PR that improves the performance of `updateController` in addition to fixing the tests that don't use the same clock source as `updateController` (which is why I think they depend on a certain amount of time passing). What do you think if that idea?

> Lastly, this function is called a few lines below, so it seems required to call update a few times for [these](https://github.com/ros-controls/ros2_controllers/blob/77a2c0658d89267eb7ee39a8aeb70d9bc39d89c6/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp#L382) checks to pass, is it not?

The tests run just fine with two calls to update. I believe this is because I'm essentially setting up a step function in `updateController` that has a period set before and after the trajectory message is active and so the update function calculates the control to make the step function achieve the desired period.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -199,47 +199,35 @@ TEST_P(TrajectoryControllerTestParameterized, activate)
...
+  executor.add_node(traj_node->get_node_base_interface());
+
+  auto state = traj_controller_->configure();
+  ASSERT_EQ(state.id(), State::PRIMARY_STATE_INACTIVE);
+
+  ActivateTrajectoryController();
+  ASSERT_EQ(traj_node->get_current_state().id(), State::PRIMARY_STATE_ACTIVE);
+
+  // send msg
+  auto duration = rclcpp::Duration::from_nanoseconds(1);
```
destogl, line 216: Try to set it to 0 or remove it completely

jaron-l:
On further inspection, I'm not sure I see the value of adding this test back in. As I see it, the test mainly has two components: 
1. It tests that the controller gets activated appropriately.
2. It tests that a command is received and that the controller position goes to that position

These two parts are already tested in two different unit tests. The first one is tested in `TrajectoryControllerTestParameterized.activate` and the second one is tested in `TrajectoryControllerTestParameterized.correct_initialization_using_parameters`. What if I just deleted this test?

PR #446, opened 2022-10-10
Add build_export_dependency on parameter_traits for controllers that …

fmauch:
…use generate_parameter_library

The generated paramters.hpp file will include <parameter_traits/parameter_traits.hpp> and will itself be included by the controller.hpp files, this is indeed a dependency that should get exported to packages including this controller.hpp.

This currently breaks building [`ur_controllers`](https://build.ros2.org/job/Hbin_uJ64__ur_controllers__ubuntu_jammy_amd64__binary/26/console) on the buildfarm.

### DIFF
--- a/force_torque_sensor_broadcaster/package.xml
+++ b/force_torque_sensor_broadcaster/package.xml
@@ -13,6 +13,8 @@
 
   <build_depend>generate_parameter_library</build_depend>
 
+  <build_export_depend>parameter_traits/</build_export_depend>
+
   <depend>controller_interface</depend>
   <depend>geometry_msgs</depend>
   <depend>hardware_interface</depend>
--- a/joint_trajectory_controller/package.xml
+++ b/joint_trajectory_controller/package.xml
@@ -16,6 +16,8 @@
   <build_depend>pluginlib</build_depend>
   <build_depend>realtime_tools</build_depend>
 
+  <build_export_depend>parameter_traits/</build_export_depend>
+
   <exec_depend>angles</exec_depend>
   <exec_depend>pluginlib</exec_depend>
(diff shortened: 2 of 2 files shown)


destogl:
I had the same issue. @tylerjw and I manage to mitigate it, and it should be fixed in the new release. Let me check again how I fixed this. It was very similar to this fix. Around evening I'll check this.

tylerjw:
I'm sorry for this. I believe I fixed this bug and with the next sync from the build farm it should go away. You should be able to get the new version from the testing repo of you want to try that to make sure it fixes this for you.

fmauch:
@tylerjw  Since the error mentioned above comes from the buildfarm which is using the testing repo, I don't believe that the next sync will fix that. As you can see in the [linked build](https://build.ros2.org/job/Hbin_uJ64__ur_controllers__ubuntu_jammy_amd64__binary/27/console), the package parameter_traits never gets installed, which is not a surprise, as the dependency is not exported.

Is there anything wrong with my understanding of the problem? From my point of view it is quite obvious that the problem is the missing exported dependency and this change should fix that. But I might be missing something.

tylerjw:
I was hoping that it would be picked up as a transitive dependency with my recent changes. I'll work on this today to see if there is some way I can express it so users of generate_parameter_library don't have to know about it.

tylerjw:
@fmauch locally I tested this change and it got the build of ur_controllers further but did not get it to fully build for me (it just runs into the next dependency of the parameter library) so I created this PR which I believe fixes this issue in a cleaner way by not exposing the implementation details of generate_parameter_library in the package.xml files here: https://github.com/ros-controls/ros2_controllers/pull/448

PR #448, opened 2022-10-10
Fix parameter library export

tylerjw:
@fmauch

This PR is directed towards fixing the issue experienced by ur_controllers in this PR: #446

In my local testing the changes in #446 did not cuase the ur_controllers code to complete building, it did get it further than the error experienced in the buildfarm.

Instead of using the parameters.xml to export the transitive dependency of `parameter_traits` from the parmaeter library this updates the cmake to export the generated pareamters library and export a dependency on `generate_parameter_library` which exports the dependency on `parameter_traits` in addition to a handful of other dependencies of the parameters interface.

I tested this by building the following workspace on Rolling.
```yaml
repositories:
  Universal_Robots_ROS2_Driver:
    type: git
    url: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver.git
    version: main
  control_msgs:
    type: git
    version: humble
  cpp_polyfills:
    type: git
    url: https://github.com/PickNikRobotics/cpp_polyfills.git
    version: main
  generate_parameter_library:
    type: git
    url: https://github.com/PickNikRobotics/generate_parameter_library.git
    version: main
  realtime_tools:
    type: git
    url: https://github.com/ros-controls/realtime_tools.git
    version: master
  ros2_control:
    type: git
    url: https://github.com/ros-controls/ros2_control.git
    version: master
  ros2_controllers:
    type: git
    url: https://github.com/tylerjw/ros2_controllers.git
    version: fix_gpl_export
```

For reference on the cmake changes for exporting libraries I used this: https://docs.ros.org/en/rolling/How-To-Guides/Ament-CMake-Documentation.html#building-a-library

Recently I went through fixing a bunch of my other library code and had to learn how all this works again.

### DIFF
--- a/admittance_controller/CMakeLists.txt
+++ b/admittance_controller/CMakeLists.txt
@@ -38,15 +38,15 @@ foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
   find_package(${Dependency} REQUIRED)
 endforeach()
 
-add_library(${PROJECT_NAME} SHARED src/admittance_controller.cpp)
-target_include_directories(${PROJECT_NAME} PRIVATE include)
-generate_parameter_library(${PROJECT_NAME}_parameters src/admittance_controller_parameters.yaml)
-target_link_libraries(${PROJECT_NAME} admittance_controller_parameters)
-ament_target_dependencies(${PROJECT_NAME} ${THIS_PACKAGE_INCLUDE_DEPENDS})
+add_library(admittance_controller SHARED src/admittance_controller.cpp)
+target_include_directories(admittance_controller PRIVATE include)
+generate_parameter_library(admittance_controller_parameters src/admittance_controller_parameters.yaml)
+target_link_libraries(admittance_controller admittance_controller_parameters)
+ament_target_dependencies(admittance_controller ${THIS_PACKAGE_INCLUDE_DEPENDS})
 
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
-target_compile_definitions(${PROJECT_NAME} PRIVATE "ADMITTANCE_CONTROLLER_BUILDING_DLL")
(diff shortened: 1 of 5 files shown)


fmauch:
This looks quite sane, but I'm unsure how to actually test that locally, since the problem only occurred on the buildfarm when the JTC was installed as binary package. As the dependencies would have been installed already, if we built ros2_controllers from source on the same machine, this would not be quite the same test.

tylerjw:
I was able to build from source locally and that produced the same failure as in the logs in the buildfarm. I built using colcon with this command:

```
colcon build --mixin ccache rel-with-deb-info lld shared compile-commands
```

fmauch:
I build my workspace locally and it builds fine also without the change introduced here. However, we experienced problems with ccache also in our CI pipeline when migrating the API to the parameter interface, see [this commit](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/492/commits/c8fe2754e5d6fc692b247876269ae7d7e46f6d90) and [this failing pipeline before disabling the cache](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/actions/runs/3192271038/jobs/5214825410).

That's why I am rather certain that it should not be a matter of building everything from source, but the problem is actually the dependency simply not being installed, as it was only a build dependency. As this PR adds `generate_parameter_library` as a full dependency to the JTC package, I am almost 100% certain that this should fix the issue on the buildfarm, as well.

tylerjw:
> As this PR adds generate_parameter_library as a full dependency to the JTC package

I wanted to do it this way because the library generated has more dependencies than `parameter_traits` and adding `parameter_traits` as a dependency only gets the build a little further and it breaks on the next dependency.

destogl (approved):
Thanks for updating this!

destogl:
@tylerjw It's OK that some builds are failing (unfortunately). We have to get `kinematics_interface` package merged.

bmagyar (approved):

PR #452, opened 2022-10-17
Change units of velocity feedback in diff_drive_controller #behaviour-change

### LINKED ISSUE #411: Units of velocity feedback in diff_drive_controller
Currently velocity feedback in diff_drive_controller is expected to be in m/s. Taking into account that position feedback is in radians and velocity commands are in rad/s it seems quite odd. I think that it will be better to also use rad/s there, following change should be enough:
```
odometry_.updateFromVelocity(
        left_feedback_mean * left_wheel_radius * period.seconds(), right_feedback_mean * right_wheel_radius * period.seconds(), time);
```

macstepien:
Closes #411 
Adds multiplying velocity feedback by wheel radius. When the `position_feedback` option is used, feedback is multiplied by radius (it is expected to be in `rad`). This PR changes velocity feedback to be also multiplied by radius, so that units used will be more consistent.


bmagyar:
Aka feedback is in m/s? Isn't there a 2* missing?

macstepien:
On the current master branch velocity feedback has to be in `m/s` for it to work properly. After this change it will be in `rad/s` and by multiplying by radius it will be converted to `m/s` for proper handling in the `updateFromVelocity` function.

I think that it is correct, we multiply angular velocity by radius to get linear velocity. Am I missing any other reason why there should be `2*`?

bmagyar (approved):

destogl (approved):
Great!

Issue #451, opened 2022-10-11
Fix param export following #448


bmagyar:
#448 applied some fixes to controllers using the generate parameters lib when it was merged. Anything merged after that should also keep track of this.

* [ ] #396 
* [ ] #398 
* [ ] #386 
* [ ] #399

tylerjw:
Pushed a change to #396 based on this.

bmagyar:
Thanks, merged!

tylerjw:
Cleaned up https://github.com/ros-controls/ros2_controllers/pull/398

tylerjw:
Cleaned up #386 CMake usage, had to create new PR because that was @pac48's PR: https://github.com/ros-controls/ros2_controllers/pull/453

tylerjw:
Cleaned up CMake in https://github.com/ros-controls/ros2_controllers/pull/399

PR #386, opened 2022-07-21
[DiffDriveController] Use generate parameter library

pac48:
This is an example PR of converting the parameter usage in differential drive controller to use https://github.com/PickNikRobotics/generate_parameter_library

PR #453, opened 2022-10-17
Generate Parameter Library for diff_drive_controller

tylerjw:
- Generate parameter library for diff drive
- CMake export diff_drive_controller

Replacement of https://github.com/ros-controls/ros2_controllers/pull/386

- Squashed @pac48's commits into one from https://github.com/ros-controls/ros2_controllers/pull/386
- Rebased on master
- Cleaned up the CMake per https://github.com/ros-controls/ros2_controllers/issues/451

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
@@ -42,6 +42,9 @@
 #include "realtime_tools/realtime_publisher.h"
 #include "tf2_msgs/msg/tf_message.hpp"
 
+#include "diff_drive_controller_parameters.hpp"
+
+
 namespace diff_drive_controller
 {
 class DiffDriveController : public controller_interface::ControllerInterface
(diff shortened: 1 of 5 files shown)


diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
```diff
@@ -101,35 +104,21 @@ class DiffDriveController : public controller_interface::ControllerInterface
     const std::string & side, const std::vector<std::string> & wheel_names,
     std::vector<WheelHandle> & registered_handles);
 
-  std::vector<std::string> left_wheel_names_;
-  std::vector<std::string> right_wheel_names_;
+//  std::vector<std::string> left_wheel_names_;
+//  std::vector<std::string> right_wheel_names_;
```
tylerjw, line 108: Why did we leave these commented out?

diff_drive_controller/src/diff_drive_controller_parameter.yaml
```diff
@@ -0,0 +1,177 @@
...
+        type: bool,
+        default_value: false,
+      }
+      has_jerk_limits: {
+        type: bool,
+        default_value: false,
+      }
+      max_velocity: {
+        type: double,
+        default_value: .NAN,
```
tylerjw, line 153: NAN in a parameter breaks ROS2 python handling of parameters so commands like `ros2 param dump` stop working and throw a python exception. We probably should not use it as a default value.

PR #399, opened 2022-07-29
Generate parameters for IMU Sensor Broadcaster

tylerjw:
Signed-off-by: Tyler Weaver <tyler@picknik.ai>

generate_parameter_library for imu_sensor_bradcaster

PR #455, opened 2022-10-21
[IMU Broadcaster] Added parameters for definition of static covariances.

destogl:
Sits on top of #399 that should be merged first.


imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
```diff
@@ -76,6 +75,15 @@ controller_interface::CallbackReturn IMUSensorBroadcaster::on_configure(
 
   realtime_publisher_->lock();
   realtime_publisher_->msg_.header.frame_id = params_.frame_id;
+  // convert double vector to fixes size array in the message
```
bmagyar, line 78:
```suggestion
  // convert double vector to fixed-size array in the message
```

bmagyar (approved):
me gusta

PR #464, opened 2022-11-15
test: :white_check_mark: improve speed of joint_trajectory_controller tests

jaron-l:
### Summary:
This PR mainly focuses on changing `joint_trajectory_controller/test/test_trajectory_controller_utils.hpp:updateController` to be synchronous for reliability which has the added benefit of making them run faster. On my system, this change brings the `joint_trajectory_controller` tests from 26.3s to 5.24s.

### Specific points of interest and requests for feedback:
- `joint_trajectory_controller/src/joint_trajectory_controller.cpp::publish_state`
  - I made the `state_update_rate` parameter not rate limit when set to 0. This is a breaking change for that parameter if someone uses it to disable the controller. My argument for this change is that [other controllers behave that way](https://github.com/ros-controls/ros2_controllers/blob/12160de6d9c132b99b6c6f1cf975f6296006db7c/diff_drive_controller/src/diff_drive_controller.cpp#L230) and that is a user wants to disable the controller, they can do that through the lifecycle functionality and not through the update rate limiter. This was done because it makes the tests easier to work with if you don't have to arbitrarily wait for the filter to allow updates.
- `joint_trajectory_controller/test/test_trajectory_controller.cpp::TrajectoryControllertestParameterized.zero_state_publish_rate`
  - I removed this test as it checks for the above point.
- `joint_trajectory_controller/test/test_trajectory_controller.cpp::TrajectoryControllertestParameterized.test_state_publish_rate`
  - I changed this test to not be parameterized as it doesn't change behavior in response to the parameters. I suspect that it was originally made that way because almost all of the tests are parameterized and the author was just following the other test examples; however, since this test requires a sleep for it to work, it is inefficient to run it several times when each time is identical.

### Additional notes:
- This PR is intentionally held back from the tip of master at this point in time because master is unstable and not building nor passing tests.


joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
```diff
@@ -355,12 +356,30 @@ class TrajectoryControllerTest : public ::testing::Test
...
-    const auto end_time = start_time + wait_time;
-    while (clock.now() < end_time)
+    
+    rclcpp::Time now_time = rclcpp::Clock(RCL_STEADY_TIME).now();
+    traj_controller_->update(now_time, wait_time);
+    traj_controller_->update(now_time + wait_time, wait_time);
+  }
+  
+  /* Returns true if timed out, else returns false */
+  bool waitForSubscriberToBeReady(rclcpp::Duration timeout = rclcpp::Duration::from_seconds(.2))
```
destogl, line 366: I know it's longer, but find it easier to read and clearer.

```suggestion
  bool waitForSubscriberToBeReady(rclcpp::Duration timeout = rclcpp::Duration::from_seconds(0.2))
```

destogl (approved):
Lookg good! Thanks!

jaron-l:
This will likely need something like https://github.com/ros-controls/realtime_tools/pull/73 as I believe the tests are failing now because the realtime loop fails at trylock. Will move to draft for now.

christophfroehlich:
I'd say the update method was changed similarly with #858: But I made it updating with a fixed update rate instead of calling update() only twice as proposed here. 
I avoided the problem with trylock by still using the correct clock for the message-tests, and don't rely on the publisher for the other tests.

The part with the wait_set might be an improvement, but currently there isn't any problem with the subscriber. I'll keep that in mind if any problem arises.

PR #465, opened 2022-11-16
[AdmittanceController] Add missing dependecies for the tests

destogl:
We need a concrete implementation of `kinematics_interface` for tests to work. We use `kinematics_interface_kdl` implementation in the tests.


bmagyar (approved):
:facepalm:

PR #467, opened 2022-11-17
Include <string> to fix compilation error on macOS

light-tech:
On macOS (maybe generalized to systems using `libc++` for the STL), the inclusions of standard headers `algorithm` and `stdexcept` do not include `string` indirectly which leads to compilation error in `tricycle_controller/src/steering_limiter.cpp` when trying to instantiate an instance `const std::string error`.

Simply include `<string>` improves the compatibility with `libc++`.

### DIFF
--- a/tricycle_controller/src/steering_limiter.cpp
+++ b/tricycle_controller/src/steering_limiter.cpp
@@ -16,9 +16,9 @@
  * Author: Tony Najjar
  */
 
-#include <string>
 #include <algorithm>
 #include <stdexcept>
+#include <string>
 
 #include "rcppmath/clamp.hpp"
 #include "tricycle_controller/steering_limiter.hpp"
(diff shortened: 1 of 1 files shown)


DasRoteSkelett:
lgtm

DasRoteSkelett (approved):

tricycle_controller/src/steering_limiter.cpp
```diff
@@ -16,6 +16,7 @@
  * Author: Tony Najjar
  */
 
+#include <string>
 #include <algorithm>
 #include <stdexcept>
```
destogl, line 21:
```suggestion
#include <algorithm>
#include <stdexcept>
#include <string>
```

bmagyar (approved):

PR #468, opened 2022-11-17
[TricycleController] Removed “publish period” functionality ⏱ #abi-break #behavior-break

Robotgir:
Per discussion in Control WG meeting from Nov. 16th 2022

### DIFF
--- a/tricycle_controller/include/tricycle_controller/tricycle_controller.hpp
+++ b/tricycle_controller/include/tricycle_controller/tricycle_controller.hpp
@@ -167,11 +167,6 @@ protected:
   TractionLimiter limiter_traction_;
   SteeringLimiter limiter_steering_;
 
-  // publish rate limiter
-  double publish_rate_ = 50.0;
-  rclcpp::Duration publish_period_ = rclcpp::Duration::from_nanoseconds(0);
-  rclcpp::Time previous_publish_timestamp_{0};
-
   bool is_halted = false;
   bool use_stamped_vel_ = true;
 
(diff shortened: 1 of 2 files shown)


destogl (approved):
Only flaky tests are breaking.

bmagyar (approved):

tonynajjar:
> Per discussion in Control WG meeting from Nov. 16th 2022

I can't find any minutes of that meeting, could someone quickly explain why this change is needed?

destogl:
> I can't find any comments on that topic in the minutes of that meeting, could someone quickly explain why this change is needed?

Here is the discussion on this topic on JTC: https://docs.google.com/document/d/1818AoYucI2z82awL_-8sAA5pMCV_g_wXCJiM6SQmhSQ/edit?disco=AAAAlmxhM3k

I also wrote the decision just now into the comments to make it clear (@bmagyar can you please confirm?)


The reason is that we had issues on the test with this functionality, and is repeatable across the controller. So, the idea would be to make this somehow reusable in `helpers.hpp` header or somewhere. But then actually people at the meeting realized that the functionality is rather confusing and not actually necessary since there is a data loss on the topic which is not what one usually needs/wants. Therefore, we decided to make this logic simpler, reduce the code and remove features which value is not clear.

bmagyar:
https://github.com/ros-controls/ros2_controllers/issues/473

tonynajjar:
Thanks for the references, I commented my follow up question in #473

Issue #470, opened 2022-11-23
Add velocity limit for wheel in diff_drive_controller


arslogavitabrevis:
In a differential drive controller, each motor individually have a velocity limit. Even if the velocity is limited for linear speed and for angular speed of the robot, the combination of the angular speed and linear speed may lead to overspeed of one motor. If the speed is limited in the hardware interface or by the motor itself, the heading of the robot won't follow the command velocity.

**Describe the solution you'd like**
To keep the heading in the good direction, the speed of the other motor also have to be scaled down with the same factor. The velocity limit of the motor can be added to the parameters. When the update function is called, the speed of each motor is compared to the limits.

**Describe alternatives you've considered**

**Additional context**
I tried the solution on the galactic branch, I prepared a pull request.

PR #471, opened 2022-11-23
[DiffDriveController] Add velocity limits for individual wheels

arslogavitabrevis:
Pull request for feature request #470

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
@@ -115,6 +115,12 @@ protected:
     double separation_multiplier = 1.0;
     double left_radius_multiplier = 1.0;
     double right_radius_multiplier = 1.0;
+    bool left_wheel_has_velocity_limits = false;
+    bool right_wheel_has_velocity_limits = false;
+    double left_wheel_max_velocity = NAN;
+    double right_wheel_max_velocity = NAN;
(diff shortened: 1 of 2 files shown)


bmagyar:
Could you please add some testing too?

arslogavitabrevis:
I just had a look at the test that are already there.

I think I should bring the code added to the update function and the code that I should add to the on_configure function in a subclass like if was done for the speed_limiter to keep it clean. Do you think I shculd create a new class wheel_speed_limiter that or should I enhance the actual speed_limiter class with the wheel limit feature?

destogl (commented):
Can we somehow use the pre-defined structure for joint limits to declare and read the parameters? (see [JointLimitsParam class](https://github.com/ros-controls/ros2_control/blob/master/joint_limits/include/joint_limits/joint_limits_rosparam.hpp)

Can this limiter be useful for other controllers?

A year ago we were working on a ["SimpleJointLimiter"](https://github.com/destogl/ros2_control/pull/6/files#diff-e5cd0ef0a6b09b54041b737552f84bdd8876b0072a7eee9161c4930d7aa4e6a1) that would be in the same package as above limiter and reusable across controllers and other packages.

arslogavitabrevis:
Yes I think it would be interesting to use the JointLimitsParam. I will try use it.

Do you think I could pause this pull request in the meantime?

christophfroehlich:
This is not needed any more as the RM is handling this, recently added with https://github.com/ros-controls/ros2_control/pull/2047

PR #476, opened 2022-12-03
[JTC] Remove deprecation from parameters validation file.

destogl:
P.S. Failing CI is missing humble sync (and flaky admittance controller tests)

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/validate_jtc_parameters.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/validate_jtc_parameters.hpp
@@ -16,12 +16,17 @@
 #define JOINT_TRAJECTORY_CONTROLLER__VALIDATE_JTC_PARAMETERS_HPP_
 
 #include <string>
+#include <vector>
 
 #include "parameter_traits/parameter_traits.hpp"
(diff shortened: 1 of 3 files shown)


joint_trajectory_controller/include/joint_trajectory_controller/validate_jtc_parameters.hpp
```diff
@@ -16,12 +16,17 @@
 #define JOINT_TRAJECTORY_CONTROLLER__VALIDATE_JTC_PARAMETERS_HPP_
 
 #include <string>
+#include <vector>
 
 #include "parameter_traits/parameter_traits.hpp"
+#include "rclcpp/parameter.hpp"
+#include "rsl/algorithm.hpp"
+#include "tl_expected/expected.hpp"
```
bmagyar, line 24: How do we go about these two "new" dependencies? they aren't declared in the package.xml but we get them transitively. A general rule of thumb is if you *explicitly* use code from a dependency, you should declare it as your own, not a transitive one.

destogl: Agree! I'll add this

PR #479, opened 2022-12-06
Don't run humble workflows from master branch

bmagyar:


destogl:
If we remove all of those, then we will lose scheduled builds for those versions. Only the default branch can execute scheduled builds. This means that master needs to have all actions, which are then filtered by triggers at the beginning of the definition.

bmagyar:
I've retargeted all humble workflows to the humble branch.

destogl (approved):
Thanks!

moriarty:
@Mergifyio backport humble

PR #484, opened 2022-12-21
Steering odometry library and controllers

petkovich:
We have noticed that the current implementations of steering mobile robot controllers (tricycle and ackermann) share much of the functionality but implement it differently. This PR proposes using a base `SteeringController` class/file and an implementation class/file for each mobile robot configuration. The `SteeringOdometry` is used for calculating open and closed loop odometry, as well as commands that need to be sent to the hardware. The new odometry fully supports different velocities of traction wheels.

This PR does not encapsulate some features that probably should be implemented in the future, for example:
- support ackermann_msg reference messages


steering_controllers/include/steering_controllers/steering_odometry.hpp
```diff
@@ -0,0 +1,209 @@
...
+  double y_;        //   [m]
+  double heading_;  // [rad]
+
+  /// Current velocity:
+  double linear_;   //   [m/s]
+  double angular_;  // [rad/s]
+
+  double wheel_separation_;
+  double wheelbase_;
+  double wheel_radius_;
```
willcbaker, line 198: can we allow for different sized wheel radii? at least front and rear axle, but I wonder if there is an even more portable solution.
I am not sure, but how might different front/rear wheel_separations also play into the downstream odom calculations?

steering_controllers/src/steering_odometry.cpp
```diff
@@ -0,0 +1,294 @@
...
+  return true;
+}
+
+bool SteeringOdometry::update_from_velocity(
+  const double rear_wheel_vel, const double front_steer_pos, const double dt)
+{
+  // (right_wheel_est_vel + left_wheel_est_vel) * 0.5;
+  double linear_velocity = rear_wheel_vel * wheel_radius_;
+
+  //const double angular = (right_wheel_est_vel - left_wheel_est_vel) / wheel_separation_w_;
```
willcbaker, line 157: admittedly without following the math, I am confused by these comments (repeated elsewhere)

petkovich: They were leftovers from the previous implementations and are removed now

steering_controllers/include/steering_controllers/steering_odometry.hpp
petkovich: I have enabled different sized radii and track lengths using name conventions as here: https://en.wikipedia.org/wiki/Axle_track#/media/File:Wheelbase_and_Track.png

Also, rear steering is now enabled by setting `front_steering` parameter to false. I still need to check if different wheel tracks influence odometry equations.

petkovich:
I have made some updates to the features, the most important being:

- Parameters are now spread across multiple files. The `steering_controllers.yaml` has parameters that are shared between implementations and each implementation uses its own specific parameters. This means that different wheel radii and axle lengths (wheel tracks) are now fully supported.
- Rear steering is now enabled, as well as an arbitrary number of wheels in front and in back
- Forward kinematics for each "canonical" implementation has been implemented with a straightforward way to implement custom inverse and forward kinematics in the `steering_odometry.cpp`.

bmagyar:
This PR so far adds a ton of code. Could you also show what it removes? Aka the usage of these classes

steering_controllers/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+  LIBRARY DESTINATION lib
+)
+
+install(
+  DIRECTORY include/
+  DESTINATION include/${PROJECT_NAME}
+)
+
+if(BUILD_TESTING)
+  # TODO
```
bmagyar, line 85: This'd be pretty pretty important to have ;)

bmagyar:
Food for thought:

This conundrum has been around starting from shortly after the first `diff_drive_controller` implementation. Yes indeed there's the intuitive solution of not duplicating code and "bringing all X controllers under one umbrella" but controllers do not need to have multiple layers of complexity in them, the interface designs are straightforward to allow people to implement things mostly in-place. With ROS1 `ros_control` we learned the hard lesson of allowing multiple layers of inheritance into the code and how that affects maintainability and extendability. ANother point to mention is the lack of separation of dependencies for the not-so-different-but-still-different controllers for which you've already brought an example: `ackermann_steering_msgs`. Also [here](https://discourse.ros.org/t/is-twist-still-a-good-velocity-command-interface/13218/1) is a discourse discussion about unifying mobile base controller input messages by not unifying them at all ( :D ).


You'll have to work harder than dumping 2k lines of code here to convince me that this is a good way forward ;)

destogl:
> This conundrum has been around starting from shortly after the first `diff_drive_controller` implementation. Yes indeed there's the intuitive solution of not duplicating code and "bringing all X controllers under one umbrella" but controllers do not need to have multiple layers of complexity in them, the interface designs are straightforward to allow people to implement things mostly in-place. 

Not really. There is one method different in all controllers and they are actually the same. I will add some more explanations.

ANother point to mention is the lack of separation of dependencies for the not-so-different-but-still-different controllers for which you've already brought an example: `ackermann_steering_msgs`.

Don't understand the problem here? All controller should/can have dependency to Ackermann msgs because they are exactly that. TBH how "Ackermann" is used is really bad. During research for this I actually understood that Ackerman is commonly misused in robotics.

Long story short, all this steering controllers are what people usually call "Ackermann". We differentiate here more strictly between different number of joints those controller command and listen to. This ended up being very important in simulation. 

P.S. "Ackermann" controller from ros_control should not have this nice because it is **actually bicycle controller.

After merging this we will remove tricycle controller.


> You'll have to work harder than dumping 2k lines of code here to convince me that this is a good way forward ;) 


This needs a bit more work, still it will cover 6 different kinematics possible with two axes.

ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  bool wait_for_commands(
+    rclcpp::Executor & executor,
+    const std::chrono::milliseconds & timeout = std::chrono::milliseconds{500})
+  {
+    return wait_for_command(executor, ref_subscriber_wait_set_, timeout);
+  }
+
+  // TODO(anyone): add implementation of any methods of your controller is needed
```
bmagyar, line 111: probably not needed

ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
```diff
@@ -0,0 +1,281 @@
...
+    }
+    ASSERT_GE(max_sub_check_loop_count, 0) << "Test was unable to publish a message through "
+                                              "controller/broadcaster update loop";
+
+    // take message from subscription
+    rclcpp::MessageInfo msg_info;
+    ASSERT_TRUE(subscription->take(msg, msg_info));
+  }
+
+  // TODO(anyone): add/remove arguments as it suites your command message type
```
bmagyar, line 209: copy pasta?

steering_controllers_library/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+  TARGETS
+  ${PROJECT_NAME}
+  RUNTIME DESTINATION bin
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+)
+
+install(
+  DIRECTORY include/
+  #DESTINATION include/${PROJECT_NAME}   # TODO: it should be this, but then it doesn't compile (?)
```
bmagyar, line 62: hello?

steering_controllers_library/src/steering_odometry.cpp
```diff
@@ -0,0 +1,338 @@
...
+    // TODO(petkovich): implement this
+    // double wanted_turning_radius = wheelbase_ / std::tan(alpha);
+    double alpha_r = alpha;
+    double alpha_l = alpha;
+    steering_commands = {alpha_r, alpha_l};
+    return std::make_tuple(traction_commands, steering_commands);
+  }
+  else
+  {
+    throw std::runtime_error("Config not implemented");
```
VX792, line 290: Is it necessary to throw exceptions that we don't catch at all in these scenarios? I might be wrong, but calling std::terminate seems a bit overkill to me when a service is unavailable or the config is missing.

From a quick search I've seen something similar in the diff drive controller's code though.

steering_controllers_library/CMakeLists.txt
destogl: I added this, sir. You wanted to investigate this mister, have you? This is related to the install paths to overloading packages. I have a problem with using this includes in other controllers.

steering_controllers_library/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+  DESTINATION include
+)
+
+if(BUILD_TESTING)
+  # TODO
```
destogl, line 67: @petkovich maybe adding a timeout test would make sense. @Robotgir can you help here, please?

```suggestion
  # Tested implicitly by controllers implementations.
```

destogl (changes requested):
I have reviewed Steering Library and Ackermann in detail and added numerous suggestions. Please apply those to other controllers too.

ackermann_steering_controller/src/ackermann_steering_controller.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+controller_interface::CallbackReturn AckermannSteeringController::configure_odometry()
+{
+  ackermann_params_ = ackermann_param_listener_->get_params();
+
+  const double front_wheels_radius = ackermann_params_.front_wheels_radius;
+  const double rear_wheels_radius = ackermann_params_.rear_wheels_radius;
+  const double front_wheel_track = ackermann_params_.front_wheel_track;
+  const double rear_wheel_track = ackermann_params_.rear_wheel_track;
+  const double wheelbase = ackermann_params_.wheelbase;
```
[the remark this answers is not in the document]
petkovich: I consulted this page when making a decision for wheel track and wheelbase.:
https://en.wikipedia.org/wiki/Wheelbase

If it's not clear enough, we can rename it.

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -0,0 +1,576 @@
...
+    if (!std::isnan(current_ref->twist.linear.x) && !std::isnan(current_ref->twist.angular.z))
+    {
+      reference_interfaces_[0] = current_ref->twist.linear.x;
+      reference_interfaces_[1] = current_ref->twist.angular.z;
+    }
+  }
+  else
+  {
+    reference_interfaces_[0] = std::numeric_limits<double>::quiet_NaN();
+    reference_interfaces_[1] = std::numeric_limits<double>::quiet_NaN();
```
Robotgir, line 418:
```suggestion
    reference_interfaces_[0] = std::numeric_limits<double>::quiet_NaN();
    reference_interfaces_[1] = std::numeric_limits<double>::quiet_NaN();
```


if (
      !std::isnan(current_ref->twist.linear.x) && !std::isnan(current_ref->twist.angular.z))
    {
      reference_interfaces_[0] = 0.0;
      reference_interfaces_[1] = 0.0;

      current_ref->twist.linear.x = std::numeric_limits<double>::quiet_NaN();
      current_ref->twist.angular.z = std::numeric_limits<double>::quiet_NaN();
    }

Robotgir, line 418:
```suggestion
    reference_interfaces_[0] = std::numeric_limits<double>::quiet_NaN();
    reference_interfaces_[1] = std::numeric_limits<double>::quiet_NaN();
```


if (
      !std::isnan(current_ref->twist.linear.x) && !std::isnan(current_ref->twist.angular.z))
    {
      reference_interfaces_[0] = 0.0;
      reference_interfaces_[1] = 0.0;

      current_ref->twist.linear.x = std::numeric_limits<double>::quiet_NaN();
      current_ref->twist.angular.z = std::numeric_limits<double>::quiet_NaN();
    }

Robotgir, line 418:
```suggestion
if (
      !std::isnan(current_ref->twist.linear.x) && !std::isnan(current_ref->twist.angular.z))
    {
      reference_interfaces_[0] = 0.0;
      reference_interfaces_[1] = 0.0;

      current_ref->twist.linear.x = std::numeric_limits<double>::quiet_NaN();
      current_ref->twist.angular.z = std::numeric_limits<double>::quiet_NaN();
    }
```

steering_controllers_library/src/steering_odometry.cpp
petkovich: The exception will be triggered only if the config_type_ isn't set to bicycle/tricycle/ackermann. The idea here was to ensure all non-canonical configurations implemented on top of the `steering_controllers_library` need to add and set their own `config_type_` if their fwd kinematics needs customization. Otherwise, they need to set `config_type_` to one of the three canonical configs. 
I'm leaning towards @destogl last idea because only returning NaN could lead the developer towards debug divisions and multiplications while the root cause is the unset parameter. So, the warning would be printed in the `steering_odometry.cpp` and I can return NaN tuple to avoid throwing exceptions?

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -0,0 +1,572 @@
...
+    }
+  }
+  else
+  {
+    if (!std::isnan(current_ref->twist.linear.x) && !std::isnan(current_ref->twist.angular.z))
+    {
+      reference_interfaces_[0] = 0.0;
+      reference_interfaces_[1] = 0.0;
+      current_ref->twist.linear.x = std::numeric_limits<double>::quiet_NaN();
+      current_ref->twist.angular.z = std::numeric_limits<double>::quiet_NaN();
```
Robotgir, line 435:
```suggestion
     {
      if(params_.position_feedback == hardware_interface::HW_IF_VELOCITY)
      {
        reference_interfaces_[0] = 0.0;
        reference_interfaces_[1] = 0.0;
        current_ref->twist.linear.x = std::numeric_limits<double>::quiet_NaN();
        current_ref->twist.angular.z = std::numeric_limits<double>::quiet_NaN();  
      } else 
      {
        reference_interfaces_[0] = std::numeric_limits<double>::quiet_NaN();
        reference_interfaces_[1] = std::numeric_limits<double>::quiet_NaN();
        current_ref->twist.linear.x = std::numeric_limits<double>::quiet_NaN();
        current_ref->twist.angular.z = std::numeric_limits<double>::quiet_NaN();
      }
      
    }
```

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -0,0 +1,576 @@
...
+    {
+      reference_interfaces_[0] = current_ref->twist.linear.x;
+      reference_interfaces_[1] = current_ref->twist.angular.z;
+    }
+  }
+  else
+  {
+    reference_interfaces_[0] = std::numeric_limits<double>::quiet_NaN();
+    reference_interfaces_[1] = std::numeric_limits<double>::quiet_NaN();
+  }
```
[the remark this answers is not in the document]
Robotgir: now covered both the cases from params_.position_feedback

steering_controllers_library/doc/userdoc.rst
```diff
@@ -0,0 +1,22 @@
+.. _steering_controllers_userdoc:
+
+steering_controllers
+=====================
+
+Controllers for mobile robots with steering drive.
+Input for control are robot base_link twist stamped commands which are translated to traction and steering commands for the drive base. Odometry is computed from hardware feedback and published.
+
```
destogl, line 8:
```suggestion

Nomenclature used for the controller is used from `wikipedia <https://en.wikipedia.org/wiki/Wheelbase>`_.

```

ackermann_steering_controller/CMakeLists.txt
```diff
@@ -0,0 +1,106 @@
+cmake_minimum_required(VERSION 3.8)
```
destogl, line 1: Change this to the new version / structure

ackermann_steering_controller/src/ackermann_steering_controller.yaml
```diff
@@ -0,0 +1,40 @@
...
+      default_value: 0.0,
+      description: "Rear wheel track length.",
+      read_only: false,
+    }
+
+  wheelbase:
+    {
+      type: double,
+      default_value: 0.0,
+      description: "Distance between front and rear wheels.",
```
destogl, line 22:
```suggestion
      description: "Distance between front and rear wheels. For details see: https://en.wikipedia.org/wiki/Wheelbase",
```

ackermann_steering_controller/src/ackermann_steering_controller.yaml
```diff
@@ -0,0 +1,40 @@
+ackermann_steering_controller:
+  front_wheel_track:
+    {
+      type: double,
+      default_value: 0.0,
+      description: "Front wheel track length.",
```
destogl, line 6:
```suggestion
      description: "Front wheel track length. For details see: https://en.wikipedia.org/wiki/Wheelbase",
```

ackermann_steering_controller/src/ackermann_steering_controller.yaml
```diff
@@ -0,0 +1,40 @@
...
+      read_only: false,
+    }
+
+  rear_wheel_track:
+    {
+      type: double,
+      default_value: 0.0,
+      description: "Rear wheel track length.",
```
destogl, line 14:
```suggestion
      description: "Rear wheel track length. For details see: https://en.wikipedia.org/wiki/Wheelbase",
```

ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
```diff
@@ -0,0 +1,320 @@
...
+using ackermann_steering_controller::CMD_TRACTION_LEFT_WHEEL;
+using ackermann_steering_controller::CMD_TRACTION_RIGHT_WHEEL;
+
+namespace
+{
+constexpr auto NODE_SUCCESS = controller_interface::CallbackReturn::SUCCESS;
+constexpr auto NODE_ERROR = controller_interface::CallbackReturn::ERROR;
+const double COMMON_THRESHOLD = 1e-6;
+}  // namespace
+// namespace
```
destogl, line 59:
```suggestion
```

bicycle_steering_controller/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
+cmake_minimum_required(VERSION 3.8)
```
destogl, line 1: Adjust to the new structure

bicycle_steering_controller/src/bicycle_steering_controller.yaml
```diff
@@ -0,0 +1,24 @@
+bicycle_steering_controller:
+  wheelbase:
+    {
+      type: double,
+      default_value: 0.0,
+      description: "Distance between front and rear wheel.",
```
destogl, line 6:
```suggestion
      description: "Distance between front and rear wheel. For details see: https://en.wikipedia.org/wiki/Wheelbase",
```

ros2_controllers.rolling.repos
```diff
@@ -9,8 +9,8 @@ repositories:
     version: master
   control_msgs:
     type: git
-    url: https://github.com/ros-controls/control_msgs.git
-    version: humble
+    url: https://github.com/StoglRobotics-forks/control_msgs.git
```
destogl, line 12: Merge this and replace

steering_controllers_library/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
+cmake_minimum_required(VERSION 3.16)
```
destogl, line 1: Update this to the new structure

tricycle_steering_controller/CMakeLists.txt
```diff
@@ -0,0 +1,115 @@
+cmake_minimum_required(VERSION 3.8)
```
destogl, line 1: Update to new structure

tricycle_steering_controller/include/tricycle_steering_controller/visibility_control.h
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt)
+// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt) (template)
```
destogl, line 2:
```suggestion
```

tricycle_steering_controller/src/tricycle_steering_controller.yaml
```diff
@@ -0,0 +1,32 @@
...
+      default_value: 0.0,
+      description: "Wheel track length.",
+      read_only: false,
+    }
+
+  wheelbase:
+    {
+      type: double,
+      default_value: 0.0,
+      description: "Distance between front and rear wheels.",
```
destogl, line 14:
```suggestion
      description: "Distance between front and rear wheels. For details see: https://en.wikipedia.org/wiki/Wheelbase",
```

tricycle_steering_controller/src/tricycle_steering_controller.yaml
```diff
@@ -0,0 +1,32 @@
+tricycle_steering_controller:
+  wheel_track:
+    {
+      type: double,
+      default_value: 0.0,
+      description: "Wheel track length.",
```
destogl, line 6:
```suggestion
      description: "Wheel track length. For details see: https://en.wikipedia.org/wiki/Wheelbase",
```

tricycle_steering_controller/test/test_load_tricycle_steering_controller.cpp
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt)
+// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt) (template)
```
destogl, line 2:
```suggestion
```

tricycle_steering_controller/test/test_tricycle_steering_controller.cpp
```diff
@@ -0,0 +1,277 @@
+// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt)
+// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt) (template)
```
destogl, line 2:
```suggestion
```

tricycle_steering_controller/test/test_tricycle_steering_controller_preceeding.cpp
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt)
+// Copyright (c) 2023, Stogl Robotics Consulting UG (haftungsbeschränkt) (template)
```
destogl, line 2:
```suggestion
```

destogl (changes requested):
Small changes needed

destogl (approved):
Cool! Thanks!

steering_controllers_library/src/steering_odometry.cpp
```diff
@@ -0,0 +1,341 @@
...
+  {
+    return 0;
+  }
+  return std::atan(theta_dot * wheelbase_ / Vx);
+}
+
+std::tuple<std::vector<double>, std::vector<double>> SteeringOdometry::get_commands(
+  double Vx, double theta_dot)
+{
+  // desired velocty and steering angle of the midle of traction and steering axis
```
christophfroehlich, line 227:
```suggestion
  // desired velocity and steering angle of the middle of traction and steering axis
```

steering_controllers_library/doc/userdoc.rst
```diff
@@ -0,0 +1,88 @@
...
+
+
+Description of controller's interfaces
+--------------------------------------
+
+References (from a preceding controller)
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+- <controller_name>/linear/velocity      [double]  # in [m/s]
+- <controller_name>/angular/position     [double]  # in [rad]
```
christophfroehlich, line 39: As I understand the library, in a simple single-controller setup the twist (linear+angular velocity) inputs are calculated to traction wheel velocity and steering axis angle via inverse kinematics. Why do we have a position interface here? Of what type is the preceding controller in this case? Maybe you can add a paragraph on these basics, even if it is common-sense for roboticists ;) (Honestly, I never worked with these type of robots.)

christophfroehlich (changes requested):
As far as I can review it this all makes sense to me. 
Some feedback related to documentation: The code is rather easy to read, but from a user-perspective I'd need a little bit more information (see the questions in the steering lib userdoc).

steering_controllers_library/src/steering_odometry.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+/**
+   * \brief Other possible integration method provided by the class
+   * \param linear
+   * \param angular
+   */
+void SteeringOdometry::integrate_exact(double linear, double angular)
+{
+  if (fabs(angular) < 1e-6)
+    integrate_runge_kutta_2(linear, angular);
```
VX792, line 323:
```suggestion
{
    integrate_runge_kutta_2(linear, angular);
}
```

Because I'm not sure if the Format job will eat this without brackets

steering_controllers_library/src/steering_odometry.cpp
```diff
@@ -0,0 +1,341 @@
...
+  angular_acc_.accumulate(angular / dt);
+
+  linear_ = linear_acc_.getRollingMean();
+  angular_ = angular_acc_.getRollingMean();
+
+  return true;
+}
+
+bool SteeringOdometry::update_from_position(
+  const double traction_wheel_pos, const double steer_pos, const double dt)
```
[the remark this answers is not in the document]
bmagyar: they totally should

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+// Tests controller update_reference_from_subscribers and
+// its two cases for position_feedback true/false behavior
+// when too old msg is sent i.e age_of_last_command > ref_timeout case
+TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
+{
+  SetUpController();
+
+  rclcpp::executors::MultiThreadedExecutor executor;
+  executor.add_node(controller_->get_node()->get_node_base_interface());
```
bmagyar, line 88: These test implementations tend to be pretty flaky. Why do you need an entire node environment for testing the business logic? Simple unit tests for *a library* are much more powerful IMO. I'd do the brunt of the testing against steering_odometry in pure C++

steering_controllers_library/src/steering_odometry.cpp
```diff
@@ -0,0 +1,341 @@
...
+ */
+
+#include "steering_controllers_library/steering_odometry.hpp"
+
+#include <cmath>
+#include <iostream>
+
+namespace steering_odometry
+{
+SteeringOdometry::SteeringOdometry(size_t velocity_rolling_window_size)
```
bmagyar, line 29: basic unit tests would be nice. This class has 0 coverage on it's own

bmagyar (changes requested):

steering_controllers_library/src/steering_odometry.cpp
destogl: This is intentionally. Since we would need some fake implementation of virtual class which we basically have in the specific implementations. If we test everything, then it's the same as we test implementations itself.

steering_controllers_library/test/test_steering_controllers_library.cpp
destogl: yes, we need this since we are loading all parameters through the node interface each time. What is the value in implementing fake things if we anyway have that things are working with provided parameters.

ackermann_steering_controller/test/test_ackermann_steering_controller.cpp
```diff
@@ -0,0 +1,294 @@
...
+    msg.linear_velocity_command.data[CMD_TRACTION_RIGHT_WHEEL], 0.22222222222222224,
+    COMMON_THRESHOLD);
+  EXPECT_NEAR(
+    msg.linear_velocity_command.data[CMD_TRACTION_LEFT_WHEEL], 0.22222222222222224,
+    COMMON_THRESHOLD);
+  EXPECT_NEAR(msg.steering_angle_command.data[0], 1.4179821977774734, COMMON_THRESHOLD);
+  EXPECT_NEAR(msg.steering_angle_command.data[1], 1.4179821977774734, COMMON_THRESHOLD);
+}
+
+int main(int argc, char ** argv)
```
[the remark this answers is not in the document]
destogl: Not really. We have to have this because we are using an integrated test with generate_parameter_library. This enables us to have exemplary parameter sets “on board” that confirm functionality, and one can simply copy those files and adjust them. I think this is a neat end-user life improvement having such a thing. Also, one can simply put their parameters into tests to check the behavior if needed – good for debugging so strange setups.

steering_controllers_library/src/steering_odometry.cpp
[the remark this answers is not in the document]
destogl: No, they shouldn't. Because it depends on the wheels and configuration, how optometry is calculated. And we need support for the calculating odometry when we have only position interfaces towards hardware.

steering_controllers_library/doc/userdoc.rst
destogl: What is you are controlling the position of your joints. That the point. you can control position or velocity of specific joint. This setup enables all combinations.

steering_controllers_library/src/steering_odometry.cpp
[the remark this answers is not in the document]
bmagyar: I addressed this

ackermann_steering_controller/test/test_ackermann_steering_controller.cpp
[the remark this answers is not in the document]
bmagyar: This piece of code has none of what you mention. Please elaborate.

bicycle_steering_controller/test/test_load_bicycle_steering_controller.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+  std::shared_ptr<rclcpp::Executor> executor =
+    std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+
+  controller_manager::ControllerManager cm(
+    std::make_unique<hardware_interface::ResourceManager>(
+      ros2_control_test_assets::minimal_robot_urdf),
+    executor, "test_controller_manager");
+
+  ASSERT_NE(+
```
bmagyar, line 37:
```suggestion
  ASSERT_NE(
```

ARK3r:
I would like to start testing the ackermann controller (for personal and community gains hopefully) to expedite the process of getting these changes pushed to the repo.

I'm looking at [f1tenth](https://f1tenth.org/) car with the [urdf provided here](https://github.com/f1tenth-dev/simulator).

The model looks like this:
Screenshot from 2023-05-09 16-37-33

So I'm wondering 
- firstly if the apache 2.0 is a good license for proceeding with getting this possibly integrated into the ros2_control demos as well and if not what other available models should I go ahead with,
- and secondly I tried to build [the odometry library branch of ros2_controllers from stogl-robotics-forks](https://github.com/StoglRobotics-forks/ros2_controllers/tree/ackermann-odometry-library) with ros2_control main branch and it was failing. What configuration should I use to have it build properly or are there any dockers that help me with the build?

ARK3r:
In the process of preparing a docker image for the ackermann controller test I'm installing humble from source and I had a build error of 

`steering_odometry.hpp:28:10: fatal error: rcpputils/rolling_mean_accumulator.hpp: No such file or directory`.

Might [rcpputils](https://github.com/ros2/rcpputils) be a dependency for the package or am I running into this simply because I'm building from source?

christophfroehlich:
> * firstly if the apache 2.0 is a good license for proceeding with getting this possibly integrated into the ros2_control demos as well and if not what other available models should I go ahead with,

Apache 2.0 is the same license as others in ros-controls, I think you can use this one.

> * and secondly I tried to build [the odometry library branch of ros2_controllers from stogl-robotics-forks](https://github.com/StoglRobotics-forks/ros2_controllers/tree/ackermann-odometry-library) with ros2_control main branch and it was failing. What configuration should I use to have it build properly or are there any dockers that help me with the build?

The problem is that the joint_trajectory_controller does not build since we updated the controller_state message without deprecation period in the rolling branch. Try to merge the main into this branch and it should run, unless @petkovich pushes it to the PR.

steering_controllers_library/doc/userdoc.rst
```diff
@@ -0,0 +1,88 @@
...
+
+The controller uses velocity input, i.e., stamped or unstamped Twist messages where linear ``x`` and angular ``z`` components are used.
+Angular component under
+Values in other components are ignored.
+In the chain mode the controller provides two reference interfaces, one for linear velocity and one for steering angle position.
+Other relevant features are:
+
+  - support for front and rear steering configurations;
+  - odometry publishing as Odometry and TF message;
+  - input command timeout based on a parameter.
```
christophfroehlich, line 22:
```suggestion
* support for front and rear steering configurations;
* odometry publishing as Odometry and TF message;
* input command timeout based on a parameter.
```

I updated the template from control.ros.org, now I suggest using an asterisk without spaces in front, e.g., see
https://github.com/ros-controls/ros2_controllers/blob/6851df137cf2118d1a7166921d34eb759c225526/joint_trajectory_controller/doc/userdoc.rst?plain=1#L34

steering_controllers_library/doc/userdoc.rst
```diff
@@ -0,0 +1,88 @@
...
+
+The command for the wheels are calculated using ``odometry`` library where based on concrete kinematics traction and steering commands are calculated.
+Currently implemented kinematics in corresponding packages are:
+
+  - :ref:`Bicycle <bicycle_steering_controller_userdoc>` - with one steering and one drive joints;
+  - :ref:`Tricylce <tricycle_steering_controller_userdoc>` - with one steering and two drive joints;
+  - :ref:`Ackermann <ackermann_steering_controller_userdoc>` - with two seering and two drive joints.
```
christophfroehlich, line 29:
```suggestion
* :ref:`Bicycle <bicycle_steering_controller_userdoc>` - with one steering and one drive joints;
* :ref:`Tricylce <tricycle_steering_controller_userdoc>` - with one steering and two drive joints;
* :ref:`Ackermann <ackermann_steering_controller_userdoc>` - with two seering and two drive joints.
```

christophfroehlich, line 29:
```suggestion
* :ref:`Bicycle <bicycle_steering_controller_userdoc>` - with one steering and one drive joints;
* :ref:`Tricylce <tricycle_steering_controller_userdoc>` - with one steering and two drive joints;
* :ref:`Ackermann <ackermann_steering_controller_userdoc>` - with two seering and two drive joints.
```

steering_controllers_library/doc/userdoc.rst
christophfroehlich: Sorry, I still do not understand it 😕 But if you sure about it, I believe you :wink:

steering_controllers_library/doc/userdoc.rst
```diff
@@ -0,0 +1,88 @@
...
+
+``front_steering == false``
+
+- <front_wheels_names[i]>/velocity     [double]  # in [m/s]
+- <rear_wheels_names[i]>/position      [double]  # in [rad]
+
+States
+,,,,,,,
+``position_feedback == true`` --> ``TRACTION_FEEDBACK_TYPE = position``
+``position_feedback == false`` --> ``TRACTION_FEEDBACK_TYPE = velocity``
```
christophfroehlich, line 56:
```suggestion
Depending on the ``position_feedback``, different feedback types are expected

* ``position_feedback == true`` --> ``TRACTION_FEEDBACK_TYPE = position``
* ``position_feedback == false`` --> ``TRACTION_FEEDBACK_TYPE = velocity``
```

steering_controllers_library/doc/userdoc.rst
```diff
@@ -0,0 +1,88 @@
...
+- <rear_wheels_names[i]>/position                   [double]  # in [rad]
+
+Subscribers
+,,,,,,,,,,,,
+Used when controller is not in chained mode (``in_chained_mode == false``).
+
+- <controller_name>/reference  [geometry_msgs/msg/TwistStamped]
+  **NOTE**: Parameter ``use_stamped_vel`` is ``true``.
+- <controller_name>/reference_unstamped   [geometry_msgs/msg/Twist]
+  **NOTE**: Parameter ``use_stamped_vel`` is ``true``.
```
christophfroehlich, line 75: This list is not rendered properly in rst.

What does the note mean? It expects it to be true, or it will force it to true?

```suggestion
- <controller_name>/reference  [geometry_msgs/msg/TwistStamped]
- <controller_name>/reference_unstamped   [geometry_msgs/msg/Twist]

.. note::
  Parameter ``use_stamped_vel`` is ``true``.
```

christophfroehlich (changes requested):

ARK3r:
I built the docker image with the steering controllers and made a bare bone ros2 package for the f1tenth urdf. I'm thinking my next step would be to port the urdf to gazebo and configure the Ackermann controller then get it to work with gazebo. 

Any possible suggestions how I should alter this course of action? Should I use something besides gazebo classic? I've never used gazebo nor have I set up ros2 control before.

christophfroehlich:
Then I'd suggest that first you should get familiar with the demos from ros2_control_demos. Have a look on example_2, which uses a simple simulation of the DiffBot within the hardware component. You could adapt this component to test the library of this PR.
Because this is rather off topic to this PR, please open a new issue or post it on answers.ros.org

bmagyar (approved):
Marking this mammoth to be auto-merged

destogl:
@mergifyio backport humble

PR #486, opened 2022-12-28
[JTC] Use the same variables for state feedback and PID loop

christophfroehlich:
Before this change, the published `state.position.error` was normalized by `angles::shortest_angular_distance` but the position error for the PID loop was not.


destogl (approved):
Generally yes, when CI is also happy.

Could it be that `state_error_` is calculated after this lines? Maybe this is the reason of failing CI at the moment.

christophfroehlich:
The only thing failing is the binary build, or am I missing something?

>  ERROR: the following rosdeps failed to install
>   dnf: Failed to detect successful installation of [ros-rolling-generate-parameter-library]
>   dnf: Failed to detect successful installation of [ros-rolling-rsl]

christophfroehlich:
#491 includes this change and does not change the default behavior (in contrast to this PR).

bmagyar:
closing in favour of #491

PR #490, opened 2023-01-06
[JTC] Allow ff_velocity_scale=0 without deprecated warning

christophfroehlich:
Otherwise I always got this warning when I used a PID controller without feedforward gain.


bmagyar:
I'm even tempted to say we could remove the deprecation warning completely and not accept the old format. We shouldn't be dealing with Param handling that's not with the generate params library.

christophfroehlich:
ok for me, I didn't want to question Denis' deprecation warning ;)

bmagyar:
The point of deprecation warnings is to eventually break what they warn
people about

destogl:
> ok for me, I didn't want to question Denis' deprecation warning ;)

Feel free to remove it :)

It was not very long in there, but hopefully at least one release cycle.

bmagyar (approved):

PR #491, opened 2023-01-06
[JTC] Configurable joint positon error normalization behavior

christophfroehlich:
This PR addresses an optional configuration if the position error used as input for the PID controller should be normalized between -pi/pi or not (the latter is the default). 
This normalization is useful if endless rotary joints are controlled, where always the shortest relative distance is of interest. But if linear joints or limited rotary joints are used, it might be very dangerous.

I'm open for suggestions to a better naming of the parameter, and if it should be situated below the gain-structure or in a separate one.


joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -101,6 +101,11 @@ joint_trajectory_controller:
         default_value: 0.0,
         description: "Feed-forward scaling of velocity."
       }
+      normalize_error: {
```
destogl, line 104: I am even open to have this here. It is not fully off - it is quite related to the PID stuff in JTC...

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -483,6 +487,211 @@ TEST_P(TrajectoryControllerTestParameterized, state_topic_consistency)
...
+    EXPECT_NEAR(points[0][2], joint_pos_[2], allowed_delta);
+  }
+
+  if (traj_controller_->has_velocity_command_interface())
+  {
+    // check command interface
+    EXPECT_LE(0.0, joint_vel_[0]);
+    EXPECT_LE(0.0, joint_vel_[1]);
+
+    // use_closed_loop_pid_adapter_?
```
destogl, line 661:
```suggestion
    // use_closed_loop_pid_adapter_
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -483,6 +487,211 @@ TEST_P(TrajectoryControllerTestParameterized, state_topic_consistency)
...
+  EXPECT_EQ(n_joints, state_msg->desired.positions.size());
+  EXPECT_EQ(n_joints, state_msg->actual.positions.size());
+  EXPECT_EQ(n_joints, state_msg->error.positions.size());
+
+  // are the correct desired positions used?
+  EXPECT_NEAR(points[0][0], state_msg->desired.positions[0], allowed_delta);
+  EXPECT_NEAR(points[0][1], state_msg->desired.positions[1], allowed_delta);
+  EXPECT_NEAR(points[0][2], state_msg->desired.positions[2], 3 * allowed_delta);
+
+  // no normalization of position error
```
destogl, line 637: Shouldn't this here be normalized?

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: fixed

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: The comments were misleading, maybe it's clearer now

destogl (approved):
Looking good!

bmagyar (approved):

christophfroehlich:
@mergifyio backport humble

bmagyar:
@mergifyio backport humble

PR #494, opened 2023-01-11
add publisher topic parameter to forward_position_controller

mamueluth:


ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
```diff
@@ -25,14 +25,14 @@ class PublisherForwardPosition(Node):
     def __init__(self):
         super().__init__("publisher_forward_position_controller")
         # Declare all parameters
-        self.declare_parameter("controller_name", "forward_position_controller")
+        self.declare_parameter("publish_topic", "/position_commands")
```
destogl, line 28: I would rather go with empty and force users to set it. What do you think?

```suggestion
        self.declare_parameter("publish_topic", "")
```

destogl (changes requested):

ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
bmagyar: yes but if there are any tests relying on that those sh ould also be updated (in demos I guess)

bmagyar:
In general isn't this what remapping is for?

ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
destogl: Demos are currently independent of this since they use old implementation in the `ros2_control_demos` repository.

destogl:
> In general isn't this what remapping is for?

actually no in this case. It would be better that we enable setting of argument in YAML file to give path to custom configuration. This way users don't have to copy the launch file and adjust it, but just include it and set new configuration. Also convinient for CMD calls.

bmagyar:
Ok I agree, config file is good!

bmagyar (approved):

PR #507, opened 2023-01-21
Changing to_chrono to use nanoseconds

danwahl:
Previously std::chrono::seconds, which rounds to 0 if action_monitor_period_ < 1


nbbrooks:
A [related bugfix with JTC](https://github.com/ros-controls/ros2_controllers/pull/437) also resets the timer pointer so you don't end up creating multiple ones. It's a shame noone caught the pattern being used here in gripper controller.

destogl (approved):
Thanks for this fix!

mcbed (approved):

bmagyar (approved):
Nice catch, thanks for the follow-up!

bmagyar:
@mergifyio backport to humble

PR #515, opened 2023-02-04
Fix JTC segfault on unload

VX792:
Based on the discussion in #423 - the traj_home_point_ptr_ is a nullptr until the node is activated, so unloading from configured state (if no activation happened before) results in a segfault.


bijoua29 (approved):

DasRoteSkelett (commented):
The change makes total sense to me. I would however love to see a small test case, and the git messages could be squashed and the message could be a little bit more verbose. 

Just my 2 cents.

destogl (approved):
Great!

It would be cool to have small test case, but otherwise cool!

VX792:
Done!

Commits are squashed and I also added a small test case.

arne48 (approved):

bmagyar (approved):

bmagyar:
@mergifyio backport to humble

PR #516, opened 2023-02-05
Fix Interpolation Logic

swiz23:
Within the `from_string` function, there is logic being performed to determine what interpolation method to use for the Joint Trajectory Controller. Currently, the logic to check if 'splines' is being used seems to say... if the specified interpolation method is not equal to splines, then return the splines interpolation method. This doesn't make sense to me. I think the check should really be _if the specified interpolation method is equal to splines, then return the splines interpolation method._

Since there are only two options currently, this does not cause any breaking change (as we first check to see if 'None' is specified). However, if there were three or more interpolation methods, then the check in its current state will return either 'None' or Splines only.

@AndyZe, @mechwiz


AndyZe (approved):
Yep, this makes sense to me. Good find.

destogl (approved):
Thanks! Good catch!

bmagyar (approved):

bmagyar:
@mergifyio backport humble

PR #517, opened 2023-02-06
Don't set `interpolation_method_` twice

AndyZe:
Previously it was set in `on_init()` L61 and also in `on_configure()` L698.


destogl (approved):

bmagyar (approved):
thank you, here's a :fortune_cookie: for your trouble!! :D

Issue #473, opened 2022-11-30
Remove update loop filtering for publishing controller state


jaron-l:
## Background

Per the discussion in the working group meeting on Nov. 16th, all the controllers should not be filtering their update loops using a publisher period parameter. For example, the `joint_trajectory_controller` filers in the [`publish_state`](https://github.com/ros-controls/ros2_controllers/blob/09e50139882b31a21e2c54ec1c4bf691bf58ef3d/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L894) function using the associated [parameter](https://github.com/ros-controls/ros2_controllers/blob/09e50139882b31a21e2c54ec1c4bf691bf58ef3d/joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml#L53). Just remove the parameter and the check for it. Always publish the update when called.

Check the rest of the controllers and look for similar functionality to remove. It will always be in their update function or in a function called by the update function (like in the `joint_trajectory_controller`).

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_controllers/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/foxy/Tutorials/Workspace/Creating-A-Workspace.html), for Step3 use "Download Source Code" section with [these instructions](https://ros-controls.github.io/control.ros.org/getting_started.html#compiling).

[...]

PR #520, opened 2023-02-08
[JTC] ⏰ remove state publish rate parameter and logic and publish on each update.

destogl:


arne48 (approved):
I assume this is implementing the changes mentioned in #473.
To me the changes look good and complete.

Edit: I don't know if there is an intended pattern for this already, but maybe it could be considered to add an RCLCPP_INFO message that the parameter was removed if we still find it in the yaml file.
This could help users who don't read the docs to often (or don't remember the log output) explicitly why after an update a parameter has no effect anymore.
This could reduce the cluttering of config files in the field.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -936,23 +922,12 @@ controller_interface::CallbackReturn JointTrajectoryController::on_shutdown(
...
-  {
-    return;
-  }
-
-  if (state_publisher_ && state_publisher_->trylock())
+  if (state_publisher_->trylock())
   {
-    last_state_publish_time_ = get_node()->now();
-    state_publisher_->msg_.header.stamp = last_state_publish_time_;
+    state_publisher_->msg_.header.stamp = time;
```
jaron-l, line 930: It seems like the only time when it would be useful to change the time in the header from now to a specified time would be in testing. Would it be better to just leave it as is and then create a standard test function that uses this publish state function correctly?

destogl: This is still good to have in any case. We are getting time from controller manager to have it consistent. So I would keep this as it is sinc this is the control time

destogl:
> I don't know if there is an intended pattern for this already, but maybe it could be considered to add an RCLCPP_INFO message that the parameter was removed if we still find it in the yaml file.
> This could help users who don't read the docs too often (or don't remember the log output) explicitly why after an update a parameter has no effect anymore.
> This could reduce the cluttering of config files in the field.

We will do that for Humble. On the Rolling we just kick it out for the simplicity. People using Rolling should follow what is happening. Is this reasonable, or you think that would confuse too many people?

jaron-l (approved):

arne48:
> > I don't know if there is an intended pattern for this already, but maybe it could be considered to add an RCLCPP_INFO message that the parameter was removed if we still find it in the yaml file.
> > This could help users who don't read the docs too often (or don't remember the log output) explicitly why after an update a parameter has no effect anymore.
> > This could reduce the cluttering of config files in the field.
> 
> We will do that for Humble. On the Rolling we just kick it out for the simplicity. People using Rolling should follow what is happening. Is this reasonable, or you think that would confuse too many people?

This addresses my question pretty well. And I agree that "Rolling" users are (should be) more likely to keep track on the changelogs.
If an information will be prompted to the "normal distribution" users I think that would be totally sufficient.
Thank you @destogl.

bmagyar (approved):

Issue #521, opened 2023-02-10
Wrong frame_id in odometry message of diff_drive_controller


tverbele:
When using a namespace with the diff_drive_controller, the frame ids in the odometry message are constructed wrongly. Concretely, the namespace contains a starting "/" and is appended twice. So if I declare `odom_frame_id` as `odom` and I have `controller_namespace` being `/robot` it creates a frame id as `/robot//robot/odom` instead of the expected `robot/odom`

bobbleballs:
Hi @destogl, I've also been looking into this. 
However my setup is using individual TF topics for each namespace, e.g `/robot_n/tf`
The way that diff_drive_controller does this is different to every other node I use. Others use underscores to add prefixes to frame ID's or do nothing and expect individual TF topics as described in this issue on Geometry2 /ros2/geometry2/issues/433.
Would it be possible to add an option to add namespacing, or at the very least change the `/` to a `_`?
Thanks.

bmagyar:
The version 2.17.1 in humble should have this fix. Do you experience it still? Or in rolling?

bobbleballs:
Hi yes,
Sorry I posted this in the wrong place. #482 is more related to what I'm talking about. TF frame names should be prefixed with `_` not `/` to keep in line with how a lot of other packages do it.

PR #522, opened 2023-02-10
Fix frame id in diff_drive_controller (#521)

tverbele:
Strips first "/" from namespace and only append once.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -387,15 +387,15 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
   }
   else
   {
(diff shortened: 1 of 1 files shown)


destogl (commented):
Thanks! The old code seems odd. Now, sure, if there were a specific reason to do that.

I agree, but the code looks strange/complicated in general. Can we write a test – to understand how this is used at all.

bmagyar:
... or at least document how you can get the current setup to fail

bmagyar (approved):
TBH the erase line could do with a comment to document it a bit but I think that the appending of namespaces is the result of a conflict resolution gone wrong @destogl , those lines are just plain dumb, we should merge this asap

destogl (approved):
Thanks!

PR #526, opened 2023-02-11
Porting of the 4-wheel steering controller 

Fixit-Davide:
Hi all, 
I am currently working on a port for the 4-wheel steering controller from ROS1 to ROS2 Humble. 
It is still a work in progress but the basic functionality has been tested through a Gazebo simulation. 
While odometry is working it is still missing the implementation of velocity limiters.


destogl (commented):
Thanks for this!

Can we somehow reuse "steering_controllers_library" from #484?

Regarding limiters, you don't need them for the first version since they will be modified anyway soon. Still if you have capacity we are glad to take them.

Fixit-Davide:
I think it is possible since the basic mechanism of the 4ws is practically a double Ackermann.  I can try in my spare time to make the necessary changes to implement the library.

Fixit-Davide:
@destogl I carefully tried to undestand how steering_controllers_library works... I saw the boolean parameter "front_steering", but for this controller it wouldn't make any sense since all wheels are controlled both in steer and in traction (nothing passive or fixed)... It would need to be bypassed somehow in steering_controllers_library.cpp or via an override of its "command_interface_configuration" method in the then derived "four_wheels_steering_controller". 
Concerning the odometry, I didn't noticed any parameter describing an eventual offset between the wheel and its steering axis, and that's something fundamental for this controller (since not all robots are bound to have the steering axis exactly over the related wheel). That may apply as well to the already existing ackermann_controller (but didn't have the time to check yet properly).

Any opinion? I may try to fit everything in the "steering_controllers_library" schema, but some work and modification / expansion is needed IMHO.

destogl:
@Fixit-Davide feel free to do any changes to steering odometery library you need. Then we can discuss how to optimize if needed.

Also, if you find other optimization options feel free to open (a separate) PR

Issue #514, opened 2023-02-01
JointTrajectoryController fails to hold position on effort control


c-rizz:
JointTrajectoryController fails to hold position when used with only an effort command interface. I believe this is because when the controller aborts because of a tolerance violation, it sets an empty goal and stops updating the joint_interfaces. In position control this means the position is kept constant, but not when using effort control.
In fact in my gazebo simulation the robot falls on itself when it hits a tolerance violation.

This could be fixed by having the update loop set the trajectory message to the current position in case of an abort, the PID should then manage to keep the position. Could this be done in all the cases in which the goal trajectory is empty? What is the intended behavior when the trajectory is empty?

c-rizz:
I implemented a fix at [https://github.com/c-rizz/ros2_controllers/tree/humble](https://github.com/c-rizz/ros2_controllers/tree/humble). For now I did it on humble, but I verified the problem is also present on master. With the current fix the controller keeps the position correctly in my use case. Also it now keeps the position on startup, I am not sure if that is desirable.
If it is fine I could submit a PR.

christophfroehlich:
The same applies for canceling goals while using velocity command interface. Your patch solved this also for me, thanks!!
I also like the idea of automatically holding the start position, I solved that now always with sending a goal manually at startup. 

I was investigating this issue because I also needed a solution for that -> I've allowed myself to cherrypick your commit and open a PR for the master branch. We then can backport it to humble.

christophfroehlich:
Edit: now I realized the already existing PR

christophfroehlich:
Fixed with #558

PR #531, opened 2023-02-21
JTC: maintain current configuration as hold position

peterdavidfagan:
This pr adds logic to ensure that the current robot configuration is used as a goal when the JTC hasn't already been passed a goal: https://github.com/frankaemika/franka_ros2/pull/10#issuecomment-1438763747. 

In particular this pr adds changes made by @marcbone in https://github.com/frankaemika/franka_ros2/commit/8671588f2fe91f925848c8aa112cc050acc770e1#diff-14bf2d9185d549c84286510ef2771e8dd01c2e6d8e05b8dae3fbeb52313c401b.

@destogl I can update the master branch too and perform a backport if needed.

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -350,6 +350,12 @@ controller_interface::return_type JointTrajectoryController::update(
       }
     }
   }
+  else
+  {
+    // TODO(peterdavidfagan) temporary solution to keep the initial position
+    // before the first trajectory goal arrives
+    set_hold_position();
+  }
 
   publish_state(state_desired_, state_current_, state_error_);
   return controller_interface::return_type::OK;
@@ -1261,10 +1267,19 @@ void JointTrajectoryController::preempt_active_goal()
 
 void JointTrajectoryController::set_hold_position()
 {
-  trajectory_msgs::msg::JointTrajectory empty_msg;
-  empty_msg.header.stamp = rclcpp::Time(0);
+  trajectory_msgs::msg::JointTrajectory msg;
+  msg.header.stamp = rclcpp::Time(0);
+  msg.joint_names = params_.joints;
+  trajectory_msgs::msg::JointTrajectoryPoint point;
+  for (const auto & joint_position : last_commanded_state_.positions)
+  {
+    point.velocities.push_back(0);
+    point.accelerations.push_back(0);
+    point.positions.push_back(joint_position);
+  }
+  msg.points.push_back(point);
 
-  auto traj_msg = std::make_shared<trajectory_msgs::msg::JointTrajectory>(empty_msg);
+  auto traj_msg = std::make_shared<trajectory_msgs::msg::JointTrajectory>(msg);
   add_new_trajectory_msg(traj_msg);
 }
 


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -350,6 +350,12 @@ controller_interface::return_type JointTrajectoryController::update(
       }
     }
   }
+  else
+  {
+    // TODO(peterdavidfagan) temporary solution to keep the initial position
```
marcbone, line 355: There is nothing more permanent then a temporary solution ;P
I put the comment there as I was unsure what kind of implications this has for other use-cases.

peterdavidfagan: I am happy to continue working on this pr based on feedback from ros-controls developers and my evolving understanding of the scripts.

destogl (commented):
This doesn't feel like the right solution for the issue. First my understanding and then the proposed solution.

## Issue
When JTC is started, there is no input message, i.e., no command to the robot. This makes issues in effort controlled robots because the gravity is not compensated. Is this correct?

## Solution
In `on_activate` method, we have to make sure that initial values are set properly for the robot. In most use-cases is that the states are set as new command for a robot. I believe this should be sufficient. If you are using the robot with `effort` command interface, then we have to think what initial values are sensible.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -350,6 +350,12 @@ controller_interface::return_type JointTrajectoryController::update(
...
+    // before the first trajectory goal arrives
+    set_hold_position();
```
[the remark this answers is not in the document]
peterdavidfagan: Thanks for the feedback @destogl I will review the current set of commits based on the above information. It may be tomorrow before I can make updates due to an item I am finishing this evening.

marcbone:
> When JTC is started, there is no input message, i.e., no command to the robot. This makes issues in effort controlled robots because the gravity is not compensated. Is this correct?

Almost. No input message results in sending zero torques to the robot. Gravity is compensated, so the robot does not move away in an ideal scenario. However, you can just push the robot around as the PID controller is not running. Also, if you have a weight attached to the EE, the robot will move down. 
And this is exactly what I want to avoid.

Here is the scenario:
* The user has opened the brakes of our robot using our proprietary web interface. Now our internal controllers kick in to keep the robot at its current position.
* User grasps an object with the gripper.
* Then the user starts our MoveIt launch file that uses the JTC with the effort interface. If we now only do gravity compensation (zero torques) the robot would move away, while the user is actually just trying to plan a motion over MoveIt. He would not expect the robot to move during that phase. Thats why I definitely need the PID controller of the JTC to be active right from the start.

destogl:
OK, so if I understand correctly, initial position setup should be set to the current position so that PID kicks in and robot is kept in place. Is this correct?

c-rizz:
Hi, at this point just posting this as a reference as it may still be useful. I was thinking of making a pull request out of my solution of issue [https://github.com/ros-controls/ros2_controllers/issues/514#issuecomment-1412462098](https://github.com/ros-controls/ros2_controllers/issues/514#issuecomment-1412462098) from a few weeks ago. But I noticed there is now this other pull request that seems to tackle the same problem. My solution was actually very similar, I think the only difference is that I used the actual current joint state instead of the last commanded state. Also, I set the new trajectory from the update  method as it seemed safer from a synchronization perspective. Next time I will directly open a PR as it seems the issue got overlooked.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -350,6 +350,12 @@ controller_interface::return_type JointTrajectoryController::update(
       }
     }
   }
+  else
```
AndyZe, line 353: I would move this `else` before the `if` since it's just a few lines. Best to get it out of the way quickly for readability, rather than requiring a couple hundred lines of scrolling.

bmagyar:
I know you probably need this in humble but we operate with a backports-only policy. First we need a PR into the master branch then we can backport.

christophfroehlich:
> Hi, at this point just posting this as a reference as it may still be useful. I was thinking of making a pull request out of my solution of issue [#514 (comment)](https://github.com/ros-controls/ros2_controllers/issues/514#issuecomment-1412462098) from a few weeks ago. But I noticed there is now this other pull request that seems to tackle the same problem. My solution was actually very similar, I think the only difference is that I used the actual current joint state instead of the last commanded state. Also, I set the new trajectory from the update method as it seemed safer from a synchronization perspective.

I was looking (and overseeing this PR first) for a solution for handling the canceling of goals. Having that in mind I vote for the solution of @c-rizz taking the current joint_state instead of the last command. Think of some manual abort because something did not go as planned (e.g, grasping using generous trajectory tolerances), then you don't want the robot still trying to reach the last commanded position. 

I created already a PR from @c-rizz to the master branch with #558, what do the others think? @peterdavidfagan @AndyZe @destogl ?

peterdavidfagan:
Apologies for being inactive on this pr for a quite a while. As a result of https://github.com/ros-controls/ros2_controllers/pull/558 being merged I am going to close this pr.

PR #533, opened 2023-02-27
[DiffDriveController] Optional tf namespace prefixes instead of using node namespace

bobbleballs:
Added an additional parameter called tf_frame_prefix that will be appended to the TF frame ID's when set.
It uses underscores instead of '/' to separate the prefix from the ID as discussed in #482. 
This allows for other the TF topic itself to be namespaced which is how some other ROS2 packages such as Nav2 do it.


diff_drive_controller/src/diff_drive_controller_parameter.yaml
```diff
@@ -39,6 +39,11 @@ diff_drive_controller:
     default_value: 1.0,
     description: "Correction factor when radius of right wheels differs from the nominal value in ``wheel_radius`` parameter.",
   }
+  tf_frame_prefix: {
+    type: string,
+    default_value: "",
+    description:  "Prefix to be appended to the tf frames, will be added to odom_id and base_frame_id before publishing",
```
destogl, line 45:
```suggestion
    description:  "(optional) Prefix to be appended to the tf frames, will be added to odom_id and base_frame_id before publishing. If the parameter is empty, controller's namespace will be used.",
```

destogl (changes requested):
I would like that we make this optional and that per-default controller's namespace is used.

Explaining more your use-case could help to understand how to solve this the best - since many options here are valid.

diff_drive_controller/src/diff_drive_controller_parameter.yaml
bobbleballs: The main problem is that I don't want a namespace **at all** in my TF tree. 

I subscribe to the method of namespacing that the Nav2 package uses which remaps the whole TF topic to `/namespace/tf`. If by default it appends a namespace then it messes up the TF tree for those packages as suddenly I have a random `namespace/odom -> namespace/base_link` (Which also messes up styling as TF frame names shouldn't have `/` in them.
I think this method works best, because if you need a namespace appended to the frame name you can just add it in the params as a prefix when you launch the controller.

Sorry about the humble base, I'm a bit confused because humble and master are significantly different. #522 which is around the same sort of issue merged against humble also...

bobbleballs: @destogl Rebased to Master

destogl: > Sorry about the humble base, I'm a bit confused because humble and master are significantly different. #522 which is around the same sort of issue merged against humble also...

No problem. Thanks for referencing this. It seems that I made a mistake by merging this into humble. This is not how it should be…

> I subscribe to the method of namespacing that the Nav2 package uses which remaps the whole TF topic to /namespace/tf. 

Sorry, I am not certain that I understand the issue completely. Do you have one or multiple robots in your system? If you have only one robot, then why use a controller under a specific namespace in the first place? Why don't you just use remapping of the `~/odometry_tf` topic? Would this work for you?

> If by default it appends a namespace then it messes up the TF tree for those packages as suddenly I have a random namespace/odom -> namespace/base_link (Which also messes up styling as TF frame names shouldn't have / in them.

I see your confusion here. This future targets multi-robot systems where automatic kind of prefixing is essential since you merge all `TF` messages into one topic or even is you use multiple ones, the visualization is very, very tricky to configure. This approach increases readability. As far as I understand, TF names should not have "/“ at the beginning of the name, but otherwise they are just strings. We decided to use "/" instead of "_" to increase readability when you are dealing with multiple robots. "_" is very often used in naming, therefore it is hard to see fast the difference between TF messages as when using "/" being more natural namespace delimiter.

bobbleballs: Yes I'm using multiple robots in my setup. 
In actual implementation of using a multi-robot system, having every robot publish into a single TF topic is bad practise due to communication issues that could arise (I'm dealing with 10-20 robots on WiFi here).
As such I treat every robot as an "island" with all the topics including `/tf` are in their own namespaces. It can only see itself and forwards messages via a broker, for debugging I relay parts of the TF tree such as `base link `and `odom` frames periodically with their namespace appended so that I can view the robot fleet as a whole. A 2 robot example is shown here:
image
(View TF is a bit wide, but you get the idea...)

I get the argument between `/` and `_` and it doesn't really matter, apart from you end up with odd things such as having to make sure you remember to strip leading `/` as seen in #522 to adhere to the TF2 standard. 

What does matter is that I need to have the option of having a node in a namespace without appending the namespace to my TF frame names otherwise things get messy with the tree. Below shows what was happening in my setup with the current Humble implementation:
image
As you can see the tree is broken. 
My fix allows for there to be no `r0/` appended to the `odom` and `base_link` frames:
image
Or if you want a prefix on the TF then you can do that too:
image

bobbleballs: @destogl I've just pushed a change that should address your comments. 

- I've switched back to `/` delimiters.
- Added a new tf_frame_prefix enable param that is true by default. 
- The tf_frame_prefix is blank by default. 
- And the default behaviour is to use the namespace if the tf_frame_prefix is blank.
This way I can turn off the prefix altogether and this solves my issue. 


I removed the `controller_namespace = controller_namespace.erase(0, 1) + "/";` from #522 in favour of a check to see if there's a single `/` as this seems to do the job and the erase was confusing.

Hope this resolves this!

bmagyar (approved):

bmagyar:
@bobbleballs some form of testing would be nice

also please run `pre-commit run --all` and commit the changes to fix the format stage

bmagyar:
Also please don't remove the auto-assigned reviewers. They are there for a reason. I'll add them back

bobbleballs:
> @bobbleballs some form of testing would be nice
> 
> also please run `pre-commit run --all` and commit the changes to fix the format stage

Cool, I've run pre-commit.
Sorry about the reviewers wasn't sure if they were auto added or I'd done something wrong.
I'll look into adding a test.

hyunseok-yang:
Hello, I've been struggling with frame id in TF when I add namespace. 
Is there any plan to merge this fix into humble branch? (or backport merging?)

bmagyar:
Yeah I think this could be an easy addition to humble. We are waiting for tests.

bobbleballs:
@destogl @bmagyar Any update on this one?

bmagyar:
Thanks for pinging! I'll check this out tomorrow morning

diff_drive_controller/test/test_diff_drive_controller.cpp
```diff
@@ -68,6 +68,17 @@ class TestableDiffDriveController : public diff_drive_controller::DiffDriveContr
     }
     return false;
   }
+
+  /**
+  * @brief Used to get the real_time_odometry_publisher to verify its contents
+  *
+  * @return Copy of realtime_odometry_publisher_ object
+  */
+  std::shared_ptr<realtime_tools::RealtimePublisher<nav_msgs::msg::Odometry>> get_rt_odom_publisher(void){
```
bmagyar, line 77:
```suggestion
  std::shared_ptr<realtime_tools::RealtimePublisher<nav_msgs::msg::Odometry>> get_rt_odom_publisher(){
```

PR #546, opened 2023-03-16
Removed auto param decl

guihomework:
`automatically_declare_parameters_from_overrides=True` in tests hinders from dynamically loading external libraries declaring their own parameters (see https://github.com/ros-controls/ros2_control/issues/531 for original discussion and reasons)

Test still pass without it


destogl (approved):
Thanks!

bmagyar (approved):

PR #551, opened 2023-03-27
[JTC] Disable use of closed-loop PID adapter if controller is used in open-loop mode.

destogl:
As the name says. It is not sensible to close loop using PID adapter if controller should run in the open-loop.


christophfroehlich:
I uploaded a proposal for an update to the docs: https://github.com/StoglRobotics-forks/ros2_controllers/pull/6

bmagyar (approved):

christophfroehlich (approved):

PR #552, opened 2023-03-28
Update docs

christophfroehlich:
Add missing controllers and fix wrong forward_command controller doc

PR #554, opened 2023-03-29
[Docs] Fix broken links

christophfroehlich:
I realized that I used wrong relative pahts in the links with #552. Furthermore, links to `.rst` files does not make sense because sphinx does not replace them with html links?! I tried it with the `:ref:` in my local sphinx installation, at least here it works. Or are there other ways to make proper links?


bmagyar (approved):
Not that I know of and sphinx keeps disappointing us

PR #557, opened 2023-03-31
[JTC] Implement new ~/controller_state message

christophfroehlich:
This implements a state message for JTC following https://github.com/ros-controls/roadmap/pull/57

Needs https://github.com/ros-controls/control_msgs/pull/86


christophfroehlich:
I'm not sure how to handle the deprecations here. 

Acc. to the roadmap, the topic should be  `~/controller_state`. How should we deal with the transition phase? Making a parameter with the topic name, defaulting to the old `state` and adding a deprecated warning if it is not changed to `controller_state`?

christophfroehlich:
We decided last WG meeting that we don't want to support the old msg format in rolling/iron. I'll add new PR for humble with proper deprecation warnings.

destogl (approved):

bmagyar (approved):

Issue #469, opened 2022-11-22
JTC does not stop velocity command after goal succeed


tonynajjar:
**Describe the bug**

I'm using the JTC with a velocity command interface. After the trajectory goal succeeds, a small velocity command is still continuously sent to the hardware interface.

**To Reproduce**
setup JTC with velocity command and send a goal

**Expected behavior**

I would expect the velocity command to be reset to 0.

**Environment (please complete the following information):**
 - OS: Ubuntu Jammy
 - Version: master

**Additional context**

tonynajjar:
I noticed that this only happens when `use_closed_loop_pid_adapter_ = true`, so when there is a only a velocity command, which is my case

christophfroehlich:
Even worse (?): I realized now that JTC with `use_closed_loop_pid_adapter_` does not stop the movement after a goal gets cancelled, but the last velocity is maintained.

christophfroehlich:
We have now two PRs addressing the hold-position at startup or after canceling a goal. I think now, that the issue of this PR has a different cause because `set_hold_position()` is not called after reaching the goal:
https://github.com/ros-controls/ros2_controllers/blob/b70a8d0f685574f7e9647f8f4b4518097d505214/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L323-L333

I tried different action goals, with velocity command interface (`use_closed_loop_pid_adapter_` is true). I could not reproduce the behavior you described if the action goal has only position entries.
But if I add velocities to the points, where the last point is nonzero, the controller does not stop after reaching the goal. (behavior is depending on the combination of `ff_velocity_scale` and feedback gains)
https://github.com/christophfroehlich/gazebo_ros2_control/blob/66eabd30851847da2143714a586ec9524d79a73a/gazebo_ros2_control_demos/examples/example_position_with_velocity.cpp#L132-L137

Might that be your case? I'm not sure what the controller should do if the last received commanded velocity is different from zero. Should we call `set_hold_position()` then?

tonynajjar:
I moved on from this and don't use JTC anymore unfortunately, I won't be able to help

christophfroehlich:
Ok. If the problem was a nonzero velocity of the trajectory point at the end, it's maybe not a bug but some missing (optional) feature to automatically set it zero.

christophfroehlich:
Fixed with #558

PR #558, opened 2023-03-31
[JTC] Explicitly set hold position

christophfroehlich:
Using velocity or effort command interface: After canceling a goal (tolerances or manually by cancelling the action) the movement is not stopped, see #514. Thanks to @c-rizz for proposing a solution fixing #514 

**Edit:** 

The behavior after merging this PR is the following:
- `set_hold_position` is called at
   -  Action interface:
      - [x] state tolerance violation
      - [x] goal_time_tolerance violation
      - [x] cancel of action
      - [x] reaching the goal successfully
   -  Topic interface: 
      - [x] state tolerance violation
   - [x] Activation: (optionally) The position which is read the first time from the hardware is actively held
-  `set_hold_position` means
   - [x]  position command interface: set the current position as command
   - [x]  velocity (open loop), acceleration command interface: set command to zero
   - [x]  command interface velocity + effort in closed loop: position control on current position
- Deactivation: 
   - [x]  position command interface: set the current position as command
   - [x]  velocity + effort command interface: set zero
- [x] Tests ensuring this behavior

This should fix
- #469 
- #514
- #683


christophfroehlich:
Duplicate of #531

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -155,6 +155,11 @@ open_loop_control (boolean)
 
   Default: false
 
+start_holding (bool)
```
destogl, line 158: For more clarity:

```suggestion
start_with_holding (bool)
```

destogl: Although, I know that I wasn't sure about this – but this might be default behavior without the parameter. I think I confused something before.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -116,6 +116,29 @@ controller_interface::return_type JointTrajectoryController::update(
     return controller_interface::return_type::OK;
   }
 
+  // current state update
+  state_current_.time_from_start.set__sec(0);
+  read_state_from_hardware(state_current_);
+
+  if (start_holding_)
```
destogl, line 123: Shouldn't we here check which interfaces are used?

Moreover, what I am missing here is if starting holding will work when trajectory is aborted.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1390,11 +1412,8 @@ void JointTrajectoryController::preempt_active_goal()
 
 void JointTrajectoryController::set_hold_position()
 {
-  trajectory_msgs::msg::JointTrajectory empty_msg;
-  empty_msg.header.stamp = rclcpp::Time(0);
-
-  auto traj_msg = std::make_shared<trajectory_msgs::msg::JointTrajectory>(empty_msg);
-  add_new_trajectory_msg(traj_msg);
+  // TODO(c-rizz) Should I use writeFromNonRT?
```
destogl, line 1415: This depends from where is this called. If from a callback than, yes, we should use RealtimeBuffer to make sure things work as they should.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -548,6 +548,40 @@ TEST_P(TrajectoryControllerTestParameterized, state_topic_consistency)
...
+TEST_P(TrajectoryControllerTestParameterized, hold_on_startup)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+
+  rclcpp::Parameter start_holding_parameter("start_holding", true);
+  SetUpAndActivateTrajectoryController(executor, true, {start_holding_parameter});
+
+  constexpr auto FIRST_POINT_TIME = std::chrono::milliseconds(250);
+  updateController(rclcpp::Duration(FIRST_POINT_TIME));
+  ASSERT_TRUE(traj_controller_->carryingOutTrajectory());
```
destogl, line 580: Can you add more comments to better understand the tests.

destogl (commented):
I am uncertain if I understand the approach sufficiently here. Am I missing some context because the target branch is not master.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
destogl: Just discussing this with @bmagyar, can we move this part into `set_hold_position`. The idea is that `set_hold_position` sets an adequate trajectory to enable this functionality. Ideally, we shouldn't have to change anything in the `update()` (or just a bit).

I would really like to avoid this `ıf`, i.e., avoid having another logic path – of course, if this is possible.

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: done

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: I'll have a look

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: Let's discuss this after the other changes if it is still necessary.

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: I added comments now, is this clearer?
Now I see that tests are missing if it is really holding after a call of `set_hold_position`. But this should be an easy one.

christophfroehlich:
Tests fail -> I change this PR to draft in the meantime.

I summarized the desired behavior I have in mind in the initial description above. 

> The behavior after merging this PR is the following:
> 
> * `set_hold_position` is called at
>   
>   * Action interface:
>     
>     * state tolerance violation
>     * goal_time_tolerance violation
>     * cancel of action
>     * reaching the goal successfully
>   * Topic interface:
>     
>     * state tolerance violation
> 
> * `set_hold_position` means
>   
>   * position command interface: set the current position as command
>   * velocity (open loop), acceleration command interface: set command to zero
>   * command interface velocity + effort in closed loop: position control on current position
> 
> * (optionally) After startup, the position which is read from the hardware is actively held via `set_hold_position`
> 
> * Tests ensuring this behavior

Do you agree with that @bmagyar @destogl?  

And should  we change the behavior here?

> Open questions with topic interface:
> 
>    * what should we do when reaching the last point successfully?
>    * what should we do when not reaching the point on time (goal_time violation)?

saikishor:
> Do you agree with that @bmagyar @destogl?

I agree with @destogl and @bmagyar that a part of the code makes sense inside the `set_hold_position` method. 
 
> Open questions with topic interface:
> 
>     * what should we do when reaching the last point successfully?
If the last point is reached successfully, I don't think you should call `set_hold_position` as there might be a case where your last command point has a non-zero velocity and it contradicts this especially. If it's successful, better not to do anything and have trust in the commands
 
>     * what should we do when not reaching the point on time (goal_time violation)? See
I personally think, in this specific case, we have to call `set_hold_position` method as there is some limitation that is not allowing you to reach it in specific time, and continuing here doesn't make much sense.

However, what will happen if you deactivate the controller while the last command point is a non-zero velocity, will it continue with the velocity? or will it stop as the controller is deactivated?. I'm not sure about this part, but if the controller continues then this will be a big issue, so better to set_hold_position and then deactivate the controller. I personally think this should be the behaviour both at the activation and deactivation of the controller. 

What's your opinion on this?

christophfroehlich:
> I agree with @destogl and @bmagyar that a part of the code makes sense inside the `set_hold_position` method.

I also agree with that. Just wanted a confirmation if the proposed behavior makes sense!

> > Open questions with topic interface:
> > ```
> > * what should we do when reaching the last point successfully?
> > ```
> 
> If the last point is reached successfully, I don't think you should call `set_hold_position` as there might be a case where your last command point has a non-zero velocity and it contradicts this especially. If it's successful, better not to do anything and have trust in the commands

This is the point of #607 and #567. I understand that there might be a use-case with non-zero velocity, and `allow_nonzero_velocity_at_trajectory_end` has to true to allow it. So we could use this also to check if `set_hold_position` shall be called?

> 
> > ```
> > * what should we do when not reaching the point on time (goal_time violation)? See
> > ```
> 
> I personally think, in this specific case, we have to call `set_hold_position` method as there is some limitation that is not allowing you to reach it in specific time, and continuing here doesn't make much sense.

Ok.

> 
> However, what will happen if you deactivate the controller while the last command point is a non-zero velocity, will it continue with the velocity? or will it stop as the controller is deactivated?. I'm not sure about this part, but if the controller continues then this will be a big issue, so better to set_hold_position and then deactivate the controller. I personally think this should be the behaviour both at the activation and deactivation of the controller.
> 
> What's your opinion on this?

Never thought about this but this is a good question. Read+Write methods of hw-components are called even if no controller is active I guess.

saikishor:
> This is the point of #607 and #567. I understand that there might be a use-case with non-zero velocity, and `allow_nonzero_velocity_at_trajectory_end` has to true to allow it. So we could use this also to check if `set_hold_position` shall be called?

I don't think it is needed right?, well even if we use the method over with the condition, won't it be redundant as we know that if it is successful that it has reached it's terminal state (which is basically that it is holding position)?


> Never thought about this but this is a good question. Read+Write methods of hw-components are called even if no controller is active I guess.

I think setting holding position upon activation and deactivation should be a default behavior and maybe it need not be parameterized. What's your thoughts on this @destogl and @bmagyar ?

christophfroehlich:
> > This is the point of #607 and #567. I understand that there might be a use-case with non-zero velocity, and `allow_nonzero_velocity_at_trajectory_end` has to true to allow it. So we could use this also to check if `set_hold_position` shall be called?
> 
> I don't think it is needed right?, well even if we use the method over with the condition, won't it be redundant as we know that if it is successful that it has reached its terminal state (which is basically that it is holding position)?

If you use PID control but feed-forward gains only, then you send the velocity point over and over to the hardware, even if the position has reached the goal.

bmagyar:
> I think setting holding position upon activation and deactivation should be a default behavior and maybe it need not be parameterized. What are your thoughts on this @destogl and @bmagyar?

The intuitive answer is yes, it shouldn't be needed.
*However* this feature smells exactly like the type of stuff that will break someone's setup without that flag ;) The default behaviour should be this but there should be a way to opt-out.

christophfroehlich:
> I think setting holding position upon activation and deactivation should be a default behavior and maybe it need not be parameterized. 

At activation, first run of the `update` method: This PR already adds this behavior (optionally)
`on_deactivate` I had a look now, this was implemented already:
> - Deactivation: 
>    -  position command interface: set the current position as command
>    -  velocity + effort command interface: set zero

bmagyar:
Big 👍 on the latest writeup in the opening post

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -417,50 +410,51 @@ TEST_P(TestTrajectoryActionsTestParameterized, test_state_tolerances_fail)
...
-    point2.positions[2] = 9.0;
+    point2.positions = points_positions.at(1);
     points.push_back(point2);
 
     gh_future = sendActionGoal(points, 1.0, goal_options_);
   }
   controller_hw_thread_.join();
 
   EXPECT_TRUE(gh_future.get());
+  // TODO(christophfroehlich) sometimes the trajectory is overwritten by
```
christophfroehlich, line 442: How can this happen here? `Holding position due to state tolerance violation` is in the if branch if there is no active goal.
```
4: [ RUN      ] PositionTrajectoryControllersActions/TestTrajectoryActionsTestParameterized.test_state_tolerances_fail/2
4: [WARN] [1689668022.296384234] [test_joint_trajectory_controller]: [Deprecated]: "allow_nonzero_velocity_at_trajectory_end" is set to true. The default behavior will change to false.
4: [INFO] [1689668022.297142446] [test_joint_trajectory_controller]: No specific joint names are used for command interfaces. Using 'joints' parameter.
4: [INFO] [1689668022.297155740] [test_joint_trajectory_controller]: Command interfaces are [position] and state interfaces are [position velocity acceleration].
4: [INFO] [1689668022.297162969] [test_joint_trajectory_controller]: Using 'splines' interpolation method.
4: [INFO] [1689668022.297494705] [test_joint_trajectory_controller]: Action status changes will be monitored at 20.00 Hz.
4: [INFO] [1689668022.599148369] [test_joint_trajectory_controller]: Received new action goal
4: [INFO] [1689668022.599193964] [test_joint_trajectory_controller]: Accepted new action goal
4: [ERROR] [1689668022.599247504] [test_joint_trajectory_controller]: Holding position due to state tolerance violation
4: [INFO] [1689668022.599315419] [test_joint_trajectory_controller]: Goal reached, success!
4: /workspaces/ros2_rolling_ws/src/ros2_controllers/joint_trajectory_controller/test/test_trajectory_actions.cpp:446: Failure
4: Expected equality of these values:
4:   rclcpp_action::ResultCode::ABORTED
4:     Which is: 1-byte object <06>
4:   common_resultcode_
4:     Which is: 1-byte object <04>
4: /workspaces/ros2_rolling_ws/src/ros2_controllers/joint_trajectory_controller/test/test_trajectory_actions.cpp:447: Failure
4: Expected equality of these values:
4:   control_msgs::action::FollowJointTrajectory_Result::PATH_TOLERANCE_VIOLATED
4:     Which is: -4
4:   common_action_result_code_
4:     Which is: 0
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: I refactored it -> looks way cleaner now

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: I use the RT-safe `add_new_trajectory_msg` now, no need for this hold-variable.

christophfroehlich:
The tests do not fail at my local machine, but the CI jobs have strange errors (not receiving action feedback etc). Any suggestions?

christophfroehlich:
It seems that I fixed the race conditions by introducing the `rt_has_pending_goal_` boolean :partying_face:  It happened that the action goal was not updated but the trajectory, and vice versa. The boolean is now updated simultaneously with the trajectory itself, and does not depend on the processing of the `goal_handle_timer_`.

The only error I get now locally (not always, but with `--retest-until-fail 10`) with the `test_cancel_hold_position` test:
```
1: terminate called after throwing an instance of 'std::runtime_error'
1:   what():  Taking data from action server but nothing is ready
1: Stack trace (most recent call last) in thread 246877:
1: #13   Object "", at 0xffffffffffffffff, in 
1: #12   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f1abb6309ff, in 
1: #11   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f1abb59eb42, in 
1: #10   Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7f1abb82e2b2, in 
1: #9    Object "/opt/ros/rolling/lib/librclcpp.so", at 0x7f1abbb21971, in rclcpp::executors::MultiThreadedExecutor::run(unsigned long)
1: #8    Object "/opt/ros/rolling/lib/librclcpp.so", at 0x7f1abbb18a7e, in rclcpp::Executor::get_next_ready_executable_from_map(rclcpp::AnyExecutable&, std::map<std::weak_ptr<rclcpp::CallbackGroup>, std::weak_ptr<rclcpp::node_interfaces::NodeBaseInterface>, std::owner_less<std::weak_ptr<rclcpp::CallbackGroup> >, std::allocator<std::pair<std::weak_ptr<rclcpp::CallbackGroup> const, std::weak_ptr<rclcpp::node_interfaces::NodeBaseInterface> > > > const&)
1: #7    Object "/opt/ros/rolling/lib/librclcpp_action.so", at 0x7f1abbc6b785, in 
1: #6    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7f1abb800517, in __cxa_throw
1: #5    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7f1abb8002b6, in std::terminate()
1: #4    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7f1abb80024b, in 
1: #3    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7f1abb7f4bbd, in 
1: #2    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f1abb5327f2, in abort
1: #1    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f1abb54c475, in raise
1: #0    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f1abb5a0a7c, in pthread_kill
1: Aborted (Signal sent by tkill() 246776 1000)
```
I'm not sure if the action client or server is causing this.

joint_trajectory_controller/test/test_trajectory_actions.cpp
christophfroehlich: Fixed with `rt_has_pending_goal_` boolean

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1417,13 +1444,34 @@ void JointTrajectoryController::preempt_active_goal()
...
+    rclcpp::Time(0.0, 0.0, get_node()->get_clock()->get_clock_type());  // start immediately
+  current_pose_msg.joint_names = params_.joints;
+  current_pose_msg.points.push_back(state_current_);
+  current_pose_msg.points[0].velocities.clear();
+  current_pose_msg.points[0].accelerations.clear();
+  current_pose_msg.points[0].effort.clear();
+  if (has_velocity_command_interface_)
+  {
+    current_pose_msg.points[0].velocities.resize(
+      dof_, 0.0);  // ensure no velocity (PID will fix this)
```
bmagyar, line 1462: read a little better
```suggestion
    // ensure no velocity (PID will fix this)
    current_pose_msg.points[0].velocities.resize(dof_, 0.0);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1417,13 +1444,34 @@ void JointTrajectoryController::preempt_active_goal()
...
+  }
+  if (has_acceleration_command_interface_)
+  {
+    current_pose_msg.points[0].accelerations.resize(dof_, 0.0);  // ensure no acceleration
+  }
+  if (has_effort_command_interface_)
+  {
+    current_pose_msg.points[0].effort.resize(
+      dof_, 0.0);  // ensure no explicit effort (PID will fix this)
```
bmagyar, line 1471:
```suggestion
    // ensure no explicit effort (PID will fix this)
    current_pose_msg.points[0].effort.resize(dof_, 0.0);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1417,13 +1444,34 @@ void JointTrajectoryController::preempt_active_goal()
...
+  current_pose_msg.points[0].accelerations.clear();
+  current_pose_msg.points[0].effort.clear();
+  if (has_velocity_command_interface_)
+  {
+    current_pose_msg.points[0].velocities.resize(
+      dof_, 0.0);  // ensure no velocity (PID will fix this)
+  }
+  if (has_acceleration_command_interface_)
+  {
+    current_pose_msg.points[0].accelerations.resize(dof_, 0.0);  // ensure no acceleration
```
christophfroehlich, line 1466:
```suggestion
    ensure no acceleration
    current_pose_msg.points[0].accelerations.resize(dof_, 0.0);
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1417,13 +1444,35 @@ void JointTrajectoryController::preempt_active_goal()
...
+  current_pose_msg.points[0].accelerations.clear();
+  current_pose_msg.points[0].effort.clear();
+  if (has_velocity_command_interface_)
+  {
+    // ensure no velocity (PID will fix this)
+    current_pose_msg.points[0].velocities.resize(dof_, 0.0);
+  }
+  if (has_acceleration_command_interface_)
+  {
+    ensure no acceleration
```
christophfroehlich, line 1466:
```suggestion
    // ensure no acceleration
```

bmagyar (approved):
Thanks so much for this! THe refactoring definitely added quite a few changes but was worth it

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -424,6 +418,45 @@ TEST_P(TrajectoryControllerTestParameterized, state_topic_consistency)
...
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+
+  rclcpp::Parameter start_with_holding_parameter("start_with_holding", true);
+  SetUpAndActivateTrajectoryController(executor, true, {start_with_holding_parameter});
+
+  constexpr auto FIRST_POINT_TIME = std::chrono::milliseconds(250);
+  updateController(rclcpp::Duration(FIRST_POINT_TIME));
+  // after startup with start_with_holding being set, we expect an active trajectory:
+  ASSERT_THAT(traj_controller_->has_active_traj());
```
bmagyar, line 452:
```suggestion
  ASSERT_TRUE(traj_controller_->has_active_traj());
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -424,6 +418,44 @@ TEST_P(TrajectoryControllerTestParameterized, state_topic_consistency)
...
+TEST_P(TrajectoryControllerTestParameterized, hold_on_startup)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+
+  rclcpp::Parameter start_with_holding_parameter("start_with_holding", true);
+  SetUpAndActivateTrajectoryController(executor, true, {start_with_holding_parameter});
+
+  constexpr auto FIRST_POINT_TIME = std::chrono::milliseconds(250);
+  updateController(rclcpp::Duration(FIRST_POINT_TIME));
+  // after startup with start_with_holding being set, we expect an active trajectory:
```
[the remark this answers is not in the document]
christophfroehlich: `expectHoldingPoint` checks for active trajectory, but the additional assertion is ok ;)

BarisYazici:
I just had the time to test on our FR3 robot. And it still didn't solve the hold position issue. I didn't dive deep in debugging, but basically the robot successfully goes to the goal position and either free falls down or drags slowly sideways. We are only using effort commands. I will test today again.

christophfroehlich:
effort command with actice PID control?

christophfroehlich:
I think I've found a possible issue, do you have set a `goal_time_tolerance>0`? If the time tolerance is exceeded, the PID controller will be deactivated even in hold position mode.

BarisYazici:
I didn't explicitly set a `goal_time` tolerance. So I guess this was not checked. I used the panda_moveit_config launch file in our franka_ros2 repo to test the new `JTC`. Maybe it's been implicitly set there but I doubt it.

christophfroehlich:
I checked that now, false alarm. It could happen only if the PID can't hold the position (wrong tuning, effort limit,..), then it might enter a loop setting a new hold position again and again.

BarisYazici:
I think the problem is set_hold_position() is never get called in our case. When I add this else [motified_jtc](https://github.com/frankaemika/franka_ros2/blob/humble/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L353), it works

BarisYazici:
in our case `goal_time` is default 0 . And the set_hold_position is never get called when the robot reaches the target.

christophfroehlich:
Do you use the topic or action interface? In case of the latter, do you get a success action result?

christophfroehlich:
@BarisYazici do you have a public repo with which I could try your setup?

BarisYazici:
yes, here is the launch file I used to try JTC - https://github.com/frankaemika/franka_ros2/blob/humble/franka_moveit_config/launch/moveit.launch.py  . All settings are default. I removed the old JTC in our repo and added the one in humble branch from ros2_controllers.

christophfroehlich:
> yes, here is the launch file I used to try JTC - https://github.com/frankaemika/franka_ros2/blob/humble/franka_moveit_config/launch/moveit.launch.py . All settings are default. I removed the old JTC in our repo and added the one in humble branch from ros2_controllers. Devcontainer in the repo should reflect my setup.

This PR hasn't been backported to humble yet, which means you were still testing the old version of JTC. Could you give it a try with [this manual cherrypick](https://github.com/christophfroehlich/ros2_controllers/tree/backport558) please?

BarisYazici:
> This PR hasn't been backported to humble yet, which means you were still testing the old version of JTC. Could you give it a try with [this manual cherrypick](https://github.com/christophfroehlich/ros2_controllers/tree/backport558) please?

You are right, my mistake. I tested it, and it works as expected. We will remove the old JTC from our repo in the next update.

christophfroehlich:
thanks for the feedback

PR #560, opened 2023-04-03
Add filter to AdmittanceController

### LINKED ISSUE #424: Admittance Controller should use filters from `control_toolbox` for gravity compensation and smoothing.
Please use filter for gravity compensation. If needed, the filter can be extended. And not only that, we want to use filter for other things too, like noise reduction and similar.

_Originally posted by @destogl in https://github.com/ros-controls/ros2_controllers/pull/370#discussion_r917714081_

guihomework:
Re-introduced filters to the admittance_controller as originally done in https://github.com/ros-controls/ros2_controllers/pull/382. Improved testing, variable names and comments.

Depends on https://github.com/ros-controls/control_toolbox/pull/152 and https://github.com/ros-controls/control_toolbox/pull/153

Tests currently fail due to compensation already moving the robot after the first update step. This is expected but was not expected in past test. So a discussion should occur to decide how to change the test.


guihomework:
@destogl  I discover that "allowing edits" means allow others to write to my branch. I did not knot this was possible. Isn't it better to always rebase rather than do a merge from the master (which is then not a straight line merging for the PR anymore) ? I can force-push a rebased PR now if desired to get a cleaner history.

guihomework:
Merging master into this branch introduced a test failure (another one than the known one). This came from some invalid parameters that cannot be tested in `on_init` because they are only used `on_configure`

Additionally I should have fixed the other test failing due to gravity compensation. In fact this test (`receive_message_and_publish_updated_status`) should not pass when there is a mass at eef because gravity compensation (the one of the filter in PR or the hard-coded one in existing code) will necessarily make the posture after one iteration deviate from the start. 
Why that particular test passed before is because a smoothing (equivalent to the new LP filter) was hard-coded and applied **after** gravity compensation, which meant that after one iteration almost no gravity compensation was seen by the eef (0.1 instead 23 N). It makes no sense to filter the gravity compensation, only the noisy input data from the sensor should be filtered and that is what the proposed implementation does (LP then GC filter)

I think CI fails as it relies on 2 other PRs.

admittance_controller/src/admittance_controller_parameters.yaml
```diff
@@ -65,44 +65,17 @@ admittance_controller:
...
       }
-    filter_coefficient: {
-      type: double,
-      default_value: 0.05,
-      description: "Specifies the filter coefficient for the sensor's exponential filter."
-    }
-
-  control:
-    frame:
+    meas_frame:
```
Juliaj: nit, `measurement_frame` is more descriptive IMO.

admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
```diff
@@ -271,9 +269,14 @@ bool AdmittanceRule::calculate_admittance_rule(AdmittanceState & admittance_stat
 
   // zero out any forces in the control frame
   Eigen::Matrix<double, 6, 1> F_control;
+  // rotate to control frame
   F_control.block<3, 1>(0, 0) = rot_base_control.transpose() * F_base.block<3, 1>(0, 0);
   F_control.block<3, 1>(3, 0) = rot_base_control.transpose() * F_base.block<3, 1>(3, 0);
+
+  // select the axis to apply the zeroing to
```
Juliaj:
```suggestion
  // zero out forces/torques for axes not in selected_axes
```

Juliaj (approved):
Minor comments, overall LGTM.

admittance_controller/src/admittance_controller_parameters.yaml
christophfroehlich: done

christophfroehlich (dismissed):

christophfroehlich (changes requested):
We need to deprecate the old config, and we need to add migration notes.

PR #562, opened 2023-04-05
[CI] Fixup the repos file for humble and rolling.

destogl:
Needs backport to humble.

### DIFF
--- a/ros2_controllers-not-released.rolling.repos
+++ b/ros2_controllers-not-released.rolling.repos
@@ -1,5 +1,6 @@
 repositories:
-  generate_parameter_library:
-    type: git
-    url: https://github.com/picknikrobotics/generate_parameter_library.git
-    version: main
+  ## EXAMPLE DEPENDENCY
+#   <some_ros_package>:
+#     type: git
+#     url: git@github.com:<some_github_namespace>/<some_ros_package>.git
+#     version: master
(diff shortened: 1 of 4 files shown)


bmagyar:
So is the failure we see on the binary builds related to not having gen parameters lib from source?

destogl:
> So is the failure we see on the binary builds related to not having gen parameters lib from source?

Puh, I would also like to know that… It shouldn't be because it is released...

ros2_controllers-not-released.rolling.repos
```diff
@@ -1,5 +1,6 @@
 repositories:
-  generate_parameter_library:
-    type: git
-    url: https://github.com/picknikrobotics/generate_parameter_library.git
-    version: main
+  ## EXAMPLE DEPENDENCY
+#   <some_ros_package>:
+#     type: git
+#     url: git@github.com:<some_github_namespace>/<some_ros_package>.git
+#     version: master
```
destogl, line 6:
```suggestion
  generate_parameter_library:
    type: git
    url: https://github.com/picknikrobotics/generate_parameter_library.git
    version: main
```

destogl:
Ok, now it should work and we can continue, since this blocks PIDController and other controllers

bmagyar:
Thanks. Let's investigate that issue... it's a little disturbing but indeed the current way this PR is it's already improving things so this is getting merged

bmagyar (approved):

PR #566, opened 2023-04-08
[JTC] Import docs from wiki.ros.org

christophfroehlich:
I found some useful documentation on [wiki.ros.org](http://wiki.ros.org/joint_trajectory_controller), which I think is still correct for ROS 2. I added the information with citation to the original source.

I was not sure if the ROS 2 implementation has the same behavior on two details, see comments below.


joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -101,6 +113,105 @@ A yaml file for using it could be:
...
+  Query controller state at any future time.
+
+Preemption policy [#f1]_
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Only one action goal can be active at any moment, or none if the topic interface is used. Path and goal tolerances are checked only for the trajectory segments of the active goal.
+
+When an active action goal is preempted by another command coming from the action interface, the goal is canceled and the client is notified.
+
+Sending an empty trajectory message from the topic interface (not the action interface) will override the current action goal and not abort the action.
```
christophfroehlich, line 154: I think this changed from ROS 1, is the described behavior correct (and intentional?)

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -101,6 +113,105 @@ A yaml file for using it could be:
...
+
+~/joint_trajectory (input topic) [trajectory_msgs::msg::JointTrajectory]
+  Topic for commanding the controller.
+
+Both use the ``trajectory_msgs/JointTrajectory`` message to specify trajectories, and require specifying values for all the controller joints (as opposed to only a subset) if ``allow_partial_joints_goal`` is not set to ``True``.
+
+The primary way to send trajectories is through the action interface, and should be favored when execution monitoring is desired.
+Action goals allow to specify not only the trajectory to execute, but also (optionally) path and goal tolerances.
+When no tolerances are specified, the defaults given in the parameter interface are used (see :ref:`parameters`).
+If tolerances are violated during trajectory execution, the action goal is aborted, the client is notified, and the current position is held.
```
christophfroehlich, line 133: I think this changed from ROS 1, is the described behavior correct (and intentional?)

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -101,6 +113,105 @@ A yaml file for using it could be:
...
+The scenario is that of a controller executing the trajectory from the previous example (shown in red),
+and receiving a new command (shown in green) with a trajectory start time set to either zero (start now),
+a future time, or a time in the past.
+
+.. image:: trajectory_replacement_future.png
+  :alt: Trajectory start time in the future.
+
+|
+
+.. image:: trajectory_replacement_now.png
```
christophfroehlich, line 200: Is this the current behavior? I don't think so: At t=81 and t=89 I sent the action once, at t=100 I sent the action before the first one was finished and it does not interpolate from the old one?
image

tested with the gazebo demos [with a custom action client](https://github.com/christophfroehlich/gazebo_ros2_control/blob/jtc_tests/gazebo_ros2_control_demos/examples/example_position_with_velocity.cpp)

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: See https://github.com/ros-controls/ros2_controllers/blob/fa7499f0eb28ccca11408e703b8d5fce4f4deb30/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L959-L964

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -3,23 +3,35 @@
...
 
     Quintic: Position, velocity and acceleration are specified: Guarantees continuity at the acceleration level.
 
-Hardware interface type
------------------------
+Hardware interface type [#f1]_
+-------------------------------
 
-The controller is templated to work with multiple hardware interface types. Currently joints with position, velocity and effort interfaces are supported. For position-controlled joints, desired positions are simply forwarded to the joints; while for velocity (effort) joints, the position+velocity trajectory following error is mapped to velocity (effort) commands through a PID loop. Example controller configurations can be found here.
+The controller is templated to work with multiple hardware interface types.
```
destogl, line 30:
```suggestion
```

joint_trajectory_controller/doc/userdoc.rst
destogl: It should be. Maybe we broke it, but we want to have this behavior

joint_trajectory_controller/doc/userdoc.rst
destogl: I think it will be overridden, but it should not be. The explanation is correct, but I  don't think it is desirable.

joint_trajectory_controller/doc/userdoc.rst
destogl: This looks strange... maybe we have a bug here. I was debugging few things before, but not sure right now.

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -224,18 +335,6 @@ gains.<joint_name>.normalize_error (bool)
 
   Default: false
 
-ROS2 interface of the controller
```
destogl, line 227: I would keep this here

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -257,10 +356,16 @@ The following version of the Joint Trajectory Controller are available mapping t
 ..   - velocity_controllers::JointTrajectoryController
 ..     - Input: position, [velocity, [acceleration]]
 ..     - Output: velocity
-.. TODO(anyone): would it be possible to output velocty and acceleration?
+.. TODO(anyone): would it be possible to output velocity and acceleration?
```
destogl, line 359: This should go because it is not possible in the current implemenation

joint_trajectory_controller/doc/userdoc.rst
destogl: I would actually extend this to be more like other new controllers, describing all interfaces in detail.

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: I'll investigate this one, maybe I just did not understand the nested if-else clauses.

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -101,6 +113,105 @@ A yaml file for using it could be:
...
+
+Further interfaces are:
+
+~/state (output topic) [control_msgs::msg::JointTrajectoryControllerState]
+  Topic publishing internal states with the update-rate of the controller manager.
+
+~/query_state (service) [control_msgs::srv::QueryTrajectoryState]
+  Query controller state at any future time.
+
+Preemption policy [#f1]_
```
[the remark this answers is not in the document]
christophfroehlich: done

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: I applied the structure of the PID controller, what do you think?

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: I removed it. How does the namespacing work as described in this section? I did not understand this up to now..

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: I removed this section from this PR until we figured out how this really works.

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: Ok, so this is fine for this PR -> I created a new issue based on your comment.

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: see #574

joint_trajectory_controller/doc/userdoc.rst
christophfroehlich: I checked this one now -> it aborts the action.
> Aborted due to state tolerance violation

You meant that the abort is the desired behavior? And not the ROS 1 behavior, where client is only notified but the action is still active?

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -3,23 +3,35 @@
...
 
-Hardware interface type
------------------------
+Hardware interface type [#f1]_
+-------------------------------
 
-The controller is templated to work with multiple hardware interface types. Currently joints with position, velocity and effort interfaces are supported. For position-controlled joints, desired positions are simply forwarded to the joints; while for velocity (effort) joints, the position+velocity trajectory following error is mapped to velocity (effort) commands through a PID loop. Example controller configurations can be found here.
+The controller is templated to work with multiple hardware interface types.
+Currently joints with position, velocity and effort interfaces are supported.
+For position-controlled joints, desired positions are simply forwarded to the joints;
```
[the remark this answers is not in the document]
christophfroehlich: I added details about the necessary state interfaces depending on the command interfaces.

joint_trajectory_controller/doc/userdoc.rst
bmagyar: https://github.com/ros-controls/ros2_controllers/issues/573

bmagyar (approved):

joint_trajectory_controller/doc/userdoc.rst
destogl: Yes I think we should abort then

bmagyar:
@mergifyio backport humble

PR #567, opened 2023-04-08
[JTC] Reject trajectories with nonzero terminal velocity

christophfroehlich:
As suggested by @bmagyar, the new trajectory should be even ignored if the last velocity point is nonzero. 
- This is valid for both interfaces (topic + actions).
- I think there might be use-cases, where this is intentionally (maybe receiving topics from a MPC with moving horizon, where the end of the trajectory is not a steady state). Therefore, I added a parameter to be able to deactivate this behavior. To avoid breaking changes, I deactivated this by default now but give a deprecation warning if `allow_nonzero_velocity_stop==true`.

I also added a test for that, please review #603 first.

### DIFF
--- a/joint_trajectory_controller/doc/userdoc.rst
+++ b/joint_trajectory_controller/doc/userdoc.rst
@@ -175,6 +175,9 @@ open_loop_control (boolean)
 
   Default: false
 
+allow_nonzero_velocity_stop (boolean)
+  If not set true, the last velocity point has to be zero or the goal will be rejected.
+
 constraints (structure)
   Default values for tolerances if no explicit values are states in JointTrajectory message.
 
--- a/joint_trajectory_controller/doc/userdoc.rst
+++ b/joint_trajectory_controller/doc/userdoc.rst
@@ -175,8 +175,10 @@ open_loop_control (boolean)
 
   Default: false
 
-allow_nonzero_velocity_stop (boolean)
+allow_nonzero_velocity_at_trajectory_end (boolean)
   If not set true, the last velocity point has to be zero or the goal will be rejected.
+
+  Default: true
 
 constraints (structure)
   Default values for tolerances if no explicit values are states in JointTrajectory message.
--- a/joint_trajectory_controller/doc/userdoc.rst
+++ b/joint_trajectory_controller/doc/userdoc.rst
@@ -176,7 +176,7 @@ open_loop_control (boolean)
   Default: false
 
 allow_nonzero_velocity_at_trajectory_end (boolean)
-  If not set true, the last velocity point has to be zero or the goal will be rejected.
+  If false, the last velocity point has to be zero or the goal will be rejected.
 
   Default: true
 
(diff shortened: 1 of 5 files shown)


destogl:
> As suggested by @bmagyar, the new trajectory should be even ignored if the last velocity point is nonzero.

I don't agree with this at all. If we have velocity only input, we want to continue moving between trajectories. This is not so unusual scenario, e.g., when we have joystick input. I would rather like to see a timeout if we have velocity-only input to stop the input.

christophfroehlich:
Ok, I did not think of using JTC with Joystick, but rejecting it always did not seem to make sense for me too.
So you mean we should even not hold the position automatically but only after some timeout reached?

destogl:
> So you mean we should even not hold the position automatically but only after some timeout reached?

Exactly, I think we want to avoid run away if only velocity is used.

christophfroehlich:
now I'm confused :D maybe we should discuss this in the next WG meeting
I think there are different use-cases, and you have others in mind than I do..

BarisYazici:
any update merging this PR?

christophfroehlich:
We recently decided to merge several smaller changes (#567 #558 and some optional time-out for topic interface) for JTC into a feature branch, test them thoroughly together, and merge them then together into master. 
I fear that this will take some more weeks to happen.

christophfroehlich:
@BarisYazici you could help out by reviewing (and hopefully approving) this PR. I think everyone can do this without special permissions (goto Files changed -> Review Changes on top right corner)

christophfroehlich:
I've split this PR into two, the more important one is #608

christophfroehlich:
the order listed in #607 would be fine

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -175,6 +175,9 @@ open_loop_control (boolean)
 
   Default: false
 
+allow_nonzero_velocity_stop (boolean)
```
destogl, line 178: Proposal:
```suggestion
allow_nonzero_velocity_at_trajectory_end (boolean)
```

Current name sounds to me to have inverse logic, then the description says.

christophfroehlich: You are right -> I pushed a commit renaming the parameter.

saikishor (approved):

christophfroehlich:
The failing test is not related to this PR, but shows the flakiness of the JTC tests 😕

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -66,6 +66,11 @@ joint_trajectory_controller:
       one_of<>: [["splines", "none"]],
     }
   }
+  allow_nonzero_velocity_at_trajectory_end: {
+    type: bool,
+    default_value: true,
+    description: "If not set true, the last velocity point has to be zero or the goal will be rejected",
```
bmagyar, line 72:
```suggestion
    description: "If false, the last velocity point has to be zero or the goal will be rejected",
```

joint_trajectory_controller/doc/userdoc.rst
```diff
@@ -175,6 +175,11 @@ open_loop_control (boolean)
 
   Default: false
 
+allow_nonzero_velocity_at_trajectory_end (boolean)
+  If not set true, the last velocity point has to be zero or the goal will be rejected.
```
bmagyar, line 179:
```suggestion
  If false, the last velocity point has to be zero or the goal will be rejected.
```

PR #569, opened 2023-04-12
Renovate load controller tests

bmagyar:
Should fix https://github.com/ros-controls/ros2_control/issues/825


bmagyar:
So what this test renovation uncovered is that the load admittance controller was never going to work with this setup or has been broken for quite a while.

* A bunch of params that need to be set are not set in the config. It's funny that the config for the loading test even says `  # contains minimal parameters that need to be set to load controller` :D 
* Once using the full `test_admittance_controller` config for the load test, things still fail as it is now missing the robot description. Is it really necessary to require all these things at *load time* ? I mean the robot description is an easy ask but grabbing parameters etc are meant to be in `configure`, not `load`.

destogl:
> Once using the full test_admittance_controller config for the load test, things still fail as it is now missing the robot description. Is it really necessary to require all these things at load time ?

Yes, yes and yes, we have caught this few weeks ago, and it should be solved herein #547.

Isn't it?

In #560 @guihomework was updating tests and afaik with fix from #547 things should be fine.

destogl:
To clarify the process.

1. Parameter structures are created in `on_init`.
2. Parameters are read / updated in `on_configure`.

What here may be confusing (but I am not sure about it). `generate_parameter_library` is potentially expecting all parameters to be available when parameter object is created. I think that then all internal parameters checks are done. Maybe we have to fix this logic there and do parameters check when we update them.

admittance_controller/test/test_load_admittance_controller.cpp
```diff
@@ -33,8 +33,9 @@ TEST(TestLoadAdmittanceController, load_controller)
       ros2_control_test_assets::minimal_robot_urdf),
     executor, "test_controller_manager");
 
-  ASSERT_NO_THROW(
-    cm.load_controller("load_admittance_controller", "admittance_controller/AdmittanceController"));
+  ASSERT_NE(
+    cm.load_controller("load_admittance_controller", "admittance_controller/AdmittanceController"),
+    nullptr);
```
bmagyar, line 38:
```suggestion
  ASSERT_NE(
    cm.load_controller("load_admittance_controller", "admittance_controller/AdmittanceController"),
    nullptr);
```

gripper_controllers/test/test_load_gripper_action_controllers.cpp
```diff
@@ -32,10 +32,14 @@ TEST(TestLoadGripperActionControllers, load_controller)
...
-  ASSERT_NO_THROW(cm.load_controller(
-    "test_gripper_action_effort_controller", "effort_controllers/GripperActionController"));
+  ASSERT_NE(
+    cm.load_controller(
+      "test_gripper_action_position_controller", "position_controllers/GripperActionController"),
+    nullptr);
+  ASSERT_NE(
+    cm.load_controller(
+      "test_gripper_action_effort_controller", "effort_controllers/GripperActionController"),
+    nullptr);
```
destogl, line 42: Why twice here?

admittance_controller/test/test_load_admittance_controller.cpp
```diff
@@ -33,8 +33,9 @@ TEST(TestLoadAdmittanceController, load_controller)
       ros2_control_test_assets::minimal_robot_urdf),
     executor, "test_controller_manager");
 
-  ASSERT_NO_THROW(
-    cm.load_controller("load_admittance_controller", "admittance_controller/AdmittanceController"));
+  ASSERT_EQ(
```
destogl, line 36: Please also open an follow-up

destogl (approved):

admittance_controller/test/test_load_admittance_controller.cpp
bmagyar: Already done: https://github.com/ros-controls/ros2_controllers/issues/583

gripper_controllers/test/test_load_gripper_action_controllers.cpp
```diff
@@ -32,10 +32,10 @@ TEST(TestLoadGripperActionControllers, load_controller)
...
     executor, "test_controller_manager");
 
-  ASSERT_NO_THROW(cm.load_controller(
-    "test_gripper_action_position_controller", "position_controllers/GripperActionController"));
-  ASSERT_NO_THROW(cm.load_controller(
-    "test_gripper_action_effort_controller", "effort_controllers/GripperActionController"));
+  ASSERT_NE(
+    cm.load_controller(
+      "test_gripper_action_position_controller", "position_controllers/GripperActionController"),
+    nullptr);
```
bmagyar, line 38:
```suggestion
    nullptr);
  ASSERT_NE(
    cm.load_controller(
      "test_gripper_action_position_controller", "effort_controllers/GripperActionController"),
    nullptr);
```

bmagyar: This had the other variant in here too, hence the "duplication"

gwalck:
> To clarify the process.
> 
>     1. Parameter structures are created in `on_init`.
> 
>     2. Parameters are read / updated in `on_configure`.
> 
> 
> What here may be confusing (but I am not sure about it). `generate_parameter_library` is potentially expecting all parameters to be available when parameter object is created. I think that then all internal parameters checks are done. Maybe we have to fix this logic there and do parameters check when we update them.

Hi, I am @guihomework  with a new account. While working on a new controller generation using `RosTeamWorkspace`, I faced the issue of a controller not passing the **renovated load test** due to a **mandatory** parameter missing.

My opinion on the problem is that the previous `ASSERT_NO_THROW` did not catch the param loading issues, but it still asserted that the plugin was found and could be loaded, otherwise there is a throw (I just faced pluginlib not finding the lib due a badly set xml lib path). The missing param generates an exception that is caught in the `on_init` which then fails with a `nullptr` but does not throw. 

I don't think the _load test_ should require the minimum mandatory parameters and additionally succeed the `on_init` call, since the loading of the plugin is what we want to test there, and `on_init` is checked in a second test that has the mandatory parameters.

christophfroehlich:
@bmagyar should this be backported also to humble? #670 has conflicts because of this PR missing in humble.

bmagyar:
let's try!

bmagyar:
@mergifyio backport humble

Issue #549, opened 2023-03-22
diff_drive_controller_parameter.yaml "open_loop" description is wrong


cangozpi:
I belive the description of the _open\_loop_ parameter is given wrong. It says __"If set to false the odometry of the robot will be calculated from the commanded values and not from feedback."__. I believe it supposed to mean the opposite, and it is misleading. 

* _open\_loop_ parameters description can be found here: https://github.com/ros-controls/ros2_controllers/blob/master/diff_drive_controller/src/diff_drive_controller_parameter.yaml
* How the code actually behaves can be found in: https://github.com/ros-controls/ros2_controllers/blob/master/diff_drive_controller/src/diff_drive_controller.cpp in and in line 145: it follows by 
``` cpp
if (params_.open_loop)
  {
    odometry_.updateOpenLoop(linear_command, angular_command, time);
  }
  else
  {
...}
```
which I believe does the opposite of what is described.

I believe a correction to this would require a fix on the demo code for the differential_control by negating the value of the _open\loop_ parameter in the configuration yaml file.

PR #570, opened 2023-04-14
[DiffDriveController] Adjusted open_loop parameter description

muritane:
The issue is described here: #549
The description does not reflect the code:
Description: "If set to false the odometry of the robot will be calculated from the commanded values and not from feedback."
The code: https://github.com/ros-controls/ros2_controllers/blob/master/diff_drive_controller/src/diff_drive_controller.cpp#L145-L152

open_loop_param


destogl (approved):
Thanks!

bmagyar (approved):

bmagyar:
@mergifyio backport humble

muritane:
@bmagyar thanks for merging, can you please close then the corresponding issue #549

Issue #462, opened 2022-11-14
Test all JTC parameters to be set properly from YAML file when using `generate_parameters_library`


destogl:
## Background

Add all parameters from JTC into [the example/test YAML file](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/test/config/test_joint_trajectory_controller.yaml).

To add all parameters check the definition of parameters in [this YAML file](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml) and add them with non-default values.

Also, don't forget to adjust test to still work correctly.

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_controllers/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/foxy/Tutorials/Workspace/Creating-A-Workspace.html), for Step3 use "Download Source Code" section with [these instructions](https://ros-controls.github.io/control.ros.org/getting_started.html#compiling).

- [ ] 🍴 **Fork the repository** using the handy button at the top of the repository page and **clone** it into `~/ws_ros2_control/src/ros-controls/ros2_controllers`, [here is a guide that you can follow](https://guides.github.com/activities/forking/) (You will have to remove or empty the existing `ros2_controllers` folder before cloning your own fork)

- [ ] **Checkout a new branch** using `git checkout -b <branch_name>`

- [ ] 🤖 **Apply `pre-commit`** auto formatting, by running `pip3 install pre-commit` and running `pre-commit install` in the ros2_control repo.

- [ ] 💾 **Commit and Push** your changes

[...]

PR #572, opened 2023-04-16
added JTC test parameters to yaml file

RamCharanThota:
Hi, I have added JTC test parameters to yam file here and submitting for your review. you can find the link to respective issue here #462


joint_trajectory_controller/test/config/test_joint_trajectory_controller.yaml
```diff
@@ -10,3 +10,33 @@ test_joint_trajectory_controller:
...
+  open_loop_control: true
+
+  allow_integration_in_goal_trajectories: true
+
+  action_monitor_rate: 30.0
+
+  interpolation_method: "splines"
+
+  gains:
+    __map_joints:
```
christophfroehlich, line 29: `__map_joints` is only a keyword for the GenerateParameterLibrary, you have to create a map for every controlled joint. E.g., see the online docs https://control.ros.org/master/doc/ros2_controllers/joint_trajectory_controller/doc/userdoc.html

joint_trajectory_controller/test/config/test_joint_trajectory_controller.yaml
```diff
@@ -10,3 +10,33 @@ test_joint_trajectory_controller:
...
+        i: 1.0
+        d: 1.0
+        i_clamp: 10.0
+        ff_velocity_scale: 0.0
+        normalize_error: 0.02
+
+  constraints:
+     stopped_velocity_tolerance: 0.03
+     goal_time: 0.012
+     __map_joints:
```
christophfroehlich, line 40: same as above

joint_trajectory_controller/test/config/test_joint_trajectory_controller.yaml
```diff
@@ -10,3 +10,33 @@ test_joint_trajectory_controller:
   state_interfaces:
     - position
     - velocity
+
+  command_joints:
```
christophfroehlich, line 14: Shouldn't throw this an error, if the size of `command_joints` is different than `joints`? IMHO controller chaining is not implemented for JTC yet, you might have a look if this parameter is used and what is expected there.

christophfroehlich (changes requested):
Congratulations to your first PR ;) 

Please add a link to the issue you are working on in the PR description.

christophfroehlich:
I think the purpose of the issue was also to test if the parameters were set correctly the the given non-default values (@destogl ?)
One possibility: You might add a new test in `test_trajectory_controller.cpp`, maybe called 
`TEST_P(TrajectoryControllerTestParameterized, test_non_default_parameters)` 
and see if the parameters were set correctly. `Params params_` is a protected variable and should be accessible from `std::shared_ptr<TestableJointTrajectoryController> traj_controller_`

bmagyar:
> I think the purpose of the issue was also to test if the parameters were set correctly the the given non-default values (@destogl ?)

Be mindful of overtesting too. We don't necessarily need to test that gen params library works correctly in the controller. I think general controller behaviour with no defaults is the key here

joint_trajectory_controller/test/config/test_joint_trajectory_controller.yaml
RamCharanThota: Thank you @christophfroehlich . I am updating it based on your suggestion. I have a question, Do I need add joints given under "command_joints" too if those joints are different from "joints". 
For example:
joints:
    - joint1
    - joint2
command_joints:
    - joint3
in this case, do i need to create map for joint1,joint2 and joint3?. I am confused between joints and command_joints. I really appreciate if you can explain a bit. Thank you

RamCharanThota: I defined the parameters as follows. 
```
gains:
    __map_joints:
        joint1:
          p: 10.0
          i: 1.0
          d: 1.0
          i_clamp: 10.0
          ff_velocity_scale: 0.0
          normalize_error: 0.02
        joint2:
          p: 10.0
          i: 1.0
          d: 1.0
          i_clamp: 10.0
          ff_velocity_scale: 0.0
          normalize_error: 0.02
```

is it correct format?

christophfroehlich: Not exactly, it works like this way: (and normalize_error is a boolean)

```
gains:
    joint1:
      p: 10.0
      i: 1.0
      d: 1.0
      i_clamp: 10.0
      ff_velocity_scale: 0.0
      normalize_error: false
    joint2:
      p: 10.0
      i: 1.0
      d: 1.0
      i_clamp: 10.0
      ff_velocity_scale: 0.0
      normalize_error: false
```

christophfroehlich: > Thank you @christophfroehlich . I am updating it based on your suggestion. I have a question, Do I need add joints given under "command_joints" too if those joints are different from "joints". For example: joints: - joint1 - joint2 command_joints: - joint3 in this case, do i need to create map for joint1,joint2 and joint3?. I am confused between joints and command_joints. I really appreciate if you can explain a bit. Thank you

have a look at this code here. It has to have the same size.

https://github.com/ros-controls/ros2_controllers/blob/48cc202108d4c91af2556176cae0c796f31d40fa/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L559-L572

I'm not sure actually if the tests use the yaml or hardcode the parameters in every case
https://github.com/ros-controls/ros2_controllers/blob/48cc202108d4c91af2556176cae0c796f31d40fa/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp#L129

If your yaml file does not throw an error: I suggest you investigating this to be sure, starting here:
https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/test/test_trajectory_controller.cpp#L368

RamCharanThota: @christophfroehlich I went through the files you mentioned earlier. I think I have got an Idea of what's going on here. I want to double check with you whether my understand is correct or not. In this particular issue, we need to read and assign the trajectory controller param values from yaml file. We are using generate_param_library to generate header files for this parameters in yaml and we are using these param header files to retrieve and assign the values to trajectory controller parameter. And our test need to verify whether assigned trajectory controller parameters values properly matched with non default values given in the respective yaml file.

And i think you are right, we need to create new test case as you mentioned below. Also, you mentioned that I can use "param_" variable  to retrieve the parameter values using traj_controller_ object. But, I think, First, we have to  assign the values from header file to traj_controller_ using the following function in our test case
https://github.com/ros-controls/ros2_controllers/blob/48cc202108d4c91af2556176cae0c796f31d40fa/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp#L195-L216

I really appreciate it if you can have look and let me whether my understand is right or wrong. If  my approach is right
1) how can i create " std::vector<rclcpp::Parameter> & parameters = {}" argument in the below function  by using yaml parameters header files created by generate param library
https://github.com/ros-controls/ros2_controllers/blob/48cc202108d4c91af2556176cae0c796f31d40fa/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp#L195

2) And also how can i retrieve protected "param_" variable using traj_controller_ object when there is no getter function for that variable.

Thank you and looking forward to your reply. have a good day and sorry for long question :)

christophfroehlich: As @bmagyar has pointed out, we don't have to test if the values are read correctly from generate_parameter_library (we assume that this dependency works as expected), but test if JTC works with non-default parameters. 
Maybe @destogl can comment in the original issue which tests exactly he had in mind.
As a prerequisite for that we have to understand how we can load/use the yaml file in the tests -> I honestly also have to see and play around how this works.

bmagyar:
I naively marked this ready for review, is that ok?

christophfroehlich:
> I naively marked this ready for review, is that ok?

I still don't know what exactly was the idea of @destogl for the (missing?) tests and how to parse the yaml file from the tests.

@RamCharanThota The CI tests fail, because it does not even compile. Furthermore, please have a look at the [contribution guidelines](https://github.com/ros-controls/ros2_control/blob/master/CONTRIBUTING.md#contributing-via-pull-requests) on how to install pre-commit hooks for automatic formating.

RamCharanThota:
> > I naively marked this ready for review, is that ok?
> 
> I still don't know what exactly was the idea of @destogl for the (missing?) tests and how to parse the yaml file from the tests.
> 
> @RamCharanThota The CI tests fail, because it does not even compile. Furthermore, please have a look at the [contribution guidelines](https://github.com/ros-controls/ros2_control/blob/master/CONTRIBUTING.md#contributing-via-pull-requests) on how to install pre-commit hooks for automatic formating.

I already installed the pre-commit hooks. But, I will double check. Thank you

RamCharanThota:
> > I naively marked this ready for review, is that ok?
> 
> I still don't know what exactly was the idea of @destogl for the (missing?) tests and how to parse the yaml file from the tests.
> 
> @RamCharanThota The CI tests fail, because it does not even compile. Furthermore, please have a look at the [contribution guidelines](https://github.com/ros-controls/ros2_control/blob/master/CONTRIBUTING.md#contributing-via-pull-requests) on how to install pre-commit hooks for automatic formating.

Whenever I tried to commit. I am getting the following error

amet_cpp_error

any idea how can i solve it. I am not doing anything complex. I am just declaring a function
function_declaration

christophfroehlich:
I always use `export AMENT_CPPCHECK_ALLOW_SLOW_VERSIONS=true` to get that running.

joint_trajectory_controller/test/config/test_joint_trajectory_controller.yaml
RamCharanThota: Hi @christophfroehlich , I am still struggling to understand what needs to be done exactly in this test and how to retrieve the parameters. I wondering whether you get the chance to have a look at it?

RamCharanThota:
> This pull request is in conflict. Could you fix it @RamCharanThota?

Hi @mergify

PR #577, opened 2023-04-20
[CI] Switch to recommended versions of ros-tooling

destogl:
See here: https://github.com/ros-tooling/action-ros-lint#run-generic-and-python-linters


bmagyar:
These used to be quite flaky back then. They ok now?

PR #578, opened 2023-04-23
[JTC] Implement new ~/controller_state message

christophfroehlich:
This is the humble version of #557, where the old msg format and topic is kept for legacy reasons.

I added a deprecation warning if one subscribes to the old `~/state` topic

Needs https://github.com/ros-controls/control_msgs/pull/87


destogl (approved):
Very nice! Thanks!!!

bmagyar (commented):
Thank you!

PR #585, opened 2023-04-27
Fix wrong publish timestamp initialization

Noel215:
With the current implementation the `previous_publish_timestamp_` variable was initialized in the `on_configure` method.

However, when using the simulation, the time source has not been always updated to the ros time and it's initialized to the system time, so the condition for publishing the odometry and the related tf is never satisfied:

`system_time` + `publish_period_` is always  > `simulation_time`

With this fix, `previous_publish_timestamp_` starts without an uninitialized clock and whenever the time source changes, it is initialized/changed.

The odometry will be published in 2 cases:
- Whenever the time source is changed
- If the publish_rate condition is satisfied

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -185,7 +185,23 @@ controller_interface::return_type DiffDriveController::update(
   tf2::Quaternion orientation;
   orientation.setRPY(0.0, 0.0, odometry_.getHeading());
 
-  if (previous_publish_timestamp_ + publish_period_ < time)
+  bool should_publish = false;
+  try
+  {
+    if (previous_publish_timestamp_ + publish_period_ < time)
+    {
+      previous_publish_timestamp_ += publish_period_;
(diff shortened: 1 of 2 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -185,7 +185,23 @@ controller_interface::return_type DiffDriveController::update(
...
+  bool should_publish = false;
+  try
+  {
+    if (previous_publish_timestamp_ + publish_period_ < time)
+    {
+      previous_publish_timestamp_ += publish_period_;
+      should_publish = true;
+    }
+  }
+  catch (const std::runtime_error)
```
bmagyar, line 197: could you please document that this is to handle time source changes & exceptions from that?

Noel215: Better now ?

bmagyar (approved):
Thanks a lot!

bmagyar:
@mergifyio backport humble

PR #586, opened 2023-04-28
Fix wrong publish timestamp initialization

jmguerreroh:
Using the solution proposed in this PR by @Noel215 to fix timestamp initialization in humble branch: https://github.com/ros-controls/ros2_controllers/pull/585

Regards


christophfroehlich:
Thanks for your contribution, but we normally accept PRs to the master branch only, and backport them after merge. So there is no need to do a manual backport to other branches.

bmagyar:
As per #593

PR #595, opened 2023-05-04
switch from dash to underscore in setup.cfg

moriarty:
This fixes the warning which was introduced here:
https://github.com/pypa/setuptools/commit/a2e9ae4cb


bmagyar (approved):
Thanks for the fix and also the reference!

bmagyar:
@mergifyio backport humble

PR #596, opened 2023-05-04
Clear registered handles of DiffDriveController on deactivate

Noel215:
Hi,

With the current implementation, the wheel handles are not being reset when running `on_deactivate` method. Then, if the controller is deactivated and then activated again the handles are accumulated and that can lead to issues (See https://github.com/ros-controls/ros2_control_demos/issues/239)

`on_activate` and `on_deactivate` are the opposite, so everything that is enabled in `on_activate` have to be disabled in `on_deactivate` (or reset at the beginning of the `on_activate`). Following that, I propose two possible solutions, let me know which one fits better:

**Current Solution**:  Reset the handles with `registered_handles.clear();` in the `configure_side()` method.

**Alternative Solution**:  Clear each of the handles in `on_deactivate` method:
```
controller_interface::CallbackReturn DiffDriveController::on_deactivate(
{
  registered_left_wheel_handles_.clear();
  registered_right_wheel_handles_.clear();
  return controller_interface::CallbackReturn::SUCCESS;
}
```

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -481,6 +481,7 @@ controller_interface::CallbackReturn DiffDriveController::on_deactivate(
   const rclcpp_lifecycle::State &)
 {
   subscriber_is_active_ = false;
+  halt();
   registered_left_wheel_handles_.clear();
   registered_right_wheel_handles_.clear();
   return controller_interface::CallbackReturn::SUCCESS;
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -482,6 +482,11 @@ controller_interface::CallbackReturn DiffDriveController::on_deactivate(
 {
   subscriber_is_active_ = false;
   halt();
+    if (!is_halted)
+    {
+      halt();
+      is_halted = true;
+    }
   registered_left_wheel_handles_.clear();
   registered_right_wheel_handles_.clear();
   return controller_interface::CallbackReturn::SUCCESS;
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -481,12 +481,11 @@ controller_interface::CallbackReturn DiffDriveController::on_deactivate(
   const rclcpp_lifecycle::State &)
 {
   subscriber_is_active_ = false;
-  halt();
-    if (!is_halted)
-    {
-      halt();
-      is_halted = true;
-    }
+  if (!is_halted)
+  {
+    halt();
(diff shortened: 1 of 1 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -558,6 +558,7 @@ controller_interface::CallbackReturn DiffDriveController::configure_side(
   }
 
   // register handles
+  registered_handles.clear();
```
bmagyar, line 561: should this happen on deactivate?

Noel215: That is being executed on activate, before creating the handles.
I can change it to use the other solution if you think it's better.

bmagyar:
Well those handles should be returned from loan on deactivate so I think
it's only appropriate ;)

bmagyar (approved):

bmagyar:
Houston, we have a problem. Some tests are failing

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -481,6 +481,8 @@ controller_interface::CallbackReturn DiffDriveController::on_deactivate(
   const rclcpp_lifecycle::State &)
 {
   subscriber_is_active_ = false;
+  registered_left_wheel_handles_.clear();
```
christophfroehlich, line 484:
```suggestion
    if (!is_halted)
    {
      halt();
      is_halted = true;
    }
  registered_left_wheel_handles_.clear();
```

christophfroehlich (changes requested):
Thanks for pointing in this direction, I was not able to find this on my own ;) 

The tests fail now, because it is assumed that the commands are zero after deactivation
https://github.com/ros-controls/ros2_controllers/blob/546422d8639082fea5d224511505e281e8566ecf/diff_drive_controller/test/test_diff_drive_controller.cpp#L424-L425

This is now not the case, because `DiffDriveController::halt()` has no effect if no handles are registered.
https://github.com/ros-controls/ros2_controllers/blob/546422d8639082fea5d224511505e281e8566ecf/diff_drive_controller/src/diff_drive_controller.cpp#L109

Let it explicitly halt before releasing the handles?

diff_drive_controller/src/diff_drive_controller.cpp
Noel215: oh! I've just pushed, I haven't seen the comment. 
I only called `halt()`, so I'm going to apply your suggestion

christophfroehlich (approved):

bmagyar (approved):

bmagyar (approved):

bmagyar:
@mergifyio backport humble

bmagyar:
@mergifyio backport humble
#

PR #601, opened 2023-05-08
Use generate_parameter_library for all params of diff_drive_controller

christophfroehlich:
I ran over some parameters, which were parsed with `get_node()->get_parameter()` instead of using the already parsed ones from `params_` structure by generate_parameter_library. Or are there any good reasons for that?


bmagyar (approved):
Thx!

bmagyar:
@mergifyio backport humble

PR #602, opened 2023-05-08
Do not validate admittance controller parameters until configure

sea-bass:
The root cause is that the `admittance_controller/AdmittanceController` parameters do not have default values for all required parameters.

Without this change, if you try to load an admittance controller without parameters (e.g., by using `ControllerManager::load_controller()` or the corresponding ROS service), the `generate_parameter_library` validation will fail because the value of the `joints` parameter is not set rather than an empty string vector. And same for other follow-on params.

This PR now changes the logic so that we're not validating parameters and dynamically allocating data based on number of joints until `on_configure()` -- it previously happened on `on_init()`.

### DIFF
--- a/admittance_controller/src/admittance_controller.cpp
+++ b/admittance_controller/src/admittance_controller.cpp
@@ -33,31 +33,6 @@ namespace admittance_controller
 {
 controller_interface::CallbackReturn AdmittanceController::on_init()
 {
-  // initialize controller config
-  try
-  {
-    parameter_handler_ = std::make_shared<admittance_controller::ParamListener>(get_node());
-    admittance_ = std::make_unique<admittance_controller::AdmittanceRule>(parameter_handler_);
-  }
-  catch (const std::exception & e)
-  {
-    RCLCPP_ERROR(
-      get_node()->get_logger(), "Exception thrown during init stage with message: %s \n", e.what());
-    return controller_interface::CallbackReturn::ERROR;
-  }
-
-  // number of joints in controllers is fixed after initialization
-  num_joints_ = admittance_->parameters_.joints.size();
-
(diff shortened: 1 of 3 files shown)


aprotyas (approved):

sea-bass:
Sorry, I did this wrong. Actually, many other parameters require defaults.

bmagyar:
Hmm.... I think default values are only sensible if we can have good, usable values there. Joint names or srte/command interfaces being empty doesn't really leaves us with a functional controller... 

Some of these defaults (numerical ones) are probably ok but empty strings/empty vectors likely not.

destogl:
Here is the problem somewhere else. That is in the `generate_parameter_library`. In `on_init` (called after load) we only create objects to access parameters. They are accessed later in  `on_configure`. @tylerjw can we add an option to not get and check parameters when creating the object? I see that this is convenient in many applications, but when using lifecycle it can be a disturbance.

destogl:
@sea-bass This is great that you are doing that, but it's not a solution we will merge. Because is just hides issues behind. We should focus on the fixing issues behind not their consequences.

sea-bass:
I agree, and I'm sure there is a better way of doing this.

To get back to my issue which is why I attempted this PR, is there a way to run `ControllerManager::load_controller()` in a way that allows me to specify a parameters file (or even manually specified parameters) when the admittance controller is initialized?

destogl:
Usually, you use a spawner with `--load-only` and then also send parameter file to the spawner.

sea-bass:
I see, thanks!

I was looking for a possible solution that was free of launch files since I wanted more of a unit test than an integration test... but we found a different workaround.

Will go ahead and close this.

PR #604, opened 2023-05-10
Fix github links on control.ros.org

christophfroehlich:
This is a companion PR to https://github.com/ros-controls/control.ros.org/pull/85. More details can be found there.


bmagyar (approved):
New files without this or ill-defined paths still won't break the page, but only the URL, right?

christophfroehlich:
Exactly, if `github_url` is not set it uses the autogenerated one for the link

https://github.com/readthedocs/sphinx_rtd_theme/blob/03a86b803cf7cf281748ef94345ba7c35953952a/sphinx_rtd_theme/breadcrumbs.html#L35-L40

christophfroehlich:
@bmagyar pls backport to humble, too

bmagyar:
https://github.com/Mergifyio backport humble

PR #510, opened 2023-01-26
Fix overriding of install

tylerjw:
This is related to:
- ros-controls/kinematics_interface#13
- ros-controls/ros2_control#926
- ros-controls/realtime_tools#105

Here are the colcon docs on overriding: https://colcon.readthedocs.io/en/released/user/overriding-packages.html#how-to-make-it-easier-for-your-users-to-override

Here is the ament_cmake user guide: https://docs.ros.org/en/rolling/How-To-Guides/Ament-CMake-Documentation.html

PR #605, opened 2023-05-11
Fix overriding of install (backport of #510)

fmauch:
As done in #510 this installs the headers correctly on Humble. I resolved one merge request, but this PR does update the required CMake version, as well. I hope this is fine, otherwise please let me know.

As discussed with @destogl this would be nice to have in Humble, as well, see https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/656.


bmagyar (approved):
Thanks for taking the time!

Issue #407, opened 2022-08-07
[JTC] Add new test to ensure that controller goes into position holding when tolerances are violated 


bmagyar:
[...] Would it be possible to write a small test for this behavior? We would then feel better in the future to know that behavior is not broken (again).

_Originally posted by @destogl in https://github.com/ros-controls/ros2_controllers/pull/368#pullrequestreview-1019716551_

PR #607, opened 2023-05-11
[JTC] Fix tests and set_hold_position

christophfroehlich:
This a summary PR for the `jtc_features` branch, where individual PRs shall be tested together before merging into master

Independent of other PRs (order of merge is not relevant)
- #558
- #613 

Completed:
- #699

These should fix (amongst others)
- #407 
- #469 
- #514
- #683

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_controller.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller.cpp
@@ -65,6 +65,8 @@ TEST_P(TrajectoryControllerTestParameterized, configure_state_ignores_commands)
 {
   rclcpp::executors::MultiThreadedExecutor executor;
   SetUpTrajectoryController(executor);
+  traj_controller_->get_node()->set_parameter(
+    rclcpp::Parameter("allow_nonzero_velocity_at_trajectory_end", true));
 
   // const auto future_handle_ = std::async(std::launch::async, spin, &executor);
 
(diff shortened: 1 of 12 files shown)


joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -338,7 +225,10 @@ TEST_P(TrajectoryControllerTestParameterized, cleanup)
 
   state = traj_controller_->get_node()->cleanup();
   ASSERT_EQ(State::PRIMARY_STATE_UNCONFIGURED, state.id());
+
   // update for 0.25 seconds
+  // TODO(anyone): should the controller even allow calling update() when it is not active?
+  // update loop receives a new msg and updates accordingly
   updateController(rclcpp::Duration::from_seconds(0.25));
```
destogl, line 232: This should be clearly removed. This is something that it doesn't happen in the real world using CM - so we should not test it at all.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -428,25 +318,28 @@ TEST_P(TrajectoryControllerTestParameterized, correct_initialization_using_param
...
   }
 
-  // cleanup
-  state = traj_controller_->get_node()->cleanup();
-  ASSERT_EQ(State::PRIMARY_STATE_UNCONFIGURED, state.id());
+  // reactivated
+  // wait so controller process the third point when reactivated
+  std::this_thread::sleep_for(std::chrono::milliseconds(3000));
+  // TODO(anyone) test copied from ROS 1: it fails now!
+  //  should the old trajectory really be processed after reactivation?
```
destogl, line 325: No, I think after reactivation, we should remove the trajectory that was commanded. It is basically “a reset”

christophfroehlich: Fixed with bceaf96e93dd2b14c3f1ee3ec00847ac4d41f5cb in #705

bmagyar:
Question: now that #705 is merged, are there any changes/functionality on this branch that needs to be merged prior to #558? I'm thinking we may not need the `jtc-features` branch anymore at all...

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: And even more with 0bf984488008d8a4e1bd2b26fd319fc870338cde in #558

PR #608, opened 2023-05-11
[JTC] Hold the position after the goal was reached successfully

christophfroehlich:
JTC with command interface `velocity`: As described in #469, JTC does not stop if last velocity point is not zero.

I propose calling `set_hold_position` if the goal was successfully reached. 

Optional rejection of nonzero velocity is proposed with #567 

- [ ]  Changes from #558 are necessary to have an effect.
- [ ]  Add tests to check this behavior


christophfroehlich:
merged into #558

Issue #168, opened 2021-04-08
Lock-free Write to Buffer


matthew-reynolds:
A long-standing shortcoming in the realtime_tools::RealtimeBuffer is an inability to provide lock-free writes from the RT context. (See https://github.com/ros-controls/realtime_tools/issues/14).

The `joint_trajectory_controller` in particular is currently using `writeFromNonRT()` within the RT loop due to this limitation. Either through an upstream change (maybe https://github.com/ros-controls/realtime_tools/pull/73? I haven't had a chance to look into it yet) or some restructuring of the JTC, we need a lock-free write.

PR #609, opened 2023-05-12
[JTC] Add time-out for trajectory interfaces

christophfroehlich:
This adds an optional time-out parameter for the trajectory interfaces (topic or action).

If using velocity as command interface, and the last velocity point is not zero -> the robot would move forever without stopping if no new trajectory is received.


christophfroehlich:
Just wondering now, should the timeout be counted from the reception of the message, or from the time of the end of the trajectory (i.e., `before_last_point==false`)?

bmagyar:
Good question... It's intuitive to do it from receiving the message but since this particular message has explicit timing in it I think we should only allow timeouts defined relative to the trajectory

destogl:
We should define it from the end for the trajectory (time of the last point)

bmagyar:
@christophfroehlich this PR shouldn't really target the jtc-features branch no?

christophfroehlich:
> @christophfroehlich this PR shouldn't really target the jtc-features branch no?

It is somehow independent, but why not?

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -66,6 +66,13 @@ joint_trajectory_controller:
       one_of<>: [["splines", "none"]],
     }
   }
+  cmd_timeout: {
+    type: double,
+    default_value: 0.0, # seconds
+    description: "Timeout after which the input command is considered stale.
+     Timeout is counted from the end of the trajectory (the last point).
```
saikishor, line 73: Hello!
Thanks for the PR.
In the code, you are verifying the command timeout at every sample point right? Why not w.r.t to the total trajectory duration w.r.t to current trajectory duration time

Moreover, why not rename it to `trajectory_timeout`?, because `cmd_timeout` might give an impression about every command point of the sent goal or message.

What do you think?

christophfroehlich: The requested feature here is to stop the robot from doing dangerous things if, e.g., the communication is broken suddenly or the node sending the new trajectories is crashed
> If using velocity as command interface, and the last velocity point is not zero -> the robot would move forever without stopping if no new trajectory is received.

This is very close to the `goal_time` tolerance, where additionally it is checked if the goal itself is reached. 

For both timeouts the same time information 
```c++
      const rclcpp::Time traj_start = (*traj_point_active_ptr_)->get_trajectory_start_time();
      const rclcpp::Time traj_end = traj_start + start_segment_itr->time_from_start;
```
is used, where the iterators are updated with every call of the `sample` method. When the sample method reaches the last point, `traj_end` is the time instance of the _current trajectory duration time_ as you mentioned. This means only, that this time was never calculated explicitly. Or am I missing something?

Acc. to the naming: As the point of this PR is the lack of a new command coming in, I'll stick to `cmd_timeout`. Furthermore, this is similar to the `cmd_vel_timeout` parameter of the _diff_drive_controller_

saikishor: @christophfroehlich Thanks for the clear explanation. Yes, now the `cmd_timeout` makes sense. 

If this is the case, then I have another question. Why not use `allow_nonzero_velocity_at_trajectory_end` parameter from [your other PR](https://github.com/ros-controls/ros2_controllers/pull/567) along with a condition of `cmd_timeout > 0`, just not to trigger it if the `goal_time_tolerance` is set higher than the `cmd_timeout`.

Ideally, the timeout needs to be higher than the goal tolerance. So, maybe we need to either document it better or have a warning printed when `cmd_timeout` is lower than `goal_time_tolerance` or maybe we can consider the command timeout adding goal_time_tolerance.

christophfroehlich: You are right, both timeouts can be competing each other. I updated the docs, added a check in the `on_configure` method and test it.

saikishor: Awesome. Now, it looks good. Thank you. :)

saikishor (approved):

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
christophfroehlich: I was thinking again if my explanation is correct and extended the timeout test. It failed first, because I missed that I must not check the timeout before the last point of the message was reached. (time difference is counting from the last point reached and not from the to-be-reached point).
Fix + extended tests are pushed.

saikishor: Perfect! Now, it makes sense after using the before_last_point

christophfroehlich:
CI fails because of flaky JTC tests, not related to this PR

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -211,12 +211,32 @@ controller_interface::return_type JointTrajectoryController::update(
...
+      // have we reached the end, are not holding position, and is a timeout configured?
+      // Check independently of other tolerances
+      if (
+        !before_last_point && *(rt_is_holding_.readFromRT()) == false && cmd_timeout_ > 0.0 &&
+        time_difference > cmd_timeout_)
+      {
+        RCLCPP_WARN(get_node()->get_logger(), "Aborted due to command timeout");
+
+        traj_msg_external_point_ptr_.reset();
+        traj_msg_external_point_ptr_.initRT(set_hold_position());
```
bmagyar, line 237: do we always have to call both? I'd imagine re-setting the value resets things properly

joint_trajectory_controller/doc/parameters.rst
```diff
@@ -68,6 +68,15 @@ allow_nonzero_velocity_at_trajectory_end (boolean)
 
   Default: true
 
+cmd_timeout (double)
+  Timeout after which the input command is considered stale.
+  Timeout is counted from the end of the trajectory (the last point).
+  ``cmd_timeout`` must be greater than ``constraints.goal_time``.
```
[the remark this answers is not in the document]
bmagyar:
```suggestion
  ``cmd_timeout`` must be greater than ``constraints.goal_time``,
  otherwise ignored.
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -565,6 +565,147 @@ TEST_P(TrajectoryControllerTestParameterized, position_error_not_normalized)
...
+ */
+TEST_P(TrajectoryControllerTestParameterized, accept_correct_cmd_timeout)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+  // zero is default value, just for demonstration
+  double cmd_timeout = 3.0;
+  rclcpp::Parameter cmd_timeout_parameter("cmd_timeout", cmd_timeout);
+  rclcpp::Parameter goal_time_parameter("constraints.goal_time", 2.0);
+  SetUpAndActivateTrajectoryController(
+    executor, true, {cmd_timeout_parameter, goal_time_parameter}, false);
```
bmagyar, line 579:
```suggestion
    executor, {cmd_timeout_parameter, goal_time_parameter}, false);
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -565,6 +565,147 @@ TEST_P(TrajectoryControllerTestParameterized, position_error_not_normalized)
...
+ * @brief cmd_timeout must be greater than constraints.goal_time
+ */
+TEST_P(TrajectoryControllerTestParameterized, decline_false_cmd_timeout)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+  // zero is default value, just for demonstration
+  rclcpp::Parameter cmd_timeout_parameter("cmd_timeout", 1.0);
+  rclcpp::Parameter goal_time_parameter("constraints.goal_time", 2.0);
+  SetUpAndActivateTrajectoryController(
+    executor, true, {cmd_timeout_parameter, goal_time_parameter}, false);
```
bmagyar, line 594:
```suggestion
    executor, {cmd_timeout_parameter, goal_time_parameter}, false);
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -565,6 +565,147 @@ TEST_P(TrajectoryControllerTestParameterized, position_error_not_normalized)
...
+
+/**
+ * @brief check if no timeout is triggered
+ */
+TEST_P(TrajectoryControllerTestParameterized, no_timeout)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+  // zero is default value, just for demonstration
+  rclcpp::Parameter cmd_timeout_parameter("cmd_timeout", 0.0);
+  SetUpAndActivateTrajectoryController(executor, true, {cmd_timeout_parameter}, false);
```
bmagyar, line 607:
```suggestion
  SetUpAndActivateTrajectoryController(executor, {cmd_timeout_parameter}, false);
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -565,6 +565,147 @@ TEST_P(TrajectoryControllerTestParameterized, position_error_not_normalized)
...
+/**
+ * @brief check if timeout is triggered
+ */
+TEST_P(TrajectoryControllerTestParameterized, timeout)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+  constexpr double cmd_timeout = 0.1;
+  rclcpp::Parameter cmd_timeout_parameter("cmd_timeout", cmd_timeout);
+  double kp = 1.0;  // activate feedback control for testing velocity/effort PID
+  SetUpAndActivateTrajectoryController(executor, true, {cmd_timeout_parameter}, false, kp);
```
bmagyar, line 659:
```suggestion
  SetUpAndActivateTrajectoryController(executor, {cmd_timeout_parameter}, false, kp);
```

bmagyar (approved):
Thank you, I'll apply the comments to fix the test compilation (due to the updated test API)

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -83,6 +83,14 @@ joint_trajectory_controller:
     default_value: true,
     description: "If false, the last velocity point has to be zero or the goal will be rejected",
   }
+  cmd_timeout: {
+    type: double,
+    default_value: 0.0, # seconds
+    description: "Timeout after which the input command is considered stale.
+     Timeout is counted from the end of the trajectory (the last point).
+     cmd_timeout must be greater than constraints.goal_time.
```
bmagyar, line 91:
```suggestion
     cmd_timeout must be greater than constraints.goal_time, otherwise ignored.
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: this is a consequence of #168, I didn't know how to perform this because there is no set-method. But probably the reset() is not necessary before.

PR #612, opened 2023-05-14
[JTC] Extend tests

christophfroehlich:
In preparation for writing new tests for the upcoming changes:
- I reactivated commented tests in test_trajectory_controller
- Added parameterized tests for test_trajectory_actions.

If tests were deactivated for other reasons than being flaky, please let me know ;) 

Three tests are failing, and I don't know if the test logic or JTC behavior is broken. I added comments below

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_actions.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_actions.cpp
@@ -197,7 +197,24 @@ public:
   }
 };
 
-TEST_F(TestTrajectoryActions, test_success_single_point_sendgoal)
+// From the tutorial: https://www.sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest
+class TestTrajectoryActionsTestParameterized
+: public TestTrajectoryActions,
+  public ::testing::WithParamInterface<
+    std::tuple<std::vector<std::string>, std::vector<std::string>>>
+{
+public:
+  virtual void SetUp()
+  {
+    TestTrajectoryActions::SetUp();
+    command_interface_types_ = std::get<0>(GetParam());
+    state_interface_types_ = std::get<1>(GetParam());
+  }
+
+  static void TearDownTestCase() { TrajectoryControllerTest::TearDownTestCase(); }
+};
+
+TEST_P(TestTrajectoryActionsTestParameterized, test_success_single_point_sendgoal)
 {
   SetUpExecutor();
   SetUpControllerHardware();
@@ -223,12 +240,15 @@ TEST_F(TestTrajectoryActions, test_success_single_point_sendgoal)
   EXPECT_TRUE(gh_future.get());
   EXPECT_EQ(rclcpp_action::ResultCode::SUCCEEDED, common_resultcode_);
 
-  EXPECT_EQ(1.0, joint_pos_[0]);
-  EXPECT_EQ(2.0, joint_pos_[1]);
-  EXPECT_EQ(3.0, joint_pos_[2]);
+  if (traj_controller_->has_position_command_interface())
+  {
+    EXPECT_EQ(1.0, joint_pos_[0]);
+    EXPECT_EQ(2.0, joint_pos_[1]);
+    EXPECT_EQ(3.0, joint_pos_[2]);
+  }
 }
 
-TEST_F(TestTrajectoryActions, test_success_multi_point_sendgoal)
+TEST_P(TestTrajectoryActionsTestParameterized, test_success_multi_point_sendgoal)
 {
   SetUpExecutor();
   SetUpControllerHardware();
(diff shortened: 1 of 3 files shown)


joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -428,25 +318,28 @@ TEST_P(TrajectoryControllerTestParameterized, correct_initialization_using_param
     EXPECT_LE(0.0, joint_eff_[2]);
   }
 
-  // cleanup
-  state = traj_controller_->get_node()->cleanup();
-  ASSERT_EQ(State::PRIMARY_STATE_UNCONFIGURED, state.id());
+  // reactivated
+  // wait so controller process the third point when reactivated
+  std::this_thread::sleep_for(std::chrono::milliseconds(3000));
+  // TODO(anyone) test copied from ROS 1: it fails now!
```
christophfroehlich, line 324: I merged the old reactivation test into that one, should the old trajectory really be processed after reactivation? (it doesn't)

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -739,826 +627,847 @@ TEST_P(TrajectoryControllerTestParameterized, position_error_normalized)
...
+  // Wait until the end start and end of partial traj
+
+  expected_actual.positions = {partial_traj.back()[0], partial_traj.back()[1], full_traj.back()[2]};
+  expected_desired = expected_actual;
+
+  waitAndCompareState(
+    expected_actual, expected_desired, executor, rclcpp::Duration(delay * (2 + 2)), 0.1);
+}
+
+// TODO(destogl) this test fails with errors
```
christophfroehlich, line 1176: @destogl maybe you can have a quick look on that?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -739,826 +627,847 @@ TEST_P(TrajectoryControllerTestParameterized, position_error_normalized)
...
+    // Finally the first goal will be commanded/reached
+    updateController(rclcpp::Duration::from_seconds(1.1));
+    EXPECT_NEAR(first_goal[0], joint_pos_[0], COMMON_THRESHOLD);
+  }
+
+  executor.cancel();
+}
+#endif
+
+// TODO(destogl) this test fails
```
christophfroehlich, line 1281: @destogl maybe you can have a quick look on that?

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
```diff
@@ -363,7 +365,13 @@ class TrajectoryControllerTest : public ::testing::Test
     const auto end_time = start_time + wait_time;
     while (clock.now() < end_time)
     {
-      traj_controller_->update(clock.now(), clock.now() - start_time);
+      // TODO(christophfroehlich): use the node's clock here for internal comparison
+      // if using RCL_STEADY_TIME ->
+      //   C++ exception with description
+      //   "can't compare times with different time sources" thrown in the test body.
+      // traj_controller_->update(clock.now(), clock.now() - start_time);
```
christophfroehlich, line 372: should we create the node_ with `NodeOptions`, `clock_type = RCL_STEADY_TIME`?

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -197,7 +197,24 @@ class TestTrajectoryActions : public TrajectoryControllerTest
   }
 };
 
-TEST_F(TestTrajectoryActions, test_success_single_point_sendgoal)
+// From the tutorial: https://www.sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest
+class TestTrajectoryActionsTestParameterized
+: public TestTrajectoryActions,
+  public ::testing::WithParamInterface<
```
bmagyar, line 203: We also did tests this way at work a few years ago when we got fascinated by the idea of tuples in C++ but you can get much more readable and maintainable code with good old structs (also sometimes referred to as named tuples ;))

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
bmagyar: Some of these may be from a time we didn't have any clock solution, worth reassessing if needed

joint_trajectory_controller/test/test_trajectory_controller.cpp
bmagyar: It shouldn't imo and we should have a test to ensure it

bmagyar:
Conforming here that the only reason I remember tests were disabled was flakiness

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: yes, I agree. Would be strange if it does. Writing the test for that is easy, because this one fails already :D

joint_trajectory_controller/test/test_trajectory_actions.cpp
christophfroehlich: Regarding tests I'm only a copy and paste engineer :yum: I just copied it from the other test file, but if you give me a hint on how it can be improved, I can change that..

bmagyar (approved):

PR #613, opened 2023-05-14
[JTC] Tolerance tests + Hold on time violation

### LINKED ISSUE #407: [JTC] Add new test to ensure that controller goes into position holding when tolerances are violated

christophfroehlich:
I realized that if using the topic interface: The current trajectory will never be stopped if the goal isn't reached within time. 

New tests fix #407 but need
- [x] #558 to succeed, because `set_hold_position` does not set hold position now

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -376,7 +376,7 @@ controller_interface::return_type JointTrajectoryController::update(
         !before_last_point && !within_goal_time && *(rt_has_pending_goal_.readFromRT()) == false)
       {
         RCLCPP_ERROR(
-          get_node()->get_logger(), "Holding position due goal_time_tolerance exceeding");
+          get_node()->get_logger(), "Exceeded goal_time_tolerance: holding position...");
 
         traj_msg_external_point_ptr_.reset();
         traj_msg_external_point_ptr_.initRT(set_hold_position());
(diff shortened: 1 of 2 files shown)


christophfroehlich:
CI fails because of flaky JTC tests, not related to this PR

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -362,19 +362,28 @@ controller_interface::return_type JointTrajectoryController::update(
...
         RCLCPP_ERROR(get_node()->get_logger(), "Holding position due to state tolerance violation");
 
         traj_msg_external_point_ptr_.reset();
         traj_msg_external_point_ptr_.initRT(set_hold_position());
       }
+      else if (
+        !before_last_point && !within_goal_time && *(rt_has_pending_goal_.readFromRT()) == false)
+      {
+        RCLCPP_ERROR(
+          get_node()->get_logger(), "Holding position due goal_time_tolerance exceeding");
```
bmagyar, line 379:
```suggestion
          get_node()->get_logger(), "Exceeded goal_time_tolerance: holding position...");
```

bmagyar (approved):

PR #616, opened 2023-05-16
[Parameters] Avoid deprecation warnings."

destogl:
Maybe not suitable since ABI breaking…


bmagyar (approved):
IMO this should be fine. Controllers are not really intended to be inherited, even though it happens from time to time.

Issue #642, opened 2023-06-02
[rqt jtc] fails to receive state since new jtc ctrl msgs


gwalck:
**Describe the bug**

following https://github.com/ros-controls/ros2_controllers/pull/557 jtc is not working anymore to read feedback, and failed at sending commands

```
KeyError: 'position'
Traceback (most recent call last):
  File "~/workspaces/ws/build/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py", line 423, in _update_cmd_cb
    pos = self._joint_pos[name]["position"]
```

**Expected behavior**
RQT JTC should get its information from the new "controller_state" topic in the new JTC message.

**Environment (please complete the following information):**
 - OS: ubuntu 22.04
 - Version master of ros2_controllers against rolling

A PR is ready. standby

bmagyar:
Ah thank you! I started looking into reproducing this a few days ago

bmagyar:
Do we also have this issue in humble?

gwalck:
@bmagyar  if jtc new control messages were ported back to humble, maybe, I did not test.

gwalck:
as far as I can see when looking at code in humble and testing a modified rrbot launch file to start a jtc instead of the forward controller on humble branch of `ros2_control_demos`
```
ros2 launch ros2_control_demo_example_1 rrbot.launch.py
```
the jtc controller of humble is publishing to both topics (state and controller_state) with a message containing both old and new structures, so the RQT JTC in humble is fine with it (well the rrbot has no joint limits I had to add some to get he RQT JTC to accept display the sliders).

destogl:
We could also update `humble` but this could also confuse people...

destogl:
Maybe to add a note that RQT plugin is using only new topic because we want people using humble to transition to the new one

gwalck:
What I understood is that humble was supposed to handle both, so we should not break RQT_JTC there. But now RQT_JTC is broken in rolling because in rolling there is only one topic accepted and one type of msg accepted.

destogl:
@mpollayil this looks like a different issue. Can you give us more informations about your setup?

What are you using from binaries, what are you using from source. This might points us to know what is happening.

mpollayil:
Hi Denis,

thanks for your prompt reply.

> @mpollayil this looks like a different issue. Can you give us more informations about your setup?

Would you like me to open a separate thread for this?
 
[...]

PR #643, opened 2023-06-02
Adapted rqt_jtc to newest control_msgs for jtc

gwalck:
solves #642


destogl (approved):
Thanks this looks good!

bmagyar (approved):
Not only looks good but I found the same issue last week and also tested this against my old setup. Thanks!

PR #644, opened 2023-06-02
[bugfix] diff_drive_controller publish rate is half of what it is supposed to be

julescarpentier:
Hello dear reviewer,

The variable `previous_publish_timestamp_` is incremented twice by the `publish_period_` in the `update` function in `diff_drive_controller.cpp`. This effectively halved the `publish_rate` defined in the node parameters.


bmagyar (commented):
good catch! thank you!

bmagyar (approved):

PR #664, opened 2023-06-12
Fix controllers index

christophfroehlich:
Sorry, there was a remnant of the merge conflict by #661


bmagyar (approved):
I wonder how many times we all fixed those by now and broke it again in unexpected ways :)

christophfroehlich:
maybe we can add a CI job for the docs in the main repos? ;)

destogl:
> maybe we can add a CI job for the docs in the main repos? ;)

We probably should...

destogl:
We can actually use external workflows if we convert one for the checks of the documentation into reusable workflow.

PR #667, opened 2023-06-13
Add -Wconversion flag to protect future developments

gwalck:
Same reasons as https://github.com/ros-controls/ros2_control/pull/1053

explicitly catching errors of time being zero in `forward_command_controller`.
I think there might be an ABI breaking in JTC with one argument changed from `int` to `size_t`. I did not know what to do there, and went for ABI breaking but I could also make it a `static_cast` if preferred.

### DIFF
--- a/tricycle_controller/src/tricycle_controller.cpp
+++ b/tricycle_controller/src/tricycle_controller.cpp
@@ -67,7 +67,7 @@ CallbackReturn TricycleController::on_init()
     auto_declare<bool>("enable_odom_tf", odom_params_.enable_odom_tf);
     auto_declare<bool>("odom_only_twist", odom_params_.odom_only_twist);
 
-    auto_declare<int>("cmd_vel_timeout", cmd_vel_timeout_.count());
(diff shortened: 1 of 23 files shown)


tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -244,8 +244,8 @@ controller_interface::return_type TricycleController::update(
     auto & realtime_ackermann_command = realtime_ackermann_command_publisher_->msg_;
     // speed in AckermannDrive is defined desired forward speed (m/s) but we use it here as wheel
     // speed (rad/s)
-    realtime_ackermann_command.speed = Ws_write;
-    realtime_ackermann_command.steering_angle = alpha_write;
+    realtime_ackermann_command.speed = static_cast<float>(Ws_write);
```
bmagyar, line 247: this is ok but I'm tempted to say we should change the messages to use float64 here: https://github.com/ros-drivers/ackermann_msgs/blob/ros2/msg/AckermannDrive.msg

next time...

christophfroehlich:
@mergifyio backport humble iron

Issue #673, opened 2023-06-16
[JTC] Jerky motion with Franka Emika Panda - Effort interface


tingelst:
**Describe the bug**
I have a Franka Emika Panda setup working with Humble. With the original `joint_effort_trajectory_controller` I get smooth motions in MoveIt and when using `rqt_joint_trajectory_controller`. However, when using the `joint_trajectory_controller` here on the `humble` branch I get jerky motions.

**To Reproduce**
1. Run this PR on your system: https://github.com/frankaemika/franka_ros2/pull/10
2. Start MoveIt or `rqt_joint_trajectory_controller` and move the robot
3. Robot moves with jerky motions

**Expected behavior**
The robot should move smoothly.

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04
 - Version: Humble

tingelst:
Seems like I am not the only one having this issue: https://github.com/frankaemika/franka_ros2/pull/10#issuecomment-1594526860

tingelst:
Thank you, @JafarAbdi!  I can confirm that I got the expected behavior, that is, the robot motion smooth, when I checked out the parent commit: https://github.com/ros-controls/ros2_controllers/commit/ce0aeb29e8b42a030248c95a6ee3ec0e99c5e80b. I will continue testing tomorrow.

tingelst:
@JafarAbdi, I have found the source of the jerky/wobbly motion. The bug was introduced in https://github.com/ros-controls/ros2_controllers/commit/ce0aeb29e8b42a030248c95a6ee3ec0e99c5e80b as you expected. The velocity error term is not computed when using an effort command interface, only for velocity command interfaces. This can be fixed by changing [this line](https://github.com/ros-controls/ros2_controllers/blob/6c979091a7e6477963226765a23ecadd33a70d3a/joint_trajectory_controller/src/joint_trajectory_controller.cpp#LL136C9-L136C38) from
```cpp
    if (has_velocity_state_interface_ && has_velocity_command_interface_)
```
to
```cpp
    if (has_velocity_state_interface_ && (has_velocity_command_interface_ || has_effort_command_interface_))
```

I'll submit at PR for this.

I also have a problem with non-zero effort commands commanded after execution as mentioned in issue https://github.com/ros-controls/ros2_controllers/issues/671#issue-1757801230, but I'll take that discussion there.

JafarAbdi:
@tingelst This should be closed since the bug is fixed in the main branch, right?

[...]

PR #679, opened 2023-06-20
Compute velocity errors when using an effort command interface

tingelst:
This PR is in reference to the discussion in issue https://github.com/ros-controls/ros2_controllers/issues/673#issue-1760444492. The changes in https://github.com/ros-controls/ros2_controllers/commit/8e6697b introduced a bug in that the velocity error terms was not computed when using an effort command interface, resulting in a wobbly/jerky motion. This PR fixes this issue.


bmagyar:
Do you have a good idea for adding a test for this? To avoid making this mistake in the future

christophfroehlich (commented):
I think this is fine. But I'm not sure if #491 really introduced this, the velocity error wasn't calculated before with effort command interface?

tingelst:
The velocity error was calculated before, but it was done directly in the `computeCommand` as seen [here](https://github.com/ros-controls/ros2_controllers/pull/491/commits/54509970861c86f413959f92d1a2da49adc0bd94#diff-8dd2135dc1f4abe6382cc7671ef9ce939196b2d8adf18ac3a77b5c0fddf1a25fL247).

christophfroehlich:
> The velocity error was calculated before, but it was done directly in the `computeCommand` as seen [here](https://github.com/ros-controls/ros2_controllers/pull/491/commits/54509970861c86f413959f92d1a2da49adc0bd94#diff-8dd2135dc1f4abe6382cc7671ef9ce939196b2d8adf18ac3a77b5c0fddf1a25fL247).

I see, sorry. I thought I removed the redundant difference calculation only, but hadn't had the effort interface in mind.

christophfroehlich:
I've written a test and made the PR #690 against _jtc-features_ branch, where the effort tests are active already. It failed before I cherry-picked the commit from this PR. Let the maintainers decide if we merge the test directly to _master_ or wait for _jtc-features_ to be merged.

christophfroehlich:
@destogl this is already included in #705 (I squashed the changes, this is why it shows new commits but no changes)

destogl:
OK, I'll then close this, OK?

PR #681, opened 2023-06-20
Fix file name for include guard

christophfroehlich:
Otherwise cpplint complains

> Using '--root=/workspaces/ros2_rolling_ws/src/ros2_controllers/admittance_controller/test' argument
> 
> /workspaces/ros2_rolling_ws/src/ros2_controllers/admittance_controller/test/6d_robot_description.hpp:15:  #ifndef header guard has wrong style, please use: 6D_ROBOT_DESCRIPTION_HPP_  [build/header_guard] [5]
> /workspaces/ros2_rolling_ws/src/ros2_controllers/admittance_controller/test/6d_robot_description.hpp:313:  #endif line should be "#endif  // 6D_ROBOT_DESCRIPTION_HPP_"  [build/header_guard] [5]

and the preprocessor macro cannot start with a number.

### DIFF
--- a/admittance_controller/test/test_admittance_controller.hpp
+++ b/admittance_controller/test/test_admittance_controller.hpp
@@ -27,7 +27,6 @@
 
 #include "gmock/gmock.h"
 
-#include "6d_robot_description.hpp"
 #include "admittance_controller/admittance_controller.hpp"
 #include "control_msgs/msg/admittance_controller_state.hpp"
 #include "geometry_msgs/msg/transform_stamped.hpp"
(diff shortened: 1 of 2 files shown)


christophfroehlich:
@bmagyar do you see more info why the ament_lint_cmake job failed? I can't access any docs.

After `pre-commit autoupdate` I get
```
 $ pre-commit run ament_lint_cmake --all
ament_lint_cmake.........................................................Passed
```

bmagyar:
It was just flaky, happened a few times recently

Issue #683, opened 2023-06-23
[JTC] set_hold_position() doesn't actually stop the robot


egordon:
**Describe the bug**
Upon aborting a goal or some other unrecoverable error, trajectory execution is stopped in the realtime loop by calling `set_hold_position()`, which sets an empty trajectory. This doesn't actually stop the robot. Instead, the empty trajectory does not have any valid point, so nothing at all is commanded to the hardware interface. This means that the last-commanded hardware state remains in place, which has no guarantee of actually stopping the robot.

**To Reproduce**
1. Start up JTC with e.g. velocity or acceleration control.
2. Start any trajectory.
3. Cancel the trajectory goal mid-execution so the trajectory is aborted
4. Observe that the robot continues to execute the last-commanded velocity or acceleration or effort.

This issue isn't visible with position-only control.

**Expected behavior**
We should handle this like a transition to the `deactivate` lifecycle state. Namely: `set_hold_position()` should echo the current position back to the robot, and velocity/acceleration should be commanded to 0.

I'll submit a PR to this effect.

**Screenshots**
N/A
But I can provide video of our physical robot (a modified Kinova JACO2) if necessary.

**Environment (please complete the following information)**

* OS: Ubuntu 22.04 Jammy
* ROS Version Humble
* With current package ros-humble-joint-trajectory-controller v2.20.0-1jammy.20230522.072811

**Additional Context**

Note the existing TODO with regards to effort interfaces, it is unclear how to stop an effort-only interface in a robot-agnostic way.

christophfroehlich:
Duplicate of #514? At least the same behavior causing both issues IMHO.

PR #684, opened 2023-06-23
Have set_hold_position() actually stop the robot

egordon:
Addresses https://github.com/ros-controls/ros2_controllers/issues/683

In almost all cases, that a trajectory doesn't return a valid point implies that `set_hold_position()` (which defines an empty trajectory with no valid points) has been called somewhere. We should stop the robot in this case.

Note the existing TODO that it is unclear how to stop an effort-controlled robot.

Note also that we should not reach this code under any other circumstance, since all incoming trajectories are validated before copied to the realtime pointer. Therefore, reaching this code without `set_hold_position()` implies a bug in our validation code, and I would argue stopping the robot is reasonable behavior.

- [x] Fork the repository.

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -364,7 +364,6 @@ controller_interface::return_type JointTrajectoryController::update(
     {
       /// Likely an Empty trajectory from set_hold_position()
       /// Command hold position
-      // TODO(anyone): How to halt when using effort commands?
       for (size_t index = 0; index < dof_; ++index)
       {
         if (has_position_command_interface_)
@@ -377,7 +376,8 @@ controller_interface::return_type JointTrajectoryController::update(
         {
           joint_command_interface_[1][index].get().set_value(0.0);
         }
-
+        
+        // TODO(anyone): How to halt when using effort commands?
         if (has_effort_command_interface_)
         {
           joint_command_interface_[3][index].get().set_value(0.0);
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -360,6 +360,30 @@ controller_interface::return_type JointTrajectoryController::update(
         RCLCPP_ERROR(get_node()->get_logger(), "Holding position due to state tolerance violation");
       }
     }
+    else
+    {
+      /// Likely an Empty trajectory from set_hold_position()
+      /// Command hold position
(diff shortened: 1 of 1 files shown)


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -360,6 +360,30 @@ controller_interface::return_type JointTrajectoryController::update(
...
+          joint_command_interface_[0][index].get().set_value(
+            joint_command_interface_[0][index].get().get_value());
+        }
+
+        if (has_velocity_command_interface_)
+        {
+          joint_command_interface_[1][index].get().set_value(0.0);
+        }
+
+        if (has_effort_command_interface_)
```
bmagyar, line 381:
```suggestion
      for (size_t index = 0; index < dof_; ++index)
      {
        if (has_position_command_interface_)
        {
          joint_command_interface_[0][index].get().set_value(
            joint_command_interface_[0][index].get().get_value());
        }

        if (has_velocity_command_interface_)
        {
          joint_command_interface_[1][index].get().set_value(0.0);
        }
        
        // TODO(anyone): How to halt when using effort commands?
        if (has_effort_command_interface_)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -360,6 +360,30 @@ controller_interface::return_type JointTrajectoryController::update(
         RCLCPP_ERROR(get_node()->get_logger(), "Holding position due to state tolerance violation");
       }
     }
+    else
+    {
+      /// Likely an Empty trajectory from set_hold_position()
+      /// Command hold position
```
bmagyar, line 366:
```suggestion
      /// Likely an Empty trajectory from set_hold_position()
      /// Command hold position
      RCLCPP_DEBUG(get_node()->get_logger(), "Likely an Empty trajectory from set_hold_position(). Holding position...");
```

bmagyar: What do you think of this version @egordon ?

bmagyar (approved):

christophfroehlich:
I think that #558 already adresses this issue?

bmagyar:
Partial to #558 as it has tests...

christophfroehlich:
I had a detailed look: I think this one here does not solve the problem for all interface combinations: it sets command-interfaces of velocity/effort to zero at the end of the update method -> a position hold is not guaranteed, because the PIDs' outputs will be overriden.

destogl:
Closing in favor of #613

Issue #688, opened 2023-06-26
[JTC] Segmentation fault with action tests


christophfroehlich:
**Describe the bug**
Included with #607, the action tests get activated again. But with the latest changes to the master branch, the [tests fail with a segmentation fault](https://github.com/ros-controls/ros2_controllers/actions/runs/5381557578/jobs/9765782001?pr=607#step:6:6374).
Before the rebase from @bmagyar today everything was fine. Might this be a result of #682?

Locally I get the following stack trace:

```
4: [ RUN      ] OnlyEffortTrajectoryControllers/TestTrajectoryActionsTestParameterized.test_success_single_point_sendgoal/0
4: [INFO] [1687809125.124157051] [test_joint_trajectory_controller]: No specific joint names are used for command interfaces. Using 'joints' parameter.
4: [INFO] [1687809125.124258126] [test_joint_trajectory_controller]: Command interfaces are [effort] and state interfaces are [position velocity].
4: [INFO] [1687809125.124306425] [test_joint_trajectory_controller]: Using 'splines' interpolation method.
4: [INFO] [1687809125.125600011] [test_joint_trajectory_controller]: Action status changes will be monitored at 20.00 Hz.
4: [INFO] [1687809125.431236591] [test_joint_trajectory_controller]: Received new action goal
4: [INFO] [1687809125.431359484] [test_joint_trajectory_controller]: Accepted new action goal
4: [INFO] [1687809125.931574512] [test_joint_trajectory_controller]: Goal reached, success!
4: Stack trace (most recent call last) in thread 78980:
4: #6    Object "", at 0xffffffffffffffff, in 
4: #5    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f867bb059ff, in 
4: #4    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f867ba73b42, in 
4: #3    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7f867bd032b2, in 
4: #2    Source "/workspaces/ros2_rolling_ws/src/ros2_controllers/joint_trajectory_controller/test/test_trajectory_actions.cpp", line 99, in operator() [0x55e624efb8fd]
4:          96:         auto end_time = start_time + wait;
4:          97:         while (clock.now() < end_time)
4:          98:         {
4:       >  99:           traj_controller_->update(clock.now(), clock.now() - start_time);
4:         100:         }
4:         101:       });
[...]

PR #689, opened 2023-06-27
Run workflow on feature branches

christophfroehlich:
Rolling build workflows now should run on `workflow_dispatch` and `pull_request` events on branches matching `*feature*` or `*feature/**`.

These jobs fail now, because of #688

### DIFF
--- a/.github/workflows/rolling-binary-build-main.yml
+++ b/.github/workflows/rolling-binary-build-main.yml
@@ -6,9 +6,13 @@ on:
   workflow_dispatch:
     branches:
       - master
+      - '*feature*'
+      - '*feature/**'
   pull_request:
     branches:
       - master
+      - '*feature*'
+      - '*feature/**'
   push:
     branches:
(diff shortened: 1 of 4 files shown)


destogl:
Is this something we really need? Usually, maintainers can restart CI quite without problems, just opening the workflow.

christophfroehlich:
Now it would be handy with the jtc-features branch, because the PRs targeting into it aren't checked (can you manually run them on an arbitrary PR?). For this maybe the change of `workflow_dispatch` is not really necessary, but `pull_request` would be.
But I don't now if this feature branch is an extraordinary exception, or if that might be a best-practice example for the future.

bmagyar (approved):
I like the idea!!

destogl:
Should we revert this, or do we want to merge those changes into master too?

christophfroehlich:
I vote for bringing this to master for other new feature branches in the future.

PR #693, opened 2023-07-06
Fixed period computation in test

gwalck:
As discussed and confirmed in https://github.com/ros-controls/ros2_controllers/pull/686 there seem to be a wrong computation of the period (dt)  passed to `traj_controller->update` in the `UpdateController` test helper function.

Fixing it does not make tests fail, but is, first necessary to avoid keeping wrong computations, and second for future up-coming improvement that relies on dt to update the state interface as well.

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
@@ -373,9 +373,13 @@ public:
     auto clock = rclcpp::Clock(RCL_STEADY_TIME);
     const auto start_time = clock.now();
     const auto end_time = start_time + wait_time;
+    auto previous_time = start_time;
+
     while (clock.now() < end_time)
     {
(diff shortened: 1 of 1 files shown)


bmagyar (approved):
well-spotted!

bmagyar:
Please run `pre-commit run --all` in the repo and fix the highlighted warnings for the CI to be happy too :)

gwalck:
> Please run `pre-commit run --all` in the repo and fix the highlighted warnings for the CI to be happy too :)

Sorry, I know this, did it several times, and for the PR (which is only a single change compared to many I did), I did a cherry-pick and apparently this introduced a line. I forgot to re-run. I think I had a reason to not "install" the pre-commit but I forgot which one.

bmagyar:
I find precommit on every commit to be slowing my work down and be generally annoying so I also never actually run `pre-commit install` :)

christophfroehlich (approved):
thanks!

destogl:
@mergifyio backport humble

PR #699, opened 2023-07-14
[JTC] Reject messages with effort fields

### LINKED ISSUE #688: [JTC] Segmentation fault with action tests

christophfroehlich:
- Trajectory class does not interpolate effort fields
- But [JointTrajectoryPoint message ](https://github.com/ros2/common_interfaces/blob/f4eac72f0bbd70f7955a5f709d4a6705eb6ca7e8/trajectory_msgs/msg/JointTrajectoryPoint.msg#L20-L23)has an effort field.

We decided in the last WG meeting to reject incoming trajectory messages having effort entries for now. If someone needs this, we have to discuss how to interpolate effort.

Together with 0f8402423edc63410ee94723610568686f006be2 this closes #688

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -289,14 +289,7 @@ controller_interface::return_type JointTrajectoryController::update(
         }
         if (has_effort_command_interface_)
         {
-          if (use_closed_loop_pid_adapter_)
-          {
-            assign_interface_from_point(joint_command_interface_[3], tmp_command_);
-          }
-          else
-          {
-            assign_interface_from_point(joint_command_interface_[3], state_desired_.effort);
-          }
+          assign_interface_from_point(joint_command_interface_[3], tmp_command_);
         }
 
         // store the previous command. Used in open-loop control mode
@@ -1267,8 +1260,9 @@ bool JointTrajectoryController::validate_trajectory_point_field(
   if (joint_names_size != vector_field.size())
   {
     RCLCPP_ERROR(
-      get_node()->get_logger(), "Mismatch between joint_names (%zu) and %s (%zu) at point #%zu.",
-      joint_names_size, string_for_vector_field.c_str(), vector_field.size(), i);
+      get_node()->get_logger(),
+      "Mismatch between joint_names size (%zu) and %s (%zu) at point #%zu.", joint_names_size,
+      string_for_vector_field.c_str(), vector_field.size(), i);
     return false;
   }
   return true;
@@ -1384,9 +1378,15 @@ bool JointTrajectoryController::validate_trajectory_msg(
       !validate_trajectory_point_field(joint_count, points[i].positions, "positions", i, false) ||
       !validate_trajectory_point_field(joint_count, points[i].velocities, "velocities", i, true) ||
       !validate_trajectory_point_field(
-        joint_count, points[i].accelerations, "accelerations", i, true) ||
-      !validate_trajectory_point_field(joint_count, points[i].effort, "effort", i, true))
+        joint_count, points[i].accelerations, "accelerations", i, true))
     {
+      return false;
+    }
+    // reject effort entries
+    if (points[i].effort.size() > 0)
+    {
+      RCLCPP_ERROR(
+        get_node()->get_logger(), "Trajectories with effort fields are currently not supported.");
       return false;
     }
   }
--- a/joint_trajectory_controller/test/test_trajectory_controller.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller.cpp
@@ -1050,10 +1050,9 @@ TEST_P(TrajectoryControllerTestParameterized, invalid_message)
   traj_msg.points[0].accelerations = {1.0, 2.0};
   EXPECT_FALSE(traj_controller_->validate_trajectory_msg(traj_msg));
 
-  // Incompatible data sizes, too few efforts
+  // Effort is not supported in trajectory message
   traj_msg = good_traj_msg;
-  traj_msg.points[0].positions.clear();
-  traj_msg.points[0].effort = {1.0, 2.0};
+  traj_msg.points[0].effort = {1.0, 2.0, 3.0};
   EXPECT_FALSE(traj_controller_->validate_trajectory_msg(traj_msg));
 
   // Non-strictly increasing waypoint times


gavanderhoorn:
What sort of error / result code would clients see?

A single `error` line in the controller's terminal is probably not sufficient to prevent surprises / confused users.

christophfroehlich:
> What sort of error / result code would clients see?
> 
> A single `error` line in the controller's terminal is probably not sufficient to prevent surprises / confused users.

The message will be rejected, which means that nothing moves and the user will have a look in the command prompt.
Using the topic interface, there is no chance for returning error codes. Using action interface, `rclcpp_action::GoalResponse::REJECT` is the result for the action client.

What else do you suggest?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1384,11 +1378,17 @@ bool JointTrajectoryController::validate_trajectory_msg(
...
       !validate_trajectory_point_field(joint_count, points[i].velocities, "velocities", i, true) ||
       !validate_trajectory_point_field(
-        joint_count, points[i].accelerations, "accelerations", i, true) ||
-      !validate_trajectory_point_field(joint_count, points[i].effort, "effort", i, true))
+        joint_count, points[i].accelerations, "accelerations", i, true))
     {
       return false;
     }
+    // reject effort entries
+    if (points[i].effort.size() > 0)
```
aprotyas, line 1386: Optional nit, I feel it reads better this way

```suggestion
    if (!points[i].effort.empty())
```

aprotyas (approved):

gavanderhoorn:
> The message will be rejected, which means that nothing moves and the user will have a look in the command prompt.

I guess I would phrase this as:

> The message will be rejected, which means that nothing moves and the user will **have to** look in the command prompt.

that's what I meant when I wrote:

> A single error line in the controller's terminal is probably not sufficient to prevent surprises / confused users.

I've been such a confused user in the past, hence my comment.

> Using action interface, `rclcpp_action::GoalResponse::REJECT` is the result for the action client.

Last time I checked, there was no support for setting the [error_code](https://github.com/ros-controls/control_msgs/blob/c9ce53dfb5e87b3dfd60bdd85b4826b67989ced4/control_msgs/action/FollowJointTrajectory.action#L33) or the [error_string](https://github.com/ros-controls/control_msgs/blob/c9ce53dfb5e87b3dfd60bdd85b4826b67989ced4/control_msgs/action/FollowJointTrajectory.action#L49) as part of the rejection, as you could in ROS 1 (https://github.com/ros2/rclc/issues/271).

That complicates things, but a work-around I've seen is to accept the goal, then 'abort' by never processing it and immediately sending the final result, in which you set the `error_code` and `error_string` to a meaningful value.

Clearly a work-around, and I still consider the limitation discussed in the linked issue a serious design flaw, but it seems without some work, the best you can do.

The advantage of doing something like this is that clients would have machine readable information about what happened to their goal.

"the human needs to look at the command line" is not a viable error handling / recovery strategy in many cases, so should be avoided as much as possible. The `error_code` field could be used by clients to implement automated recovery, or at the very least promote the error returned by the JTC in a way easily visible to users (using some form of UI, or perhaps still an `ERROR` in a log somewhere, but at least not hidden in the rest of the output).

christophfroehlich:
The check introduced with this PR just follows the way the trajectory message gets rejected with any other errors. So I suggest discussing your point in a separate issue, because we then would have to change the overall behavior of the `validate_trajectory_msg` method, which is not linked to the particular issue with the effort fields.

destogl (approved):
Thanks! Especially like negative sum of changes!!!

bmagyar (approved):
Thanks!

bmagyar:
Thanks @gavanderhoorn for the insightful comments, it's worth tracking that on a separate issue.

PR #708, opened 2023-07-17
[JTC] Change clock_type for tests and further bugfixes

christophfroehlich:
I extracted changes to the JTC tests from #558 and #609, which are not directly linked to the behavior changes being introduced there.

- While implementing the action tests for #558, I struggled with different time sources and c++ exceptions of the type  _can't compare times with different time sources_. Now consequently `RCL_STEADY_TIME` (introduced with #427) is used everywhere for the tests, also as clock source for the node itself. I had to change the trajectory class as well because I found no other way to get the correct time source for `trajectory_start_time_` for every case in the tests -> **ABI breaking**
- `has_active_trajectory` etc will be used in tests of #558 and #609 later on.
- I think the logic behind `test_ignore_partial_old_trajectory` was wrong: It **should** accept the new goal and end up with the last of the new points.

CI tests are failing due to the admittance controller issues.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -248,6 +248,9 @@ protected:
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC
   bool reset();
 
+  JOINT_TRAJECTORY_CONTROLLER_PUBLIC
+  bool has_active_trajectory();
+
   using JointTrajectoryPoint = trajectory_msgs::msg::JointTrajectoryPoint;
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC
   void publish_state(
--- a/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
@@ -54,7 +54,9 @@ public:
     const trajectory_msgs::msg::JointTrajectoryPoint & current_point);
 
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC
-  void update(std::shared_ptr<trajectory_msgs::msg::JointTrajectory> joint_trajectory);
+  void update(
+    std::shared_ptr<trajectory_msgs::msg::JointTrajectory> joint_trajectory,
+    rcl_clock_type_t clock_type = RCL_ROS_TIME);
 
   /// Find the segment (made up of 2 points) and its expected state from the
   /// containing trajectory.
--- a/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
@@ -133,6 +133,9 @@ public:
   bool has_trajectory_msg() const;
 
   JOINT_TRAJECTORY_CONTROLLER_PUBLIC
+  bool has_nontrivial_msg() const;
+
+  JOINT_TRAJECTORY_CONTROLLER_PUBLIC
   std::shared_ptr<trajectory_msgs::msg::JointTrajectory> get_trajectory_msg() const
   {
     return trajectory_msg_;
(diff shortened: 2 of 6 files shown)


gwalck:
@christophfroehlich , I am sorry that I could yet find time to work on my own attempt of fixing the clock.

I encountered the compare time issue when trying to move to time+dt sampling, because at new goal, there must be an initialization of time (start_time is set to the time sent by the update in the first call to sample), but this could be a problem if the previous initialization of the trajectory_msgs was with a different clock. 
The compare time issue was happening [there](https://github.com/ros-controls/ros2_controllers/blob/148687d51d6651ef4c9879110a627b5444fe1a0a/joint_trajectory_controller/src/trajectory.cpp#L89C30-L89C30) 

On my side I fixed it with a time initialization AFTER set_point_before_trajectory_msg is called. Then the comparison error disappeared (without changing clocks), because the time_before was then initialized with the time passed by update

I obviously did not dig into your issue, maybe my info gives you hints.  

Personally, I believe one should not need clocks to be passed around. The only time one needs to compare time that might be from different clocks is when a new goal is validated (https://github.com/ros-controls/ros2_controllers/blob/148687d51d6651ef4c9879110a627b5444fe1a0a/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1301) to be in the future or in the past isn't it ? but not in update, when we go in update, the start_time is using the time past from update (at first sample and next ones)

Where exactly did your compare time issues occur ? (what line causes that exception ?). I had to use some log error to find where, because I did not match to set breakpoints when called with pytest.

christophfroehlich:
@AndyZe introduced `RCL_STEADY_TIME` for good reasons I think. The current master branch is ok, but adding new tests always introduces problems with time sources.

The last one which I could not fix other than passing it to the `trajectory::update` method was:

If `time_from_start` is zero, the update method sets `trajectory_start_time_` from `sample_time` having the correct clock type.
https://github.com/ros-controls/ros2_controllers/blob/8436ed549becce0b694dfe23cfa053436f2b6d42/joint_trajectory_controller/src/trajectory.cpp#L80
But if it is set to a specific time, e.g. with `test_ignore_old_trajectory`, it is set with the operator= of `rclcpp::Time`
https://github.com/ros-controls/ros2_controllers/blob/8436ed549becce0b694dfe23cfa053436f2b6d42/joint_trajectory_controller/src/trajectory.cpp#L56
and there the [clock source](https://github.com/ros2/rclcpp/blob/6e97990a32afa17ab98ef4c800783c5cce801786/rclcpp/include/rclcpp/time.hpp#L93-L101) is `RCL_ROS_TIME` causing the c++ exception here:
https://github.com/ros-controls/ros2_controllers/blob/8436ed549becce0b694dfe23cfa053436f2b6d42/joint_trajectory_controller/src/trajectory.cpp#L98

gwalck:
thanks for clarifying, it is exactly at the same spot, and indeed, I did not try for a time different than zero. Instead of passing the clock type to update, isn't there a way to pass it once only, and use it for the rest of the existence of JTC ? clock types are usually at start thing (sim or walltime typically). 

In any case, I read too fast yesterday night and you don't pass the clock but the clock type which is probably fine and in theory, this should still permit me to maybe get rid of the steady_clock completely (now needed due to the way state_msg is passed around that only works with steady_clock)

christophfroehlich:
I wanted to give a hint why I had to change the clock_type of the node, which introduced the error described above with the tests from `test_trajectory_controller.cpp`. I reverted the changes, and guess what? No exceptions this time... I don't know how much time I spent yesterday on that with clean rebuilds etc. Now it just works without.
If I can't reproduce it, I'll close this PR and make a minimal one without the clock stuff.

gwalck:
Oh I hate when this happens. I feel your frustration. Could it be that you were in a situation with some race condition that made the first point (the one before msg) to not be initialized yet when doing the comparison ?

gwalck:
I will really try to reserve some time soon to progress on a PR for the `state_msgs` to not be used (except in one test which evaluates that everything it passed fine there).

christophfroehlich:
I can't reproduce the c++ exceptions now, which made the changes to the clock type necessary. I tried rebuilds with different build types (Release, Debug,..), but it does not occur any more. Saving this PR for later, if it occurs again.

PR #712, opened 2023-07-18
[JTC] Fix test_ignore_partial_old_trajectory

christophfroehlich:
I think the logic behind `test_ignore_partial_old_trajectory` was wrong: It **should** accept the new goal and end up with the last of the new points.


joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
```diff
@@ -123,6 +124,13 @@ class TestableJointTrajectoryController
 
   bool is_open_loop() { return params_.open_loop_control; }
 
+  bool has_active_traj() { return has_active_trajectory(); }
+
+  bool has_trivial_traj()
```
saikishor, line 129:
```suggestion
  bool has_active_traj() const { return has_active_trajectory(); }

  bool has_trivial_traj() const
```

christophfroehlich:
@gwalck now I've found again why I had to introduce the changes to the clock: The new goal is not accepted because the time is different `test_ignore_partial_old_trajectory`. 
```cpp
  rclcpp::Time new_traj_start =
    rclcpp::Clock(RCL_STEADY_TIME).now() - delay - std::chrono::milliseconds(100);
```
It has the same clock source (the rclcpp::Time standard), but useless values (compare logging time with the [times in the log message](https://github.com/ros-controls/ros2_controllers/actions/runs/5586950643/jobs/10211637632?pr=712#step:6:3705)).
https://github.com/ros-controls/ros2_controllers/blob/317969489c0477bf8a1e45fcd46ad55cc1d22a90/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1298-L1317

When I now change the corresponding time in the test to 

```cpp
  rclcpp::Time new_traj_start =
    traj_controller_->get_node()->get_clock()->now() - delay - std::chrono::milliseconds(100);
```
or 
```cpp
  auto now = static_cast<builtin_interfaces::msg::Time>(traj_controller_->get_node()->get_clock()->now());
  rclcpp::Time new_traj_start =
    static_cast<rclcpp::Time>(now) - delay - std::chrono::milliseconds(100);
```
then I get the c++ exceptions. 
I don't know why it did not fail tomorrow morning, my build chain has some memory it seems. I pushed it now to see if it throws the same errors on the CI job.

gwalck:
> Any suggestions?

@christophfroehlich  I would like to look deeper as well but I am under the water the last days, no personal time at all and no allocated time slot on control this week.

Issue #574, opened 2023-04-16
[JTC] Trajectory replacement does not interpolate from active trajectory


christophfroehlich:
**Describe the bug**
While copying documentation from [ROS 1 Wiki](http://wiki.ros.org/joint_trajectory_controller/UnderstandingTrajectoryReplacement) we realized that the trajectory replacement [does not work as expected](https://github.com/ros-controls/ros2_controllers/pull/566#discussion_r1161112731). 

**To Reproduce**
1. Checkout the jtc_tests branch from my [gazebo_ros2_control](https://github.com/christophfroehlich/gazebo_ros2_control/tree/jtc_tests) fork
2. Run [this launch file](https://github.com/christophfroehlich/gazebo_ros2_control/blob/jtc_tests/gazebo_ros2_control_demos/launch/cart_example_velocity_jtc.launch.py)
`ros2 launch gazebo_ros2_control_demos cart_example_velocity_jtc.launch.py `
3. Run [this action client](https://github.com/christophfroehlich/gazebo_ros2_control/blob/jtc_tests/gazebo_ros2_control_demos/examples/example_position_with_velocity_nonzero.cpp)
`ros2 run gazebo_ros2_control_demos example_position_with_velocity_nonzero`
4. Run it again before the old goal was reached.

Used JTC configuration:
```
joint_trajectory_controller:
  ros__parameters:
    joints:
      - slider_to_cart
    command_interfaces:
      - velocity
    state_interfaces:
      - position
      - velocity
    allow_nonzero_velocity_stop: true
    gains.slider_to_cart.ff_velocity_scale: 1.0
```

Action goal:
* position + velocity points are sent -> cubic spline.
* `traj start` is left zero, which means "start now" acc to [ROS 1 Wiki](http://wiki.ros.org/joint_trajectory_controller#Trajectory_replacement) and as far as I understand the [code here](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/trajectory.cpp#L80).

**Expected behavior**
The trajectory should be interpolated from the current state (or current desired values) to the new one, as described in the [ROS 1 Wiki](http://wiki.ros.org/joint_trajectory_controller/UnderstandingTrajectoryReplacement)

**Screenshots**
At 9s a new action goal was sent and immediately overridden by the new one:
![image](https://user-images.githubusercontent.com/3367244/232339735-e5da203c-470c-49b9-bfda-336d6541b3c6.png)

[...]

PR #714, opened 2023-07-21
[JTC] Update trajectory documentation

christophfroehlich:
## Trajectory Representation
I added several plots to the docs, which should help to understand what is going on here with the different combinations of the given trajectory.

I moved this part as well as the parameters to a subpage, feel free to comment if you have a different suggestion for the doc structure.

For reference: I wrote a Python module using pybind to visualize the outputs of the Trajectory class with matplotlib, see [here](https://github.com/christophfroehlich/test_jtc_trajectories).

## Trajectory Replacement
As of the outcome from #574 I re-added the trajectory replacement section discarded from #566 

## Bugfixes
Also fixes a broken link https://github.com/ros-controls/control.ros.org/issues/131


joint_trajectory_controller/doc/parameters.rst
```diff
@@ -0,0 +1,132 @@
+:github_url: https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/joint_trajectory_controller/doc/parameters.rst
+
+.. _parameters:
```
bmagyar, line 3: don't we generate this nwoadays?

bmagyar (approved):

PR #716, opened 2023-07-24
[JTC] Process tolerances sent with action goal

### LINKED ISSUE #249: JointTrajectoryController ignores tolerances sent via action client
**Would be nice to have**
`JointTrajectoryController` ignores `goal_time_tolerance` set via [FollowJointTrajectory.Goal.goal_time_tolerance](http://docs.ros.org/en/api/control_msgs/html/action/FollowJointTrajectory.html) together with `path_tolerance` and `goal_tolerance` .

**Potential ROS1 inspiration**
Implement something like a ROS1 version [`updateSegmentTolerances`](https://github.com/ros-controls/ros_controllers/blob/aa9c32b8ea8f83cb578cde82619f78e720fef89c/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.h#L232) which is called every time new goal is received through [`updateTrajectoryCommand`](https://github.com/ros-controls/ros_controllers/blob/aa9c32b8ea8f83cb578cde82619f78e720fef89c/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L604) or whenever [new trajectory is received](https://github.com/ros-controls/ros_controllers/blob/aa9c32b8ea8f83cb578cde82619f78e720fef89c/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L169).
The parameter propagation is continued on the call of [`initJointTrajectory`](https://github.com/ros-controls/ros_controllers/blob/aa9c32b8ea8f83cb578cde82619f78e720fef89c/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L532) which in the end uses `updateSegmentTolerances` [here](https://github.com/ros-controls/ros_controllers/blob/aa9c32b8ea8f83cb578cde82619f78e720fef89c/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h#L318)

This is needed to be able to overwrite constraints that are set via parameter server on the jtc start.

christophfroehlich:
From [here](https://github.com/ros-controls/ros2_controllers/issues/249#issue-1019945122):

> JointTrajectoryController ignores goal_time_tolerance set via [FollowJointTrajectory.Goal.goal_time_tolerance](http://docs.ros.org/en/api/control_msgs/html/action/FollowJointTrajectory.html) together with path_tolerance and goal_tolerance .

This PR proposes a way how to process the tolerances: The tolerances from one action goal are not saved for the next one, but the default ones will be used if no tolerances are set with the following action goals.

(Temporary) deactivation is also implemented like documented in the [msg definition](https://github.com/ros-controls/control_msgs/blob/b4354a10acef592bd0cabf6092f95b843b7df50e/control_msgs/msg/JointTolerance.msg#L7-L10).

From the node parameter we cannot set velocity or acceleration tolerances (except for a single `stopped_velocity_tolerance` for the goal tolerances of all joints). Should we add them as parameters as well, to have the same structure like the action message?

This new feature could break some existing projects, because the tolerances were just ignored and might now be breaking behaviors. Should we introduce a temporary parameter to opt-in?

Fixes #249


joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -106,16 +109,149 @@ SegmentTolerances get_segment_tolerances(Params const & params)
...
+
+  active_tolerances.goal_time_tolerance = rclcpp::Duration(goal.goal_time_tolerance).seconds();
+  auto logger = rclcpp::get_logger("tolerance");
+  RCLCPP_DEBUG(logger, "%s %f", "goal_time", active_tolerances.goal_time_tolerance);
+
+  // from
+  // https://github.com/ros-controls/control_msgs/blob/master/control_msgs/msg/JointTolerance.msg #
+  // There are two special values for tolerances: #  * 0 - The tolerance is unspecified and will
+  // remain at whatever the default is #  * -1 - The tolerance is "erased".  If there was a default,
+  // the joint will be #         allowed to move without restriction.
```
bmagyar, line 148:
```suggestion
  // There are two special values for tolerances:
  // * 0 - The tolerance is unspecified and will remain at whatever the default is 
  // * -1 - The tolerance is "erased".  If there was a default, the joint will be  allowed to move without restriction.
```

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -106,16 +109,149 @@ SegmentTolerances get_segment_tolerances(Params const & params)
...
+        logger, "%s",
+        ("joint " + joint + "specified in goal.path_tolerance does not exist").c_str());
+      return active_tolerances;
+    }
+    auto i = std::distance(params.joints.cbegin(), it);
+    if (joint_tol.position > 0.0)
+    {
+      active_tolerances.state_tolerance[i].position = joint_tol.position;
+    }
+    else if (joint_tol.position == -1.0)
```
bmagyar, line 168: we could use an ERASE_TOLERANCE constant defined in this file and you can remove the comments

bmagyar (changes requested):
A few small comments, otherwise looks great!

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -106,16 +109,149 @@ SegmentTolerances get_segment_tolerances(Params const & params)
...
+    }
+
+    if (goal_tol.acceleration > 0.0)
+    {
+      active_tolerances.goal_state_tolerance[i].acceleration = goal_tol.acceleration;
+    }
+    else if (goal_tol.acceleration == -1.0)
+    {  // -1 means erase
+      active_tolerances.goal_state_tolerance[i].acceleration = 0.0;
+    }
```
saikishor, line 239: I've a question here, why only -1 is considered as erase tolerance?. I see that if it is positive, we are considering the tolerance. So why not like any value less than zero is considered as no active tolerances?

saikishor (changes requested):
Just a minor comment. Rest looks good.

Thanks

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
christophfroehlich: because it is described [here](https://github.com/ros-controls/control_msgs/blob/b4354a10acef592bd0cabf6092f95b843b7df50e/control_msgs/msg/JointTolerance.msg#L7-L10), I thought this might be the same behavior as with ros-controls?

saikishor: It could be from ros_control as you said. 
For me it makes sense to rule out all negative values and count them as disabling the tolerance. 

What's your opinion on this?

christophfroehlich: I don't have a strong opinion on that. @bmagyar?

bmagyar: I actually have an opinion on that. In general any random number that's negative should be an *invalid value* but we are making an exception with `-1.0` which is one of an infinite number of negative floating point number to do a specific thing.

bmagyar (approved):

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
christophfroehlich: Do you want me to add an error message if an arbitrary negative value is sent?

bmagyar: Yes please

christophfroehlich: Done, and added more tests

```
$ colcon test --packages-select joint_trajectory_controller
Starting >>> joint_trajectory_controller
[Processing: joint_trajectory_controller]                   
[Processing: joint_trajectory_controller]                           
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
Finished <<< joint_trajectory_controller [5min 1s]                  

Summary: 1 package finished [5min 2s]
vscode ➜ /workspaces/ros2_iron_ws (master ✗) $ colcon test-result
Summary: 608 tests, 0 errors, 0 failures, 0 skipped
```

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -106,16 +110,219 @@ SegmentTolerances get_segment_tolerances(Params const & params)
...
+    {
+      active_tolerances.goal_state_tolerance[i].position = 0.0;
+    }
+    else if (goal_tol.position < 0.0)
+    {
+      RCLCPP_ERROR(
+        logger, "%s",
+        ("joint '" + joint +
+         "' specified in goal.goal_tolerance has a negative position tolerance. "
+         "Use default tolerances.")
```
saikishor, line 269:
```suggestion
         "' specified in goal.goal_tolerance has a invalid position tolerance. "
         "Using default tolerances.")
```

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -106,16 +110,219 @@ SegmentTolerances get_segment_tolerances(Params const & params)
...
+    {
+      active_tolerances.goal_state_tolerance[i].velocity = 0.0;
+    }
+    else if (goal_tol.velocity < 0.0)
+    {
+      RCLCPP_ERROR(
+        logger, "%s",
+        ("joint '" + joint +
+         "' specified in goal.goal_tolerance has a negative velocity tolerance. "
+         "Use default tolerances.")
```
saikishor, line 288:
```suggestion
         "' specified in goal.goal_tolerance has a invalid velocity tolerance. "
         "Using default tolerances.")
```

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -106,16 +110,219 @@ SegmentTolerances get_segment_tolerances(Params const & params)
...
+    {
+      active_tolerances.goal_state_tolerance[i].acceleration = 0.0;
+    }
+    else if (goal_tol.acceleration < 0.0)
+    {
+      RCLCPP_ERROR(
+        logger, "%s",
+        ("joint '" + joint +
+         "' specified in goal.goal_tolerance has a negative acceleration tolerance. "
+         "Use default tolerances.")
```
saikishor, line 307:
```suggestion
         "' specified in goal.goal_tolerance has a invalid acceleration tolerance. "
         "Using default tolerances.")
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -224,23 +225,25 @@ controller_interface::return_type JointTrajectoryController::update(
         if (
           (before_last_point || first_sample) && *(rt_is_holding_.readFromRT()) == false &&
           !check_state_tolerance_per_joint(
-            state_error_, index, default_tolerances_.state_tolerance[index],
-            true /* show_errors */))
+            state_error_, index, active_tol->state_tolerance[index], true /* show_errors */) &&
+          *(rt_is_holding_.readFromRT()) == false)
```
saikishor, line 229:
```suggestion
            state_error_, index, active_tol->state_tolerance[index], true /* show_errors */))
```

Do we need this `rt_is_holding_` check again after this?. I see it before the `check_state_tolerance_per_joint` call

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -224,23 +225,25 @@ controller_interface::return_type JointTrajectoryController::update(
...
         // past the final point, check that we end up inside goal tolerance
         if (
           !before_last_point && *(rt_is_holding_.readFromRT()) == false &&
           !check_state_tolerance_per_joint(
-            state_error_, index, default_tolerances_.goal_state_tolerance[index],
-            false /* show_errors */))
+            state_error_, index, active_tol->goal_state_tolerance[index],
+            false /* show_errors */) &&
+          *(rt_is_holding_.readFromRT()) == false)
         {
```
saikishor, line 240:
```suggestion
            false /* show_errors */))
         {
```
Same here

saikishor (approved):
Few minor comments. Rest looks good to me. 
Thank you for the changes!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: I'm not sure if I understand your comment, but yes we need the check here otherwise the tolerances will be checked with the holding position/final point after success.

saikishor:
```
         if ( 
            (before_last_point || first_sample) && *(rt_is_holding_.readFromRT()) == false && 
            !check_state_tolerance_per_joint( 
             state_error_, index, default_tolerances_.state_tolerance[index], 
             true /* show_errors */)) 
             state_error_, index, active_tol->state_tolerance[index], true /* show_errors */) && 
           *(rt_is_holding_.readFromRT()) == false)
```

The `*(rt_is_holding_.readFromRT()) == false)` check is repeated. That's what I mean.

christophfroehlich: ah you are right.. I had a look on the wrong branch yesterday :see_no_evil: must come from some merge of master

joint_trajectory_controller/test/test_tolerances_utils.hpp
destogl, line 1: Not sure that we need a header file here. Maybe just add this on top of the `.cpp` file. Those are not too big.

destogl (approved):
Very good PR! Thanks. I have a few minor comments, but we can also merge without resolving them. They are mostly cosmetic nature.

> From the node parameter we cannot set velocity or acceleration tolerances (except for a single `stopped_velocity_tolerance` for the goal tolerances of all joints). Should we add them as parameters as well, to have the same structure like the action message?

This would be nice to have in the follow-up PR.

> This new feature could break some existing projects, because the tolerances were just ignored and might now be breaking behaviors. Should we introduce a temporary parameter to opt-in?

No. Just to add a note to migration notes.

joint_trajectory_controller/test/test_tolerances_utils.hpp
christophfroehlich: I use it in test_tolerances.cpp and test_trajectory_actions.cpp
I can add it to test_trajectory_controller_utils.hpp if you prefer

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -106,16 +110,219 @@ SegmentTolerances get_segment_tolerances(Params const & params)
...
 
   return tolerances;
 }
 
+/**
+ * \brief Populate trajectory segment tolerances using data from an action goal.
+ *
+ * \param default_tolerances The default tolerances to use if the action goal does not specify any.
+ * \param goal The new action goal
+ * \param params The ROS Parameters
```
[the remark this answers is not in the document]
christophfroehlich: The only field we need is the params.joints actually, I can change it to have a std::vector as argument instead?

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -95,6 +97,8 @@ SegmentTolerances get_segment_tolerances(Params const & params)
 
   SegmentTolerances tolerances;
   tolerances.goal_time_tolerance = constraints.goal_time;
+  auto logger = rclcpp::get_logger("tolerance");
```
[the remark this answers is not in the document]
christophfroehlich: I haven't had a better idea as passing the JTC logger as argument to the `get_segment_tolerances` methods. There is no persistent class with a configuration/constructor method to pass the logger only once.

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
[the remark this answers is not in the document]
christophfroehlich: done

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -106,16 +110,219 @@ SegmentTolerances get_segment_tolerances(Params const & params)
...
+ * \return Trajectory segment tolerances.
+ */
+SegmentTolerances get_segment_tolerances(
+  const SegmentTolerances & default_tolerances,
+  const control_msgs::action::FollowJointTrajectory::Goal & goal, const Params & params)
+{
+  SegmentTolerances active_tolerances(default_tolerances);
+
+  active_tolerances.goal_time_tolerance = rclcpp::Duration(goal.goal_time_tolerance).seconds();
+  auto logger = rclcpp::get_logger("tolerance");
```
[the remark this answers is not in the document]
christophfroehlich: see above

joint_trajectory_controller/test/test_tolerances_utils.hpp
christophfroehlich: I removed it.

christophfroehlich:
> Very good PR! Thanks. I have a few minor comments, but we can also merge without resolving them. They are mostly cosmetic nature.
> 
> > From the node parameter we cannot set velocity or acceleration tolerances (except for a single `stopped_velocity_tolerance` for the goal tolerances of all joints). Should we add them as parameters as well, to have the same structure like the action message?
> 
> This would be nice to have in the follow-up PR.

I opened a follow-up issue.

> > This new feature could break some existing projects, because the tolerances were just ignored and might now be breaking behaviors. Should we introduce a temporary parameter to opt-in?
> 
> No. Just to add a note to migration notes.

Done. 

@destogl I addressed all your comments, please have a look once again.

Test still are fine.
```
$ colcon test --packages-select joint_trajectory_controller
Starting >>> joint_trajectory_controller
[Processing: joint_trajectory_controller]                   
[Processing: joint_trajectory_controller]                           
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
[Processing: joint_trajectory_controller]                             
Finished <<< joint_trajectory_controller [4min 49s]                   

Summary: 1 package finished [4min 50s]
```

christophfroehlich:
@bmagyar do we want to backport this to humble/iron? Then I'd update the release notes accordingly

PR #721, opened 2023-07-26
Fix out of bound access in admittance controller

Abishalini:
If the number of robot joints is more than 6, admittance controller currently segfaults with an out-of-bound access error.
This PR moves a few state updates out of the loop which iterates over the number of joint states.

### DIFF
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -28,6 +28,9 @@
 
 namespace admittance_controller
 {
+
+constexpr auto NUM_CARTESIAN_DOF = 6;  // (3 translation + 3 rotation)
+
 /// Configure admittance rule memory for num joints and load kinematics interface
 controller_interface::return_type AdmittanceRule::configure(
   const std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node, const size_t num_joints)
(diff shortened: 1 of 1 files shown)


admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
```diff
@@ -331,13 +331,17 @@ void AdmittanceRule::process_wrench_measurements(
...
   for (size_t i = 0; i < parameters_.joints.size(); ++i)
   {
     state_message_.joint_state.name[i] = parameters_.joints[i];
     state_message_.joint_state.position[i] = admittance_state_.joint_pos[i];
     state_message_.joint_state.velocity[i] = admittance_state_.joint_vel[i];
     state_message_.joint_state.effort[i] = admittance_state_.joint_acc[i];
-    state_message_.stiffness.data[i] = admittance_state_.stiffness[i];
     state_message_.damping.data[i] = admittance_state_.damping[i];
     state_message_.selected_axes.data[i] = static_cast<bool>(admittance_state_.selected_axes[i]);
     state_message_.mass.data[i] = admittance_state_.mass[i];
```
sea-bass, line 347: Sorry I didn't catch this earlier, but these 3 lines also need to go in the newly added loop, as these all have 6 elements necessarily.

Abishalini:
I have retargeted the changes to go in the master branch. 
Can this be backported to humble when this PR is merged?

sea-bass (approved):

admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
```diff
@@ -331,16 +331,20 @@ void AdmittanceRule::process_wrench_measurements(
 
 const control_msgs::msg::AdmittanceControllerState & AdmittanceRule::get_controller_state()
 {
+  for (size_t i = 0; i < 6; ++i)
```
bmagyar, line 334:
```suggestion
  for (size_t i = 0; i < NUM_MAX_SUPPORTED_JOINTS; ++i)
```

in the interest of keeping it somewhat readable and maintainable, let's use a named constant instead of a magic number

sea-bass: Using a constant is fine, but this 6 isn't the number of max supported joints, it's the number of degrees of freedom for the admittance controller (3 translation + 3 rotation)

bmagyar: Oooops! Even more reason for a better name ;)

Abishalini: Done

bmagyar (approved):
Thanks very much!

PR #740, opened 2023-08-07
[JTC] Disable use of PID adapter if in open-loop mode (backport #551)

christophfroehlich:
A manual backport of #551 without merge conflicts.


christophfroehlich:
Failing CI is due to flaky tests, but imho not directly related to this change because open_loop is never set

bmagyar (approved):
thanks!

bmagyar:
Tests were at first failing on JTC but now they are all cleared up, the flaky admittance controller test won't block merging

PR #748, opened 2023-08-16
[JTC] Fix typos, implicit cast, const member functions

christophfroehlich:
Fix typos, implicit cast warnings, and declare const member functions as const


saikishor (approved):
LGTM

saikishor:
It is weird that CI says that the changes break ABI. I don't see any ABI breaking changes. 

Thanks for the changes

christophfroehlich:
> It is weird that CI says that the changes break ABI. I don't see any ABI breaking changes.

I guess that the const qualifiers change ABI?

`has_active_trajectory` from #558 has not been released until now, and the other methods are for testing only -> don't see any problem changing this now.

bmagyar (approved):
Very good, thank you!

PR #751, opened 2023-08-20
Update docs for diff drive controller

christophfroehlich:
While having a look on https://github.com/ros-controls/ros2_control_demos/pull/350 I realized, that the diff drive controller documentation was in a rather incomplete state.

I updated it now (and applied another ROS-space-2 flaw)


bmagyar (approved):
Thank you for the lots of extra space! :D

PR #752, opened 2023-08-20
[JTC] Adding tests for acceleration command interface

christophfroehlich:
I found todos concerning acceleration tests:
- Adds checks for acceleration command interfaces to  
  - `test_jumbled_joint_order` and
  - `test_partial_joint_list`
- Use variable for `test_jumbled_joint_order` goal. 
- Change logic of `is_same_sign` to `is_same_sign_or_zero`

Minor changes
- use `std` namespace for `copysign`

If someone is thinking about the test trajectory points and the expected outputs of `test_jumbled_joint_order`
image


bmagyar (approved):
grand!

PR #758, opened 2023-08-24
[JTC] Fix hold position mode with goal_time>0

### LINKED ISSUE #757: [JTC] Hold position & goal_time>0: Infinite loop
I think I've found a possible issue, do you have set a `goal_time_tolerance>0`? If the time tolerance is exceeded, the PID controller will be deactivated even in hold position mode.

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_controllers/issues/558#issuecomment-1691565433_

This could happen if the JTC can't hold the position (wrong PID tuning, effort limit,..), then it might enter a loop setting a new hold position again and again.

christophfroehlich:
- In hold position mode with goal_time>0: If the JTC can't hold the position (wrong PID tuning, effort limit,..), then it might enter a loop setting a new hold position again and again. Fixes #757
I had to introduce a new class variable to know if we are in holding mode, which breaks ABI. This is also used for #609 to avoid deadlocks with timeouts.
- I improved the test criterion for the cases with no position command interface: We can check only with active PID mode if the correct position is held.
- `set_hold_position` generated invalid trajectory messages (use `validate_trajectory_msg(current_pose_msg);` to test that). I changed that to have the correct fields after sampling (with `spline` and also `none`), acc. to the [docs of the trajectory representation](https://control.ros.org/master/doc/ros2_controllers/joint_trajectory_controller/doc/trajectory.html)
- `set_hold_position` should be now realtime-safe: there is no reason to do memory allocation now where DOF and interface types are fixed


joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1623,6 +1624,13 @@ TEST_P(TrajectoryControllerTestParameterized, test_goal_tolerances_fail)
 
   // it should have aborted and be holding now
   expectHoldingPoint(joint_state_pos_);
+
+  // what happens if we wait longer bit it harms the tolerance again?
```
bmagyar, line 1628:
```suggestion
  // what happens if we wait longer but it harms the tolerance again?
```

bmagyar (approved):

PR #761, opened 2023-09-01
[JTC] Activate update of dynamic parameters

### LINKED ISSUE #754: Dynamically tuning [joint_trajectory_controller] PID gains
Just wondering whether `ros2_controllers`  PID gains can dynamically be tuned using `rqt_reconfigure`. As far as I know, it works in ROS1 but not sure about ROS2.

I want to tune dynamically PID gains.[p/i/d] for each of the joints while the `joint_trajectory_controller` is run. I tried using `rqt_reconfigure` to tune each PID gain but nothing happened.

christophfroehlich:
This fixes #754: Parameters are now updated with on_activate (especially default_tolerances_) as well as at the beginning of the update loop (only if there are any changes)


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1530,6 +1532,26 @@ bool JointTrajectoryController::has_active_trajectory() const
   return traj_external_point_ptr_ != nullptr && traj_external_point_ptr_->has_trajectory_msg();
 }
 
+void JointTrajectoryController::update_pids()
+{
+  for (size_t i = 0; i < dof_; ++i)
+  {
+    const auto & gains = params_.gains.joints_map.at(params_.joints[i]);
+    if (pids_.at(i))
```
saikishor, line 1540:
```suggestion
    if (pids_[i])
```
Why not use `pids_[i]` directly here?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1530,6 +1532,26 @@ bool JointTrajectoryController::has_active_trajectory() const
...
+    {
+      // update PIDs with gains from ROS parameters
+      pids_[i]->setGains(gains.p, gains.i, gains.d, gains.i_clamp, -gains.i_clamp);
+    }
```
saikishor, line 1544: Here we don't want to update `ff_velocity_scale` right?. I think it makes sense. However, just wanted to confirm😅

saikishor (approved):
LGTM

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: This was an intermediate step while I was looking for the cause of a segfault ;) Is it bad practice to use .at() method of `std::vector`?

saikishor: Well it is not a bad practice, but accessing by .at() is a bit slower than the other way.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: `ff_velocity_scale` will be set in every case. I just had to differentiate if a new instance of PID has to be created, or just its parameters have to be set.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: Anyway you are already resizing your vector at the configure phase, I think it shouldn't cause seg fault.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: Ahh yes, you are right. I missed the ending bracket. Thanks

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: It did without the check if  `use_closed_loop_pid_adapter_==true`, because the PID vector is empty then. As I said, I was looking for the source of the segfault, now it should be save to skip.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -415,6 +415,42 @@ TEST_P(TrajectoryControllerTestParameterized, state_topic_consistency)
...
 }
 
+/**
+ * @brief check if dynamic parameters are updated
+ */
+TEST_P(TrajectoryControllerTestParameterized, update_dynamic_parameters)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+
+  SetUpAndActivateTrajectoryController(executor, true);
```
bmagyar, line 425:
```suggestion
  SetUpAndActivateTrajectoryController(executor);
```

bmagyar:
Heads up @fmauch , this may break things on the scaled JTC

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1007,6 +998,28 @@ controller_interface::CallbackReturn JointTrajectoryController::on_activate(
...
+    {
+      cmd_timeout_ = params_.cmd_timeout;
+    }
+    else
+    {
+      // deactivate timeout
+      RCLCPP_WARN(
+        get_node()->get_logger(),
+        "Command timeout must be higher than goal_time tolerance (%f vs. %f)",
+        params_.cmd_timeout, default_tolerances_.goal_time_tolerance);
```
bmagyar, line 1014:
```suggestion
        "Command timeout must be higher than goal_time tolerance (%f vs. %f)", params_.cmd_timeout,
        default_tolerances_.goal_time_tolerance);
```

christophfroehlich:
@bmagyar should be fine now. The single failing test is the usual flakiness.

bmagyar (approved):

destogl (approved):
Thanks! Very needed feature!

PR #769, opened 2023-09-08
[gripper] Add test for effort gripper controller

### LINKED ISSUE #764: [gripper] Add tests for effort interface
We should definitely add tests for that to improve the gripper_controller. @chama1176 Could you have a look on that? I suggest copying the parameterized tests from joint_trajectory_controller, so that we run all tests with position and effort interface.

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_controllers/pull/756#pullrequestreview-1613036898_

chama1176:
Add test for effort gripper controller.

Fix #764 

GripperController uses template parameters, so I used type-parameterized tests.
https://github.com/google/googletest/blob/main/docs/advanced.md#type-parameterized-tests
https://en.cppreference.com/w/cpp/types/integral_constant


bmagyar (approved):
Thank you!

I used gmock for many years but never came across typed tests, you learn something new every day! This is very elegant and clean, thanks!

Issue #394, opened 2022-07-27
Discourage changing the number of DoFs / joints for controller after those are initialized?


destogl:
Discussing parameters handling with @tylerjw and @pac48 it seems that resetting the number of DoF in `on_configure` in some controllers only complicates the management of those. This does not make much sense because if one needs to change the number of DoF in a controller, another instance of the controller should be used. 

Recommended resolution for all (standard) controller in this package:
1. Read the number of `joints`/`joint_names` parameters in the `on_init` method.
2. Remove updates of parameters from `on_configure` method.

AndyZe:
This will also improve realtime performance. If you look at the JTC, all of the dynamic memory allocation occurs because it is possible for `dof` to change size. (Noted in #393)

bmagyar:
There should be a whole bunch of parameters I'd discourage people from changing, including DOF

PR #771, opened 2023-09-09
[JTC] Make most parameters read-only

christophfroehlich:
This is the first part of implementing the suggestions from #394 to JTC. 

I made all parameters read-only, which aren't updated anyways or shouldn't be reconfigureable like DoF. 

Remaining dynamic parameters (open for discussion):
- allow_partial_joints_goal
- allow_integration_in_goal_trajectories
- allow_nonzero_velocity_at_trajectory_end
- start_with_holding (aren't updated yet but well be updated in on_active once  #761 is merged)
- gains: aren't updated yet, see #761 
- constraints (tolerances): aren't updated yet, see #761 

Changes to get the tests running again:
- The tests failed because they heavily relied on parameter reconfiguration -> Now use parameter overrides before initializing the node.
- I removed the yaml file from the test/config folder because it wasn't used at all
- I removed the use_local_param from the testable class, I think it made no difference (any more?)


bmagyar (approved):

saikishor (approved):
LGTM

PR #772, opened 2023-09-11
[JTC] Rename normalize_error to angle_wraparound parameter

### LINKED ISSUE #85: Port over "joint wraparound" to joint_trajectory_controller
As impemented in [ROS Control 1](https://github.com/ros-controls/ros_controllers/blob/melodic-devel/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h#L161)

I believe https://github.com/ros-controls/ros2_controllers/issues/84 needs to be implemented first or alongside this.

christophfroehlich:
I found a very old issue and realized that `normalize_error` was called `angle_wraparound` in ROS 1. I renamed the parameter because the old naming is more descriptive from a user perspective (and some user might be familiar with the old name)

Closes #85

### DIFF
--- a/joint_trajectory_controller/doc/parameters.rst
+++ b/joint_trajectory_controller/doc/parameters.rst
@@ -101,7 +101,7 @@ gains (structure)
 
      u = k_{ff} v_d + k_p e + k_i \sum e dt + k_d (v_d - v)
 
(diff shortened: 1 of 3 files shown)


bmagyar (approved):
Nicely done, thanks!

Issue #548, opened 2023-03-22
JTC action result never returns


livanov93:
https://github.com/PickNikRobotics/moveit_studio/pull/2391

**Describe the bug**
When  `constraints.goal_time = 0.0` and there is `constraints.<joint_name>.goal` that is different than zero and can't be satisfied we always end up [here](https://github.com/ros-controls/ros2_controllers/blob/82a4e2434c8f766c22c23fc11ffb38d54404c212/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L347-L348) and the action result is never returned.

I am not sure this is desired behavior.

 
**To Reproduce**
1. Use https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver and change [this part](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/dfa1410653ebcb57f3522491f61a4d6344425056/ur_robot_driver/config/ur_controllers.yaml#L95-L100) into:
```
      shoulder_pan_joint: { trajectory: 0.2, goal: 0.001 }
      shoulder_lift_joint: { trajectory: 0.2, goal: 0.001 }
      elbow_joint: { trajectory: 0.2, goal: 0.001 }
      wrist_1_joint: { trajectory: 0.2, goal: 0.001 }
      wrist_2_joint: { trajectory: 0.2, goal: 0.001 }
      wrist_3_joint: { trajectory: 0.2, goal: 0.001 }
```
2. Use https://github.com/UniversalRobots/Universal_Robots_ROS2_Description and change [this part](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/blob/e53ee294b0fa86485ad6cf97b4642c7f9cc60f4c/urdf/ur.ros2_control.xacro#L34) to:
```
<param name="position_state_following_offset">-0.005</param>
```
which will make sure you never reach goal.

3.  Send action goal and see the result not returning.

mechwiz:
I could see this being the intended behavior since if you set the goal_time param to zero, that is effectively saying, wait an indefinite amount of time until the goal tolerance is met before considering the action complete. If you want to set a timeout, then the goal_time param should be set to be nonzero. that said, I could hear an argument to maybe default the goal_time param to just be some high number value if it's set to zero or not set at all just so that the action returns eventually.

olivier-stasse:
[...]

PR #773, opened 2023-09-11
[JTC] Add note on goal_time=0 in docs

christophfroehlich:
I added a note on the special case of  `goal_time=0` as suggested with #548.

### DIFF
--- a/joint_trajectory_controller/doc/parameters.rst
+++ b/joint_trajectory_controller/doc/parameters.rst
@@ -78,6 +78,7 @@ constraints.stopped_velocity_tolerance (double)
 
 constraints.goal_time (double)
   Maximally allowed tolerance for not reaching the end of the trajectory in a predefined time.
+  If set to zero, the controller will wait an indefinite amount of time.
 
   Default: 0.0 (not checked)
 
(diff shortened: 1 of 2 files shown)


joint_trajectory_controller/doc/parameters.rst
```diff
@@ -78,6 +78,7 @@ constraints.stopped_velocity_tolerance (double)
 
 constraints.goal_time (double)
   Maximally allowed tolerance for not reaching the end of the trajectory in a predefined time.
+  If set to zero, the controller will wait an indefinite amount of time.
```
bmagyar, line 81: Indefinite or potentially infinite?

christophfroehlich: both :D

christophfroehlich, line 81:
```suggestion
  If set to zero, the controller will wait an potentially inifinte amount of time.
```

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -117,7 +117,8 @@ joint_trajectory_controller:
     goal_time: {
       type: double,
       default_value: 0.0,
-      description: "Time tolerance for achieving trajectory goal before or after commanded time.",
+      description: "Time tolerance for achieving trajectory goal before or after commanded time.
+        If set to zero, the controller will wait an indefinite amount of time.",
```
christophfroehlich, line 121:
```suggestion
        If set to zero, the controller will wait a potentially infinite amount of time.",
```

bmagyar (approved):
Thx :D

PR #780, opened 2023-09-26
[JTC] Use time of the last command for set_point_before_trajectory_msg in open-loop mode

destogl:
This PR tackles the issues that happen in open-loop mode where on replacing trajectories the last command is repeated.
The figures below show the results of the output trajectories before and after this functionality. The trajectory that uses the time of the last command as point before trajectory messages gets smoother output.

Functionality right now:
When using current time

Functionality with this PR:
With using previous time with previous command

P.S. The similar thing would be useful for trajectory replacement in closed loop mode, but then we would need to know exactly when the state is read from the hardware. This is just something to think about in the future.

P.P.S. I will add tests too.


christophfroehlich (commented):
Looks good for me (even more if the mentioned tests are added)
concerning closed-loop mode: I'm not sure if this is necessary, shouldn't the trajectory generator be aware of the current state? But I guess there might be configurations where this could improve something.

christophfroehlich:
I'm asking myself if it is desirable at all to interpolate from measured state instead of the last command with closed-loop config. but that really depends on the algorithm generating the trajectory

christophfroehlich (changes requested):
`test_no_jump_when_state_tracking_error_not_updated` fails

christophfroehlich (approved):
LGTM, I fixed the test now.

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1679,9 +1679,9 @@ TEST_P(TrajectoryControllerTestParameterized, test_jump_when_state_tracking_erro
   publish(
     time_from_start, points, rclcpp::Time(0.0, 0.0, RCL_STEADY_TIME), {}, first_goal_velocities);
   traj_controller_->wait_for_trajectory(executor);
-  updateControllerAsync(rclcpp::Duration::from_seconds(1.1));
+  auto end_time = updateControllerAsync(rclcpp::Duration::from_seconds(1.1));
 
-  // JTC is executing trajectory in open-loop therefore:
+  // JTC is NOT executing trajectory in open-loop therefore:
```
Thieso, line 1684: `open-loop` needs to be `interpolate_from_desired_state`

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1783,7 +1783,7 @@ TEST_P(TrajectoryControllerTestParameterized, test_no_jump_when_state_tracking_e
   std::vector<std::vector<double>> points{{first_goal}};
   publish(time_from_start, points, rclcpp::Time(0.0, 0.0, RCL_STEADY_TIME));
   traj_controller_->wait_for_trajectory(executor);
-  updateControllerAsync(rclcpp::Duration::from_seconds(1.1));
+  auto end_time = updateControllerAsync(rclcpp::Duration::from_seconds(1.1));
 
   // JTC is executing trajectory in open-loop therefore:
```
Thieso, line 1788: `open-loop` needs to be `interpolate_from_desired_state`

Thieso (changes requested):
Looks good to me, just need to change `open-loop`  to be `interpolate_from_desired_state` according to #1525  but I do not have the permissions for this

christophfroehlich:
for this PR it is fine with open-loop, if you approve we can merge that immediately. we then can change that in your PR, where we have to wait for a review of another maintainer anyways

Thieso (approved):

Issue #784, opened 2023-09-29
Steering controllers library open_loop mode


christophfroehlich:
**Describe the bug**
The open_loop mode of the steering controllers library is not working.

**To Reproduce**
1. Checkout the CarLike demo https://github.com/ros-controls/ros2_control_demos/pull/316
2. Set the `open_loop` parameter to true
3. Check the results, the robot is not moving.

**Expected behavior**
Ideal odometry

**Environment (please complete the following information):**
 - Version : Rolling

**Additional context**
I think the issue is that the variables here are never set
https://github.com/ros-controls/ros2_controllers/blob/f7c13ad25938246609b724104054754ec7fad18d/steering_controllers_library/include/steering_controllers_library/steering_controllers_library.hpp#L135-L137
but used here
https://github.com/ros-controls/ros2_controllers/blob/f7c13ad25938246609b724104054754ec7fad18d/ackermann_steering_controller/src/ackermann_steering_controller.cpp#L61
https://github.com/ros-controls/ros2_controllers/blob/f7c13ad25938246609b724104054754ec7fad18d/bicycle_steering_controller/src/bicycle_steering_controller.cpp#L59
https://github.com/ros-controls/ros2_controllers/blob/f7c13ad25938246609b724104054754ec7fad18d/tricycle_steering_controller/src/tricycle_steering_controller.cpp#L59

PR #785, opened 2023-09-29
[Steering controllers lib] Improve docs

christophfroehlich:
Fixing some typos and adding additional context. However, `open_loop` mode seems to be broken see #784


bmagyar (approved):

bmagyar:
Thanks for flagging up the bug

PR #794, opened 2023-10-09
[JTC] Add option to hold position on last command

### LINKED ISSUE #759: [JTC] set_hold_position results into not exactly reaching desired target
**Describe the bug**

In #558  (included in rolled out version 3.14.0)  `set_hold_position` was added which commands the current position as the target after sampling the last point of the trajectory (https://github.com/ros-controls/ros2_controllers/blob/1dac309979fb9392dc12feb6b46555582fed6fc6/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L346).

This results into not reaching the target exactly. (In our case a UR16e arm).
As far as I see there is no way to disable or tune this behavior

**To Reproduce**

Take some robot arm. Move rather fast from A to B. Check if the desired target has been reached within a tolerance of for example 0.001rad 

**Expected behavior**

In case the checks (e.g. arm is within goal tolerance) were successful set the last point of the trajectory as target

- iron
- ros2control : 3.14.0
-

christophfroehlich:
As suggested with #759, there is now an option to hold the position of the last command instead of the state read from the hardware. 

I added a new parameter, but we could also use the already existing `open_loop_control`? @firesurfer Is there a use case for `open_loop_control==false` but one would like to hold the last commanded point?

Still missing:
- [ ] Tests

This closes #759


firesurfer:
@christophfroehlich Thanks for adding this option. 

I am not sure actually about the use case. But I think reusing the parameter could be quite confusing

christophfroehlich:
See #811, we will use a different solution.

PR #698, opened 2023-07-14
[ForceTorqueSensorBroadcaster] Create ParamListener and get parameters on configure

Noel215:
Hi, 

After the last sync of humble, the `ForceTorqueSensorBroadcaster` stopped working for me. After looking for the issue, I've found that there is a validation of the parameter `frame_id` (added [here](https://github.com/ros-controls/ros2_controllers/commit/04218c269521df574a4d285d90b81189c09ca7df#diff-8e9b8a23a01e45384f62b4ebdd0e51191907e858809db4a6ffc0cf3269bc7c05R6-R8)), that is failing always because the parameters have not been loaded into the node yet and then, the parameter is empty at that point.

I have the parameter declared in the yaml file that I'm loading in the node but the error I'm having is the following:
```
[spawner-9] [INFO] [1689253918.601573973] [spawner_ft_sensor_controller]: Set controller type to "force_torque_sensor_broadcaster/ForceTorqueSensorBroadcaster" for ft_sensor_controller
[gzserver-1] [INFO] [1689253918.602927505] [controller_manager]: Loading controller 'ft_sensor_controller'
[gzserver-1] Exception thrown during init stage with message: Invalid value set during initialization for parameter 'frame_id': Parameter 'frame_id' cannot be empty 
[gzserver-1] [ERROR] [1689253918.613905153] [controller_manager]: Could not initialize the controller named 'ft_sensor_controller'
[INFO] [spawn_entity.py-3]: process has finished cleanly [pid 56778]
[spawner-9] [FATAL] [1689253918.815438355] [spawner_ft_sensor_controller]: Failed loading controller ft_sensor_controller
[ERROR] [spawner-9]: process has died [pid 56791, exit code 1, cmd '/home/user/exchange/alum/emergency_button/install/controller_manager/lib/controller_manager/spawner ft_sensor_controller --controller-manager controller_manager --controller-type force_torque_sensor_broadcaster/ForceTorqueSensorBroadcaster --param-file /home/user/exchange/alum/emergency_button/install/tiago_controller_configuration/share/tiago_controller_configuration/config/ft_sensor_controller.yaml --ros-args'].
```

By creating the `ParamListener` and reading the parameters on configure instead of on init is working for me. I would like to have also a backport for `humble`.

Thanks! :)

PR #795, opened 2023-10-10
[IMUSensorBroadcaster] Create ParamListener and get parameters on configure

YueErro:
Hi,

It seems there is a very similar issue as in https://github.com/ros-controls/ros2_controllers/pull/698. The `IMUSensorBroadcaster` stopped working for me. 
After looking for the issue, I've found that there is a validation of the parameter `sensor_name` (added [here](https://github.com/ros-controls/ros2_controllers/blob/f62fc3ac5f480b5a74765e642cd33d17600f7f5a/imu_sensor_broadcaster/src/imu_sensor_broadcaster_parameters.yaml#L8-L10)), that is failing always because the parameters have not been loaded into the node yet and then, the parameter is empty at that point.

I have the parameter declared in the yaml file that I'm loading in the node but the error I'm having is the following:
```bash
[spawner-4] [INFO] [1696928866.880159937] [spawner_imu_sensor_broadcaster]: Set controller type to "imu_sensor_broadcaster/IMUSensorBroadcaster" for imu_sensor_broadcaster
[gzserver-1] [INFO] [1696928866.881161921] [controller_manager]: Loading controller 'imu_sensor_broadcaster'
[gzserver-1] [ERROR] [1696928866.897606928] [imu_sensor_broadcaster]: Exception thrown during init stage with message: Invalid value set during initialization for parameter 'sensor_name': Parameter 'sensor_name' cannot be empty 
[gzserver-1] 
[gzserver-1] [ERROR] [1696928866.897674246] [controller_manager]: Could not initialize the controller named 'imu_sensor_broadcaster'
[spawner-4] [FATAL] [1696928866.904203561] [spawner_imu_sensor_broadcaster]: Failed loading controller imu_sensor_broadcaster
[ERROR] [spawner-4]: process has died [pid 61163, exit code 1, cmd '/opt/pal/alum/lib/controller_manager/spawner imu_sensor_broadcaster --controller-manager controller_manager --controller-type imu_sensor_broadcaster/IMUSensorBroadcaster --param-file /home/yueerro/git/my_dockers/shared/alum/staging/pmb3_ws/install/pmb3_controller_configuration/share/pmb3_controller_configuration/config/imu_sensor_broadcaster.yaml --ros-args'].
```
I guess the same happens with the validation of the parameter `frame_id` (added [here](https://github.com/ros-controls/ros2_controllers/blob/f62fc3ac5f480b5a74765e642cd33d17600f7f5a/imu_sensor_broadcaster/src/imu_sensor_broadcaster_parameters.yaml#L16-L18)).

By creating the `ParamListener` and reading the parameters on configure instead of on init is working for me:
```bash
[spawner-4] [INFO] [1696930457.282406584] [spawner_imu_sensor_broadcaster]: Set controller type to "imu_sensor_broadcaster/IMUSensorBroadcaster" for imu_sensor_broadcaster
[gzserver-1] [INFO] [1696930457.283297464] [controller_manager]: Loading controller 'imu_sensor_broadcaster'
[gzserver-1] [INFO] [1696930457.295975680] [controller_manager]: Setting use_sim_time=True for imu_sensor_broadcaster to match controller manager (see ros2_control#325 for details)
[spawner-4] [INFO] [1696930457.302856674] [spawner_imu_sensor_broadcaster]: Loaded imu_sensor_broadcaster
[spawner-4] [INFO] [1696930457.305764498] [spawner_imu_sensor_broadcaster]: Loaded parameters file "/home/yueerro/git/my_dockers/shared/alum/staging/pmb3_ws/install/pmb3_controller_configuration/share/pmb3_controller_configuration/config/imu_sensor_broadcaster.yaml" for imu_sensor_broadcaster
[spawner-4] [INFO] [1696930457.306153768] [spawner_imu_sensor_broadcaster]: Loaded /home/yueerro/git/my_dockers/shared/alum/staging/pmb3_ws/install/pmb3_controller_configuration/share/pmb3_controller_configuration/config/imu_sensor_broadcaster.yaml into imu_sensor_broadcaster
[gzserver-1] [INFO] [1696930457.306952999] [controller_manager]: Configuring controller 'imu_sensor_broadcaster'
[spawner-4] [INFO] [1696930457.333095661] [spawner_imu_sensor_broadcaster]: Configured and activated imu_sensor_broadcaster
[spawner-4] Set parameter sensor_name successful
[spawner-4] Set parameter frame_id successful
[spawner-4] Set parameter publish_rate successful
[INFO] [spawner-4]: process has finished cleanly [pid 66228]
```

I would like to have also a backport for `humble`.

Thank you in advance!

### DIFF
--- a/imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
+++ b/imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
@@ -25,6 +25,12 @@ namespace imu_sensor_broadcaster
 {
 controller_interface::CallbackReturn IMUSensorBroadcaster::on_init()
 {
+  return CallbackReturn::SUCCESS;
+}
+
+controller_interface::CallbackReturn IMUSensorBroadcaster::on_configure(
+  const rclcpp_lifecycle::State & /*previous_state*/)
+{
   try
   {
     param_listener_ = std::make_shared<ParamListener>(get_node());
@@ -33,17 +39,9 @@ controller_interface::CallbackReturn IMUSensorBroadcaster::on_init()
   catch (const std::exception & e)
   {
     RCLCPP_ERROR(
-      get_node()->get_logger(), "Exception thrown during init stage with message: %s \n", e.what());
+      get_node()->get_logger(), "Exception thrown during configure stage with message: %s \n", e.what());
     return CallbackReturn::ERROR;
   }
-
-  return CallbackReturn::SUCCESS;
-}
-
-controller_interface::CallbackReturn IMUSensorBroadcaster::on_configure(
-  const rclcpp_lifecycle::State & /*previous_state*/)
-{
-  params_ = param_listener_->get_params();
 
   imu_sensor_ = std::make_unique<semantic_components::IMUSensor>(
     semantic_components::IMUSensor(params_.sensor_name));


christophfroehlich (commented):
Thanks for pointing this out, it seems that I introduced the issue. 

But I'm not sure why this fails, having a look at JTC https://github.com/ros-controls/ros2_controllers/blob/f62fc3ac5f480b5a74765e642cd33d17600f7f5a/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L53-L59
the param_listener is created and get_params() called also in the on_init method.

@bmagyar @destogl where is the correct place doing this? Is this documented somewhere?
https://github.com/ros-controls/ros2_controllers/blob/f62fc3ac5f480b5a74765e642cd33d17600f7f5a/doc/writing_new_controller.rst?plain=1#L55-L58
vs
https://github.com/ros-controls/ros2_control/blob/e07736011bf28895596cf10dcb7cbcdb40d86daa/hardware_interface/doc/writing_new_hardware_component.rst?plain=1#L48-L52

bmagyar:
I think you are onto something!

Even recent code such as the toy controller I created for the roscon workshop does the same in the `on_init()` function. That being said, those parameters had a default value and no validator.

I think we need to take a closer look at generate_parameter_library and find where validation happens.

On first read I don't see anything wrong with moving those two lines into `on_configure()` and it'd probably be good to future-proof all our controllers if this is the right solution.

christophfroehlich:
Maybe @tylerjw can help to answer how to use generate_parameter_library properly here?

christophfroehlich:
thanks @tylerjw, best-practice would be then doing it in on_configure of all lifecycle nodes?

tylerjw:
Maybe, I'm not confident on the best way to use a lifecycle node.

saikishor:
Hello!

As far as I have checked, It seems like we can declare params in the `init` method, however, we can only retrieve them in the configure phase as we tend to register the callbacks of the node post `on_init` method.

https://github.com/ros-controls/ros2_control/blob/bd6911d663abb687dc956bf19a104d03648f8985/controller_interface/src/controller_interface_base.cpp#L34-L55

I think for the same reason, we get the parameters of the `update_rate` and `is_async` in the `on_configure` method
https://github.com/ros-controls/ros2_control/blob/bd6911d663abb687dc956bf19a104d03648f8985/controller_interface/src/controller_interface_base.cpp#L85-L89

christophfroehlich:
I discussed this with @saikishor 

- In general, parameters should be declared in `on_init` to be able to set parameters (e.g. via CLI) before the controller is configured. This is also the issue with the failing tests of this PR -> they are not declared and, e.g., 
`imu_broadcaster_->get_node()->set_parameter({"sensor_name", sensor_name_});` fails,.
- If generate_parameter_library is used, this is done by the constructor of the `ParamListener` by calling the `declare_params` method.

But I still don't know why `get_params` is failing for @YueErro here. Could you check if the creation of `ParamListener` in the `on_init()` but leaving `get_params()` in the `on_configure()` works for you? Then the CI should be happy here too.

saikishor:
Hello @YueErro!

Can you please check if the fix in the above PR address your issue or not?

bmagyar:
Fyi I'm happy to merge this PR if you get a green CI

destogl:
@tylerjw and @christophfroehlich will your fix of those issues in Generate Parameter Library be beckported to humble? If yes, then there is not need for fixing this.

tylerjw:
Yes, I will backport to humble.

erickisos (approved):

PR #796, opened 2023-10-10
[JTC] Angle wraparound for first segment of trajectory

christophfroehlich:
I experienced a problem with the angle_wraparound of continuous joints if the current position and the first point of the trajectory have an offset of 2*pi.

I forgot that case when I implemented angle_wraparound, and saw now that this was already [part of ROS 1](https://github.com/ros-controls/ros_controllers/blob/7df8be7eaf8c6b90d35f477c5cea0669d78d8f3b/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h#L434). So I ported the [existing implementation](https://github.com/ros-controls/ros_controllers/blob/7df8be7eaf8c6b90d35f477c5cea0669d78d8f3b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_segment.h#L223) and updated the tests (I think now that they were wrong before anyways).

This wraparound of `state_before_traj_msg_.positions` could also be done inside the update method of the controller instead of the `trajectory` class (avoiding ABI break). Comment if you think it should be changed.

I added new tests for the trajectory class to check if `wraparound_joint` works.  `position_error_not_angle_wraparound` and `position_error_angle_wraparound` could also be improved, I opened #802


joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -44,10 +45,31 @@ Trajectory::Trajectory(
...
+
+        // Deal with singularity at M_PI shortest distance
+        if (std::abs(std::abs(dist) - M_PI) < 1e-9)
+        {
+          dist = next_position > prev_position ? std::abs(dist) : -std::abs(dist);
+        }
+        state_before_traj_msg_.positions[i] = next_position - dist;
+      }
+    }
+  }
```
bmagyar, line 72: could we please maintain a similar form for this as in the `ros_control` code? we could unit test that one function in a cleaner way more extensively than through the full controller code

bmagyar:
I think it's more consistent to keep this change to the Trajectory class than avoiding ABI breakages.
Also... better to not load more things into `update()`

joint_trajectory_controller/src/trajectory.cpp
christophfroehlich: you mean moving the code into a separate function like `wraparoundJointOffset`, but call that in `set_point_before_trajectory_msg`?

bmagyar: yes!

christophfroehlich: roger

christophfroehlich: done

christophfroehlich:
@bmagyar do you agree with the changes now and might want to approve it?

destogl (approved):
Looking good! Thanks

joint_trajectory_controller/src/trajectory.cpp
bmagyar: thank you!

bmagyar (approved):
Very good! I only split the tests into the separate cases

PR #797, opened 2023-10-10
[JTC] Fix state offset tests

christophfroehlich:
I tried to write tests for #794 and copied the initialization of `joint_pos_` from `test_hw_states_has_offset_first_controller_start`/`test_hw_states_has_offset_later_controller_start`: This did not work because `ActivateTrajectoryController` always has set it to the constant values from the utils-header -> I fixed that.

Tests failed now, and thinking about the test criteria itself, considering #189, I rewrote the tests -> succeed now.
> add approach to change controller without "jumps" from the command to state values:
> 
>    * when started HW interface has to set command_interfaces to N aN
>    * controller checks if command interfaces are NaN --> then is a controller started for the first time, so read states
>    * if command interface have a value --> so a controller was running already (supports switching between controllers)

@destogl You have added this part, before it got changed a bit by #340 by @AndyZe and @mechwiz 
* Why should we set `state_desired_` to the value of `read_state_from_command_interfaces`? If there is no active trajectory (`start_with_holding` is false), this value will only be published on the state-topic. If there is a new trajectory, it will be overwritten by the sampled trajectory.
* Setting `state_current` to the command read from the hardware interface is fine: It is only used for setting the set_hold_position if `start_with_holding` is set. It will be overwritten by the first call of `update`.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -263,6 +263,12 @@ protected:
 
   void read_state_from_hardware(JointTrajectoryPoint & state);
 
+  /** Assign values from the command interfaces as state.
+   * This is only possible if command AND state interfaces exist for the same type,
+   *  therefore needs check for both.
+   * @param[out] state to be filled with values from command interfaces.
+   * @return true if all interfaces exists and contain non-NaN values, false otherwise.
+   */
   bool read_state_from_command_interfaces(JointTrajectoryPoint & state);
   bool read_commands_from_command_interfaces(JointTrajectoryPoint & commands);
 
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -261,7 +261,7 @@ protected:
     const rclcpp::Time & time, const JointTrajectoryPoint & desired_state,
     const JointTrajectoryPoint & current_state, const JointTrajectoryPoint & state_error);
 
-  void read_state_from_hardware(JointTrajectoryPoint & state);
+  void read_state_from_state_interfaces(JointTrajectoryPoint & state);
 
   /** Assign values from the command interfaces as state.
    * This is only possible if command AND state interfaces exist for the same type,
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -951,12 +951,8 @@ controller_interface::CallbackReturn JointTrajectoryController::on_activate(
 
   subscriber_is_active_ = true;
 
-  // Initialize current state storage if hardware state has tracking offset
-  read_state_from_hardware(state_current_);
-  read_state_from_hardware(state_desired_);
-  read_state_from_hardware(last_commanded_state_);
-  // Handle restart of controller by reading from commands if
-  // those are not nan
+  // Handle restart of controller by reading from commands if those are not nan (a controller was
+  // running already)
   trajectory_msgs::msg::JointTrajectoryPoint state;
   resize_joint_trajectory_point(state, dof_);
   if (read_state_from_command_interfaces(state))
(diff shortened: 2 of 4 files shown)


bmagyar (approved):
let's wait for @destogl to reply too

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -999,6 +996,13 @@ controller_interface::CallbackReturn JointTrajectoryController::on_activate(
     state_desired_ = state;
     last_commanded_state_ = state;
   }
+  else
+  {
+    // Initialize current state storage from hardware
+    read_state_from_hardware(state_current_);
+    read_state_from_hardware(state_desired_);
+    read_state_from_hardware(last_commanded_state_);
```
destogl, line 1004: When we are already fixing this. I would propose to rename the method to `read_state_from_state_interfaces` – it is a longer but clearer name. Read state from hardware is a bit unclear. Especially now with having chaining of state interfaces, this might be confusing.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -265,6 +265,12 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
 
   void read_state_from_hardware(JointTrajectoryPoint & state);
```
destogl, line 266: When we are already fixing this. I would propose to rename the method to read_state_from_state_interfaces – it is a longer but clearer name. Read state from hardware is a bit unclear. Especially now with having chaining of state interfaces, this might be confusing.

```suggestion
  void read_state_from_state_interfaces(JointTrajectoryPoint & state);
```

destogl (commented):
Just a small proposal for optimization since we are discussing those functions.

## Answers to comments

> * Why should we set `state_desired_` to the value of `read_state_from_command_interfaces`? If there is no active trajectory (`start_with_holding` is false), this value will only be published on the state-topic. If there is a new trajectory, it will be overwritten by the sampled trajectory.

Can it be that we change this when fixing the “hold on start” feature? I think the reason was that we have immediate output to the command interface even if there is no reference. If we can achieve that by using `start_with_holding` this is also fine with me. Nevertheless, I would not add additional if, even if we set the value “unnecessary”. Or do you think that this could be confusing?

> * Setting `state_current` to the command read from the hardware interface is fine: It is only used for setting the set_hold_position if `start_with_holding` is set. It will be overwritten by the first call of `update`.

OK, you are right. Then we probably should not set `state_desired` in that case.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: fine!

christophfroehlich:
> > * Why should we set `state_desired_` to the value of `read_state_from_command_interfaces`? If there is no active trajectory (`start_with_holding` is false), this value will only be published on the state-topic. If there is a new trajectory, it will be overwritten by the sampled trajectory.
> 
> Can it be that we change this when fixing the “hold on start” feature? I think the reason was that we have immediate output to the command interface even if there is no reference. If we can achieve that by using `start_with_holding` this is also fine with me. Nevertheless, I would not add additional if, even if we set the value “unnecessary”. Or do you think that this could be confusing?

I'll think that trough again if this is achieved with the hold-on-start in a fresh moment.

Optional override is already inside an if, for me it is clearer to have it in the else branch instead (but that's only my personal preference).

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
christophfroehlich: done

christophfroehlich:
> > * Why should we set `state_desired_` to the value of `read_state_from_command_interfaces`? If there is no active trajectory (`start_with_holding` is false), this value will only be published on the state-topic. If there is a new trajectory, it will be overwritten by the sampled trajectory.
> 
> Can it be that we change this when fixing the “hold on start” feature? I think the reason was that we have immediate output to the command interface even if there is no reference. If we can achieve that by using `start_with_holding` this is also fine with me. Nevertheless, I would not add additional if, even if we set the value “unnecessary”. Or do you think that this could be confusing?

The current way with this PR is:
- If we have `start_with_holding=false`, no update of the command interface until the first trajectory is received. See `has_active_trajectory()` in `update` method.
- `start_with_holding=true`: There will be an immediate update of the command interfaces, with position set from `state_current_`, which is
  - set to the values of the command interfaces, if they are not NaN or
  - set to the values read from the state interfaces otherwise.

Notes:
- I asked in #816 if we need `start_with_holding` parameter at all. 
- with #794 there would be no change, because `last_commanded_state_` is initialized identically.

@destogl What do you think?

destogl (approved):
This looks great! Thanks!

PR #798, opened 2023-10-15
[JTC] Update requirements of state interfaces

christophfroehlich:
A small adaption of the docs explaining requirements of command and state interfaces.


bmagyar (approved):
Great catch, thank you!

PR #803, opened 2023-10-22
[JTC] Cleanup comments and unnecessary checks

christophfroehlich:
- I found some old and wrong comments
- No specialization is planned any more, see #742 
- dof can't change any more #771


bmagyar (approved):
thank you!

PR #804, opened 2023-10-22
[CI] Fix coverage build and codecov stats

christophfroehlich:
- The Codecov stats were only evaluated for diff_drive_controller, see [old](https://app.codecov.io/gh/ros-controls/ros2_controllers) vs [new](https://app.codecov.io/gh/ros-controls/ros2_controllers/tree/patch-2).
- The test folders have very low coverage (why?). I excluded them from the statistics as I don't think it makes sense to include them.

### DIFF
--- a/.github/workflows/ci-coverage-build.yml
+++ b/.github/workflows/ci-coverage-build.yml
@@ -29,7 +29,7 @@ jobs:
           import-token: ${{ secrets.GITHUB_TOKEN }}
           # build all packages listed in the meta package
           package-name:
-            diff_drive_controller
+            ros2_controllers
 
           vcs-repo-file-url: |
             https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/ros2_controllers-not-released.${{ env.ROS_DISTRO }}.repos?token=${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci-coverage-build.yml
+++ b/.github/workflows/ci-coverage-build.yml
@@ -28,7 +28,7 @@ jobs:
           target-ros2-distro: ${{ env.ROS_DISTRO }}
           import-token: ${{ secrets.GITHUB_TOKEN }}
           # build all packages listed in the meta package
-          package-name:            
+          package-name:
             ackermann_steering_controller
             admittance_controller
             bicycle_steering_controller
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-coverage-build.yml
```diff
@@ -28,8 +28,24 @@ jobs:
           target-ros2-distro: ${{ env.ROS_DISTRO }}
           import-token: ${{ secrets.GITHUB_TOKEN }}
           # build all packages listed in the meta package
-          package-name:
+          package-name:
```
christophfroehlich, line 31:
```suggestion
          package-name:
```

.github/workflows/ci-coverage-build.yml
```diff
@@ -29,7 +29,23 @@ jobs:
           import-token: ${{ secrets.GITHUB_TOKEN }}
           # build all packages listed in the meta package
```
christophfroehlich, line 30:
```suggestion
          # build all packages listed here
```

christophfroehlich:
idk why the CI job fails, something is happening with [joint_state_broadcaster](https://github.com/ros-controls/ros2_controllers/actions/runs/6605604132/job/17940940887?pr=804#step:4:4199).

bmagyar:
We don't test the tests enough, right? ;)

christophfroehlich:
and the flaky tests now also affect the coverage build :/

bmagyar:
Could the codecov builds be very different? Or only the environment? Something is making some tests flakier than they usually are

christophfroehlich:
I'm not sure, but I have to save that :grin: 
image

christophfroehlich:
> Could the codecov builds be very different? Or only the environment? Something is making some tests flakier than they usually are

`ros-tooling/action-ros-ci `vs `ros-industrial/industrial_ci`
maybe the mixin makes a totally different build
https://github.com/ros-controls/ros2_controllers/blob/22356e0e9d39c726926f3f3a0ddb7afd51333295/.github/workflows/ci-coverage-build.yml#L36-L41

bmagyar:
> I'm not sure, but I have to save that 😁 image

Top contributor of the year

bmagyar (approved):

PR #807, opened 2023-10-30
[CI] Codecov 

christophfroehlich:
- Add license + codecov badge to readme (as in ros2_control)
- Remove the invalid `branches` key from the `workflow_displatch`

Do we want to activate codecov also for humble?


bmagyar (approved):
I'll go ahead with this one, please add Humble and Iron too (the latter is coming in a day or two)

PR #808, opened 2023-10-30
fixed admittance_controller/AdmittanceController::read_state_from_hardware()

tpoignonec:
__Issue:__ only one state interface is read, but at least position AND velocity should be read.

__Cause:__ "else if" statements used instead of "if" in `AdmittanceController::read_state_from_hardware()`.

__Solution:__ see edits.

### DIFF
--- a/admittance_controller/src/admittance_controller.cpp
+++ b/admittance_controller/src/admittance_controller.cpp
@@ -491,13 +491,13 @@ void AdmittanceController::read_state_from_hardware(
         state_interfaces_[pos_ind * num_joints_ + joint_ind].get_value();
       nan_position |= std::isnan(state_current.positions[joint_ind]);
     }
-    else if (has_velocity_state_interface_)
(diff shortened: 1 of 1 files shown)


tpoignonec:
N.B., actually this seems to be common to all branches (e.g., see [main branch](https://github.com/tpoignonec/ros2_controllers/blob/master/admittance_controller/src/admittance_controller.cpp#L483-L489)).

bmagyar:
I tried simply retargeting the PR but it isn't that simple unfortunately. Please rebase this PR to the master branch and we can backport it to humble.

bmagyar:
Additionally, if you could add a test that reproduces the issue that'd be great

tpoignonec:
Actually, we can close this PR, I hadn't realized the controller was made for position interface specifically.
After investigation, my issue was more due to the fact that I tried to modify the code to do something it wasn't designed to do...
Sorry for the inconvenience.

PR #809, opened 2023-10-30
[JTC] Accept larger number of `joints` than `command_joints`

christophfroehlich:
This PR softens the condition that the number of `joints` and `command_joints` have to be equal: JTC now supports having less `command_joints` than `joints`.

This gives the opportunity to track the state and error of passive joints, especially joints that are not `command_joints` but a degree-of-freedom of the physical system. 

As an example, see the cart-pole example in the following video (a pendulum on a cart). Now it is possible to send a swing-up trajectory of the pendulum including values of the cart and the pendulum joint, JTC will do the trajectory interpolation and publishes the full state including the desired values for the pendulum joint, but it finally only can command the cart-joint.

https://github.com/ros-controls/ros2_controllers/assets/3367244/d9c89a01-9940-442b-83af-fd64a7a83e12

This PR does not directly address passive joints being part of a closed-loop kinematic chain, i.e., joints not being a degree-of-freedom of the system. One could track the trajectory of all joints being part of the closed loop, but the kinematics need to be solved in the hardware interface, e.g., by means of a transmission interface.

A future PR will propose a possibility to implement different control laws than PID, e.g., a state-space controller, which is in fact necessary to control the pendulum from the example above.

More implementation details
- `command_joints` must be a subset of `joints` if they don't have the same size.
- If they have the same size, a 1:1 mapping is assumed independent of the values of `command_joints`. (see `check_interface_names_with_command_joints` test)
- `~/controller_state` topic uses a single `joint_names` vector for all fields. Hence, entries of the output field being not included in `command_joints` will be NaN.

This would break rqt_joint_trajectory_controller, I used now the required state interface instead of claimed hardware interfaces from the contoller_manager msg.


saikishor (approved):
LGTM

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -912,10 +942,12 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
     std::bind(&JointTrajectoryController::goal_accepted_callback, this, _1));
 
   resize_joint_trajectory_point(state_current_, dof_);
-  resize_joint_trajectory_point_command(command_current_, dof_);
+  resize_joint_trajectory_point_command(
+    command_current_, dof_, std::numeric_limits<double>::quiet_NaN());
```
saikishor, line 946: Shouldn't this be `num_cmd_joints_`?

```suggestion
    command_current_, num_cmd_joints_, std::numeric_limits<double>::quiet_NaN());
```

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
```diff
@@ -461,12 +461,9 @@ def _joint_widgets(self):  # TODO: Cache instead of compute every time?
 
 
 def _jtc_joint_names(jtc_info):
-    # NOTE: We assume that there is at least one hardware interface that
-    # claims resources (there should be), and the resource list is fetched
-    # from the first available interface
 
     joint_names = []
-    for interface in jtc_info.claimed_interfaces:
+    for interface in jtc_info.required_state_interfaces:
```
saikishor, line 466: Do we need this changes?

saikishor (commented):
In generl, looks very good to me.

Just some nitpicking things

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
christophfroehlich: thb, I don't remember well (this PR is 18months old). But according to the note in the initial commit, I believe yes :D

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: no. it has size of `dof_` and is later indexed by the `map_cmd_to_joints_`. The full vectory is then used for the status message publisher, where the non-commands remain NaN

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -724,12 +726,40 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
...
       logger, "No specific joint names are used for command interfaces. Using 'joints' parameter.");
   }
-  else if (command_joint_names_.size() != params_.joints.size())
+  num_cmd_joints_ = command_joint_names_.size();
+
+  if (num_cmd_joints_ > dof_)
   {
     RCLCPP_ERROR(
-      logger, "'command_joints' parameter has to have the same size as 'joints' parameter.");
+      logger, "'command_joints' parameter must not have greater size as 'joints' parameter.");
```
christophfroehlich, line 734:
```suggestion
      logger, "The 'command_joints' parameter must not exceed the size of the 'joints' parameter.");
```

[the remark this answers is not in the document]
christophfroehlich: I reformulated it a bit again, what do you think?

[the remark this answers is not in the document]
saikishor: Looks good!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: Got it.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
christophfroehlich: yes. the thing is, that you **can** send a trajectory messages including the non-command DOF now. (does not make much sense for the GUI maybe, but you will see the effect then).
should not change anything for existing configs, right?

saikishor: Thanks for the explanation

saikishor (approved):
Thanks for the explanation.

LGTM

PR #813, opened 2023-11-01
Use pre-commit fork of clang-format instead of local

bmagyar:
10/10 would use it again


bijoua29:
What exactly wasn't working - I'm curious?

aprotyas (approved):

bmagyar:
This doesn't require a certain version of clang-format to be installed
locally. Good if you want to use a different version locally but also helps
people who don't have the same setup to "just call pre-commit" for
formatting

PR #814, opened 2023-11-04
Add iron branch

bmagyar:
Adding bits to support an iron branch


.github/workflows/iron-abi-compatibility.yml
```diff
@@ -1,11 +1,11 @@
-name: Humble - ABI Compatibility Check
+name: Iron - ABI Compatibility Check
 on:
   workflow_dispatch:
     branches:
-      - humble
+      - master
```
christophfroehlich, line 5:
```suggestion
      - iron
```
should't this be iron?

.github/workflows/iron-abi-compatibility.yml
```diff
@@ -1,11 +1,11 @@
...
   pull_request:
     branches:
-      - humble
+      - master
```
christophfroehlich, line 8:
```suggestion
      - iron
```
as above

christophfroehlich (commented):
why are you renaming the humble workflow files, and not adding iron additionally?

.github/mergify.yml
```diff
@@ -8,6 +8,16 @@ pull_request_rules:
         branches:
           - humble
 
+pull_request_rules:
```
christophfroehlich, line 11:
```suggestion


```
see format job

PR #825, opened 2023-11-10
Added controller_namesapce to the left_wheel_names and right_wheel_na…

suchetanrs:
The controller namespace is added to the odom_frame_id and the base_frame_id. However, it is not done to the left_wheel_names and right_wheel_names.

The differential drive controller is often used as a gazebo plugin like below. 
```
    <gazebo>
      <plugin filename="libgazebo_ros2_control.so" name="gazebo_ros2_control">
        <namespace>${robot_namespace}</namespace>
        <robot_param>robot_description</robot_param>
        <robot_param_node>robot_state_publisher</robot_param_node>
        <parameters>package://robot_description/config/robot_control.yaml</parameters>
      </plugin>
    </gazebo>
```

Since this is defined inside a xacro, the Rewritten yaml cannot be used to add namespaces to joints. Adding the namespace in the cpp file bypasses this issue.  In the PR, the convention is followed. That is, "robot_namespace/left_wheel_name". Please let me know if I need to change it in any other place. The modified changes solves my issue and seems to run correctly on the Gazebo simulator.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -405,7 +405,16 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
 
   const auto odom_frame_id = controller_namespace + params_.odom_frame_id;
   const auto base_frame_id = controller_namespace + params_.base_frame_id;
-
+  
+  for (auto & joint_name : params_.left_wheel_names)
+  {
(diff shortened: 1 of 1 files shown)


christophfroehlich (changes requested):
LGTM. 

According to our guidelines: Please target the PR to the master branch, and we will perform backports to iron+humble.

PR #826, opened 2023-11-10
Use Generic Steering Controller State Message in Steering Controllers Library

mbharatheesha:
I was reading through the code as a part of learning how to write a new steering controller. I understood that a new steering controller inherits from the SteeringControllersLibrary. But I also found the use of `Ackermannxxx` in the base class code for a status message and the same was being used in other steering controllers such as `BicycleSteeringController` and `TricycleSteeringController`. 

If I understood the inheritance and the code structure correctly, the SteerinControllersLibrary should be agnostic to the specifics of a steering controller that derives from it. Therefore, this PR introduces the following changes:

- Remove the use of `Ackermannxxx` declarations in the `SteeringControllersLibrary` base class and rename it to `SteeringControllerStateMsg`
- Modify the tests and state publishers to use the renamed declaration
- Remove `ackermann_msgs` dependency from `steering_controllers_library` package.

All tests for the modified packages were run successfully (locally).

### DIFF
--- a/ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
+++ b/ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
@@ -34,7 +34,7 @@
 #include "rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp"
 
 using ControllerStateMsg =
-  steering_controllers_library::SteeringControllersLibrary::AckermanControllerState;
+  steering_controllers_library::SteeringControllersLibrary::SteeringControllerStateMsg;
 using ControllerReferenceMsg =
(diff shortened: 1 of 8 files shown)


christophfroehlich (changes requested):
LGTM, thanks for the cleanup.

According to our guidelines: Please target the PR to the master branch, and we will perform backports to iron+humble.

bmagyar:
For humble and iron we can only deprecate the use of these messages. Full removal can be done on rolling

christophfroehlich:
but it doesn't change anything for users, only inherited classes may be broken. we could maybe use the `[[deprecated]]` c++ qualifier if necessary.

mbharatheesha:
Apologies for the late response. Thanks for your remarks. I will submit a new PR targeted at master later today.

PR #834, opened 2023-11-14
[JTC] Remove deprecation warnings, set `allow_nonzero_velocity_at_trajectory_end` default false

christophfroehlich:
As per title
- Remove deprecation warnings (both merged since two releases)
   - #772
   - set `allow_nonzero_velocity_at_trajectory_end` default false. Might break some setups without setting it manually to true (since #609)
- fix tests
- rename class variables from normalize_error to angle_wraparound

### DIFF
--- a/joint_trajectory_controller/doc/parameters.rst
+++ b/joint_trajectory_controller/doc/parameters.rst
@@ -66,7 +66,7 @@ start_with_holding (bool)
 allow_nonzero_velocity_at_trajectory_end (boolean)
   If false, the last velocity point has to be zero or the goal will be rejected.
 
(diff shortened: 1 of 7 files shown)


bmagyar (approved):
thanks!

fmauch (approved):
lgtm, I also prepared a PR for our Scaled JTC.

destogl (approved):

christophfroehlich:
Can we backport this to iron as it is, too? see conflicht #837 
For humble, I'll do a manual backport for the variable renamings. Or can we update the deprecations there too?

PR #844, opened 2023-11-17
Ignore test code for codecov

christophfroehlich:
We forgot to backport this one from #804


bmagyar (approved):
You are already top contributor for the month :D

PR #845, opened 2023-11-17
[JTC] Remove unused home pose

### LINKED ISSUE #841: [JTC] home pose seems to be broken: explain, fix or remove?
JTC has code for a `home_pose`, but I think it is broken (and there is no test for it).

There are two shared pointers:
```c++
  std::shared_ptr<Trajectory> traj_home_point_ptr_ = nullptr;
  std::shared_ptr<trajectory_msgs::msg::JointTrajectory> traj_msg_home_ptr_ = nullptr;
```

- [`on_activate()` ](https://github.com/ros-controls/ros2_controllers/blob/3fc0f50c440ad49b2194b2c7dae8c54d6cda7b07/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L924C6-L939)
   - `traj_msg_home_ptr_` is created and set to the position of the joint_state_interface
   - `traj_home_point_ptr_` created with an empty `std::make_shared<Trajectory>()` 
- Both will be resetted at `on_deactivate()`
https://github.com/ros-controls/ros2_controllers/blob/3fc0f50c440ad49b2194b2c7dae8c54d6cda7b07/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1031-L1032
- `on_cleanup()`:  would be used if not a `nullptr`
https://github.com/ros-controls/ros2_controllers/blob/3fc0f50c440ad49b2194b2c7dae8c54d6cda7b07/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1041-L1045

Am I right that: [lifecycle nodes](https://design.ros2.org/articles/node_lifecycle.html) have no possible transition from active state to unconfigured without `on_deactivate()`, which means that within `on_cleanup()` it can only hold `nullptr` and no action is performed?

Is there any rationale behind that or can it be deleted?

christophfroehlich:
This closes #841, see the discussion in the issue.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -184,8 +184,6 @@ protected:
   rclcpp::Service<control_msgs::srv::QueryTrajectoryState>::SharedPtr query_state_srv_;
 
   std::shared_ptr<Trajectory> traj_external_point_ptr_ = nullptr;
-  std::shared_ptr<Trajectory> traj_home_point_ptr_ = nullptr;
-  std::shared_ptr<trajectory_msgs::msg::JointTrajectory> traj_msg_home_ptr_ = nullptr;
   realtime_tools::RealtimeBuffer<std::shared_ptr<trajectory_msgs::msg::JointTrajectory>>
     traj_msg_external_point_ptr_;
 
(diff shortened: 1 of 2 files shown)


saikishor (approved):
LGTM. Thank you so much for cleaning it up.

christophfroehlich:
Removing untested code is decreasing coverage, that's not fair :smiling_face_with_tear:

bmagyar (approved):

PR #855, opened 2023-11-20
added documentation on common controller parameters

saikishor:
I've add some minor documentation update on the common controller parameters for every ROS2 Controller. 
@christophfroehlich FYI


doc/controllers_index.rst
```diff
@@ -72,3 +72,12 @@ In the sense of ros2_control, broadcasters are still controllers using the same
    IMU Sensor Broadcaster <../imu_sensor_broadcaster/doc/userdoc.rst>
    Joint State Broadcaster <../joint_state_broadcaster/doc/userdoc.rst>
    Range Sensor Broadcaster <../range_sensor_broadcaster/doc/userdoc.rst>
+
+
+Common Controller Parameters
+****************************
+
+Every controller and broadcaster have a few common parameters that are necessary upon initialization. These parameters are `update_rate` and `is_async`.
```
christophfroehlich, line 80:
```suggestion
Every controller and broadcaster has a few common parameters that are optional upon initialization:
```

christophfroehlich (commented):
Thanks. These parameters are optional, not necessary? or do you mean it is necessary that they are set before init if desired?

doc/controllers_index.rst
saikishor: These parameters are needed to be set before configuring the controller.

christophfroehlich: I updated my suggestion.

I saw now that the docs lack of proper description of the lifecycle changes/transitions (nothing to link to).

doc/controllers_index.rst
```diff
@@ -72,3 +72,12 @@ In the sense of ros2_control, broadcasters are still controllers using the same
...
+
+* `update_rate`: An unsigned integer parameter representing the rate at which every controller/broadcaster runs its update cycle. When unspecified, they run at the same frequency as the controller_manager.
+* `is_async`: A boolean parameter that is needed to specify if the controller update needs to run asynchronously.
```
christophfroehlich, line 83:
```suggestion
* ``update_rate``: An unsigned integer parameter representing the rate at which every controller/broadcaster runs its update cycle. When unspecified, they run at the same frequency as the controller_manager.
* ``is_async``: A boolean parameter that is needed to specify if the controller update needs to run asynchronously.
```

doc/controllers_index.rst
saikishor: Hello, I reformatted a bit. Can you verify?

Every controller and broadcaster has a few common parameters. They are optional, but if needed they have to be set before ``onConfigure`` transition to ``inactive`` state, see `lifecycle documents <https://design.ros2.org/articles/node_lifecycle.html>`__. Once the controllers are already loaded, this transition is done using the ``configure_controller`` service of the controller_manager:

doc/controllers_index.rst
```diff
@@ -72,3 +72,12 @@ In the sense of ros2_control, broadcasters are still controllers using the same
    IMU Sensor Broadcaster <../imu_sensor_broadcaster/doc/userdoc.rst>
    Joint State Broadcaster <../joint_state_broadcaster/doc/userdoc.rst>
    Range Sensor Broadcaster <../range_sensor_broadcaster/doc/userdoc.rst>
+
+
+Common Controller Parameters
+****************************
+
+Every controller and broadcaster has a few common parameters. They are optional, but if needed they have to be set before ``onConfigure`` transition to ``inactive`` state, see `lifecycle documents <https://design.ros2.org/articles/node_lifecycle.html>`__. Once the controllers are already loaded, this transition is done using the `configure_controller` service of the controller_manager.
```
christophfroehlich, line 80:
```suggestion
Every controller and broadcaster has a few common parameters. They are optional, but if needed they have to be set before ``onConfigure`` transition to ``inactive`` state, see `lifecycle documents <https://design.ros2.org/articles/node_lifecycle.html>`__. Once the controllers are already loaded, this transition is done using the service ``configure_controller`` of the controller_manager.
```

christophfroehlich (approved):
LGTM now.

bmagyar (approved):

PR #856, opened 2023-11-25
Increase test coverage

christophfroehlich:
Adding simple test conditions with calls of
```c++
  command_interface_configuration();
  state_interface_configuration();
```
to increase test coverage of some controllers/broadcasters.


bmagyar (approved):
Ah this is super nice, thank you!

PR #857, opened 2023-11-25
[JTC] Activate checks for parameter validation

christophfroehlich:
I refactored the temporarily deactivated test for parameter validation. 

It checks now if validators are set correctly for generate_parameter_library (GPL checks for parameter itself only; during `on_init()`) and if valid combinations are checked by JTC during `on_configure()`

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_controller.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller.cpp
@@ -2016,8 +2016,8 @@ TEST_F(TrajectoryControllerTest, incorrect_initialization_using_interface_parame
   // command interfaces: empty
   command_interface_types_ = {};
   EXPECT_EQ(SetUpTrajectoryControllerLocal(), controller_interface::return_type::OK);
-  traj_controller_->get_node()->configure();
-  EXPECT_EQ(traj_controller_->get_state().id(), State::PRIMARY_STATE_UNCONFIGURED);
+  auto state = traj_controller_->get_node()->configure();
+  EXPECT_EQ(state.id(), State::PRIMARY_STATE_UNCONFIGURED);
 
   // command interfaces: bad_name
   command_interface_types_ = {"bad_name"};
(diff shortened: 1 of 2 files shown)


joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -2005,72 +2005,73 @@ INSTANTIATE_TEST_SUITE_P(
...
+  // state interfaces: acceleration without position and velocity
+  state_interface_types_ = {"acceleration"};
+  EXPECT_EQ(SetUpTrajectoryControllerLocal(), controller_interface::return_type::ERROR);
+
+  // velocity-only command interface: position - velocity not present
+  command_interface_types_ = {"velocity"};
+  state_interface_types_ = {"position"};
+  EXPECT_EQ(SetUpTrajectoryControllerLocal(), controller_interface::return_type::OK);
+  traj_controller_->get_node()->configure();
+  EXPECT_EQ(traj_controller_->get_state().id(), State::PRIMARY_STATE_UNCONFIGURED);
```
saikishor, line 2065: https://github.com/ros-controls/ros2_controllers/blob/0d3fc520831d10d4c1d16784e7378c2a0bf7b10f/joint_trajectory_controller/doc/userdoc.rst?plain=1#L37-L39

Shouldn't the velocity command interface need both position and velocity as state interfaces to configure?. If this is the case, then let's add an expect to check that the configure returns failure as per this

https://github.com/ros-controls/ros2_controllers/blob/0d3fc520831d10d4c1d16784e7378c2a0bf7b10f/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L751-L760

```suggestion
  // velocity-only command interface: position - velocity not present
  command_interface_types_ = {"velocity"};
  state_interface_types_ = {"position"};
  EXPECT_EQ(SetUpTrajectoryControllerLocal(), controller_interface::return_type::OK);
  EXPECT_EQ(traj_controller_->get_node()->configure(), controller_interface::return_type::FAILURE);
  EXPECT_EQ(traj_controller_->get_state().id(), State::PRIMARY_STATE_UNCONFIGURED);
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -2005,72 +2005,73 @@ INSTANTIATE_TEST_SUITE_P(
...
+  state_interface_types_ = {"velocity"};
+  EXPECT_EQ(SetUpTrajectoryControllerLocal(), controller_interface::return_type::ERROR);
+
+  // effort-only command interface: position - velocity not present
+  command_interface_types_ = {"effort"};
+  state_interface_types_ = {"position"};
+  EXPECT_EQ(SetUpTrajectoryControllerLocal(), controller_interface::return_type::OK);
+  traj_controller_->get_node()->configure();
```
saikishor, line 2073:
```suggestion
  EXPECT_EQ(traj_controller_->get_node()->configure(), controller_interface::return_type::FAILURE);
```

saikishor (commented):
Minor comments. The rest looks good to me. :clap:

christophfroehlich:
> Minor comments. The rest looks good to me. 👏

you are right, but the `configure()` method does not return the State message. But I use the returned state now instead of calling `get_state()` again.

saikishor (approved):
LGTM

bmagyar (approved):

Issue #821, opened 2023-11-07
[CI] joint_state_broadcaster tests are flaky with coverage build


christophfroehlich:
[TestCustomInterfaceMappingUpdate](https://github.com/ros-controls/ros2_controllers/actions/runs/6784920091/job/18442155489#step:4:12130) fails regularly.

>       /home/runner/work/ros2_controllers/ros2_controllers/ros_ws/src/j86feha78d/ros2_controllers/joint_state_broadcaster/test/test_joint_state_broadcaster.cpp:549: Failure
>       Expected equality of these values:
>         joint_state_msg.position[0]
>           Which is: nan
>         custom_joint_value_
>           Which is: 3.5

Haven't had time to dig into this.

Issue #802, opened 2023-10-21
[JTC] improve wraparound tests


christophfroehlich:
`position_error_not_angle_wraparound` and `position_error_angle_wraparound` should check the controller output (and error variable) only with constant position reference but changing position state.

- no interference with trajectory wraparound from #796
- maybe the usage of the state_msg can be removed

Issue #697, opened 2023-07-13
[JTC] conceptual issues


gwalck:
While implementing https://github.com/ros-controls/ros2_controllers/pull/696 I faced numerous issues that led me to dig deep into the implementation of JTC to understand why some "intuitive" or let say theoretical expectations I had for a JTC were not met. I did not want to create tons of separate issues as they are mostly linked together.

## Disclaimer : 
I might have some rusted theoretical knowledge in controllers, so please just point at my mistakes if any and close the issue if this is not an issue.

### dt in update Controller 

* already fixed in https://github.com/ros-controls/ros2_controllers/pull/693

###  traj sampling 

* JTC `update()` called at the `time_from_start` of a single trajectory point will generate a **desired output** that matches the trajectory point. 
* However, in the tests evaluating reached state, since no further calls to `update()`  is done , the current read from hardware state (or state_msgs.feedback) that gets evaluated will ALWAYS be the last state read from hardware and so will always be the state **before** the last update call (even with mirrored cmd/state ifs) and so will **never** be the desired state when evaluated.
*  Current tests introduced `allowed deltas` to compensate for the situation and highly depend on how tiny the delay (dt) between 2 calls of update loop (in `updateController`) is, hence variability in CI. Currently the dt is as small as a few microseconds which is unrealistic to a real robot and is variable depending on CPU load.
* I tried to remove the need for a clock and add larger dt and this really showed the problem see below with a dt = 50 ms, and the test of a trajectory with a point to be reached (and evaluated) at 250 ms.

```
Command interfaces are [position velocity] and state interfaces are [position velocity].

current pos 1.100000              -> call 1   at time 0+ (+ means some EPS due to where now is placed) for time range 0 to 0.05
current vel   0.000000              ->    current velocity not yet changed, this is normal
desired pos 1.100000              -> no desired position  yet  -> THIS seems wrong
desired vel  0.000000              -> same 

current pos 1.100000              -> call 2   at time ~0.05+  for time range 0.05 to 0.1
current vel   0.000000
desired pos 1.090396              -> first desired movement !
desired vel  -0.356063             -> first desired velocity

current pos 1.090396              -> call 3  at time ~0.1+ for time range 0.1 to 0.15
current vel -0.356063               -> current vel (and pos) mirrored from previous desired at call 2
desired pos 1.067190              
desired vel -0.544039

current pos 1.067190              -> call 4  at time ~0.15+ for time range 0.15 to 0.20
[...]

PR #858, opened 2023-11-26
[JTC] Improve update methods for tests

christophfroehlich:
I made an attempt to improve the JTC tests once more.

The ideas come from #697 and #802, and partially fix them.

- `state_topic_consistency` is now the only test subscribing to the state message publisher.
- All other tests now use the class variables used for building the state message -> no problem with the [trylock()](https://github.com/ros-controls/ros2_controllers/blob/0d3fc520831d10d4c1d16784e7378c2a0bf7b10f/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1067) of the realtime publisher, which causes problems with tests not running synchronously with the clock, like with #821 
- This enables a new `updateControllerAsync` test method, which calls the `update()` of JTC with feasible time/period arguments but asynchronous to the current clock.
- The "old" `updateController` test method has a `sleep_for` now, avoiding calling the `update()` of JTC in the while loop with almost zero duration.
- Both methods run until the end time is reached (`<= end_time` instead of `< end_time`). This allows for reducing the thresholds, because the points should be reached exactly, independent of any timing issue.
- Cherry-picking small udpates from #809 allowing to reactivate the jump-tests.

Overall: The total test time of JTC is decreased by more than 1min. And I hope they are less flaky now.

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_actions.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_actions.cpp
@@ -242,7 +242,7 @@ TEST_P(TestTrajectoryActionsTestParameterized, test_success_single_point_sendgoa
   EXPECT_EQ(rclcpp_action::ResultCode::SUCCEEDED, common_resultcode_);
 
   // run an update
-  updateController(rclcpp::Duration::from_seconds(0.01));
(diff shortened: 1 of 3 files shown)


bmagyar (approved):
Excellent!

PR #859, opened 2023-11-26
joint_state_broadcaster: Add proper subscription to TestCustomInterfaceMappingUpdate

### LINKED ISSUE #821: [CI] joint_state_broadcaster tests are flaky with coverage build

christophfroehlich:
As mentioned with #821, one tests of joint_state_broadcaster is flaky in the coverage build job. This is because the creation of the `joint_state_msg` is not ensured due to a failing `trylock()` when calling `update()` only once
https://github.com/ros-controls/ros2_controllers/blob/0d3fc520831d10d4c1d16784e7378c2a0bf7b10f/joint_state_broadcaster/src/joint_state_broadcaster.cpp#L353

I reused the wait set from other tests, which retries the update several times.

Unfortunately, I was not able to reproduce a failing test like in the CI, but this should fix #821 (other tests using the wait_set are not flaky).

### DIFF
--- a/joint_state_broadcaster/test/test_joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/test/test_joint_state_broadcaster.cpp
@@ -531,37 +531,8 @@ TEST_F(JointStateBroadcasterTest, TestCustomInterfaceMappingUpdate)
   state_broadcaster_->get_node()->set_parameter(
     {std::string("map_interface_to_joint_state.") + HW_IF_POSITION, custom_interface_name_});
 
-  // configure ok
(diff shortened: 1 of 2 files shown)


bmagyar (approved):
Thank you!

PR #860, opened 2023-11-26
[CI] Use testing repo for ABI workflows

christophfroehlich:
This should reduce the time span for failing tests after ABI breaking changes. Like currently, where there is a new release in testing but not synced to main yet.


bmagyar (approved):
Good idea!

PR #869, opened 2023-11-29
Backports to iron

christophfroehlich:
A manual backport PR to fix conflicts of open backports.

Replaces #837 #853 #862 #866


bmagyar (approved):

RobertWilbrandt:
Doing such backports and releasing them one day before sync makes it really hard to depend on these packages. For our `ur_controllers` this backport meant that we were not able to adjust to the API changes in time for the sync, and our builds now failing in the buildfarm.

christophfroehlich:
Hm, maybe we haven't thought about the API break of renaming `read_state_from_hardware` while backporting (I think this is the only break here). We should work on bringing the scaled JTC upstream into this package (#301, I guess this is the conflict here?).

bmagyar:
Sorry for the inconvenience caused guys. 

I'd suggest also following this up with the ROS Boss for Iron. Releases too close to the sync are typically held. The entire purpose of the `testing` repo is to give some mileage to new releases before they go into sync.

PR #887, opened 2023-12-07
[JTC] Fill action error_strings

christophfroehlich:
I found out that the `control_msgs::action::FollowJointTrajectory` has an error_string field, but it was not populated in some cases.

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -294,7 +294,7 @@ controller_interface::return_type JointTrajectoryController::update(
         {
           auto result = std::make_shared<FollowJTrajAction::Result>();
           result->set__error_code(FollowJTrajAction::Result::PATH_TOLERANCE_VIOLATED);
-          result->set__error_string("Aborted due to state tolerance violation");
+          result->set__error_string("Aborted due to path tolerance violation");
           active_goal->setAborted(result);
           // TODO(matthew-reynolds): Need a lock-free write here
           // See https://github.com/ros-controls/ros2_controllers/issues/168
@@ -313,7 +313,7 @@ controller_interface::return_type JointTrajectoryController::update(
           {
             auto result = std::make_shared<FollowJTrajAction::Result>();
             result->set__error_code(FollowJTrajAction::Result::SUCCESSFUL);
-            result->set__error_string("Goal reached, success!");
+            result->set__error_string("Goal successfully reached!");
             active_goal->setSucceeded(result);
             // TODO(matthew-reynolds): Need a lock-free write here
             // See https://github.com/ros-controls/ros2_controllers/issues/168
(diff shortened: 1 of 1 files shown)


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -366,17 +368,19 @@ controller_interface::return_type JointTrajectoryController::update(
           }
           else if (!within_goal_time)
           {
+            std::string error_string = "Aborted due goal_time_tolerance exceeding by " +
```
saikishor, line 371:
```suggestion
            const std::string error_string = "Aborted due goal_time_tolerance exceeding by " +
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -351,9 +352,10 @@ controller_interface::return_type JointTrajectoryController::update(
         {
           if (!outside_goal_tolerance)
           {
-            auto res = std::make_shared<FollowJTrajAction::Result>();
-            res->set__error_code(FollowJTrajAction::Result::SUCCESSFUL);
-            active_goal->setSucceeded(res);
+            auto result = std::make_shared<FollowJTrajAction::Result>();
+            result->set__error_code(FollowJTrajAction::Result::SUCCESSFUL);
+            result->set__error_string("Goal reached, success!");
```
saikishor, line 357:
```suggestion
            result->set__error_string("Goal successfully reached!");
```
Do we need to set the error string even in the case of success?

christophfroehlich: I thought the same, but there is the **error** code `FollowJTrajAction::Result::SUCCESSFUL` so I thought we should populate the string as well ;)

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -340,6 +340,7 @@ controller_interface::return_type JointTrajectoryController::update(
         {
           auto result = std::make_shared<FollowJTrajAction::Result>();
           result->set__error_code(FollowJTrajAction::Result::PATH_TOLERANCE_VIOLATED);
+          result->set__error_string("Aborted due to state tolerance violation");
```
bmagyar, line 343:
```suggestion
          result->set__error_string("Aborted due to path tolerance violation");
```
?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -371,17 +373,19 @@ controller_interface::return_type JointTrajectoryController::update(
           }
           else if (!within_goal_time)
           {
+            const std::string error_string = "Aborted due goal_time_tolerance exceeding by " +
```
bmagyar, line 376:
```suggestion
            const std::string error_string = "Aborted due to goal_time_tolerance exceeding by " +
```

bmagyar (approved):
Just minor comments

PR #902, opened 2023-12-10
[JTC] Invalidate empty trajectory messages

christophfroehlich:
I introduced a bug with #567: If an empty trajectory (no points) is received with `allow_nonzero_velocity_at_trajectory_end=true` then `trajectory.points.back()` is undefined behavior and may result in a segfault.

I added now a check to skip empty trajectories. Or is there a point for accepting them? `Trajectory::sample()` will return false with an empty trajectory.
https://github.com/ros-controls/ros2_controllers/blob/999eae5491a79f9734f8d5e4e18ffea3e7a04e2c/joint_trajectory_controller/src/trajectory.cpp#L68-L73


joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1233,8 +1238,41 @@ TEST_P(TrajectoryControllerTestParameterized, invalid_message)
...
+  good_traj_msg.joint_names = joint_names_;
+  good_traj_msg.header.stamp = rclcpp::Time(0);
+  good_traj_msg.points.resize(1);
+  good_traj_msg.points[0].time_from_start = rclcpp::Duration::from_seconds(0.25);
+  good_traj_msg.points[0].positions.resize(1);
+  good_traj_msg.points[0].positions = {1.0, 2.0, 3.0};
+  good_traj_msg.points[0].velocities.resize(1);
+  good_traj_msg.points[0].velocities = {-1.0, -2.0, -3.0};
+  // Nonzero velocity at trajectory end!
+  EXPECT_FALSE(traj_controller_->validate_trajectory_msg(good_traj_msg));
```
saikishor, line 1261: @christophfroehlich shouldn't this be True?, If I'm right, here you are parsing a valid trajectory message

christophfroehlich: No, because this test is with `allow_nonzero_velocity_at_trajectory_end=false` -> the validate method should return false for this message.

Maybe I should rename the variable `good_traj_msg`, this was just a copy & paste.

saikishor: Ok, then I would totally recommend rename the variable and also a case where this is validated and returned true. I believe this second case is already tested along with others

christophfroehlich: your are right, I renamed it now. 

There is another test right above where the same message with `allow_nonzero_velocity_at_trajectory_end=true` is checked for acceptance.

saikishor (approved):
LGTM. Thanks for the change

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1233,8 +1238,39 @@ TEST_P(TrajectoryControllerTestParameterized, invalid_message)
   EXPECT_FALSE(traj_controller_->validate_trajectory_msg(traj_msg));
 }
 
-/// With allow_integration_in_goal_trajectories parameter trajectory missing position or
-/// velocities are accepted
+/**
+ * @brief invalid_message_nonzero_vel Test invalid velocity at trajectory end
+ */
+TEST_P(TrajectoryControllerTestParameterized, invalid_message_nonzero_vel)
```
destogl, line 1244: Just a bit more clarity.
```suggestion
TEST_P(
  TrajectoryControllerTestParameterized,
  expect_invalid_when_message_with_nonzero_end_velocity_and_when_param_false)
```

destogl (approved):
Looking good to me, just a small proposal to rename test for clarity.

Maybe I tend to have tooooooooo long test names lately, but this brings clarity what is tested.

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich, line 1244:
```suggestion
 * @brief Test invalid velocity at trajectory end with parameter set to false
 */
TEST_P(TrajectoryControllerTestParameterized, expect_invalid_when_message_with_nonzero_end_velocity_and_when_param_false)
```

destogl (approved):

PR #932, opened 2023-12-15
[JTC] Add console output for tolerance checks

christophfroehlich:
This PR enables debug output from the tolerance checks, e.g.:
```
[gzserver-1] [ERROR] [1702714807.880348311] [tolerances]: State tolerances failed for joint 2:
[gzserver-1] [ERROR] [1702714807.880401620] [tolerances]: Position Error: 0.020046, Position Tolerance: 0.010000
[gzserver-1] [WARN] [1702714807.880539038] [trajectory_controllers]: Aborted due goal_time_tolerance exceeding by 1.010000 seconds
```

### DIFF
--- a/joint_trajectory_controller/doc/parameters.rst
+++ b/joint_trajectory_controller/doc/parameters.rst
@@ -81,6 +81,11 @@ cmd_timeout (double)
 constraints (structure)
   Default values for tolerances if no explicit values are states in JointTrajectory message.
 
+constraints.debug_output (bool)
+  Debug output for constraint checking. Activation may break real-time execution.
+
+  Default: false
+
 constraints.stopped_velocity_tolerance (double)
   Default value for end velocity deviation.
 
--- a/joint_trajectory_controller/doc/parameters.rst
+++ b/joint_trajectory_controller/doc/parameters.rst
@@ -81,11 +81,6 @@ cmd_timeout (double)
 constraints (structure)
   Default values for tolerances if no explicit values are states in JointTrajectory message.
 
-constraints.debug_output (bool)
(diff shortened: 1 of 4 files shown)


saikishor (commented):
@christophfroehlich why not print it all the time?. What you are printing are errors not the debug part of logging. Moreover, you can set it to always true and use `THROTTLE` instead of the basic one. I think this way you shouldn't have many problems.

christophfroehlich:
> @christophfroehlich why not print it all the time?. What you are printing are errors not the debug part of logging. Moreover, you can set it to always true and use `THROTTLE` instead of the basic one. I think this way you shouldn't have many problems.

Throttling would be an option, but I don't know what is a typical goal_time tolerance: At which interval should the throttling run?

saikishor:
> Throttling would be an option, but I don't know what is a typical goal_time tolerance: At which interval should the throttling run?

I would set it to half a second/ a second or so. I think that should be ok in general right?

christophfroehlich:
but with throtteling and the current method, one probably wouldn't see the output of the last check which is actually cancelling the goal.

christophfroehlich:
Throttling isn't possible, because it is inside a loop: I don't know a way to "reset" the macro-throttle for a different joint-index.
@saikishor what about this? It should be now printed only once in both checks: 
- with `state_tolerance` the variable `tolerance_violated_while_moving` will abort the goal
- If `default_tolerances_.goal_time_tolerance` is true and the check fails `within_goal_time`, it will be called again with `show_errors=true` and the goal will be aborted afterwards.

saikishor (approved):
This looks good to me!

bmagyar (approved):

PR #943, opened 2023-12-22
[JTC] Cleanup includes

christophfroehlich:
I cleaned-up the includes. It compiles and linters do not complain. 

Fixes some items of #944 

- I tried to not to [rely on transitive inclusion](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use) of header files 
- I removed [forward declarations](https://google.github.io/styleguide/cppguide.html#Forward_Declarations), but maybe we could let the fwd declaration of the project-intern Trajectory class?


bmagyar (approved):
Thank you, this is a good idea!

saikishor (approved):

bmagyar (approved):
I bet some of these come from older cppcheck errors, let's see how the backports will do! :D

Issue #944, opened 2023-12-22
[JTC] Cleanup TODO-List from #320


christophfroehlich:
To keep track of TODOs mentioned with #320:

- [ ] try minimizing the amount of includes #943
- [ ] check if this is needed at all https://github.com/ros-controls/ros2_controllers/blob/4c6d7236245c527f275b1e03a7e5242b658dc782/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp#L53-L61
- [x] clean the mess of trajectory msg pointers
- [x] rename `read_state_from_hardware` 
- [ ] try refactoring these into trajectory_operations.cpp or into trajectory.hpp directly https://github.com/ros-controls/ros2_controllers/blob/4c6d7236245c527f275b1e03a7e5242b658dc782/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp#L224-L241
- [ ] turn `compute_error_for_joint` into a proper function & unit test
- [ ] if "RealtimeSubscriber" existed with an API to check for new message & get message that'd make this very readable https://github.com/ros-controls/ros2_controllers/blob/4c6d7236245c527f275b1e03a7e5242b658dc782/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L164-L165
- [ ] Convert `assign_interface_from_point` into member function

christophfroehlich:
@bmagyar regarding 

> try refactoring these into trajectory_operations.cpp or into trajectory.hpp directly 

The methods

- `fill_partial_goal`
- `sort_to_local_joint_order`
- `validate_trajectory_msg`
- `add_new_trajectory_msg`

all use class variables, like `dof_`, `params_.joints` and so on. You suppose to refactor that without relying on class variables but add them as function parameters?

PR #945, opened 2023-12-22
[JTC] Convert lambda to class functions

christophfroehlich:
Fixes some items of #944:

> 
> - turn `compute_error_for_joint` into a proper function & unit test
> - Convert `assign_interface_from_point` into member function

and minor changes
- make reference_wrapper argument of `assign_interface_from_point` const
- fix format warning of `check_state_tolerance_per_joint`
- declare `sort_to_local_joint_order` as const member function

This is breaking ABI, I guess we should not backport this? Then I'd make a manual backport of the "fix format warning"


joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -461,6 +461,195 @@ TEST_P(TrajectoryControllerTestParameterized, hold_on_startup)
...
+  }
+  EXPECT_NEAR(error.positions[0], 0., EPS);
+  EXPECT_NEAR(error.positions[1], 0., EPS);
+  EXPECT_NEAR(error.positions[2], 0., EPS);
+  if (
+    traj_controller_->has_velocity_state_interface() &&
+    (traj_controller_->has_velocity_command_interface() ||
+     traj_controller_->has_effort_command_interface()))
+  {
+    // error.velocities[index] = desired.velocities[index] - current.velocities[index];
```
bmagyar, line 509: Intended to work? If yes, let's uncomment or add a ticket. If no, let's do away with the commented code

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -461,6 +461,195 @@ TEST_P(TrajectoryControllerTestParameterized, hold_on_startup)
...
+  }
+  EXPECT_NEAR(error.positions[0], desired.positions[0] - current.positions[0] - 2.0 * M_PI, EPS);
+  EXPECT_NEAR(error.positions[1], desired.positions[1] - current.positions[1], EPS);
+  EXPECT_NEAR(error.positions[2], desired.positions[2] - current.positions[2], EPS);
+  if (
+    traj_controller_->has_velocity_state_interface() &&
+    (traj_controller_->has_velocity_command_interface() ||
+     traj_controller_->has_effort_command_interface()))
+  {
+    // error.velocities[index] = desired.velocities[index] - current.velocities[index];
```
saikishor, line 541: Let's loop with the indices instead, so in case If we increase the size, we don't have to manually add new lines for the test. Same for the others

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: that's only a comment how the error is calculated by JTC. But I see this is confusing, I change that

christophfroehlich: fixed

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: at least for these trivial cases where every joint has the same expected structure, fine.

saikishor: ok

saikishor (approved):

joint_trajectory_controller/test/test_trajectory_controller.cpp
bmagyar: thx!

bmagyar (approved):

destogl (approved):
Thanks!

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: @saikishor done

saikishor: Awesome. Thanks

christophfroehlich:
@bmagyar @destogl what do you think of backporting this?

bmagyar:
Probably helps avoiding conflicts with other backports.

PR #946, opened 2023-12-22
[CI] Do not run reviewer lottery for dependabot and mergify

christophfroehlich:
Do not run reviewer lottery for PRs created by dependabot or mergify.


bmagyar (approved):
Thank you!

PR #949, opened 2023-12-27
[JTC] Parse URDF for continuous joints

### LINKED ISSUE #815: [JTC] Angle wraparound from URDF
Once https://github.com/ros-controls/ros2_control/pull/1088 gets merged, we could parse if the joint is of type _"continuous"_ and automatically set the `angle_wraparound` variable instead of manually setting it from the ROS parameters: This is the way it was done in [ROS 1 version of JTC](https://github.com/ros-controls/ros_controllers/blob/67b0e721e245fb744e33891508104a133bd2f59a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L269). 

Or do I miss something? Is there a use case where a joint with type "revolute" should have angle_wraparound, or type _"continuous"_ which should not have angle_wraparound?

christophfroehlich:
Use `urdf` package to parse the URDF string for joints. If they are of continuous type, activate `angle_wraparound` automatically.

Fixes #815


bmagyar (approved):
Very clean! Great work

Issue #931, opened 2023-12-14
Improve Steering Controllers Documentation


christophfroehlich:
> In general, I'd appreciate more documentation of the kinematics in the docs for all controllers here. Is there any publicly available reference/book for these mobile robot kinematics we could refer to? 
I'd happy to help with writing the docs, and maybe writing proper tests for the implementations. But I'm not "fluent" with these type of kinematics, I'd need a textbook for that.

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_controllers/issues/878#issuecomment-1837235341_

christophfroehlich:
For example, some simple sketches showing the system inputs of the different steering geometries. Pictures with appropriate licenses could be used and adapted, e.g., from [wikipedia](https://en.wikipedia.org/wiki/Ackermann_steering_geometry#/media/File:Ackermann_turning.svg). This might also be related with #692

bmagyar:
I'd suggest relying on Wikipedia as much as possible. The core equations are usually well explained but figures may be lacking.

christophfroehlich:
I haven't found anything appropriate to Ackermann steering for example :man_shrugging:

ARK3r:
This is a resource that I found useful: https://msl.cs.uiuc.edu/planning/node657.html

christophfroehlich:
Some literature review

- From [above](https://msl.cs.uiuc.edu/planning/node657.html) or as [PDF version](https://lavalle.pl/planning/booka4.pdf) (**Tri-Cylce, Simple Car**)
- @franzrammerstorfer mentioned [Lynch, Kevin M., and Frank C. Park. Modern Robotics: Mechanics, Planning, and Control 1st ed. Cambridge, MA: Cambridge University Press, 2017](https://hades.mech.northwestern.edu/index.php/Modern_Robotics), which has a [PDF preprint](https://hades.mech.northwestern.edu/images/2/25/MR-v2.pdf) available. (**Car-Like Robot**)
- [Siciliano et.al  Robotics: Modelling, Planning and Control](https://link.springer.com/book/10.1007/978-1-84628-642-1) (**Bicycle**)
- [Ribeiro, Lima - KINEMATICS MODELS OF MOBILE ROBOTS](http://users.isr.ist.utl.pt/~mir/cadeiras/robmovel/Kinematics.pdf) (**Tricycle**)

[...]

Issue #850, opened 2023-11-19
tricycle_controller vs tricycle_steering_controller


christophfroehlich:
We have two controllers with similar names: tricycle_controller vs tricycle_steering_controller.

I haven't used any of them, so I ask: (especially @tonynajjar who has contributed the first one, and @petkovich who created the second one)

- Is there a justification of keeping both?

If the answer is yes: 
- Let's document it.
- The "older" tricycle_controller doesn't follow our best-practices, e.g., using generate_parameter_library, and it lacks of tests (see [coverage](https://app.codecov.io/gh/ros-controls/ros2_controllers)): Let's improve that.

If the answer is no:

- Is the "newer" tricycle_steering_controller feature-equal to the "older" tricycle_controller?

If the answer is no: Let's add them.
If the answer is yes:
- Is the "newer" tricycle_steering_controller compatible to the "older" tricycle_controller? (I don't think so)
- Should we delete it after some deprecation period with some migration guidelines published in the docs?

christophfroehlich:
What I understood so far:

- **Tricycle_controller** has a single double-actuated wheel: steering and traction on the same wheel. This is basically a front-drive bicycle model, because it does not make a difference how many wheels are on the rear axle.
- This is different from **tricycle_steering_controller**, which claims having two traction wheels on the rear axle and one steering joint in the front (or vice versa). Same for **bicycle_steering_controller**: on traction wheel on the rear axle and one steering joint in the front (or vice versa)
- **Tricycle_controller** has additional features:
   - Traction limiter
   - Steering limiter
   - ackermann_msg debug topic

I tend now to include tricycle controller into the steering lib on a long term basis, see also #692
Then we don't have to decide if it should be called tricycle (one possible mechanical configuration) or bicycle (the actual kinematic model) controller ;)

Issue #692, opened 2023-07-04
Refactor to distinguish between wheel and steering joints (steering controller)


roncapat:
Pre-requirement of #648

If you have a look for example at:
https://github.com/ros-controls/ros2_controllers/blob/master/ackermann_steering_controller/test/ackermann_steering_controller_params.yaml
you can see that `front_wheels_names` and `rear_wheels_names` in fact might refer (depending on situation) to either a traction actuator or a steering actuator. 

Here:
https://github.com/ros-controls/ros2_controllers/blob/76b1ce7da5e7a102ef3a1c8c4f1b0de90ed0d137/steering_controllers_library/src/steering_controllers_library.cpp#L280-L316
a distinction is then made depending on `front_steering` param (and linked to #648, I should introduce here somehow `all_wheel_drive` parameter).

It feels to me that otherwise introducing additional implicit conventions in combination of new flags / number of passed joints and their interpretation as wheels or steers will complicate even more the `steering_controller` package...

Semantically, we have no enforcement at configuration level of what is "wheel" and what is "steer". I propose to add `front_steer_names` and `rear_steer_names`, and refactor the code in support of both #648 and #526 (which I am also involved in).

ARK3r:
Hope it's not too late to open this conversation back up. I think these changes both make sense as @roncapat suggests:
1. add `front_steer_names` and `rear_steer_names` as parameters.
2. add support for 4-wheel-drive and 4-wheel-steer.

My suggestion is to merge all models down to one controller which would result in `steering_controller`'s configuration file to determine which kinematic needs to be used:
| Vehicle Kinematic  | `front_steer_names`  |  `front_wheel_names` |   `rear_steer_names` | `rear_wheel_names`  |
|---|---|---|---|---|
| Bicycle front rear-only drive  | `[front_steer_joint]`  |  `[]` | `[]`  | `[rear_drive_joint]`  |
| Bicycle with front and rear drive and rear-steer | `[]`  | `[front_drive_joint]`  | `[rear_steer_joint]`  | `[rear_drive_joint]`  |
| Ackermann with front steer and rear-only drive  | `[front_right_steer_joint, front_left_steer_joint]`  | `[]`  | `[]`  | `[rear_right_drive_joint, rear_left_drive_joint]`  |
[...]

PR #954, opened 2023-12-30
Add mobile robot kinematics 101 and improve steering library docs

christophfroehlich:
This is the first step of #931, but I focused only on steering on the front axle for now.
Edit: I also added the diff-drive kinematics to have a complete documentation.

After peer-reviewing of the equations I'd start to rename the variables of the library to match the doc and fix possible bugs. 

# rendered docs (updated [5d6aee7](https://github.com/ros-controls/ros2_controllers/pull/954/commits/5d6aee7be2cc8767667563803cf5b7f9fc8868b3))
image


christophfroehlich:
:eyes: @aleksandrsizmailovs @roncapat @franzrammerstorfer @petkovich

christophfroehlich:
Ad odometry of ackermann steering: There is a not negligible error imho. (average is the equation implemented in the library, plotted for l = 1, w = 1; the equation from above gives the correct angle phi)
image

roncapat:
Impressive job! I think this is very valuable. At least for now everything seems ok to me, I would maybe suggest to be more explicit in the derivation of some formulas, maybe with collapsable sections.

For example, for the unicycle:
immagine this implies at least one or two intermediate steps and to make the description more appealing to readers I may tend to add detail (so the attention of the reader is shifted from understanding the formulas to trusting them) even if sometimes it seems trivial :)

The point is... there are not so many places on the internet where you can find a clean explanation. This can become one of these 👍🏻

About the "overdetermined" odometry solution of ackermann with traction, I would start to think about a set of recommendations or methods to find the best solution. It gets even worse with 4-wheel-4-steer vehicles and becomes an hell when considering mars-like rovers (6 wheeled).

christophfroehlich:
> Impressive job! I think this is very valuable. At least for now everything seems ok to me, I would maybe suggest to be more explicit in the derivation of some formulas, maybe with collapsable sections.
> 
> For example, for the unicycle: immagine this implies at least one or two intermediate steps and to make the description more appealing to readers I may tend to add detail (so the attention of the reader is shifted from understanding the formulas to trusting them) even if sometimes it seems trivial :)
> 
> The point is... there are not so many places on the internet where you can find a clean explanation. This can become one of these 👍🏻

Thanks! I was not sure where to start and where to end. I think about it.

> About the "overdetermined" odometry solution of ackermann with traction, I would start to think about a set of recommendations or methods to find the best solution. It gets even worse with 4-wheel-4-steer vehicles and becomes an hell when considering mars-like rovers (6 wheeled).

We should at least assume, that ever actuated joint has a feedback sensor (e.g., no combination of traction joints on the rear axle but speed sensors on the front axle) -> this would get very complicated to parameterize for every combination.

But as far as I understood: the claim of this lib at the current state was bicycle, tricycle, ackermann; all with front or back steering; and with the traction axle being the other than the steering axle? (the steering+traction ackermann description is not part of this then, but part of this proposal of a "general" library https://github.com/ros-controls/ros2_controllers/issues/692#issuecomment-1749748573)
This is the first real difference to the tricycle_controller for #850, which expects the steering wheel be the only traction wheel, and the two behind are just following.

christophfroehlich:
Am I right that for a trailing steering axis (`front_steering=false`) one has just to use a negative value for the wheel-base `l`, and that's it?

bmagyar:
@petkovich this may be interesting to you

MCFurry:
Very nice @christophfroehlich this looks like a very good documentation page! Nice work!
Looks alright to me.
IMHO I would finish and merge this PR. 
The documentation page will then be available to the public and more sets of eyes can look at it and possibly spot errors or recommend improvements?

christophfroehlich:
> Very nice @christophfroehlich this looks like a very good documentation page! Nice work! Looks alright to me. IMHO I would finish and merge this PR. The documentation page will then be available to the public and more sets of eyes can look at it and possibly spot errors or recommend improvements?

Thanks! Would you mind adding a review in the UI? (Below "Files Changed" tab -> Review Changes). Yes, this will be published on control.ros.org.

bmagyar (approved):

PR #957, opened 2023-12-31
[tricycle_controller] Use generate_parameter_library

christophfroehlich:
As per title.

- use generate_parameter_library, remove not anymore needed checks and variables, rewrite tests
- let the load_controller test use the yaml from the test file

I also updated the documentation a bit, see #850 

This breaks ABI because of the class variable change, is this a bad idea to backport?


tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -357,7 +291,7 @@ CallbackReturn TricycleController::on_configure(const rclcpp_lifecycle::State &
   }
 
   // initialize command subscriber
-  if (use_stamped_vel_)
+  if (params_.use_stamped_vel)
```
bmagyar, line 294: probably due to be removed soon too

tricycle_controller/src/tricycle_controller_parameter.yaml
```diff
@@ -0,0 +1,151 @@
+tricycle_controller:
+  traction_joint_name: {
+    type: string,
+    default_value: "",
+    read_only: true,
```
bmagyar, line 5: do we want read_only-s here?

christophfroehlich: after reporting #966, no..

tricycle_controller/src/tricycle_controller.cpp
christophfroehlich: in this PR or in a follow up?

bmagyar (approved):

PR #958, opened 2023-12-31
[diff_drive] Remove unused parameter and add simple validation

christophfroehlich:
- Remove unused parameter `wheels_per_side`
- Update description of names parameters (looking for joint names, not link names).
- add simple validation to names parameters, this needs some rework of the tests
- let the load_diff_drive tests use the yaml from the test file (which had two bugs)

This breaks ABI because of the class variable change, is this a bad idea to backport?


bmagyar (approved):
Wonderful, thank you!

bmagyar:
I think we should backport to Iron. Let's leave Humble as-is

PR #961, opened 2024-01-03
Add few warning flags to error

saikishor:
Applying similar flags as done in the ros2_control : https://github.com/ros-controls/ros2_control/pull/1181


saikishor:
@christophfroehlich Thank you. The linters are fixed now.

bmagyar (approved):

PR #962, opened 2024-01-04
[JTC] Cancel active goal in on_deactivate

christophfroehlich:
Currently, an active goal is not cancelled if on_deactivate is called. This PR changes this behavior now.

I also removed an unnecessary set_hold_position() call in the preempt_active_goal.


saikishor (approved):
Good catch!
LGTM once the CI passes

bmagyar (approved):
thanks!

PR #963, opened 2024-01-04
Remove robot description param from admittance YAML

Abishalini:
The following error was thrown at admittance controller initialization - 
```
Exception thrown during init stage with message: expected [string] got [not set]
```

Looks like it was coming from the robot description parameter and it is not used anywhere. Therefore, I am deleting it as part of this PR

### DIFF
--- a/admittance_controller/src/admittance_controller_parameters.yaml
+++ b/admittance_controller/src/admittance_controller_parameters.yaml
@@ -147,11 +147,6 @@ admittance_controller:
     }
 
   # general settings
-  robot_description: {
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
Good catch, thanks!

saikishor (approved):
Good catch!

olivier-stasse (approved):
Thanks. Indeed a good catch

bmagyar (approved):

destogl (commented):
Thanks

PR #985, opened 2024-01-23
[CI] debian + RHEL updates

christophfroehlich:
- Fix RHEL jobs
- Add debian jobs


christophfroehlich:
Binary builds still fail because of API breaks, but the rest is green now :factory_worker:

PR #1013, opened 2024-02-02
[CI] Use correct ref for scheduled workflows

christophfroehlich:
- workflow schedules are (only) run from master branch
- without giving an explicit ref argument the humble+iron workflows have checked out the master branch, which [obviously fails](https://github.com/ros-controls/ros2_controllers/actions/runs/7750292355/job/21136291633)
- fixed now with this if-clause in the [ref field](https://github.com/actions/checkout/tree/v4/?tab=readme-ov-file#usage)

See also https://github.com/ros-controls/ros2_control/pull/1361


bmagyar (approved):
Good catch!

PR #1032, opened 2024-02-09
[JTC] Renaming variables, reordering trajectory checks

destogl:
This PR increseas readability of the JTC by renaming trajectory varialbes to be semantically correct and self explainable.

Also it reoders checks of the trajectories done in the callback. They are grouped based on the things they are checking, i.e., 1. data available, 2. time checks; 3. trajectory point checks.

~~Additionally, it adds "last_commanded_time_" for the open loop mode to be more correct when setting point before trajectory.~~ #780


christophfroehlich (approved):
LGTM and improves readability!

christophfroehlich:
:eyes: @destogl  `test_no_jump_when_state_tracking_error_not_updated` fails now, I guess due to the `last_commanded_time_` change.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -148,29 +148,33 @@ controller_interface::return_type JointTrajectoryController::update(
...
       if (params_.open_loop_control)
       {
-        traj_external_point_ptr_->set_point_before_trajectory_msg(
-          time, last_commanded_state_, joints_angle_wraparound_);
+        if (fabs(last_commanded_time_.seconds()) < std::numeric_limits<float>::epsilon())
+        {
+          last_commanded_time_ = time;
+        }
+        current_trajectory_->set_point_before_trajectory_msg(
+          last_commanded_time_, last_commanded_state_, joints_angle_wraparound_);
```
saikishor, line 161: In this case, the `last_command_time_` would be same for the first 2 iterations. Is that okay?. If its fine, then all good.

saikishor (approved):
LGTM, apart from the following minor comment.

Now with the changes, it looks clear and good

bmagyar (dismissed):

christophfroehlich (changes requested):
This includes #780, which has the same failing test: `test_no_jump_when_state_tracking_error_not_updated`

christophfroehlich (dismissed):

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1425,9 +1411,29 @@ bool JointTrajectoryController::validate_trajectory_msg(
     }
   }
 
+  // CHECK: if trajectory end time is in the past (if start time defined)
+  const auto trajectory_start_time = static_cast<rclcpp::Time>(trajectory.header.stamp);
```
saikishor, line 1415:
```suggestion
  const rclcpp::Time trajectory_start_time = trajectory.header.stamp;
```

I don't know if this will work, but I'm guessing yes from here

https://github.com/ros2/rclcpp/blob/19773973a80d753c6fa028b0b548462fbbef122d/rclcpp/src/rclcpp/time.cpp#L109-L114

saikishor (approved):
in general, LGTM

saikishor:
@christophfroehlich it seems the downstream builds are failing from the UR drivers scaled JTC. Shall we inform Felix?

christophfroehlich:
> @christophfroehlich it seems the downstream builds are failing from the UR drivers scaled JTC. Shall we inform Felix?

The changes are simple, I can open a PR for the ur_controllers

christophfroehlich (approved):

PR #1049, opened 2024-02-17
[CI] Use reviewer lottery from ros2_control_ci

christophfroehlich:
Fixes reviewer lottery from #1047, see https://github.com/ros-controls/ros2_control/pull/1402 for more details.


bmagyar:
No access to the lottery strangely...

christophfroehlich:
as I described already, `pull_request_target` event uses the workflow from the target branch. This means, it does not use the file from this PR but the broken one from master branch

PR #1056, opened 2024-02-19
[diff_drive] Remove unused parameter and add simple validation #abi-breaking (backport #958)

christophfroehlich:
This is the 1:1 backport of #958. @bmagyar why do you think we have to do something different as you have changed with #1008?


bmagyar:
If the CI is happy we could merge this. The CI was clearly unhappy on the humble backport

christophfroehlich:
> If the CI is happy we could merge this. The CI was clearly unhappy on the humble backport

we didn't backport this to humble anyways. You tried fixing the original backport to iron.

bmagyar:
Too many parallel universes to handle :D

bmagyar (approved):

PR #1058, opened 2024-02-25
Use Eigen CMake target

christophfroehlich:
I changed the CMake file to use Eigen's CMake target as described here https://eigen.tuxfamily.org/dox/TopicCMakeGuide.html

See https://github.com/ros-controls/control_toolbox/issues/178 for the issues with RHEL 8.


saikishor (approved):
LGTM

fmauch (approved):
+1 for using modern CMake :-)

PR #1071, opened 2024-03-03
 added conditioning to have rolling tags compilable in older versions

saikishor:
This PR complements the ros2_control PR: https://github.com/ros-controls/ros2_control/pull/1422.

For this, the version conditionings is added to avoid compilation error:
* https://github.com/ros2/rcpputils/commit/1bd32c72b355761b11dd9e35abe4d2cc32b7187a (The deprecation of `rcppmath` namespace and they moved it to the `rcpputils` headers)
* https://github.com/ros2/common_interfaces/commit/05d7e1942acffefbeb40f7dd9d69ce4298119dcf (The newly added variance field of the `sensor_msgs`)

We would need the upstream version of `control_msgs` as well to compile the rolling branch in old versions

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/odometry.hpp
+++ b/diff_drive_controller/include/diff_drive_controller/odometry.hpp
@@ -25,7 +25,12 @@
 #include <cmath>
 
 #include "rclcpp/time.hpp"
+// \note The versions conditioning is added here to support the source-compatibility with Humble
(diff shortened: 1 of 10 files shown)


christophfroehlich (approved):
Thanks, LGTM and compiled on humble.

bmagyar (approved):
Great cleanup!!

PR #1084, opened 2024-03-18
Fix pid_controller build on ROS 2 Rolling on Ubuntu 24.04

clalancette:
Two things are needed to make this compile:
1.  Make sure to add a package.xml dependency on backward_ros, which is necessary to compile.
2.  Make sure to #include <rclcpp/version.h>, so the correct rclcpp::QoS object is chosen.

With both of these in place, this package compiles cleanly on Ubuntu 24.04.

This should fix the CI build like in https://build.ros2.org/view/Rbin_uN64/job/Rbin_uN64__pid_controller__ubuntu_noble_amd64__binary/13/console

Contributions via pull requests are much appreciated. Before sending us a pull request, please ensure that:

1. Limited scope. Your PR should do one thing or one set of things. Avoid adding “random fixes” to PRs. Put those on separate PRs.
2. Give your PR a descriptive title. Add a short summary, if required.
3. Make sure the pipeline is green.
4. Don’t be afraid to request reviews from maintainers.
5. New code = new tests. If you are adding new functionality, always make sure to add some tests exercising the code and serving as live documentation of your original intention.

- [x] Fork the repository.

### DIFF
--- a/pid_controller/src/pid_controller.cpp
+++ b/pid_controller/src/pid_controller.cpp
@@ -26,8 +26,8 @@
 #include "control_msgs/msg/single_dof_state.hpp"
 #include "controller_interface/helpers.hpp"
 
-#include <rclcpp/rclcpp.hpp>
-#include <rclcpp/version.h>
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/version.h"
 
 namespace
 {  // utility
--- a/pid_controller/src/pid_controller.cpp
+++ b/pid_controller/src/pid_controller.cpp
@@ -26,6 +26,9 @@
 #include "control_msgs/msg/single_dof_state.hpp"
 #include "controller_interface/helpers.hpp"
 
+#include <rclcpp/rclcpp.hpp>
+#include <rclcpp/version.h>
+
 namespace
 {  // utility
 
--- a/ackermann_steering_controller/package.xml
+++ b/ackermann_steering_controller/package.xml
@@ -16,6 +16,7 @@
 
   <build_depend>generate_parameter_library</build_depend>
 
+  <depend>backward_ros</depend>
   <depend>control_msgs</depend>
   <depend>controller_interface</depend>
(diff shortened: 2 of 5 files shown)


pid_controller/src/pid_controller.cpp
```diff
@@ -26,6 +26,9 @@
 #include "control_msgs/msg/single_dof_state.hpp"
 #include "controller_interface/helpers.hpp"
 
+#include <rclcpp/rclcpp.hpp>
+#include <rclcpp/version.h>
```
saikishor, line 30:
```suggestion
#include "rclcpp/rclcpp.hpp"
#include "rclcpp/version.h"
```
Thanks for the heads-up on the version header :)

saikishor (changes requested):
Thanks for the changes.
Regarding the `backward_ros` dependency, I've just observed the missing dependency in the following packages as well: `ackermann_steering_controller`, `bicycle_steering_controller`, `tricycle_steering_controller`. Can you also add them in the same PR?.

Thank you

clalancette:
> Regarding the `backward_ros` dependency, I've just observed the missing dependency in the following packages as well: `ackermann_steering_controller`, `bicycle_steering_controller`, `tricycle_steering_controller`. Can you also add them in the same PR?.

Yes, very good call.  Fixed now in 175c4d5.

saikishor (approved):
LGTM. Thank you for the changes :)

christophfroehlich (approved):
Thanks!

clalancette:
Rolling coverage CI seems to be failing here because the action is not installing `ros-rolling-rsl` properly in the container.  Then it later fails to build `parameter_traits`, which depends on that.

It doesn't look to be connected with this PR, so I don't think it should hold this up.

bmagyar (approved):
Brilliant, thank you!

clalancette:
A release into Rolling would be much appreciated, as it would close out this particular problem.  Thanks!

PR #1090, opened 2024-03-27
JTC trajectory end time validation fix

henrygerardmoore:
The `time_from_start` member for a `JointTrajectoryPoint` is monotonically increasing, not a delta time, so I think we can just use the last one in the vector; summing them up seems incorrect. I also moved the points empty check earlier because `.back().time_from_start` would seg fault with an empty points vector.


christophfroehlich:
Could you please check the docs if this is clearly documented how to fill the `trajectory` field? [Here](https://control.ros.org/master/doc/ros2_controllers/joint_trajectory_controller/doc/trajectory.html#trajectory-replacement) is a comment, but maybe we should add this in a more prominent place.

henrygerardmoore:
> Could you please check the docs if this is clearly documented how to fill the `trajectory` field? [Here](https://control.ros.org/master/doc/ros2_controllers/joint_trajectory_controller/doc/trajectory.html#trajectory-replacement) is a comment, but maybe we should add this in a more prominent place.

@christophfroehlich 
That page seems clear to me, the plots in there are very helpful. Perhaps it should be in the main JTC page, but where it is seems fine IMO.

christophfroehlich (changes requested):
LGTM. Just one request: Could you please add checks for the end time to `invalid_message` test?

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1384,6 +1384,11 @@ TEST_P(TrajectoryControllerTestParameterized, invalid_message)
   traj_msg = good_traj_msg;
   traj_msg.points.push_back(traj_msg.points.front());
   EXPECT_FALSE(traj_controller_->validate_trajectory_msg(traj_msg));
+
+  // End time in the past
+  traj_msg = good_traj_msg;
+  traj_msg.header.stamp = rclcpp::Time(1);
+  EXPECT_FALSE(traj_controller_->validate_trajectory_msg(traj_msg));
```
christophfroehlich, line 1391: Is this enough to test? 
Maybe we should add two tests:
- One with start_time + time_from_start < now() -> fail
- One with start_time + time_from_start > now() -> success

henrygerardmoore: Good point, I have added one such successful case as well.

christophfroehlich (approved):
LGTM, thanks!

bijoua29 (approved):

henrygerardmoore:
@christophfroehlich is there anything I can do to help get this across the finish line?

christophfroehlich:
No, we have to wait for an approval of @destogl or @bmagyar

bmagyar (approved):

PR #1092, opened 2024-04-04
Add migration notes for Jazzy

christophfroehlich:
Related to https://github.com/ros-controls/ros2_control/pull/1477

> In the last WG meeting, we decided to split the detailed release notes with the mandatory migration steps. The latter, should be mandatory for future PRs IMHO.
> 
> Feel free to comment on missing stuff, and also if you have suggestions for a better format of the list.

Needs https://github.com/ros-controls/control.ros.org/pull/265 to be merged afterwards.

## Release notes

Up to now its only a summary of the changelog, but a little bit more verbose and with hints for migration. We could also just include the CHANGELOG.rst, but this one is cluttered with CI and test-related stuff which is of no use for the average user.

image

## Migration guides
image


saikishor (approved):
Thanks

Issue #989, opened 2024-01-29
Drop non-stamped Twist support from tricycle controller


bmagyar:
probably due to be removed soon too

_Originally posted by @bmagyar in https://github.com/ros-controls/ros2_controllers/pull/957#discussion_r1449294931_

PR #1093, opened 2024-04-04
Deprecate non-stamped twist for tricycle_controller and steering_controllers

christophfroehlich:
First step of #989.

### DIFF
--- a/diff_drive_controller/doc/userdoc.rst
+++ b/diff_drive_controller/doc/userdoc.rst
@@ -46,7 +46,7 @@ Subscribers
 ,,,,,,,,,,,,
 
 ~/cmd_vel [geometry_msgs/msg/TwistStamped]
-  Velocity command for the controller, if ``use_stamped_vel=true``. The controller extracts the x component of the linear velocity and the z component of the angular velocity. Velocities on other components are ignored.
+  Velocity command for the controller. The controller extracts the x component of the linear velocity and the z component of the angular velocity. Velocities on other components are ignored.
 
 
 Publishers
--- a/steering_controllers_library/doc/userdoc.rst
+++ b/steering_controllers_library/doc/userdoc.rst
@@ -13,7 +13,7 @@ Nomenclature used for the controller is used from `wikipedia <https://en.wikiped
 Execution logic of the controller
 ----------------------------------
 
-The controller uses velocity input, i.e., stamped or unstamped Twist messages where linear ``x`` and angular ``z`` components are used.
+The controller uses velocity input, i.e., stamped Twist messages where linear ``x`` and angular ``z`` components are used.
 Angular component under
 Values in other components are ignored.
 In the chain mode the controller provides two reference interfaces, one for linear velocity and one for steering angle position.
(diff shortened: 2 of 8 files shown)


aprotyas (approved):

olivier-stasse (approved):
LGTM

Just a small remark, it will not be possible to send directly a command to a robot using DiffDrive controller from a console.

That was nice to switch easily from turtlesim to a real robot from a pedagogical point of view. But the auto completion in ROS-2 with pub topic is not working super nicely from my experience. If anybody has a nice way to have a simple example for `ros2 topic pub /cmd_vel`, it will be nice to document it.

christophfroehlich:
> Just a small remark, it will not be possible to send directly a command to a robot using DiffDrive controller from a console.
> 
> That was nice to switch easily from turtlesim to a real robot from a pedagogical point of view. But the auto completion in ROS-2 with pub topic is not working super nicely from my experience. If anybody has a nice way to have a simple example for `ros2 topic pub /cmd_vel`, it will be nice to document it.

You mean because it is not possible to auto-fill the time-stamp in the header? Or that the syntax of the nested msg is tedious?

olivier-stasse:
> You mean because it is not possible to auto-fill the time-stamp in the header? Or that the syntax of the nested msg is tedious? https://control.ros.org/master/doc/ros2_control_demos/example_2/doc/userdoc.html

Actually both. (BTW thanks again for the documentation effort this is great !) With the auto completion I never  managed to have the nice example you gave in the docs. I do not remember if there is a default acceptable value for the time stamp like 0 for all the fields.

saikishor (approved):
LGTM. Do we want to completely remove it in Jazzy or leave it with a deprecation warning? 

I agree with @olivier-stasse. I believe it should be fine to publish with 0 timestamp as it is usually considered as the  command sent with the current timestamp

christophfroehlich:
> LGTM. Do we want to completely remove it in Jazzy or leave it with a deprecation warning?
> 
Yes, I think we will completely remove it. I'm not sure if there will be a rolling release before jazzy, but I thought at least we can backport the deprecation warning to iron. 

> I agree with @olivier-stasse. I believe it should be fine to publish with 0 timestamp as it is usually considered as the command sent with the current timestamp

Is this a feature of ROS 2 CLI or does the controller just have to "ignore" zero timestamp? I'd have to test that

christophfroehlich:
> Actually both. (BTW thanks again for the documentation effort this is great !) 
:smile: 

> With the auto completion I never managed to have the nice example you gave in the docs. I do not remember if there is a default acceptable value for the time stamp like 0 for all the fields.

I also struggle always there. Sometimes it's faster to ask my AI friend for the syntax than trying around..

saikishor:
> Is this a feature of ROS 2 CLI or does the controller just have to "ignore" zero timestamp? I'd have to test that

I think the controller is the one who has to consider the zero timestamps as the recent one. Upon checking the code, this is already handled. In fact in most of the ROS/ROS2 systems it is considered like that

https://github.com/ros-controls/ros2_controllers/blob/689efe716fd97f4c79638689ececc65fe175b418/diff_drive_controller/src/diff_drive_controller.cpp#L344-L351

bmagyar:
I think the clunkiness comes from the CLI auto complete offering a one-lined yaml as opposed to the JSON format (dict with curly brackets)

bmagyar (approved):

PR #1094, opened 2024-04-05
Use the urdf_ to set the robot_description in admittance controller

SyllogismRXS:
I'm running ROS2 Iron, building ros2_control and ros2_controllers from their master branches, but the admittance controller was complaining about the lack of the `robot_description` parameter in `kinematics_interface_kdl` (https://github.com/ros-controls/kinematics_interface/blob/master/kinematics_interface_kdl/src/kinematics_interface_kdl.cpp#L34). 

The `controller_manager` now requires that the `robot_description` to not be a parameter, but come from a topic. I'm doing that. However, it seems that the admittance controller is still expecting `robot_description` as a parameter. In this PR, I set the `robot_description` parameter for the admittance controller using the `urdf_` member variable. This feels a little hacky. Am I doing something wrong? Is anyone using the latest admittance controller successfully?

Thanks!

### DIFF
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -20,6 +20,7 @@
 #include "admittance_controller/admittance_rule.hpp"
 
 #include <memory>
+#include <string>
 #include <vector>
 
 #include "rclcpp/duration.hpp"
@@ -33,7 +34,8 @@ constexpr auto NUM_CARTESIAN_DOF = 6;  // (3 translation + 3 rotation)
 
 /// Configure admittance rule memory for num joints and load kinematics interface
 controller_interface::return_type AdmittanceRule::configure(
-  const std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node, const size_t num_joints)
+  const std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node, const size_t num_joints,
+  const std::string & robot_description)
 {
   num_joints_ = num_joints;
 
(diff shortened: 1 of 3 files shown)


christophfroehlich (commented):
No I think you are totally right. But I suggest we should change the API of kinematics_interface to take the robot_description as [argument in the initalize()](https://github.com/ros-controls/kinematics_interface/blob/master/kinematics_interface/include/kinematics_interface/kinematics_interface.hpp#L43-L45) method. This could be done with a default empty-string with fallback to the node parameter to avoid an API break. What do you think @pac48 @destogl?

saikishor (commented):
I think this should be an API change on the kinematics_interface_kdl side. I agree with @christophfroehlich. I would simply add a new method with the URDF argument and parse it directly

pac48:
@christophfroehlich I agree that passing the robot description in the initializer makes sense.

destogl (approved):
This looks good!

admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
```diff
@@ -51,7 +53,7 @@ controller_interface::return_type AdmittanceRule::configure(
       kinematics_ = std::unique_ptr<kinematics_interface::KinematicsInterface>(
         kinematics_loader_->createUnmanagedInstance(parameters_.kinematics.plugin_name));
       if (!kinematics_->initialize(
-            node->get_node_parameters_interface(), parameters_.kinematics.tip))
+            node->get_node_parameters_interface(), parameters_.kinematics.tip, robot_description))
```
destogl, line 56:
```suggestion
            node->get_node_parameters_interface(), "kinematics", robot_description))
```

Propose to set here `param_base_name` so that we search for the prameters under `controller interface.kinematics`. As we are doing in [JointLimtiers](https://github.com/ros-controls/ros2_control/blob/master/joint_limits/include/joint_limits/joint_limits_rosparam.hpp)

destogl:
Closed in favor of #1247

PR #1060, opened 2024-02-26
Improved tf_prefix based on namespace

kjvanteeffelen:
Solves #1048.

When using robot namespace as `tf_prefix`, it now removes any starting '/' and only adds an ending '/' if not present already.
Starting with '/' is not common in TF2 and showed compatibility issues with for instance `robot_localization`.

PR #1102, opened 2024-04-20
[Humble Backport] tf_prefix param: fix slashes and add to IMU Broadcaster

rafal-gorecki:
Hello, 
I encountered one error related to the tf_prefix parameter. Namely, the current code works in such a way that the use of namespace does not remove the first slash. It's not typical for a tf frame to start with / so I fixed it. Additionally, I don't think it's intuitive to add a slesh every time we use tf_prefix, if only because someone might want to give a prefix that doesn't end with a slesh, although I could be wrong about that.

Additionally, I encountered one error with the option
```cpp
    validation: {
      not_empty<>: null
    }
```
it doesn't work well with namespaces and I don't know how to fix it yet. However, I see that it is not used in all packages, so I commented it temporarily.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -403,11 +403,14 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
     else
     {
       tf_prefix = std::string(get_node()->get_namespace());
-      tf_prefix.erase(0, 1);
-      if (!tf_prefix.empty())
+      if (tf_prefix != "/")
       {
-        tf_prefix = tf_prefix + '/';
+        tf_prefix += '/';
       }
+    }
+    if (tf_prefix.front() == '/')
+    {
+      tf_prefix.erase(0, 1);
     }
   }
 
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -396,22 +396,18 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
   std::string tf_prefix = "";
   if (params_.tf_frame_prefix_enable)
   {
-    if (params_.tf_frame_prefix != "")
+    if (!params_.tf_frame_prefix.empty())
     {
       tf_prefix = params_.tf_frame_prefix;
     }
     else
     {
       tf_prefix = std::string(get_node()->get_namespace());
-    }
-
-    if (tf_prefix == "/")
-    {
-      tf_prefix = "";
-    }
-    else
-    {
-      tf_prefix = tf_prefix + "/";
+      tf_prefix.erase(0, 1);
+      if (!tf_prefix.empty())
+      {
+        tf_prefix = tf_prefix + '/';
+      }
     }
   }
 
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -394,24 +394,21 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
 
   // Append the tf prefix if there is one
   std::string tf_prefix = "";
-  if (params_.tf_frame_prefix_enable)
+  if (!params_.tf_frame_prefix.empty())
   {
-    if (!params_.tf_frame_prefix.empty())
+    tf_prefix = params_.tf_frame_prefix;
+  }
+  else
+  {
+    tf_prefix = std::string(get_node()->get_namespace());
+    if (tf_prefix != "/")
     {
-      tf_prefix = params_.tf_frame_prefix;
+      tf_prefix += '/';
     }
-    else
-    {
-      tf_prefix = std::string(get_node()->get_namespace());
-      if (tf_prefix != "/")
-      {
-        tf_prefix += '/';
-      }
-    }
-    if (tf_prefix.front() == '/')
-    {
-      tf_prefix.erase(0, 1);
-    }
+  }
+  if (tf_prefix.front() == '/')
+  {
+    tf_prefix.erase(0, 1);
   }
 
   const auto odom_frame_id = tf_prefix + params_.odom_frame_id;
--- a/diff_drive_controller/test/test_diff_drive_controller.cpp
+++ b/diff_drive_controller/test/test_diff_drive_controller.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "test_diff_drive_controller.hpp"
+
 #include <gmock/gmock.h>
 
 #include <array>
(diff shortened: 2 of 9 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -396,22 +396,18 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
...
-    }
-
-    if (tf_prefix == "/")
-    {
-      tf_prefix = "";
-    }
-    else
-    {
-      tf_prefix = tf_prefix + "/";
+      tf_prefix.erase(0, 1);
```
saikishor, line 406: Why do you always want to erase the first character? If the prefix is parsed without '/' then it removes the wrong character right?

I see that you did that when you are using node namespace, however, I think you need to correct the users tf_prefix as well, if he adds a '/' as well. Try to do same logic as done in #1060

diff_drive_controller/test/test_diff_drive_controller.cpp
```diff
@@ -425,13 +425,13 @@ TEST_F(TestDiffDriveController, configure_succeeds_tf_test_prefix_true_set_names
...
-  ASSERT_EQ(test_odom_frame_id, frame_prefix + "/" + odom_id);
-  ASSERT_EQ(test_base_frame_id, frame_prefix + "/" + base_link_id);
+  ASSERT_EQ(test_odom_frame_id, frame_prefix + odom_id);
+  ASSERT_EQ(test_base_frame_id, frame_prefix + base_link_id);
 }
 
 TEST_F(TestDiffDriveController, configure_succeeds_tf_blank_prefix_true_set_namespace)
 {
-  std::string test_namespace = "/test_namespace";
+  std::string test_namespace = "test_namespace";
```
saikishor, line 434: Add test with and without slash. Maybe you can do it in a for loop

imu_sensor_broadcaster/src/imu_sensor_broadcaster_parameters.yaml
```diff
@@ -5,17 +5,19 @@ imu_sensor_broadcaster:
     description: "Defines sensor name used as prefix for its interfaces.
     Interface names are: ``<sensor_name>/orientation.x, ..., <sensor_name>/angular_velocity.x, ...,
     <sensor_name>/linear_acceleration.x.``",
-    validation: {
-      not_empty<>: null
-    }
+    # FIXME: Currently does not work with namespace
+    # validation: {
+    #   not_empty<>: null
+    # }
```
saikishor, line 11: Can you explain a bit more in what situation it doesn't work.

imu_sensor_broadcaster/src/imu_sensor_broadcaster_parameters.yaml
```diff
@@ -41,3 +43,13 @@ imu_sensor_broadcaster:
...
   }
+  tf_frame_prefix_enable: {
+    type: bool,
+    default_value: true,
+    description:  "Enables or disables appending tf_prefix to tf frame id's.",
+  }
+  tf_frame_prefix: {
+    type: string,
+    default_value: "",
+    description:  "(optional) Prefix to be appended to the tf frames, will be added to odom_id and base_frame_id before publishing. If the parameter is empty, controller's namespace will be used.",
```
saikishor, line 54:
```suggestion
    description:  "(optional) Prefix to be appended to the tf frames, will be added to the sensor's frame_id before publishing. If the parameter is empty, controller's namespace will be used.",
```

imu_sensor_broadcaster/src/imu_sensor_broadcaster_parameters.yaml
```diff
@@ -41,3 +43,13 @@ imu_sensor_broadcaster:
       fixed_size<>: [9],
     }
   }
+  tf_frame_prefix_enable: {
+    type: bool,
+    default_value: true,
+    description:  "Enables or disables appending tf_prefix to tf frame id's.",
```
saikishor, line 49: I don't understand the need for this new parameter. Can you explain me why it is needed ?

imu_sensor_broadcaster/test/test_imu_sensor_broadcaster.cpp
```diff
@@ -208,6 +209,154 @@ TEST_F(IMUSensorBroadcasterTest, SensorName_Publish_Success)
...
+  imu_broadcaster_->get_node()->set_parameter(
+    rclcpp::Parameter("tf_frame_prefix", rclcpp::ParameterValue(frame_prefix)));
+
+  ASSERT_EQ(imu_broadcaster_->on_configure(rclcpp_lifecycle::State()), NODE_SUCCESS);
+  ASSERT_EQ(imu_broadcaster_->on_activate(rclcpp_lifecycle::State()), NODE_SUCCESS);
+
+  sensor_msgs::msg::Imu imu_msg;
+  subscribe_and_get_message(imu_msg, test_namespace);
+
+  EXPECT_EQ(imu_msg.header.frame_id, frame_prefix + frame_id_);
```
saikishor, line 328: Shouldn't there be a slash in between?

saikishor (changes requested):
It seems to be a duplication of #1060, but I like the fact that you have added some tests to the cases. However, I'm not entirely convinced in some places. I've left some comments here to discuss.

imu_sensor_broadcaster/src/imu_sensor_broadcaster_parameters.yaml
rafal-gorecki: Yes sure:

I don't know exactly what's going on. However, when this is checked it then gets an error:

```bash
    [ERROR] [1713772814.609746661] [test_namespace.test_imu_sensor_broadcaster]: Exception thrown during init stage with message: Invalid value set during initialization for parameter 'sensor_name': Parameter 'sensor_name' cannot be empty
```

1. The problem is probably that adding a namespace causes the frame_id parameter to be searched for, which is no longer there (because it is ~/frame_id)
2. Something else with ControllerInterface

saikishor: I think you have misinterpreted it. This information is need to get the proper state interfaces within the controller. Please look into the following code

https://github.com/ros-controls/ros2_controllers/blob/9e97c00496c239bf4961dc043a3f461aae9b23b8/imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp#L48-L49

imu_sensor_broadcaster/src/imu_sensor_broadcaster_parameters.yaml
rafal-gorecki: I followed the example of diff_drive. I think you can get rid of it.

imu_sensor_broadcaster/test/test_imu_sensor_broadcaster.cpp
saikishor: ok if you want it to be prefixed without a '/' at the end, why don't you parse it directly already appending it to the frame_id? This frame_id is only used for the published message. I don't see a point of adding a tf_prefix, if it is not namespaced.

https://github.com/ros-controls/ros2_controllers/blob/9e97c00496c239bf4961dc043a3f461aae9b23b8/imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp#L66

What do you think about this?

rafal-gorecki:
Dear @saikishor ,
I would like to ask for verification, I hope I added all the suggestions and managed to fix the error with the parameters that occurred with the namepatch.

rafal-gorecki:
Najłatwiej było mi to zrobić w ten sposób, kontynuacji #1104

PR #1104, opened 2024-04-22
tf_prefix param: add to IMU Broadcaster and fix slashes in DiffDrive

rafal-gorecki:
Related with #1102, rabase to master.


rafal-gorecki:
@saikishor FYI

rafal-gorecki:
Hi @christophfroehlich @saikishor,
After the initial exchange of knowledge, the topic died, has something happened, are we waiting for something?

imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
```diff
@@ -62,8 +62,26 @@ controller_interface::CallbackReturn IMUSensorBroadcaster::on_configure(
...
+    if (tf_prefix != "/")
+    {
+      tf_prefix += '/';
+    }
+  }
+  if (tf_prefix.front() == '/')
+  {
+    tf_prefix.erase(0, 1);
+  }
+
```
saikishor, line 82: Hello!

While reviewing this part again, I started to question myself again, if we need this or not. I'm saying because in case of the diff_drive_controller, the `odom` message that is being published is a dynamic tf frame, but in case of the imu_state_broadcaster, as per the imu you always have a static transform link inside the URDF, I don't know if it make sense to have the `tf_prefix` in this case. 

Right now, your changes will add namespace of the controller manager to the links, I'm not sure how right is it.

Does it make sense?. Should we check if the link exists from the loaded URDF?

saikishor (changes requested):

imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
rafal-gorecki: Hi @saikishor,

Thank you, for reply! 

I'm not sure what difference it makes whether fame is dynamic or static in the case of tf_prefix. I treat it as an option that may be useful to someone. What I care about most is the ability to correctly assign a frame to a robot when many robots are launched.

I'm still wondering about the issue of `sensor_name`, because while in the case of a real robot there should be no problems, in the simulation you need to create sensor objects with different namespaces.

saikishor: Hello!

I think it is rather very important, because the sensor readings should correspond to a specific frame on the robot and this might be important. Regarding multi robot i agree that we would need this separation, then we can use the namespace, but why tf_prefix?. Moreover, i believe we have to make sure that it is a valid link.

Can you explain to me the instances that's a requirement for you?

rafal-gorecki: If you are asking specifically about tf_prefix, I don't need it, but I added it to simplify the packages.

As for checking whether a frame exists, it makes sense.

rafal-gorecki: What do you think about the problem with `sensor_name`? The simplest solution is to add a parameter that adds a namespace to `sensor_name` or add a sensor name prefix.

Generally, what I care about is being able to run many robots in simulation and on hardware while making a minimum number of changes to the code.

saikishor: Hello!

For the hardware it seems to be the best option is with the namespace, and I think it makes sense. 

Regarding the issue with sensor_name, do you mean that they need to be namespaced as well, am I right? Because the interfaces do not exist?

rafal-gorecki: When running multiple robots in a simulation, it is necessary to add a sensor with a unique name.  This name is later referred to by the sensor_name parameter.  In the case of real robots there is no such collision because everything is closed inside the topic ~/robot_description.  However, in the case of simulation there is a problem because you need to distinguish the sensor names, so I added: 
1. to urdf namespace in the sensor name.
2. I also need to add a namespace for imu, but I don't know the best way to do it.

saikishor: @rafal-gorecki ok now I got the use-case. Yes, it is tricky to make it work with simulation with this namespace. What can be done is probably, instead of using the `tf_prefix`, just add it to the `sensor_name` already. I'm not sure how to automatically configure it based on the robot. The current changes however, doesn't fix it right?

Any ideas on how to solve it are welcome

rafal-gorecki: Thank you, so much for today. I will try to add this functionality soon.  I would like to ask you to confirm whether this tf_prefix can be left for imu in order to standardize the packages, or should I get rid of it?

saikishor: Let's discuss your idea on how you are planning to tackle the idea of sensor name automatically, and then it's better to start changing

rafal-gorecki: The best solution is probably to change gz-sensor to allow the creation of sensors with the same name, provided that they are connected to different reference frames.

Workaroud may be to add the argument `add_namespace_to_sensor_name` (bool)

rafal-gorecki: Secound solution added

christophfroehlich:
@rafal-gorecki is still necessary, if yes, could you please resolve the conflicts introduced from #1420?

PR #1114, opened 2024-04-29
[CI] Add jazzy and other minor updates

christophfroehlich:
- Add Jazzy workflows (I decided to create dedicated files, because it easier to maintain if we branch for jazzy)
- Add filters to run the ROS - workflows only if code has changed, and the doc workflow only if docs have changed. (branches+path filter have to match both to run the workflow)
- Check Docs - workflow now uses the correct branch of control.ros.org (not every page is available on every distro any more).
- Run pre-commit auto-update only once per month


.github/workflows/humble-binary-build.yml
```diff
@@ -7,9 +7,19 @@ on:
   pull_request:
     branches:
       - humble
+    paths:
+      - '**.hpp'
+      - '**.h'
+      - '**.cpp'
+      - '.github/workflows/humble-binary-build.yml'
```
fmauch, line 14: It probably also makes sense to run on changes to 

- package.xml
- CMakeLists.txt
- repo files

That comment would be relevant for all build jobs.

.github/workflows/iron-rhel-semi-binary-build.yml
```diff
@@ -4,6 +4,11 @@ on:
   pull_request:
     branches:
       - iron
+    paths:
+      - '**.hpp'
+      - '**.h'
+      - '**.cpp'
+      - '.github/workflows/jazzy-debian-build.yml'
```
fmauch, line 11:
```suggestion
      - '.github/workflows/iron-rhel-semi-binary-build.yml'
```

fmauch (changes requested):

fmauch:
Also, I was wondering, why the check_docs workflows didn't have any path restrictions, now I am wondering even more while reviewing https://github.com/ros-controls/ros2_control/pull/1521

.github/workflows/humble-binary-build.yml
christophfroehlich: you are right, I'll check this

fmauch (approved):
LGTM. Thanks for all the work @christophfroehlich 

The failing CI seems unrelated.

PR #1120, opened 2024-05-07
Add parameter check for geometric values

christophfroehlich:
Inspired by [this post at RSE](https://robotics.stackexchange.com/a/111066/33182), I thought it could be a good idea to add simple checks to geometric values of the controllers for mobile robots. For example, `wheel_radius` `wheel_base` and others must not be zero :arrow_right: I added a `gt<>: [0.0]` validator for them. This also should help if a user misspelled or forgot these parameters.


bmagyar (approved):
This is great, thanks!

PR #1139, opened 2024-05-22
[Admittance] applies control frame transform to mass matrix

MarcoMagriDev:
In admittance control mass matrix is not transformed to the control frame. This will reflect in two main issues:
- Inconsistency between stiffness/damping_ratio (set from the user in control_frame) and mass (set from the user in base frame) may create confusion
- the damping value is computed from the `damping_ratio` under assumption that mass and stiffness are in the same reference frame but this at the end is not the case (mass in base and stiffness in control). This (I think) may cause an under/over-damped behavior if the masses of the two involved axis (`x` and `z` in the example) has very different values.

### How to reproduce
For simplicity, suppose to be in a configuration with the `control_frame` having the `z` axis coincident with the `x` axis of robot base like the following:
Screenshot from 2024-05-22 14-03-49
The admittance controller is configured with the following admittance parameters:
```yaml
admittance:
  mass:
    - 1.0 # x
    - 1.0 # y
    - 1.0 # z
    - 0.05 # rx
    - 0.05 # ry
    - 0.05 # rz
  
  stiffness:
    - 50.0 # x
    - 50.0 # y
    - 50.0 # z
    - 1.0 # rx
    - 1.0 # ry
    - 1.0 # rz
  
  damping_ratio: 
    - # Not relevant
```
Now, faking force readings we are able to make the robot move due to compliance:
1) constant value of force readings as [0.0, 0.0, 5.0, 0.0, 0.0, 0.0] -> the robot move along the z axis
2) constant value of force readings as [0.0, 0.0, 0.0, 0.0, 0.0, 0.0] -> the robot returns to its original position

Changing the value of the parameter `admittance.mass` to [1.0, 1.0, 100.0, 0.05, 0.05, 0.05] and repeating the previous experiment should end up in a `shorter` motion performed by the TCP but this is not the case. Instead, by setting `admittance.mass` to [100.0, 1.0, 1.0, 0.05, 0.05, 0.05] the motion amplitude is reduced.

The effect of the missing transformation can be seen clearly in this case due to the selected configuration and since we are exasperating asymmetry in desired masses.

### Proposed solution
Apply the same operations done for `stiffness` and `damping` also to the `mass_inv`

---
_Note: what about defining a new method or lambda function to avoid to repeat the same code for each matrix? Let me know so that I can open a new PR for that_

### DIFF
--- a/.github/workflows/rolling-semi-binary-build-win.yml
+++ b/.github/workflows/rolling-semi-binary-build-win.yml
@@ -27,6 +27,6 @@ jobs:
     uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@master
     with:
       ros_distro: rolling
-      pixi_dependencies: typeguard jinja2 boost compilers
+      pixi_dependencies: typeguard jinja2 boost compilers octomap
       ninja_packages: rsl
       target_cmake_args: -DBUILD_TESTING=OFF
--- a/doc/migration.rst
+++ b/doc/migration.rst
@@ -4,6 +4,11 @@ Migration Guides: Kilted Kaiju to Lyrical Luth
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 This list summarizes important changes between Kilted Kaiju (previous) and Lyrical Luth (current) releases, where changes to user code might be necessary.
 
+joint_state_broadcaster
+*****************************
+* Removed interfaces with other data types than double for publishing to ``dynamic_joint_states``. (`#2115 <https://github.com/ros-controls/ros2_controllers/pull/2115>`_).
+  Use a custom controller for publishing non-double interfaces.
+
 effort_controllers
 *****************************
 * ``effort_controllers/JointGroupEffortController`` is deprecated. Use :ref:`forward_command_controller <forward_command_controller_userdoc>` instead by adding the ``interface_name`` parameter and set it to ``effort``. (`#1913 <https://github.com/ros-controls/ros2_controllers/pull/1913>`_).
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -27,6 +27,7 @@ joint_state_broadcaster
 ************************
 * Make all parameters read-only (the never got re-evaluated after initialization anyways). (`#2064 <https://github.com/ros-controls/ros2_controllers/pull/2064>`_)
 * Added parameter ``publish_dynamic_joint_states`` to enable/disable publishing of dynamic joint states. (`#2064 <https://github.com/ros-controls/ros2_controllers/pull/2064>`_)
+* Removed interfaces with other data types than double for publishing to ``dynamic_joint_states``. (`#2115 <https://github.com/ros-controls/ros2_controllers/pull/2115>`_)
 
 omni_wheel_drive_controller
 *****************************
--- a/joint_state_broadcaster/src/joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/src/joint_state_broadcaster.cpp
@@ -236,6 +236,14 @@ bool JointStateBroadcaster::init_joint_data()
     HW_IF_POSITION, HW_IF_VELOCITY, HW_IF_EFFORT};
   for (auto si = state_interfaces_.crbegin(); si != state_interfaces_.crend(); si++)
   {
+    if (si->get_data_type() != hardware_interface::HandleDataType::DOUBLE)
+    {
+      RCLCPP_WARN(
+        get_node()->get_logger(),
+        "State interface '%s' of joint '%s' has non-double data type and will be ignored.",
+        si->get_interface_name().c_str(), si->get_prefix_name().c_str());
+      continue;
+    }
     const std::string prefix_name = si->get_prefix_name();
     // initialize map if name is new
     if (name_if_value_mapping_.count(prefix_name) == 0)
@@ -305,6 +313,10 @@ void JointStateBroadcaster::init_auxiliary_data()
   mapped_values_.clear();
   for (auto i = 0u; i < state_interfaces_.size(); ++i)
   {
+    if (state_interfaces_[i].get_data_type() != hardware_interface::HandleDataType::DOUBLE)
+    {
+      continue;
+    }
     std::string interface_name = state_interfaces_[i].get_interface_name();
     if (map_interface_to_joint_state_.count(interface_name) > 0)
     {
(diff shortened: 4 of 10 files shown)


christophfroehlich:
Maybe @pac48 can comment on this?

MarcoMagriDev:
Any update on this?

pac48:
@MarcoMagriDev Can the math be simplified? For example 

`R*M_inv*R^T*(F + R*D*R^T*Xd + R*K*R^T*X)`

We can distribute `R^T` and get the following:

`R*M_inv(R^T*F + D*R^T*Xd + K*R^T*X)`

MarcoMagriDev:
@pac48 yes i think we can.

Here is the complete reasoning behind:
- A generic admittance control law can be expressed as $`\ddot{X} = M^{-1}(F - D \dot{X} - K X)`$
- Since we would like to define matrices $`M`$, $`D`$, $`K`$ in end effector frame all the other quantities should be in expressed in the same frame obtaining: $`\ddot{X_{e}} = M^{-1}(F_{e} - D \dot{X_e} - K X_e)`$ (where the subscript `e` indicates that the vector is expressed in end effector frame) 
- By exploiting the relation $`X_{e} = R^T X_{b}`$ and $`F_{e} = R^T F_{b}`$ (where the subscript `b` indicates that the vector is expressed in base frame)  we get: $` \ddot{X_b} = R M^{-1}(R^T F_b - D R^T \dot{X_b} - K R^T X_b)`$ that is exactly your "math simplified equation"

Should I update the PR with this modification?

pac48:
@MarcoMagriDev Yes, I think it will reduce unnecessary computation and make the math more clear. Currently, each vector is rotated into the control frame, multiplied by the matrix, then rotated back to the base frame. It makes sense to do the math in the control frame and only rotate back to the base frame as the last step.

christophfroehlich:
@MarcoMagriDev Could you address the last request please?

MarcoMagriDev:
@christophfroehlich I just pushed a new commit that should address the request from @pac48. Not entirely sure if it aligns perfectly with what was expected — happy to discuss and adjust it further if needed!

admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
```diff
@@ -280,16 +272,13 @@ bool AdmittanceRule::calculate_admittance_rule(AdmittanceState & admittance_stat
...
   F_control = F_control.cwiseProduct(admittance_state.selected_axes);
-  F_base.block<3, 1>(0, 0) = rot_base_control * F_control.block<3, 1>(0, 0);
-  F_base.block<3, 1>(3, 0) = rot_base_control * F_control.block<3, 1>(3, 0);
 
   // Compute admittance control law in the base frame: F = M*x_ddot + D*x_dot + K*x
-  Eigen::Matrix<double, 6, 1> X_ddot =
-    admittance_state.mass_inv.cwiseProduct(F_base - D * X_dot - K * X);
+  Eigen::Matrix<double, 6, 1> X_ddot = rot_base_control_6d * M_inv *
+                                       (F_control - D * rot_base_control_6d.transpose() * X_dot -
+                                        K * rot_base_control_6d.transpose() * X);
```
pac48, line 281: This looks good, but it might be a good idea to update the comment "F = M*x_ddot + D*x_dot + K*x"

christophfroehlich: @MarcoMagriDev could you fix the comment as requested please?

admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
```diff
@@ -280,16 +272,13 @@ bool AdmittanceRule::calculate_admittance_rule(AdmittanceState & admittance_stat
...
   // zero out any forces in the control frame
-  Eigen::Matrix<double, 6, 1> F_control;
-  F_control.block<3, 1>(0, 0) = rot_base_control.transpose() * F_base.block<3, 1>(0, 0);
-  F_control.block<3, 1>(3, 0) = rot_base_control.transpose() * F_base.block<3, 1>(3, 0);
+  Eigen::Matrix<double, 6, 1> F_control = rot_base_control_6d.transpose() * F_base;
   F_control = F_control.cwiseProduct(admittance_state.selected_axes);
-  F_base.block<3, 1>(0, 0) = rot_base_control * F_control.block<3, 1>(0, 0);
-  F_base.block<3, 1>(3, 0) = rot_base_control * F_control.block<3, 1>(3, 0);
 
   // Compute admittance control law in the base frame: F = M*x_ddot + D*x_dot + K*x
```
christophfroehlich:
```suggestion
  // Compute admittance control law in the base frame: F_control = M*R^T*x_ddot + D*R^T*x_dot + K*R^T*x
  // with R being the rotation matrix from base to control frame
```

christophfroehlich (approved):

admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
christophfroehlich: I added my proposal to finish this.

christophfroehlich: @pac48 could you give this a quick review please?

Juliaj:
@MarcoMagriDev, please add tests covering this change or update existing tests. This helps with review and future maintenance.

christophfroehlich:
> @MarcoMagriDev, please add tests covering this change or update existing tests. This helps with review and future maintenance.

none of the involved persons seem to be active anymore. I suggest merging this, I opened a follow-up issue for the tests.

Juliaj:
@christophfroehlich, the current implementation is correct. I have a different fix which has more lines but may be more direct. Your call on which version to use. 
[alternative_fix.patch](https://github.com/user-attachments/files/24399139/alternative_fix.patch)

Tests added https://github.com/ros-controls/ros2_controllers/pull/2092

Juliaj (approved):
LGTM

saikishor (approved):

christophfroehlich:
> @christophfroehlich, the current implementation is correct. I have a different fix which has more lines but may be more direct. Your call on which version to use. [alternative_fix.patch](https://github.com/user-attachments/files/24399139/alternative_fix.patch)
> 
> Tests added #2092

I applied your patch and cherrypicked the tests into this branch. Thanks for the updates!

admittance_controller/test/test_admittance_rule_impl.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+  // Test fixture members
+  std::shared_ptr<rclcpp::Node> node_;
+  std::shared_ptr<admittance_controller::ParamListener> param_listener_;
+  std::unique_ptr<TestableAdmittanceRule> admittance_rule_;
+  MockKinematicsInterface * mock_kinematics_ptr_;
+};
+
+// Test 1: Verify that calculate_admittance_rule succeeds with rotated control frame
+// and produces motion when force is applied
```
Juliaj, line 279:
```suggestion
// and produces motion when force is applied
// cppcheck-suppress syntaxError
// To suppress the warning reported by cppcheck for complex macro expansion
```
```

Juliaj:
> it seems that you haven't installed pre-commit, the test greatly failed. I can't find the issue with the cppcheck error:
> 
> ```
> [admittance_controller/test/test_admittance_rule_impl.cpp:280]: (error: syntaxError) syntax error
> 1 errors
> ```
> 

My bad, missed the pre-commit error. I think this is a false positive from cppcheck 2.13.0 with complex macro expansions.  after adding a suppression comment, the pre-commit passed.

Juliaj:
@christophfroehlich, thanks for considering my fix. Looking into this a bit closer, my fix is more expensive computation wise since it involves more matrix multiplication than the original fix.  So recommend that we revert to the previous fix. The unit tests should work with both.

saikishor (approved):
LGTM

PR #1168, opened 2024-06-09
[Steering controllers library] Reference interfaces are body twist

### LINKED ISSUE #1167: Ackermann preceding reference command interfaces are not correct
**Describe the bug**
I am building a ackermann-controlled vehicle and would like to control the mobile base via steering angle and linear velocity instead of a twist since I have a preceding pure pursuit controller. In the documentation ([Ackermann Controller](https://control.ros.org/iron/doc/ros2_controllers/steering_controllers_library/doc/userdoc.html#references-from-a-preceding-controller) it says that the references are linear velocity and angular position of the steering. Setting these references via ForwardController reveals that the angular position command interface is in fact a angular velocity command interface. Since the steering library is worked on rn anyway: Would it be an idea to let people switch between controlling the vehicle via Twist or AckermannDrive messages? This would remove the slightly awkward setup of having to separate ForwardCommand controllers or a completely self-written controller just to set steering angle and velocity instead of a Twist.

**To Reproduce**
1. Start an Ackermann vehicle
2. Attach two forward command controllers to the Ackermann controller e.g.
 ```forward_steering_controller:
  ros__parameters:
    joints:
      - ackermann_steering_controller/angular

forward_velocity_controller:
  ros__parameters:
    joints:
      - ackermann_steering_controller/linear 
```
4. Set a steering command and a linear velocity command

**Expected behavior**
Setting a steering angle should be passed on to the angle commands and not be correlated with the velocity.

**Screenshots**
MicrosoftTeams-image(2)
The blue line is the commanded steering angle and the orange one is the true steering angle. Changing the linear velocity also influences the steering angle.

 - OS: Ubuntu 22
 - Version Iron

**Additional context**
Add any other context about the problem here, especially include any modifications to any ros2_controllers that relate to this issue.

### PR DESCRIPTION
**Pull request by christophfroehlich**
Fixes #1167 

I'm not sure if the intention of the author was to have position input, but as it is implemented we have a body twist input on reference interfaces (velocity, not position). This PR fixes this, without any attempt to deprecate the old behavior because it was just wrong.

Tests succeed on iron distro
```
$ colcon test --packages-select steering_controllers_library ackermann_steering_controller bicycle_steering_controller tricycl
e_steering_controller
Starting >>> steering_controllers_library
Finished <<< steering_controllers_library [0.48s]          
Starting >>> ackermann_steering_controller
Starting >>> bicycle_steering_controller
Starting >>> tricycle_steering_controller
Finished <<< tricycle_steering_controller [7.25s]                                                                                                  
Finished <<< ackermann_steering_controller [7.28s]
Finished <<< bicycle_steering_controller [7.33s]

Summary: 4 packages finished [8.33s]
```

### DIFF
--- a/ackermann_steering_controller/test/test_ackermann_steering_controller.cpp
+++ b/ackermann_steering_controller/test/test_ackermann_steering_controller.cpp
@@ -84,7 +84,7 @@ TEST_F(AckermannSteeringControllerTest, check_exported_interfaces)
     controller_->front_wheels_state_names_[1] + "/" + steering_interface_name_);
   EXPECT_EQ(state_if_conf.type, controller_interface::interface_configuration_type::INDIVIDUAL);
 
-  // check ref itfsTIME
+  // check ref itfs
   auto reference_interfaces = controller_->export_reference_interfaces();
   ASSERT_EQ(reference_interfaces.size(), joint_reference_interfaces_.size());
   for (size_t i = 0; i < joint_reference_interfaces_.size(); ++i)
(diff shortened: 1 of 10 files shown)


bmagyar (approved):
:thinking:

saikishor (approved):
LGTM
This basically affects only the controllers that chains with steering controller library
Thanks for the fixes @christophfroehlich

PR #1172, opened 2024-06-17
[JTC] Report on deprecated feedback

bmagyar:
Even though the `desired` and `actual` fields are deprecated, we should still fill them in.

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -1180,6 +1180,12 @@ void JointTrajectoryController::publish_state(
     state_publisher_->msg_.reference.velocities = desired_state.velocities;
     state_publisher_->msg_.reference.accelerations = desired_state.accelerations;
     state_publisher_->msg_.feedback.positions = current_state.positions;
+    // DESIRED and ACTUAL are deprecated in the message but we are still
+    // reporting on them
+    state_publisher_legacy_->msg_.desired.positions = desired_state.positions;
+    state_publisher_legacy_->msg_.desired.velocities = desired_state.velocities;
+    state_publisher_legacy_->msg_.desired.accelerations = desired_state.accelerations;
+    state_publisher_legacy_->msg_.actual.positions = current_state.positions;
     state_publisher_->msg_.error.positions = state_error.positions;
     if (has_velocity_state_interface_)
     {


saikishor (approved):
Makes perfect sense

aprotyas (approved):

bmagyar:
FYI @MarqRazz

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1180,6 +1180,12 @@ void JointTrajectoryController::publish_state(
     state_publisher_->msg_.reference.velocities = desired_state.velocities;
     state_publisher_->msg_.reference.accelerations = desired_state.accelerations;
     state_publisher_->msg_.feedback.positions = current_state.positions;
+    // DESIRED and ACTUAL are deprecated in the message but we are still
+    // reporting on them
+    state_publisher_legacy_->msg_.desired.positions = desired_state.positions;
+    state_publisher_legacy_->msg_.desired.velocities = desired_state.velocities;
+    state_publisher_legacy_->msg_.desired.accelerations = desired_state.accelerations;
+    state_publisher_legacy_->msg_.actual.positions = current_state.positions;
```
MarqRazz, line 1188: Why only `position` for `actual`?

MarqRazz (commented):
This will help with the issue I was running into but once the deprecated message is removed the new one is still missing functionality.

If I send this a trajectory you can see/plot `position, velocity, acceleration` in the `reference` value published in the `controller_state`. If I wanted to plot the `output` you can only see the values calculated for the claimed interface (i.e. `position` in my case). 

The old functionality allowed me to see/plot `position, velocity, acceleration` of the `input & output` so that I can visualize what the interpolation is doing to my trajectory.

christophfroehlich:
> The old functionality allowed me to see/plot `position, velocity, acceleration` of the `input & output` so that I can visualize what the interpolation is doing to my trajectory.

maybe I don't get what you try to achieve, but the reference field is the interpolated output?

christophfroehlich (changes requested):
I vote for deleting the fields instead. If there's something missing in the current design, let's discuss the roadmap proposal, implemented with https://github.com/ros-controls/control_msgs/pull/86

MarqRazz:
> maybe I don't get what you try to achieve, but the reference field is the interpolated output?

That is what I was looking for. Interesting you call the interpolated output `reference` I thought that was the input but now that I look again that is `desired`.

I'm all for removing them too!

Issue #1182, opened 2024-06-26
[Iron] - JTC produces jump at the end of the trajectory if chained with admittance controller


firesurfer:
**Describe the bug**

I have a setup as described in #1181 where I have a chained admittance controller with a JTC on top. 
For some reason the the JTC commands a jump in the joint positions at the end of the trajectory:

controller_state/output/positions:
![image](https://github.com/ros-controls/ros2_controllers/assets/8481806/da88caa8-3b0b-4aa2-8074-b1a67e11fcc8)

Configuration:

Note: This is the scaled_jtc from the ur package. 

```
joint_trajectory_controller_admittance_top:
  ros__parameters:
    joints:
      - ur_top/shoulder_pan_joint
      - ur_top/shoulder_lift_joint
      - ur_top/elbow_joint
      - ur_top/wrist_1_joint
      - ur_top/wrist_2_joint
      - ur_top/wrist_3_joint
    command_joints:
      - admittance_controller_top/ur_top/shoulder_pan_joint
      - admittance_controller_top/ur_top/shoulder_lift_joint
      - admittance_controller_top/ur_top/elbow_joint
      - admittance_controller_top/ur_top/wrist_1_joint
      - admittance_controller_top/ur_top/wrist_2_joint
      - admittance_controller_top/ur_top/wrist_3_joint
    command_interfaces:
      - position
    state_interfaces:
      - position
      - velocity
    state_publish_rate: 100.0
    action_monitor_rate: 20.0
    allow_partial_joints_goal: false
    constraints:
      stopped_velocity_tolerance: 0.0
      goal_time: 0.0
      admittance_controller_top/ur_top/shoulder_pan_joint: { trajectory: 0.0, goal: 0.000 }
      admittance_controller_top/ur_top/shoulder_lift_joint: { trajectory: 0.0, goal: 0.000 }
      admittance_controller_top/ur_top/elbow_joint: { trajectory: 0.0, goal: 0.000 }
      admittance_controller_top/ur_top/wrist_1_joint: { trajectory: 0.0, goal: 0.000 }
      admittance_controller_top/ur_top/wrist_2_joint: { trajectory: 0.0, goal: 0.000 }
      admittance_controller_top/ur_top/wrist_3_joint: { trajectory: 0.0, goal: 0.000 }
    use_speed_scaling_topic_instead: true                                           #Addition for the scaled jtc from the ur package for multiple arms
[...]

PR #1188, opened 2024-07-02
Ported scaled_jtc

firesurfer:
This is a remake of https://github.com/ros-controls/ros2_controllers/pull/301 given that neither @fmauch  nor @bmagyar 
are reacting why this PR has been closed.

## Differences to the original PR

1. This PR simply adds the functionality to the normal JTC
2. Instead of using the speed scaling interface a transient local speed scaling factor topic is used. This has various advantageous especially in a multiarm / multi JTC setups. (See https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/883#issuecomment-2126323231 for more information)
3. It contains a fix if a goal time is set: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/882 (Note: @fmauch  and I disagree what the correct behavior would be in that case - but he agreed to merge because otherwise the controller would not be usable) 
4. It does not store the `time_data_` field in a RealTimeBox as the access solely happens from the RT thread. 
5. The scaling factor is not clipped to [0,1] as it is really useful for sim environments to be able to upscale the execution speed

## Testing

I tested it by cherry-picking the commit back to the iron branch and running it on my setup with multiple joint motions. They were looking fine. We are using basically the same code (but added to the original scaled jtc) on our production setup for more than 3/4 of a year now.  (https://github.com/firesurfer/Universal_Robots_ROS2_Driver/tree/fixed_jtc)

## Configuration

The PR adds the following configuration entry:

>   speed_scaling_topic_name: {
      type: string,
      default_value: "~/speed_scaling_factor",
      description: "Topic name for the speed scaling factor - set to an empty string in order to disable it"
  }

For a multiarm scenario one would set this topic to the same topic name for all involved JTCs.

Up to discussion: Should it be _disabled_ per default?

## Notes

The `scaling_factor_` field does not need to be in a RealTime box (it does not matter if it is updated an update cycle earlier or later) but it might make sense to use an `std::atomic<double>` double as some platforms might not guarantee atomic access to the variable. 

It has the same issue as the original PR bei @fmauch. It does not scale velocities and efforts at the moment. But never the less I think it is a really useful addition and bringing this feature upstream will avoid issues such as #1182 where the bug has been fixed upstream but not backported yet. 

iron backport I used for testing: https://github.com/firesurfer/ros2_controllers/tree/scaled_jtc_iron

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -41,6 +41,7 @@
 #include "realtime_tools/realtime_buffer.h"
 #include "realtime_tools/realtime_publisher.h"
 #include "realtime_tools/realtime_server_goal_handle.h"
+#include "std_msgs/msg/float64.hpp"
 #include "trajectory_msgs/msg/joint_trajectory.hpp"
 #include "trajectory_msgs/msg/joint_trajectory_point.hpp"
 #include "urdf/model.h"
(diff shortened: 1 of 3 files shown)


fmauch:
Hey @firesurfer thanks for bringing this up again. I just wanted to tackle this and push it above the final barrier, you were faster than me :-)

What is missing from my point of view are

1. tests
2. documentation

I'll write some documentation and make a PR to your fork @firesurfer? Then we can get this streamlined into this PR.

bmagyar:
Thanks for the PR, we've discussed it in detail at the WG meeting today and you'll get a follow-up PR ;)

firesurfer:
@fmauch I close this PR in favor for #1191

Issue #1101, opened 2024-04-17
Test failures: subscription already associated with a wait set


christophfroehlich:
[INFO] [1713388339.950896541] [test_joint_group_velocity_controller]: activate successful
    unknown file: Failure
    C++ exception with description "subscription already associated with a wait set" thrown in the test body.

in rolling builds
https://github.com/ros-controls/ros2_controllers/actions/runs/8728644425/job/23948909083?pr=1100
and in debian source build
https://github.com/ros-controls/ros2_controllers/actions/runs/8728644426/job/23948908730

might be related to https://github.com/ros2/rclcpp/pull/2142

christophfroehlich:
:eyes: @bmagyar they confirmed in the linked PR that this is not possible any more. You said you have a different approach for the test structure?

christophfroehlich:
@dhood as you brought yourself into the topic, do you maybe have an idea how to fix that temporarily?

dhood:
yes, curiosity got the better of me, I'm a sucker for CI that needs troubleshooting :innocent:

Perhaps explicitly removing the subscription from the waitset after it's been waited on, in wait_for? might be hanging onto a reference by the time the spin_some gets called. I opened https://github.com/ros-controls/ros2_controllers/pull/1106 hoping it would run CI for me but no such luck.

Otherwise an alternate approach to avoid waitsets could be an async wait on the underlying rt_command_ptr_ to be set away from null (in combination with spin_until_future_complete), but not as nice. Maybe something else is possible with the default callback group of the node not having that parameter true that William mentioned, but I imagine it would affect something else on those nodes accidentally..?

christophfroehlich:
fixed with #1206 (kudos to @saikishor)

PR #1191, opened 2024-07-03
Scaled jtc

fmauch:
This is my take2 on #301.

I plan to finish this until Friday, @firesurfer feel free to poke me :-)

## Currently missing
- [x] tests for scaled execution
- [x] There are some ToDos in the documentation:
  - [x] Algorithmics
  - [x] Handling tolerances

## Things that will get changed, as discussed in the latest working group meeting
- [x] Allowing scaling factors above 1 ~~might get postponed to a later point.~~
- [x] Report the scaling factor in the controller state
- [x] Use a custom datatype for the scaling_factor subscriber (std_msgs are deprecated)
- [x] Change subscription to transient_local. This will require documentation, as a standard `ros2 topic pub` will not work with this.


fmauch:
Thoughts on tests I should implement:
- Execution with scaling factor of 0.5 takes twice as long (and does not fail, when no tolerances are set)
- Execution with scaling < 1 and goal time tolerance leads to an early halt and failed execution
- Execution with scaling < 1 and path tolerances set should **not** fail (if goal_time_tolerance is large enough)
- Execution with scaling > 1 should not fail independent of the tolerances set <-- is that true?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -302,8 +315,14 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   void resize_joint_trajectory_point_command(
     trajectory_msgs::msg::JointTrajectoryPoint & point, size_t size);
 
+  bool set_scaling_factor(const double scaling_factor);
+
   urdf::Model model_;
 
+  realtime_tools::RealtimeBuffer<double> scaling_factor_rt_buff_;
```
firesurfer, line 322: It would be better to use an std::atomic<double>. The RTBuffer uses a mutex which introduces unnecessary overhead.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include <chrono>
 #include <functional>
+#include <iostream>
```
firesurfer, line 19: not needed ?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -866,10 +906,33 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
...
   query_state_srv_ = get_node()->create_service<control_msgs::srv::QueryTrajectoryState>(
     std::string(get_node()->get_name()) + "/query_state",
     std::bind(&JointTrajectoryController::query_state_service, this, _1, _2));
 
+  if (
+    !has_velocity_command_interface_ && !has_acceleration_command_interface_ &&
+    !has_effort_command_interface_)
+  {
+    scaling_factor_sub_ = get_node()->create_subscription<SpeedScalingMsg>(
+      "~/speed_scaling_input", rclcpp::SystemDefaultsQoS(),
```
firesurfer, line 919: Wouldn't it make sense to make it transient local as we discussed before?

firesurfer:
hi @fmauch  Great to see this being worked on again. I added some comments to the code.

Regarding the tests:

>    Execution with scaling factor of 0.5 takes twice as long (and does not fail, when no tolerances are set)

Makes sense

>    Execution with scaling < 1 and goal time tolerance leads to an early halt and failed execution

Isn't this what we patched in: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/882. As far as I can tell you also backported this patch. (And I still think the goal time should be also scaled ;) )
    

>    Execution with scaling > 1 should not fail independent of the tolerances set <-- is that true?

Same as above. From heavily using this controller in our system I have the opinion that in most use cases it makes sense (and is what a user would expect) that the goal time should also be scaled.  But perhaps I understood your suggestion wrong.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: My main concern of making the subscriber volatile is the incompatibility with a volatile publisher: https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-compatibilities

I agree, that having a volatile subscriber has its benefits. My concern is that this is not compatible with a `ros2 topic pub` call from the command line without any further parameters which might raise a lot of issues. Maybe we could try to mitigate that in the documentation?

Regarding the topic name: Yes, in my opinion this is a prime example use case for remappings.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: yes, thanks. That was probably added automatically when I added some debugging at some point...

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
fmauch: Will using an std::atomic still allow a guaranteed non-blocking read from RT / a priority for the RT thread? I think we want to avoid having the subscriber callback block the `update()` method.

fmauch:
> From heavily using this controller in our system I have the opinion that in most use cases it makes sense (and is what a user would expect) that the goal time should also be scaled. But perhaps I understood your suggestion wrong.

Thank you for your input. That's why I put my brain dump on that here, so we can discuss this :-) And that's also why I didn't write the documentation on that, yet.

I'll wrap my head around this a little more and try to get a better feeling for the current behavior, then come back to this.

ros2_controllers-not-released.jazzy.repos
```diff
@@ -3,3 +3,8 @@ repositories:
     type: git
     url: https://github.com/picknikrobotics/generate_parameter_library.git
     version: main
+
+  control_msgs:
+    type: git
+    url: https://github.com/fmauch/control_msgs.git
+    version: scaled_jtc
```
fmauch, line 10:
```suggestion
```

ros2_controllers-not-released.rolling.repos
```diff
@@ -3,3 +3,8 @@ repositories:
     type: git
     url: https://github.com/picknikrobotics/generate_parameter_library.git
     version: main
+
+  control_msgs:
+    type: git
+    url: https://github.com/fmauch/control_msgs.git
+    version: scaled_jtc
```
fmauch, line 10:
```suggestion
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
firesurfer: std::atomic is the most lowlevel thread synchronisation type that you can use in c++. In this usecase you just want atomicity (aka. the guarantee that all bytes of the double are updated at once - see https://en.cppreference.com/w/cpp/atomic/memory_order ). 

From my point of view std::atomic will be as less blocking as it can be. (Btw. if you would just use a normal double the access would still be atomic on x86 if the memory is aligned  )

joint_trajectory_controller/src/joint_trajectory_controller.cpp
firesurfer: You are right about this. Nevertheless I think it is necessary to have it Transient local in order to avoid inconsistencies in the system if running multiple JTCs. Also there are plans to change this in the future: https://github.com/ros2/ros2/issues/1562

I agree regarding the topic name. The only argument I have for having the topic name in the parameter file is that it would allow to reconfigure the the topic at runtime (yes I sometimes do some wild stuff with controllers)... As far as I know this is not possible with remappings?!

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
fmauch: I've updated the PR accordingly. This also removes duplication of a shared buffer and a rt-only variable.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: Let's go for the transient_local in combination with some documentation.

Regarding the topic: To be able to change it during runtime, there would be some other mechanisms necessary (such as restarting the subscriber in the parameter callback). I'm not sure whether such a use case should be modeled and I think I would like to prefer leaving that discussion out of this PR for now if that's fine with you. I see your point, but as you already stated: this might be kind of an esoteric use case ;-)

fmauch:
I've took some time to think about the controller's behavior and currently, it should be like this, right @firesurfer?

- Execution with scaling factor of 0.5 takes twice as long.
- Execution with scaling factor of 2 should take 50% of the time.
- Execution with scaling factor of 0 should take as long as expected
- Independent of goal and path tolerances the action should not fail -- the goal time is scaled, as well.

The latter is the main point we discussed in https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/882. I am still fine with that approach, although I am not sure whether this covers every use case.

But with this I think I can finish the documentation and write some tests tomorrow.

firesurfer:
@fmauch 


>   Execution with scaling factor of 0.5 takes twice as long.

Yes. we sample the trajectory half as fast

>    Execution with scaling factor of 2 should take 50% of the time.

Yes we sample the trajectory twice as fast

>    Execution with scaling factor of 0 should take as long as expected

I would call this paused. I have the feeling this is one of the major usecases of this controller

>    Independent of goal and path tolerances the action should not fail -- the goal time is scaled, as well.

In case the trajectory is within the (scaled) goal time, it should not fail -> yes. Otherwise the combination of a scaled trajectory + a set goal time tolerance would always fail. 
Is there a specific use case you have in mind that would not be covered by that ?

fmauch:
My concern is mainly this:

If you have an application and you execute a trajectory on a robot that does speed scaling. The trajectory gets slowed down due to safety reasons, which means it finishes later than originally requested. This will slow down the whole application and no longer behave as programmed. This might be considered a fault. 

One could argue that

1. Depending follow-up actions should be triggered once their post-conditions are met (the motion action finished) and nit by time.
2. If execution time is critical, that can be monitored as a wrapper. MoveIt! for example does that by default.

I just quickly discussed that with a colleague and he also agreed that the behavior that we currently have is probably the expected one. He was also emphasizing the second point.

So, I guess, I'll just leave that result standing. As I said, I can definitively live with it. And it's probably even better / more appropriate than "I can live with that".

fmauch:
Update on the tests:

- testing is currently difficult in this repository, see https://github.com/ros-controls/ros2_controllers/issues/1101
- There is also a crash in one of the tests, I will open a separate issue with that
- Running the current tests produces an error in almost all tests:
  ```
  unknown file: Failure
  C++ exception with description "can't subtract times with different time sources [3 != 1]" thrown in the test body.
```
- There is currently an issue with the goal time tolerance. The action goal's tolerance always overwrites the one from he parameter. I'll add a PR for that.

**Comment by christophfroehlich** [comment]
>     * testing is currently difficult in this repository, see [Test failures: subscription already associated with a wait set #1101](https://github.com/ros-controls/ros2_controllers/issues/1101)

In the last weeks, I compiled ros2_control+controllers on jammy+humble to run the tests.

File: joint_trajectory_controller/doc/speed_scaling.rst [hunk 6]
```diff
@@ -0,0 +1,105 @@
...
+
+Methods of speed scaling
+------------------------
+
+Generally, the speed scaling feature has two separate scaling approaches in mind: On-Robot scaling
+and On-Controller scaling. They are both conceptually different and to correctly configure speed
+scaling it is important to understand the differences.
+
+On-Robot speed scaling
+~~~~~~~~~~~~~~~~~~~~~~
```
firesurfer, line 16: Question does the UR Driver currently support this mode also in the multiarm usecase ?

If I understand it correctly one needs to send the scaling factor via the corresponding command interface ?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -100,12 +100,17 @@ JointTrajectoryController::command_interface_configuration() const
       conf.names.push_back(joint_name + "/" + interface_type);
     }
   }
+  if (!params_.speed_scaling_command_interface_name.empty())
+  {
+    conf.names.push_back(params_.speed_scaling_command_interface_name);
```
firesurfer, line 105: See my comment above. I think this does not handle a multiarm robot case. But I think it would be fairly easy to support if we use a list instead of a single value here

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -116,12 +121,33 @@ JointTrajectoryController::state_interface_configuration() const
...
+  }
   return conf;
 }
 
 controller_interface::return_type JointTrajectoryController::update(
   const rclcpp::Time & time, const rclcpp::Duration & period)
 {
+  if (!params_.speed_scaling_state_interface_name.empty())
+  {
+    if (state_interfaces_.back().get_name() == params_.speed_scaling_state_interface_name)
```
firesurfer, line 139: You make the assumption that the speed scaling state interface is always the last one in the list. I think this is a rather dangerous assumption

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1563,6 +1635,45 @@ void JointTrajectoryController::resize_joint_trajectory_point_command(
...
+{
+  if (scaling_factor < 0)
+  {
+    RCLCPP_WARN(
+      get_node()->get_logger(),
+      "Scaling factor has to be greater or equal to 0.0 - Ignoring input!");
+    return false;
+  }
+
+  RCLCPP_INFO(get_node()->get_logger(), "New scaling factor will be %f", scaling_factor);
```
firesurfer, line 1648: I think it should be possible to disable this logging message. Some one could have the idea of continuously streaming the scaling factor and this would lead to spamming the log/terminal

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1563,6 +1635,45 @@ void JointTrajectoryController::resize_joint_trajectory_point_command(
...
+        get_node()->get_logger(),
+        "Setting the scaling factor while only one-way communication with the hardware is setup. "
+        "This will likely get overwritten by the hardware again. If available, please also setup "
+        "the speed_scaling_command_interface_name");
+    }
+    scaling_factor_ = scaling_factor;
+  }
+  else
+  {
+    if (get_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
```
firesurfer, line 1663: This block should be moved to the update method. 
Yo are calling set_scaling_factor from the subscription callback. This in non-rt. As far as I am aware interfaces should only be written from the rt thread. 

Additionally I would suggest to store the interface during on_activate in order to avoid looping all the command interfaces all the time. If you want to keep the explicit search I would suggest using std::find_if

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: Not necessarily, since we [put it there](https://github.com/ros-controls/ros2_controllers/pull/1191/files#diff-8dd2135dc1f4abe6382cc7671ef9ce939196b2d8adf18ac3a77b5c0fddf1a25fR105) But I would prefer a way of resolving this in a cleaner way, yes.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: We could set that output to an equality check with the current value and add a throttle to it. I think that might be a better approach than adding yet another config value.

firesurfer:
> Update on the tests:
> 
>     * testing is currently difficult in this repository, see [Test failures: subscription already associated with a wait set #1101](https://github.com/ros-controls/ros2_controllers/issues/1101)
> 
>     * There is also a crash in one of the tests, I will open a separate issue with that
> 
>     * Running the current tests produces an error in almost all tests:
>       ```
>       unknown file: Failure
>       C++ exception with description "can't subtract times with different time sources [3 != 1]" thrown in the test body.
>       ```
> 


So what about the remaining issues especially regarding the tests. 
From what I can tell so far:

1. Does not build on rolling because there the updated control_msgs are not used.  Probably it would make sense to merge the control_msgs change first. Given that the scaled JTC should/will definitely go upstream there should be no reason not to merge as soon as possible. 
2. Then in case of Jazzy where the build apparently succeeds there are multiple unrelated tests [failing](https://github.com/ros-controls/ros2_controllers/actions/runs/9839896335/job/27162931566?pr=1191)

Regarding the tests: @christophfroehlich I am not sure if your suggestion works. I had to explicitly make 1-2 small fixes to some ros2control packages in order to compile ros2control on iron. I would guess that's the same for humble. Is there currently any work in progress in order to get the test working on rolling again?

@fmauch I guess given that you can't run any of the tests at the moment you didn't start to implement any of the new tests we discussed before?

>     * There is currently an issue with the goal time tolerance. The action goal's tolerance always overwrites the one from he parameter. I'll add a PR for that.

Is there already an issue for that ? Do we need to discuss it in an extra issue?

christophfroehlich:
@firesurfer we have a CI job for that, it currently builds but some tests fail/are flaky
https://github.com/ros-controls/ros2_control_ci/actions/runs/9866959705

fmauch:
I was sick yesterday, but I'll try my best to squeeze in the tests until the end of the week.

firesurfer:
@fmauch whats the state ?

fmauch:
> @fmauch what's the current state ?

I've spent some time last week (or the week before?) on trying to get the tests running correctly. 

With that I realized that I basically ran into #697. The JTC currently samples the trajectory at the time given to the update loop. In the scaled version, however we basically sample at $t+\Delta t$. That is a conceptual difference. In the latest working group meeting we decided that I will change the JTC accordingly, as this is actually the "proper" approach.

Another issue related to that is that we move  forward in the trajectory only based on the controller periods combined with the speed scaling (which again is different to the current method which is why I will have to rewrite the tests). We can't really use the time directly, since the time we spent in the trajectory has been scaled, so we have to sum up the periods.

This will also get interesting when it comes to trajectory replacement, I think, but I haven't looked into that, yet.

So, in summary: Looks like the devil is in the details and this requires still some implementation work to do. While looking at that I did some refactoring work which I haven't pushed yet, as I want to get this together with working basic tests to make sure that the refactoring is actually behaving as expected. As I will be on vacation the next three weeks I don't know how much time I will find for that, though. But I will try.

fmauch:
I'm planning to split things up a bit into separate PRs to make things with a smaller scope, especially when it comes to adapting the tests.

- Use $t+\Delta t$ in the JTC
- Calculate the sample time based upon passed periods instead of the absolute time. This is a requirement for the scaled JTC, since we have to sum up the scaled periods to move forward in the trajectory.
- Add Speed scaling. This will basically only add the interfaces for speed scaling and change that the period the trajectory is moved forward isn't necessarily the complete last update cycle.

I'll leave out trajectory replacement for now, since this shall be implemented at a later point, anyway.

It might still make sense to merge them all in one PR, but I'll separate at least development like that.

christophfroehlich:
> It might still make sense to merge them all in one PR, but I'll separate at least development like that.

Ping me if you want me to create a `feature/scaled_jtc` branch in this repo for your individual PRs. However, the first two points seem to be independent somehow.

firesurfer:
@fmauch  

> feel free to poke me :-) 

Any updates ? :D

fmauch:
> Any updates ? :D

I almost finished the t+dT change on https://github.com/fmauch/ros2_controllers/tree/jtc_tdt. This change still makes tests fail which I have to investigate: https://github.com/fmauch/ros2_controllers/commit/f4530490e310cd5ee1aafe4cd8962015df34ffb5#diff-8dd2135dc1f4abe6382cc7671ef9ce939196b2d8adf18ac3a77b5c0fddf1a25fR287

fmauch:
Update: #1306 implements sampling at $t + \Delta T$. Furthermore, https://github.com/ros-controls/ros2_control/pull/1774 should fix incorrect controller periods reported to the controller.

With this the next step would be to propagate through the trajectory based on the accumulated periods instead of the absolute time. I expect that this will be quite a lot of work to make all tests comply with that.

fmauch:
Another update:

- [ ] Use t + Δ t in the JTC (implemented in #1306)
- [ ] Calculate the sample time based upon passed periods instead of the absolute time. This is a requirement for the scaled JTC, since we have to sum up the scaled periods to move forward in the trajectory. (Currently living ontop of the above on https://github.com/fmauch/ros2_controllers/tree/jtc_sum_periods)
- [ ] Add Speed scaling. (All merged together currently lives on https://github.com/fmauch/ros2_controllers/tree/scaled_jtc_merged)

fmauch:
This should be a state that could be started to get reviewed. There are a couple of things left that I'm not 100% happy, but they don't necessarily have to be covered in this PR in my opinion:

- I started using the feedback's `actual.time_from_start` to show the controller time passed since the trajectory started and `desired.time_from_start` for the *scaled* trajectory time since the trajectory started. This was mainly done since I needed a way to ovserve scaled time for testing purposes.
- The scaling method assumes (as mentioned in the docs) that the robot is scaling down the trajectory, so we have to scale time when looking at the beginning of the control period (How much of one period have we actually progressed in the last period?) and then we sample one full period ahead. If the robot doesn't scale down but we still want to use scaling, this will leave us with a tracking error. In that case we should scale the period for sampling the control target. My idea would be to make those two scaling methods available with a `scaling_type` parameteter with the values `HARDWARE` and `CONTROLLER`. I would also be happy to move that to a follow-up PR in order to get this one finally wrapped up. @RobertWilbrandt does this summarize the discussion we had a couple of days ago?
- There are a couple of builds failing that I don't quite understand. gpio controller is complaining about missing messages. Is this a cache thing? Because it is building on the master branch and also locally things are fine, so at least the semi-binary builds should be happy.

christophfroehlich:
>     * There are a couple of builds failing that I don't quite understand. gpio controller is complaining about missing messages. Is this a cache thing? Because it is building on the master branch and also locally things are fine, so at least the semi-binary builds should be happy.

I had a look now, I don't know tbh. The debian build fails with the same error, but this one doesn't use caches.

ros2_controllers.rolling.repos
```diff
@@ -9,8 +9,8 @@ repositories:
     version: master
   control_msgs:
     type: git
-    url: https://github.com/ros-controls/control_msgs.git
-    version: master
+    url: https://github.com/fmauch/control_msgs.git
+    version: scaled_jtc
```
christophfroehlich, line 13:
```suggestion
    url: https://github.com/ros-controls/control_msgs.git
    version: master
```

fmauch:
> ~I had a look now, I don't know tbh. The debian build fails with the same error, but this one doesn't use caches.~ Ah, I found the issue ;)

*facepalm* I thought I changed that back some time ago already.

christophfroehlich (changes requested):
Some first remarks

- Add a short description to the release_notes please.
- Add the new ROS interfaces [here](https://control.ros.org/rolling/doc/ros2_controllers/joint_trajectory_controller/doc/userdoc.html#description-of-controller-s-interfaces)

joint_trajectory_controller/doc/speed_scaling.rst
```diff
@@ -0,0 +1,105 @@
+Speed scaling
+=============
+
+The ``joint_trajectory_controller`` (JTC) supports dynamically scaling its trajectory execution speed.
+That means, when specifying a scaling factor :math:`{f}` of less than 1, execution will proceed only
```
firesurfer, line 5: Is it possible to specify a value > 1 for simulation purposes ? We are using this quite a lot in our setup. 
What would the maximum rate be? I saw the limit with the old implementation at 2.5

joint_trajectory_controller/doc/speed_scaling.rst
```diff
@@ -0,0 +1,105 @@
...
+     /joint_trajectory_controller/speed_scaling_input control_msgs/msg/SpeedScalingFactor "{factor: 0.5}"
+
+
+.. note::
+   The ``~/speed_scaling_input`` topic uses the QoS durability profile ``transient_local``. This
+   means you can restart the controller while still having a publisher on that topic active.
+
+.. note::
+   The current implementation only works for position-based interfaces.
+
```
firesurfer, line 87: Perhaps I am too tired after the ROScon. But could it be that it doesn't become 100% clear how to select a specfic mode?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -981,6 +1074,15 @@ controller_interface::CallbackReturn JointTrajectoryController::on_activate(
     cmd_timeout_ = 0.0;
   }
 
+  {
+    if (get_update_rate() == 0)
+    {
+      throw std::runtime_error("Controller's update rate is set to 0. This should not happen!");
```
firesurfer, line 1080: Is this such a common error that we need to check it ? Wouldn't it make more sense to check this some where in ros2control when loading a controller + config?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
firesurfer: Just as a reminder @fmauch  ;)

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -874,10 +904,45 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
...
+  }
+  else
+  {
+    RCLCPP_WARN(
+      logger,
+      "Speed scaling is currently only supported for position interfaces. If you want to make use "
+      "of speed scaling, please only use a position interface when configuring this controller.");
+    scaling_factor_ = 1.0;
+  }
+  if (!params_.speed_scaling_state_interface_name.empty())
```
firesurfer, line 933: Does it make sense to allow changing the scaling_factor via the topic and from the teach panel ?

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -1012,3 +1021,74 @@ INSTANTIATE_TEST_SUITE_P(
     std::make_tuple(
       std::vector<std::string>({"effort"}),
       std::vector<std::string>({"position", "velocity", "acceleration"}))));
+
+TEST_P(TestTrajectoryActionsTestScalingFactor, test_scaling_execution_time_succeeds)
```
firesurfer, line 1025: I have the feeling that just adding one test for such a complex feature is probably not enough. 

I think it is missing some parts:

1. It just checks the if the trajectory took the right amount of time -> It does not check if the trajectory was executed in the right way (e.g. had a spike what ever)

2. Executing multiple trajectories in series 
3. Proper behavior regarding state and goal tolerances (and goal time tolerances)

2/3. are motivated by: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/882

joint_trajectory_controller/doc/speed_scaling.rst
fmauch: Yes, it can be greater than 1. I haven't investigated a limit, though.

joint_trajectory_controller/doc/speed_scaling.rst
fmauch: What do you mean by "how to select a specific mode"? The current implementation will simply not do speed scaling when any other interface than position is configured for the controller.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: It should be either topic or hardware, at least that was my intention. I'll revisit that to verify that that;s what I've implemented.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: I always tend to add checks like that when I'm performing a division by a user-input variable. I don't expect this to ever happen, but it is out of my control at this point and if it ever happens, there will be a clear error.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: Thanks for the reminder, this should be resolved now.

joint_trajectory_controller/doc/speed_scaling.rst
firesurfer: If I just read the documentation I would think:

a) There are multiple modes I can choose from with the hint that on-robot scaling probably produces the better results
b) If there are multiple modes I would think: Ok somehow I have to tell the JTC to select which mode.

firesurfer:
@fmauch whats the current state of the scaled jtc efforts?

fmauch:
@RobertWilbrandt would you mind taking a review on this, as well, since you use it yourself AFAIK.

christophfroehlich (dismissed):
Thanks for adding more tests, my points of the first review were fixed! @firesurfer @RobertWilbrandt do you have time for another review round soonish?

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -46,6 +46,27 @@ joint_trajectory_controller:
...
+    default_value: "",
+    read_only: true,
+    description: "Fully qualified name of the speed scaling state interface name"
+  }
+  speed_scaling_command_interface_name: {
+    type: string,
+    default_value: "",
+    read_only: true,
+    description: "Command interface name used for setting the speed scaling factor on the hardware."
+  }
```
saikishor:
```suggestion
  speed_scaling:
    initial_scaling_factor: {
      type: double,
      default_value: 1.0,
      read_only: true,
      description: "The initial value of the scaling factor if no exchange with hardware takes place.",
      validation: {
        gt_eq<>: 0.0,
      }
    }

    state_interface: {
      type: string,
      default_value: "",
      read_only: true,
      description: "Fully qualified name of the speed scaling state interface name"
    }

    command_interface: {
      type: string,
      default_value: "",
      read_only: true,
      description: "Command interface name used for setting the speed scaling factor on the hardware."
    }
```

IMO This hierarchy is better than long naming

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -46,6 +46,27 @@ joint_trajectory_controller:
       "joint_trajectory_controller::state_interface_type_combinations": null,
     }
   }
+  scaling_factor_initial_default: {
+    type: double,
+    default_value: 1.0,
+    read_only: true,
+    description: "The initial value of the scaling factor if no exchange with hardware takes place.",
+    validation: {
+      gt_eq<>: 0.0,
```
saikishor: Is zero a valid one?

saikishor:
```suggestion
      gt<>: 0.0,
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -273,6 +283,9 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   void resize_joint_trajectory_point_command(
     trajectory_msgs::msg::JointTrajectoryPoint & point, size_t size, double value = 0.0);
 
+  bool set_scaling_factor(const double scaling_factor);
```
saikishor:
```suggestion
  bool set_scaling_factor(double scaling_factor);
```
Can you add some documentation on the return value

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -129,12 +133,34 @@ JointTrajectoryController::state_interface_configuration() const
...
 
 controller_interface::return_type JointTrajectoryController::update(
   const rclcpp::Time & time, const rclcpp::Duration & period)
 {
+  if (scaling_state_interface_.has_value())
+  {
+    scaling_factor_ = scaling_state_interface_->get().get_value();
+  }
+
+  if (get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
```
saikishor: Do we have to check for the ACTIVE status?. We know that `update` is only executed on activate right?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -938,10 +967,46 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
...
     std::bind(&JointTrajectoryController::query_state_service, this, _1, _2));
 
+  if (
+    !has_velocity_command_interface_ && !has_acceleration_command_interface_ &&
+    !has_effort_command_interface_)
+  {
+    auto qos = rclcpp::SystemDefaultsQoS();
+    qos.transient_local();
+    scaling_factor_sub_ = get_node()->create_subscription<SpeedScalingMsg>(
+      "~/speed_scaling_input", qos,
```
saikishor, line 977:
```suggestion
      "~/speed_scaling", qos,
```
May be only speed_scaling is enough? As it is a subscriber, it is implicit. Any opinion?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -938,10 +967,46 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
...
+      [&](const SpeedScalingMsg & msg) { set_scaling_factor(msg.factor); });
+    RCLCPP_INFO(
+      logger, "Setting initial scaling factor to %2f", params_.scaling_factor_initial_default);
+    scaling_factor_ = params_.scaling_factor_initial_default;
+  }
+  else
+  {
+    RCLCPP_WARN(
+      logger,
```
```suggestion
    RCLCPP_WARN_EXPRESSION(
      logger,
      params.scaling_factor_initial_default != 1.0,
```

So, it doesn't print always

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1654,6 +1756,41 @@ void JointTrajectoryController::resize_joint_trajectory_point_command(
   }
 }
 
+bool JointTrajectoryController::set_scaling_factor(const double scaling_factor)
```
```suggestion
bool JointTrajectoryController::set_scaling_factor(double scaling_factor)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1654,6 +1756,41 @@ void JointTrajectoryController::resize_joint_trajectory_point_command(
...
+  if (scaling_factor != scaling_factor_.load())
+  {
+    RCLCPP_INFO(
+      get_node()->get_logger().get_child("speed_scaling"), "New scaling factor will be %f",
+      scaling_factor);
+  }
+  scaling_factor_.store(scaling_factor);
+  if (params_.speed_scaling_command_interface_name.empty())
+  {
+    if (!params_.speed_scaling_state_interface_name.empty())
```
```suggestion
  if (params_.speed_scaling_command_interface_name.empty() && !params_.speed_scaling_state_interface_name.empty())
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1654,6 +1756,41 @@ void JointTrajectoryController::resize_joint_trajectory_point_command(
...
+    {
+      RCLCPP_WARN(
+        get_node()->get_logger(),
+        "Setting the scaling factor while only one-way communication with the hardware is setup. "
+        "This will likely get overwritten by the hardware again. If available, please also setup "
+        "the speed_scaling_command_interface_name");
+    }
```
```suggestion
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1654,6 +1756,41 @@ void JointTrajectoryController::resize_joint_trajectory_point_command(
...
+    RCLCPP_INFO(
+      get_node()->get_logger().get_child("speed_scaling"), "New scaling factor will be %f",
+      scaling_factor);
+  }
+  scaling_factor_.store(scaling_factor);
+  if (params_.speed_scaling_command_interface_name.empty())
+  {
+    if (!params_.speed_scaling_state_interface_name.empty())
+    {
+      RCLCPP_WARN(
```
```suggestion
      RCLCPP_WARN_ONCE(
```
Does it make sense for this to be ONCE, so it doesn't print every time?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -129,12 +133,34 @@ JointTrajectoryController::state_interface_configuration() const
...
     }
   }
+  if (!params_.speed_scaling_state_interface_name.empty())
+  {
+    conf.names.push_back(params_.speed_scaling_state_interface_name);
+  }
   return conf;
 }
 
 controller_interface::return_type JointTrajectoryController::update(
```
saikishor, line 143: I'm curious about one thing in the update method, if we set the initial default value, and in the first iteration of the speed scaling from the state is different, should we set the initial default one still?

I'm just curious what should be the behaviour in this case. @fmauch what should be done in your opinion?.

Thinking back may be the state interface one should superset the initial default one?.. this can be handled in a different PR

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
fmauch: It might be a valid one, yes. For UR robots, for example, it is possible to set 0.0 which is effectively pausing the robot.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
fmauch: I agree and will change that. I will not commit your suggestion as it doesn't cover the initial value as far as I can see.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
saikishor: Got it. Thanks for explanation

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
saikishor: Sure.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: I also think, the state interface should have priority, which is what the current implementation should do, right?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: As the speed scaling can also be read from a state interface, it might make sense to be explicit here. But that's not really a strong opinion, I would also be fine with having `~/speed_scaling`

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: Is it doing that?. As far as I looked, i didn't see that. I'll take a look again

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: @christophfroehlich any opinion?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: The scaling factor cmd is always set to the initial value on the first loop from the parameters right?  On the contrary, it should be the state value automatically on the first loop right?. 

This we can recheck and open an issue for future reference

fmauch: Actually, the cmd interface only gets set from the subscriber.

The state interface is always read at the beginning of `update()`. That value will be used for the cycle. That means, the initial value doesn't have an effect as soon as a state interface for reading the speed scaling is configured.

saikishor: Ok

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: let's be explicit with `speed_scaling_input` . everything will be clear from a quick `ros2 topic list` then

saikishor (approved):
LGTM

christophfroehlich (approved):
Thanks again :)

fmauch:
Sorry, for opening this again, but going through the unresolved comments, I refined the docstring requested in https://github.com/ros-controls/ros2_controllers/pull/1191#discussion_r2162069371 and extended the userdocs regarding https://github.com/ros-controls/ros2_controllers/pull/1191#discussion_r1870812258.

christophfroehlich (approved):

christophfroehlich:
@mergifyio backport jazzy

PR #1192, opened 2024-07-05
[JTC] Make goal_time_tolerance overwrite default value only if explicitly set

fmauch:
Since #716 we process the tolerances set from the action goal. This was done as described in https://github.com/ros-controls/control_msgs/blob/65814d985aa29bed0adfaed5f2ebd8cf26266056/control_msgs/action/FollowJointTrajectory.action#L6-L10 and https://github.com/ros-controls/control_msgs/blob/65814d985aa29bed0adfaed5f2ebd8cf26266056/control_msgs/msg/JointTolerance.msg#L7-L10.

However, the `goal_time_tolerance` was always overwritten with the value from the action goal. This actually changed the JTC's default behavior if

- a `goal_time_tolerance` != 0 was set in the controller config and
- no `goal_time_tolerance` or explicitly a `0.0` was given in the action goal.

In this, case, the controller would wait indefinitely until the robot reached the final target independent of the `goal_time_tolerance` configured inside the controller's default values.

This PR makes it such as

- If a `goal_time_tolerance` of `0.0` is given in the action (which is the default if left empty), the value from the controller config will be used
- If a `goal_time_tolerance` of `-1.0` is given the controller will wait at the end of the trajectory until the error gets lower than the `goal_tolerance` for each joint.
- If a `goal_time_tolerance` > 0.0 is given, the `goal_time_tolerance` from the action goal will be used.
- Any other negative value for `goal_time_tolerance` will make the controller fall back to its configured tolerances completely.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
@@ -31,6 +31,7 @@
 #define JOINT_TRAJECTORY_CONTROLLER__TOLERANCES_HPP_
 
 #include <limits>
+#include <stdexcept>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


christophfroehlich (changes requested):
Thanks for bringing this up, and fixing it immediately! But the `test_tolerances` test fails now, and we should test now the options for the goal_time_tolerances too.

```
: [ RUN      ] TestTolerancesFixture.test_deactivate_tolerances
2: /workspaces/ros2_humble_ws/src/ros2_controllers/joint_trajectory_controller/test/test_tolerances.cpp:190: Failure
2: Expected equality of these values:
2:   active_tolerances.goal_time_tolerance
2:     Which is: 0.10000000000000001
2:   0.0
2:     Which is: 0
2: [  FAILED  ] TestTolerancesFixture.test_deactivate_tolerances (0 ms)
```

bmagyar:
Waiting for new test then?

fmauch:
I've updated the `test_tolerance` tests accordingly. The jtc controller tests probably also need some updates, thoug.

bmagyar (approved):
LGTM

bmagyar:
I'll merge this so the release I'm pushing won't have behaviour-breaking changes.

saikishor (approved):
Nice! Thanks for fixing tests!

christophfroehlich:
@mergifyio backport iron humble

PR #1196, opened 2024-07-06
[Admittance] multiple state/command interfaces

MarcoMagriDev:
This PR aims to address the following issues:
1. Fix wrong `has` flags used in write and read methods 
2. Fix wrong fields of `state_commanded` used in `write_state_to_hardware`
3. Fix that multiple state interfaces are configured only the position state interface is considered due to `else if` statement.

Moreover:
1. Makes position state interface mandatory, otherwise the controller will not be able to perform computations. The controller will not be configured if not provided.
2. Adds compatibility with velocity command interface without position command interface.

### DIFF
--- a/admittance_controller/src/admittance_controller.cpp
+++ b/admittance_controller/src/admittance_controller.cpp
@@ -480,13 +480,13 @@ void AdmittanceController::read_state_from_hardware(
         state_interfaces_[pos_ind * num_joints_ + joint_ind].get_value();
       nan_position |= std::isnan(state_current.positions[joint_ind]);
     }
-    else if (has_velocity_state_interface_)
+    if (has_velocity_state_interface_)
     {
       state_current.velocities[joint_ind] =
         state_interfaces_[vel_ind * num_joints_ + joint_ind].get_value();
       nan_velocity |= std::isnan(state_current.velocities[joint_ind]);
     }
-    else if (has_acceleration_state_interface_)
+    if (has_acceleration_state_interface_)
     {
       state_current.accelerations[joint_ind] =
         state_interfaces_[acc_ind * num_joints_ + joint_ind].get_value();
@@ -532,15 +532,15 @@ void AdmittanceController::write_state_to_hardware(
       command_interfaces_[pos_ind * num_joints_ + joint_ind].set_value(
         state_commanded.positions[joint_ind]);
     }
-    else if (has_velocity_command_interface_)
+    if (has_velocity_command_interface_)
     {
       command_interfaces_[vel_ind * num_joints_ + joint_ind].set_value(
-        state_commanded.positions[joint_ind]);
+        state_commanded.velocities[joint_ind]);
     }
-    else if (has_acceleration_command_interface_)
+    if (has_acceleration_command_interface_)
     {
       command_interfaces_[acc_ind * num_joints_ + joint_ind].set_value(
-        state_commanded.positions[joint_ind]);
+        state_commanded.accelerations[joint_ind]);
     }
   }
   last_commanded_ = state_commanded;
--- a/admittance_controller/src/admittance_controller.cpp
+++ b/admittance_controller/src/admittance_controller.cpp
@@ -470,7 +470,7 @@ void AdmittanceController::read_state_from_hardware(
   bool nan_acceleration = false;
 
   size_t pos_ind = 0;
-  size_t vel_ind = pos_ind + has_velocity_command_interface_;
+  size_t vel_ind = pos_ind + has_velocity_state_interface_;
   size_t acc_ind = vel_ind + has_acceleration_state_interface_;
   for (size_t joint_ind = 0; joint_ind < num_joints_; ++joint_ind)
   {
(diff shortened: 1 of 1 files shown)


MarcoMagriDev:
Hi everyone, just a gentle reminder to review this PR when you get a chance. Let me know if anything needs clarification. Thanks!

admittance_controller/src/admittance_controller.cpp
```diff
@@ -553,8 +559,8 @@ void AdmittanceController::write_state_to_hardware(
 {
   // if any interface has nan values, assume state_commanded is the last command state
   size_t pos_ind = 0;
-  size_t vel_ind = pos_ind + has_velocity_command_interface_;
-  size_t acc_ind = vel_ind + has_acceleration_state_interface_;
+  size_t vel_ind = (has_position_command_interface_) ? has_velocity_command_interface_ : pos_ind;
```
saikishor, line 562: Can you please explain to me the logic here? 
It might be missing `pos_ind +`?

Thank you!

saikishor (commented):
Hello!

Sorry for the late review. I have a comment on a change. Rest seems fine to me.
Is this fix tested on any hardware?

admittance_controller/src/admittance_controller.cpp
```diff
@@ -553,8 +559,9 @@ void AdmittanceController::write_state_to_hardware(
 {
   // if any interface has nan values, assume state_commanded is the last command state
   size_t pos_ind = 0;
-  size_t vel_ind = pos_ind + has_velocity_command_interface_;
-  size_t acc_ind = vel_ind + has_acceleration_state_interface_;
+  size_t vel_ind =
+    (has_position_command_interface_) ? pos_ind + has_velocity_command_interface_ : pos_ind;
```
saikishor, line 563: Won't it be just pos_ind + velocity_command_interface_?

MarcoMagriDev: I don't think so. In the absence of a position command interface, doing that would result in indexing a non-existent command_interfaces_.

christophfroehlich (commented):
There exists a newer PR #1434 addressing similar issues. Does it make sense to fix everything at once, and add proper tests for the multiple interfaces?

MarcoMagriDev:
> There exists a newer PR #1434 addressing similar issues. Does it make sense to fix everything at once, and add proper tests for the multiple interfaces?

@christophfroehlich If you'd like, I can integrate those commits into my branch and push them to this PR. Let me know what you think!

christophfroehlich:
> > There exists a newer PR #1434 addressing similar issues. Does it make sense to fix everything at once, and add proper tests for the multiple interfaces?
> 
> @christophfroehlich If you'd like, I can integrate those commits into my branch and push them to this PR. Let me know what you think!

If you approve the changes and they are related, yes please merge them into your PR. Do you have an idea for adding a proper test for this use-case?

MarcoMagriDev:
Yes I think that they are somehow related to this PR since it deals with interfaces.

--- 

**Planned Tests for This PR:**
- [ ] Verify that if only the velocity `state_interface` is exported, the controller fails during configuration.  
- [ ] Ensure that if only the velocity `command_interface` is exported, there are no indexing errors in the vectors.  

**Planned Tests for the Other PR (After Integrating Changes):**
- [ ] Check that if a single `reference_interface` is exported, no errors are returned.  

@christophfroehlich Do you have any other tests in mind that we should consider?

admittance_controller/src/admittance_controller.cpp
MarcoMagriDev: I can confirm that without this fix, exporting only the velocity command interface would result in a segfault. I've added a test to cover this case and ensure it doesn't happen

christophfroehlich (dismissed):
Thanks for adding tests and the other fixes. Failing tests are not related to this PR.

admittance_controller/src/admittance_controller.cpp
saikishor: Awesome

admittance_controller/src/admittance_controller.cpp
```diff
@@ -137,7 +137,7 @@ AdmittanceController::on_export_reference_interfaces()
 
   // assign reference interfaces
   auto index = 0ul;
-  for (const auto & interface : allowed_reference_interfaces_types_)
+  for (const auto & interface : admittance_->parameters_.chainable_command_interfaces)
```
saikishor, line 140: Should we add default values of position and velocity here?

https://github.com/ros-controls/ros2_controllers/blob/c2dd18394cbcd7871f22f91c577bce58a64a65d4/admittance_controller/src/admittance_controller_parameters.yaml#L26-L32

saikishor (commented):
LGTM
Just a minor comment
Thanks for adding tests

admittance_controller/src/admittance_controller.cpp
MarcoMagriDev: Good catch! Yes, I think this should ensure that everything behaves as before with the default settings.  

If you'd like, I can push a new commit to address this.

saikishor: Sure! Yes. Please do 

Thank you

christophfroehlich (approved):

saikishor (approved):
Now, LGTM

christophfroehlich (changes requested):
`one_init_parameter_is_missing` is failing now

MarcoMagriDev:
> `one_init_parameter_is_missing` is failing now

`one_init_parameter_is_missing` was failing because we now define a default value for `chainable_command_interfaces`, whereas the test expected the controller to crash when this parameter was not explicitly provided.

I just pushed a new commit that removes this parameter from the list of tested ones in the test suite.

christophfroehlich (approved):

saikishor (approved):

PR #1198, opened 2024-07-10
Use ParamListener::try_get_params to Avoid Blocking in Real-Time Contexts

KentaKato:
Due to the potential blocking risk associated with ParamListener::get_params, it is advisable to avoid its usage in real-time contexts. The method, defined [here](https://github.com/PickNikRobotics/generate_parameter_library/blob/main/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header#L121-L124), can lead to delays that are unsuitable for time-sensitive operations.

As an alternative, I have switched to using ParamListener::try_get_params, which does not carry the risk of blocking. This non-blocking approach ensures better performance in real-time scenarios. You can find the implementation [here](https://github.com/PickNikRobotics/generate_parameter_library/blob/main/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header#L126-L135).


saikishor (commented):
@KentaKato Thank you for propagating these nice changes into the ros2_controllers. 

We would need to wait for the generate_parameter_library to be released in the distros before merging this.

Thank you

bmagyar:
Started a "tracking issue" for this ;) 

https://github.com/PickNikRobotics/generate_parameter_library/issues/209

KentaKato:
@saikishor @bmagyar Thank you!

christophfroehlich:
GPL got released, but something is very strange now with the JTC tests in the CI. Severals jobs fail with a segfault, and others have errors like `test_success_multi_point_with_velocity_sendgoal` in semi-binary (rolling, main). I ran the tests locally, here 
```
5: [  FAILED  ] OnlyEffortTrajectoryControllers/TestTrajectoryActionsTestParameterized.test_success_single_point_with_velocity_sendgoal/0, where GetParam() = ({ "effort" }, { "position", "velocity" }) (2026 ms)
```
@KentaKato could you have a look please?

KentaKato:
@christophfroehlich 

Although I haven't been able to reproduce this in my environment and haven't had time to thoroughly investigate it yet, I think the following logs are abnormal:

```
2024-10-29T21:07:04.7515029Z 5: [ERROR] [1730235957.984823612] [tolerances]: State tolerances failed for joint 0:
2024-10-29T21:07:04.7515356Z 5: [ERROR] [1730235957.984917837] [tolerances]: Position Error: 0.000000, Position Tolerance: 0.000000
2024-10-29T21:07:04.7515679Z 5: [ERROR] [1730235957.984965445] [tolerances]: Velocity Error: 0.015223, Velocity Tolerance: 0.000000
2024-10-29T21:07:04.7516030Z 5: [WARN] [1730235957.984997345] [test_joint_trajectory_controller]: Aborted due to state tolerance violation
```

This is because, according to `tolerances.hpp` ([link](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp#L303-L306)), tolerance errors should only be evaluated when `state_tolerance > 0.0`.

I'll look into this further from tomorrow onward.

bmagyar:
There was one set of tests passing on the previous run of the CI so I'm suspecting flakiness, I poked the CI again, let's see the results...

pid_controller/src/pid_controller.cpp
```diff
@@ -474,7 +474,7 @@ controller_interface::return_type PidController::update_and_write_commands(
   const rclcpp::Time & time, const rclcpp::Duration & period)
 {
   // check for any parameter updates
-  update_parameters();
+  param_listener_->try_get_params(params_);
```
christophfroehlich, line 477: We can remove update_parameters() method and use just `get_params()` in the `on_configure()` method (not in the RT loop).

bmagyar:
@christophfroehlich do we have a verdict on this?

christophfroehlich:
there was no update anymore, I'm not sure if this is needed at all

saikishor:
> I had another quick look: For me it seems that the API of `try_get_params` is not suitable here because it also returns true even if the parameters haven't changed -> stuff is updated at every update call.
> 
> ```c++
>     bool try_get_params(Params & params_in) const {
>       if (mutex_.try_lock()) {
>         if (const bool is_old = params_in.__stamp != params_.__stamp; is_old) {
>           params_in = params_;
>         }
>         mutex_.unlock();
>         return true;
>       }
>       return false;
>     }
> ```
> 
> This results in errors when I build this locally, and it might be that this is also the reason for the strange CI errors.

https://github.com/PickNikRobotics/generate_parameter_library/pull/260

This should fix it

pid_controller/src/pid_controller.cpp
christophfroehlich: done

christophfroehlich (approved):
As the upstream changes to GPL got released recently, we can go on with this PR. I updated all other occurrences as well, but kept the `try_get_params` without evaluating its return value where it is not necessary.

kilted-testing is fine, but something broke on rolling-testing [triggering the build to abort](https://build.ros2.org/job/Rbin_uN64__generate_parameter_library__ubuntu_noble_amd64__binary/).

admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
```diff
@@ -120,10 +120,8 @@ controller_interface::return_type AdmittanceRule::reset(const size_t num_joints)
 
 void AdmittanceRule::apply_parameters_update()
 {
-  if (parameter_handler_->is_old(parameters_))
-  {
-    parameters_ = parameter_handler_->get_params();
-  }
+  parameter_handler_->try_get_params(parameters_);
```
saikishor, line 123:
```suggestion
  parameter_handler_->try_update_params(parameters_);
```

Shouldn't it be `try_update_params`?

force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
```diff
@@ -144,10 +144,8 @@ controller_interface::CallbackReturn ForceTorqueSensorBroadcaster::on_deactivate
 controller_interface::return_type ForceTorqueSensorBroadcaster::update_and_write_commands(
   const rclcpp::Time & time, const rclcpp::Duration & /*period*/)
 {
-  if (param_listener_->is_old(params_))
-  {
-    params_ = param_listener_->get_params();
-  }
+  param_listener_->try_get_params(params_);
```
saikishor, line 147:
```suggestion
  param_listener_->try_update_params(params_);
```

pid_controller/src/pid_controller.cpp
saikishor:
```suggestion
  param_listener_->try_update_params(params_);
```

Am i missing something or it should be try_get_params?

saikishor (approved):

PR #1199, opened 2024-07-10
Unused header cleanup

henrygerardmoore:
This PR just removes unused headers, replacing them with different ones where necessary. It also standardizes using `
#include <gmock/gmock.h>` as opposed to `#include "gmock/gmock.h"` (both were being used). Each commit is one package for convenience. The code still builds and the same tests seem to be failing as before. I don't notice any improvement in build speed, so this is just a style change.

I used the clangd extension in vscode for this, though had to ignore a select few of its suggestions which caused build failures (particularly with `*_impl.hpp` files and similar.


henrygerardmoore:
Sorry about the force pushes, was rebasing to use the correct signing key.

bmagyar (approved):
Looking great! Thank you

christophfroehlich (approved):
LGTM :pray:

saikishor (approved):
LGTM

PR #512, opened 2023-01-27
Add Mecanum Drive Controller

destogl:
Depends on https://github.com/ros-controls/control_msgs/pull/79

PR #1201, opened 2024-07-11
Humble Mecanum Drive Controller

luis-camero:
Back-ported #512 to Humble. 

Moved implementation of the `update_reference_from_subcribers`  method to  `update_and_write_commands` to maintain timing check functionality from Jazzy.


christophfroehlich:
@luis-camero thanks for updating this for humble. But first, let us merge #512 into master. Could you leave a review comment there?
For humble, we also have to add an unstamped cmd_vel input to work with nav2.

PR #1202, opened 2024-07-12
Fix parallel gripper controller CI

saikishor:
There is a pre-formatting fix and also the failing build in the CI reported by @bmagyar and this is due to the recent change in the API of ros2_control

### DIFF
--- a/parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
+++ b/parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
@@ -322,7 +322,6 @@ controller_interface::CallbackReturn GripperActionController::on_activate(
     }
   }
 
-
   // Command - non RT version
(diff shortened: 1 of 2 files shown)


saikishor:
This is very weird that the CI is failing for some node_name stuff. I've tested it locally on my rolling docker in a separate workspace it is building and the tests are passing successfully. Could this be the cache of CI @christophfroehlich?

Moreover, if I remember correctly the CI on #1192 was successful right?. Can anyone with enough powers trigger the CI again?

saikishor:
:rotating_light: False alarm. After talking to @christophfroehlich it's just some pending releases and other issues we have caused it, so the CI is good. Thank you :)

christophfroehlich (approved):
LGTM, but I'm not sure how it could even work without the changes to the cm constructor calls

saikishor:
> LGTM, but I'm not sure how it could even work without the changes to the cm constructor calls

Yup! me neither. I don't know how it compiled and ran tests till now XD

bmagyar (approved):
Amazing, thank you!

PR #1204, opened 2024-07-13
Use a single file for migration + release notes

christophfroehlich:
we decided to show only the notes for the respective version in the docs, instead of all incremental distro upgrades

this makes backports easier and should avoid merge conflicts in the docs.

needs https://github.com/ros-controls/control.ros.org/pull/324


olivier-stasse (approved):
LGTM

olivier-stasse:
The CI is giving a strange error message while there are only 2 warnings and no visible error.

christophfroehlich:
> The CI is giving a strange error message while there are only 2 warnings and no visible error.

That's normal, the CI job considers warnings as errors and fails if it doesn't find files for example.

saikishor (approved):
LGTM!

PR #1206, opened 2024-07-15
Fix WaitSet issue in tests 

saikishor:
This PR aims to fix the WaitSet issue with the command subscription in the tests

PR #1210, opened 2024-07-16
fix parallel_gripper_controller failing tests

saikishor:
The Parallel gripper controller tests are not passing on the rolling from beginning, I've modified the tests to make it work. However, I had to remove a couple of tests as in the GPL the parameter `joint` is declared as `read_only`, and this makes most of the tests doesn't work at all, so I've cleaned them up 

https://github.com/ros-controls/ros2_controllers/blob/0bb88805713fb21cddc418bc4c502960d9fd9f39/parallel_gripper_controller/src/gripper_action_controller_parameters.yaml#L11-L14

### DIFF
--- a/parallel_gripper_controller/CMakeLists.txt
+++ b/parallel_gripper_controller/CMakeLists.txt
@@ -48,17 +48,19 @@ if(BUILD_TESTING)
   find_package(controller_manager REQUIRED)
   find_package(ros2_control_test_assets REQUIRED)
 
-  ament_add_gmock(test_load_parallel_gripper_action_controllers
-          test/test_load_parallel_gripper_action_controller.cpp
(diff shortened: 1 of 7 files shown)


christophfroehlich (commented):
Should we use GPL for validation instead of checking `params_.joint.empty()`? Then there is no need for testing it (see the deleted tests).

destogl (changes requested):
Agree with @christophfroehlich, let's use `GenerateParameterLiberry` validators instead of testing empty joints.

christophfroehlich (approved):
Thanks for the changes (haven't tested it with #1206, but I guess you did?)

destogl (approved):
Thanks!

PR #1214, opened 2024-07-16
[PID Controller] Export state interfaces for easier chaining with other controllers

saikishor:
This PR aims to export some state interfaces from the PID controllers that can be used in cascade with other controllers, this helps in not defining separate parameters for command interfaces (The one exported by the PID controller) and the state interfaces that define the same state as the PID controller, and it simplifies the design of the chained controllers


christophfroehlich:
Can't we implement this together with https://github.com/ros-controls/ros2_control/pull/1244? Instead of adding this to every chainable controller?

saikishor:
> Can't we implement this together with [ros-controls/ros2_control#1244](https://github.com/ros-controls/ros2_control/pull/1244)? Instead of adding this to every chainable controller?

@christophfroehlich #1244 is about publishing the interface information to a topic but not exposing as interfaces which is what we need here

destogl (approved):
Thanks! This looks good!

bmagyar (approved):

PR #1215, opened 2024-07-17
[ForceTorqueSensorBroadcaster] added force and torque offsets to the parameters + export state interfaces

saikishor:
This PR allows to define of the offsets of the force and torque through the parameter server. It would also be interesting to make it a chainable controller and expose the interfaces to the other controllers.

Now the offsets are removed and published and the same interfaces are exposed as state interfaces to be able to chain by other controllers


force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
```diff
@@ -141,23 +141,97 @@ controller_interface::CallbackReturn ForceTorqueSensorBroadcaster::on_deactivate
...
+
+void ForceTorqueSensorBroadcaster::apply_sensor_offset(
+  const Params & params, geometry_msgs::msg::WrenchStamped & msg)
+{
+  msg.wrench.force.x -= params.offset.force.x;
+  msg.wrench.force.y -= params.offset.force.y;
+  msg.wrench.force.z -= params.offset.force.z;
+  msg.wrench.torque.x -= params.offset.torque.x;
+  msg.wrench.torque.y -= params.offset.torque.y;
+  msg.wrench.torque.z -= params.offset.torque.z;
```
bmagyar, line 229: I don't want to be a nag but I'd reeeeeeeally expect offsets to be *added* using a `+`

saikishor: A very very fair point!
I'll try to change it ASAP

saikishor: Done. I've made the changes. Thanks for the feedback

saikishor:
Rebased the PR! :)

christophfroehlich (approved):
LGTM

force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
bmagyar: thank you :D

bmagyar (approved):
Thank you! :cookie:

PR #1221, opened 2024-07-18
[JTC] Use the internal methods instead of using the variables directly

saikishor:
These changes make the JTC use the internal method to retrieve the URDF rather than using the variables directly. I think it is safer and clear this way.

Once the https://github.com/ros-controls/ros2_control/pull/1623 is merged, we would definitely need these changes. These changes can also be merged even without 1623.

Thank you!

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -43,9 +43,10 @@ JointTrajectoryController::JointTrajectoryController()
 
 controller_interface::CallbackReturn JointTrajectoryController::on_init()
 {
-  if (!urdf_.empty())
+  const std::string & urdf = get_robot_description();
+  if (!urdf.empty())
   {
-    if (!model_.initString(urdf_))
+    if (!model_.initString(urdf))
     {
       RCLCPP_ERROR(get_node()->get_logger(), "Failed to parse URDF file");
     }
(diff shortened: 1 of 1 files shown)


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -701,7 +702,7 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
   joints_angle_wraparound_.resize(dof_);
   for (size_t i = 0; i < dof_; ++i)
   {
-    if (!urdf_.empty())
+    if (!get_robot_description().empty())
```
christophfroehlich, line 705: shouldn't we check here even for `model_` being empty? Can it be that the variable changes between `on_init` and `on_configure`? When a new robot_description is received, maybe?  but the `urdf::Model` doesn't have an empty check, we would need a persistent variable like `urdf_initialized`.

saikishor: The new `robot_description`, will be only transmitted to the new controller only upon unloading and loading the controller again, so I don't think this will change once the controller is initialized. Moreover, checking the model is valid or not can be good and be a part of a different PR.

henrygerardmoore (approved):
Allows `ros2_controllers` to build with latest `ros2_control`.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: Ok, fine for me to fix this in a different PR. I just started thinking about that reading it now (I know that I've written the current code ;)). Does it make sense to parse this in on_init at all? Currently, we use the `model_` variable only in the on_configure method.

christophfroehlich (approved):

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: Yes! As we have the URDF in the init method, we could initialize the model in the init method and reuse it

bmagyar (approved):

PR #1226, opened 2024-07-21
Add now missing includes

christophfroehlich:
https://github.com/ros-controls/ros2_control/pull/1627 removed lots of unused dependencies in the upstream packages, but now lots of packages here fail because of missing includes.

IMHO this was bad practice, so I'll fix them here instead of reverting the upstream changes.

:eyes: @henrygerardmoore


saikishor (approved):
The changes look good to me. 

We have to recheck if the downstream packages are building or not with the changes from https://github.com/ros-controls/ros2_control/pull/1627, if not we have to propagate the changes to downstream packages to avoid issues with the releases in ROS Buildfarm. 

FYI @henrygerardmoore

bmagyar (approved):
This was a surprisingly small amount to what I expected from the OP ;) thank you!

Issue #1225, opened 2024-07-21
[JTC] Refactor URDF parsing


christophfroehlich:
> Ok, fine for me to fix this in a different PR. I just started thinking about that reading it now (I know that I've written the current code ;)). Does it make sense to parse this in on_init at all? Currently, we use the `model_` variable only in the on_configure method.

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_controllers/pull/1221#discussion_r1684838474_

PR #1227, opened 2024-07-21
[JTC] Refactor URDF Model parsing 

saikishor:
This PR aims to remove the refactor of how the URDF model is parsed in the JTC, and be able to reject at the init phase if a bad robot_description is parsed, so the controller can be reloaded again when a new robot_description is updated via topic. This is helpful as if the description is bad, one should unload the controller and load it again as the `init` was reporting successful 

Closes: https://github.com/ros-controls/ros2_controllers/issues/1225
Needs: https://github.com/ros-controls/ros2_controllers/pull/1226 for the CI


christophfroehlich (approved):
This is great, thanks!

bmagyar (approved):
Very surgical! Nice job 👍

PR #1231, opened 2024-07-30
[JTC] Add Parameter to Toggle State Setting on Activation

KentaKato:
This pull request introduces a new parameter in the JointTrajectoryController that enables toggling whether the last command interface value should be set as both the current_state and last_commanded_state upon activation.

### Motivation
Consider a scenario where there is a transition from a JTC to another controller and back to a JTC. Due to the implementation specifics of the intermediary controller, it is possible that the command interface values from the first JTC remain. This can lead to dangerous situations where, upon reactivation, the second JTC inadvertently uses these stale command values to command the hardware. 

To mitigate this risk, I have added an option within the JTC that ensures command values are reliably reset to reflect the current state upon activation, enhancing system safety.


KentaKato:
Hi, I wanted to kindly ask if there’s a chance for this pull request to be reviewed. Any feedback would be greatly appreciated. Thank you!

bmagyar (approved):
Thank you!

saikishor (approved):
Thanks for the fix!

destogl:
Wait, I am not sure about this. Can you detail the scenario more. Are you using open loop?

KentaKato:
@destogl 
Thank you for your review!

---
We are not using open-loop control. We are creating a closed-loop controller called `joint_impedance_controller`. This controller only has the `effort` command_interface and does not have the `pos/vel/acc` command_interfaces.

- **JTC**
  - command_interface: pos/vel/acc/effort
- **joint_impedance_controller**
  - command_interface: effort

**Scenario:**
1. Activate JTC.
2. Switch to `joint_impedance_controller` and change the arm's position.
   - At this point, the `position` command_interface is not updated from step 1.
3. Switch to JTC again.
   - The `position` command_interface from the final position in step 1 remains. With the current implementation, this position gets set to `state_current_` and `last_command_state_`.

KentaKato:
@christophfroehlich
Thank you for your review!
I’ve added the note as you suggested. I appreciate your feedback!

bmagyar:
@christophfroehlich , are we good?

KentaKato:
@bmagyar @christophfroehlich 
friendly ping!

christophfroehlich:
@mergifyio backport iron humble

PR #1232, opened 2024-07-31
Fix admittance controller interface read/write logic

Nibanovic:
Using ROS `rolling` and gazebo `Harmonic`, I've found two bugs while trying to control a robot using _both_ `velocity` and `position` control.

1. In `write_state_to_hardware()`, `state_commanded.positions` are mistakenly being written to both `velocity` and `acceleration` command interfaces.
2. Both in `write_state_to_hardware()` and `read_state_from_hardware()`, checking whether the controller has `pos`/`vel`/`acc` interfaces is done using `if-else` clauses. This causes the block to **exit upon first true condition**, meaning, after confirming `position` interface exists, _`velocity` and `acceleration` interfaces never get written to or read from_

These fixes are verified in my use case, but seem obvious enough that no tests need to be written to verify these claims.

Nonetheless, let me know if you'd like some tests before merging. 

Cheers!

### DIFF
--- a/admittance_controller/src/admittance_controller.cpp
+++ b/admittance_controller/src/admittance_controller.cpp
@@ -476,13 +476,13 @@ void AdmittanceController::read_state_from_hardware(
         state_interfaces_[pos_ind * num_joints_ + joint_ind].get_value();
       nan_position |= std::isnan(state_current.positions[joint_ind]);
     }
-    else if (has_velocity_state_interface_)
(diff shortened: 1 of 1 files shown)


bmagyar (approved):
This makes a lot of sense sir :D

destogl (approved):

saikishor (approved):

PR #1233, opened 2024-07-31
[Joint State Broadcaster] Publish the joint_states of joints present in the URDF

saikishor:
Right now, when no `joints` parameter is set, it publishes all the entities with the `position`, `velocity` and `effort` interfaces, and the MoveIt and other plugins do not like that information and continuously spam with the warning logs, we have like 50MB of logs for 6 min run.

This PR tries to publish the `joint_states` of the joints that are within the URDF when no `joints` or `interface` params are set.


christophfroehlich:
JSB uses all joints defined in the ros2_control tag. with URDF you don't mean the ros2_control tag but the URDF-Model itself?
How can that be that someone has joints in the tag not being in the URDF model? RM does not init successfully anyways?

saikishor:
> JSB uses all joints defined in the ros2_control tag. with URDF you don't mean the ros2_control tag but the URDF-Model itself?
> How can that be that someone has joints in the tag not being in the URDF model? RM does not init successfully anyways?

You can have state interfaces from the actuators before transmission and after transmission are the joints so they are pretty different

christophfroehlich (approved):
I was not aware of the problem before. But adding gpio or sensor sections with interfaces called position
```
<gpio name="my_gpio">
  <command_interface name="position"/>
  <state_interface name="position">
    <param name="initial_value">1.0</param>
  </state_interface>
</gpio>
```
lets the JSB publish them.

As we currently don't have information about the type of the handles in the controllers (they exist only in the HardwareInfo class) we have to parse that again from the URDF. --> this PR is a straight forward solution of implementing that.

bmagyar (approved):

PR #1248, opened 2024-08-15
Fix segfault at reconfigure of AdmittanceController

### LINKED ISSUE #1181: [Iron] - Admittance controller regulary crashes with segfault on activation / configuration
**Describe the bug**

In order to change the kinematic tip and there for the admittant frame I regulary: `deactivate`  -> `set new params` -> `reconfigure` -> `activate` the controller. Sometimes the controller crashes with a segfault. 

See https://github.com/ros-controls/ros2_controllers/issues/1175 for the why I need to reconfigure the controller

Output:

> [ros2_control_node-14] [INFO] [1719385013.551660859] [controller_manager]: Configuring controller 'admittance_controller_top'
[ros2_control_node-14] [INFO] [1719385013.551826348] [admittance_controller_top]: No specific joint names are used for command interfaces. Using 'joints' parameter.
[ros2_control_node-14] [INFO] [1719385013.551938667] [admittance_controller_top]: state int types are: position
[ros2_control_node-14] 
[ros2_control_node-14] [INFO] [1719385013.552066065] [admittance_controller_top]: command int types are: position
[ros2_control_node-14] 
[ros2_control_node-14] [INFO] [1719385013.552158598] [admittance_controller_top]: chainable int types are: position
[ros2_control_node-14] 
[ros2_control_node-14] [INFO] [1719385013.552234329] [admittance_controller_top]: chainable int types are: velocity
[ros2_control_node-14] 
[ros2_control_node-14] [INFO] [1719385013.552316803] [admittance_controller_top]: Command interfaces are [position] and and state interfaces are [position].
[ros2_control_node-14] Stack trace (most recent call last) in thread 3073738:
[ros2_control_node-14] #19   Object "", at 0xffffffffffffffff, in 
[ros2_control_node-14] #18   Source "../sysdeps/unix/sysv/linux/x86_64/clone3.S", line 81, in __clone3 [0x72f6ae92684f]
[ros2_control_node-14] #17   Source "./nptl/pthread_create.c", line 442, in start_thread [0x72f6ae894ac2]
[ros2_control_node-14] #16   Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x72f6aecdc252, in 
[ros2_control_node-14] #15   Object "<my workspace>/install/rclcpp/lib/librclcpp.so", at 0x72f6af19f501, in rclcpp::executors::MultiThreadedExecutor::run(unsigned long)
[ros2_control_node-14] #14   Object "<my workspace>/install/rclcpp/lib/librclcpp.so", at 0x72f6af189253, in rclcpp::Executor::execute_any_executable(rclcpp::AnyExecutable&)
[ros2_control_node-14] #13   Object "<my workspace>/install/rclcpp/lib/librclcpp.so", at 0x72f6af187fcc, in rclcpp::Executor::execute_service(std::shared_ptr<rclcpp::ServiceBase>)
[ros2_control_node-14] #12   Object "<my workspace>/install/rclcpp/lib/librclcpp.so", at 0x72f6af185e23, in void take_and_do_error_handling<rclcpp::Executor::execute_service(std::shared_ptr<rclcpp::ServiceBase>)::{lambda()#1}, rclcpp::Executor::execute_service(std::shared_ptr<rclcpp::ServiceBase>)::{lambda()#2}>(char const*, char const*, rclcpp::Executor::execute_service(std::shared_ptr<rclcpp::ServiceBase>)::{lambda()#1}, rclcpp::Executor::execute_service(std::shared_ptr<rclcpp::ServiceBase>)::{lambda()#2}) [clone .constprop.0]
[ros2_control_node-14] #11   Object "/opt/ros/iron/lib/libcontroller_manager.so", at 0x72f6af4679ba, in 
[ros2_control_node-14] #10   Object "/opt/ros/iron/lib/libcontroller_manager.so", at 0x72f6af43d40b, in std::_Function_handler<void (std::shared_ptr<controller_manager_msgs::srv::ConfigureController_Request_<std::allocator<void> > >, std::shared_ptr<controller_manager_msgs::srv::ConfigureController_Response_<std::allocator<void> > >), std::_Bind<void (controller_manager::ControllerManager::*(controller_manager::ControllerManager*, std::_Placeholder<1>, std::_Placeholder<2>))(std::shared_ptr<controller_manager_msgs::srv::ConfigureController_Request_<std::allocator<void> > >, std::shared_ptr<controller_manager_msgs::srv::ConfigureController_Response_<std::allocator<void> > >)> >::_M_invoke(std::_Any_data const&, std::shared_ptr<controller_manager_msgs::srv::ConfigureController_Request_<std::allocator<void> > >&&, std::shared_ptr<controller_manager_msgs::srv::ConfigureController_Response_<std::allocator<void> > >&&)
[ros2_control_node-14] #9    Object "/opt/ros/iron/lib/libcontroller_manager.so", at 0x72f6af411d96, in controller_manager::ControllerManager::configure_controller_service_cb(std::shared_ptr<controller_manager_msgs::srv::ConfigureController_Request_<std::allocator<void> > >, std::shared_ptr<controller_manager_msgs::srv::ConfigureController_Response_<std::allocator<void> > >)
[ros2_control_node-14] #8    Object "/opt/ros/iron/lib/libcontroller_manager.so", at 0x72f6af405e2f, in controller_manager::ControllerManager::configure_controller(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
[ros2_control_node-14] #7    Object "/opt/ros/iron/lib/libcontroller_interface.so", at 0x72f6aef61165, in controller_interface::ControllerInterfaceBase::configure()
[ros2_control_node-14] #6    Object "<my workspace>/install/rclcpp_lifecycle/lib/librclcpp_lifecycle.so", at 0x72f6aebe27d9, in rclcpp_lifecycle::LifecycleNode::LifecycleNodeInterfaceImpl::trigger_transition(unsigned char)
[ros2_control_node-14] #5    Object "<my workspace>/install/rclcpp_lifecycle/lib/librclcpp_lifecycle.so", at 0x72f6aebe1cf1, in rclcpp_lifecycle::LifecycleNode::LifecycleNodeInterfaceImpl::change_state(unsigned char, rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn&)
[ros2_control_node-14] #4    Object "<my workspace>/install/rclcpp_lifecycle/lib/librclcpp_lifecycle.so", at 0x72f6aebe18f6, in rclcpp_lifecycle::LifecycleNode::LifecycleNodeInterfaceImpl::execute_callback(unsigned int, rclcpp_lifecycle::State const&) const
[ros2_control_node-14] #3    Object "/opt/ros/iron/lib/libadmittance_controller.so", at 0x72f6ac0c54dd, in admittance_controller::AdmittanceController::on_configure(rclcpp_lifecycle::State const&)
[ros2_control_node-14] #2    Object "/opt/ros/iron/lib/libadmittance_controller.so", at 0x72f6ac0b62a9, in admittance_controller::AdmittanceRule::configure(std::shared_ptr<rclcpp_lifecycle::LifecycleNode> const&, unsigned long)
[ros2_control_node-14] #1    Object "/opt/ros/iron/lib/libkinematics_interface_kdl.so", at 0x72f6ac044a49, in 
[ros2_control_node-14] #0    Object "/opt/ros/iron/lib/libkinematics_interface_kdl.so", at 0x72f6ac03a0bf, in 
[ros2_control_node-14] Segmentation fault (Address not mapped to object [0x72f6800fc3c8])

**To Reproduce**

Not sure. I have a rather complex setup. But basically:

1. Setup admittance controller with chained jtc
2. Run the `deactivate`  -> `set new params` -> `reconfigure` -> `activate`  steps. (I adapt the stiffness + the kinematic tip frame + the admittance frame)
3. Controller should crash in some cases

**Expected behavior**

Controller should not crash

 - OS:  Ubuntu 22.04
 - Version Iron - binary install
- But RCLCPP with race condition fix: https://github.com/cellumation/rclcpp/tree/action_iron_backport (Issue happens also with default rclcpp)

EDIT: This behavior also occurs with the normal JTC instead of the scaled JTC

firesurfer:
This PR fixes #1181.

I tested it on iron and rebased it on rolling afterwards. 

This PR simply avoid recreating the kinematics plugin, but just reinitializes it - which seems to work.

### DIFF
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -46,15 +46,17 @@ controller_interface::return_type AdmittanceRule::configure(
   {
     try
     {
-      // Do not create a new instance in case of a reinit
-      if (!kinematics_loader_)
+      // Make sure we destroy the interface first. Otherwise we might run into a segfault
+      if (kinematics_loader_)
       {
-        kinematics_loader_ =
-          std::make_shared<pluginlib::ClassLoader<kinematics_interface::KinematicsInterface>>(
-            parameters_.kinematics.plugin_package, "kinematics_interface::KinematicsInterface");
-        kinematics_ = std::unique_ptr<kinematics_interface::KinematicsInterface>(
-          kinematics_loader_->createUnmanagedInstance(parameters_.kinematics.plugin_name));
+        kinematics_.reset();
       }
+      kinematics_loader_ =
+        std::make_shared<pluginlib::ClassLoader<kinematics_interface::KinematicsInterface>>(
+          parameters_.kinematics.plugin_package, "kinematics_interface::KinematicsInterface");
+      kinematics_ = std::unique_ptr<kinematics_interface::KinematicsInterface>(
+        kinematics_loader_->createUnmanagedInstance(parameters_.kinematics.plugin_name));
+
       if (!kinematics_->initialize(
             node->get_node_parameters_interface(), parameters_.kinematics.tip))
       {
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -46,11 +46,14 @@ controller_interface::return_type AdmittanceRule::configure(
   {
     try
     {
-      kinematics_loader_ =
-        std::make_shared<pluginlib::ClassLoader<kinematics_interface::KinematicsInterface>>(
-          parameters_.kinematics.plugin_package, "kinematics_interface::KinematicsInterface");
-      kinematics_ = std::unique_ptr<kinematics_interface::KinematicsInterface>(
-        kinematics_loader_->createUnmanagedInstance(parameters_.kinematics.plugin_name));
+      //Do not create a new instance in case of a reinit
+      if(!kinematics_loader_ ){
+        kinematics_loader_ =
+          std::make_shared<pluginlib::ClassLoader<kinematics_interface::KinematicsInterface>>(
+            parameters_.kinematics.plugin_package, "kinematics_interface::KinematicsInterface");
+        kinematics_ = std::unique_ptr<kinematics_interface::KinematicsInterface>(
+          kinematics_loader_->createUnmanagedInstance(parameters_.kinematics.plugin_name));
+      }
       if (!kinematics_->initialize(
             node->get_node_parameters_interface(), parameters_.kinematics.tip))
       {
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -46,8 +46,9 @@ controller_interface::return_type AdmittanceRule::configure(
   {
     try
     {
-      //Do not create a new instance in case of a reinit
-      if(!kinematics_loader_ ){
+      // Do not create a new instance in case of a reinit
+      if (!kinematics_loader_)
+      {
         kinematics_loader_ =
           std::make_shared<pluginlib::ClassLoader<kinematics_interface::KinematicsInterface>>(
             parameters_.kinematics.plugin_package, "kinematics_interface::KinematicsInterface");


saikishor (commented):
The changes look ok, but I'm not sure if this is the right way to fix the seg fault issue you are having

firesurfer:
What would you suggest as an alternative ?

firesurfer:
My guess to what happens there is:

We basically delete and recreate the instance of the kinematics plugin.  It might be possible that at the same time there is  an access happening to the reference we have to the dynamic library -  but it is already deleted. Given that it does not happen every time is a strong  indicator that probably multiple threads are accessing the instance of the kinematics plugin at the same time. 

I think not reloading the plugin if it is already loaded is just fine.

saikishor:
> My guess to what happens there is:
> 
> We basically delete and recreate the instance of the kinematics plugin.  It might be possible that at the same time there is  an access happening to the reference we have to the dynamic library -  but it is already deleted. Given that it does not happen every time is a strong  indicator that probably multiple threads are accessing the instance of the kinematics plugin at the same time. 
> 
> I think not reloading the plugin if it is already loaded is just fine. 

If the controller is not active, the update loop is not run at all, in that case there is only one thread that is calling the configure method is the non-RT one. That's why, I'm very curious to know why only on reconfigure. Can you try resetting the kinematics_ pointer with `.reset()` before resetting the loader.

firesurfer:
I guess you were right!

I just changed to code to:

```c++
      if(kinematics_loader_ ){
        kinematics_.reset();
      }
        kinematics_loader_ =
          std::make_shared<pluginlib::ClassLoader<kinematics_interface::KinematicsInterface>>(
            parameters_.kinematics.plugin_package, "kinematics_interface::KinematicsInterface");
        kinematics_ = std::unique_ptr<kinematics_interface::KinematicsInterface>(
          kinematics_loader_->createUnmanagedInstance(parameters_.kinematics.plugin_name));
      
      if (!kinematics_->initialize(
            node->get_node_parameters_interface(), parameters_.kinematics.tip))
      {
        return controller_interface::return_type::ERROR;
      }
```

and no segfault. But it is nevertheless very interesting that this is causing an issue.

firesurfer:
@saikishor So what do you think. Shall we keep the fix as it currently or do you suggest a different approach ?

saikishor:
> @saikishor So what do you think. Shall we keep the fix as it currently or do you suggest a different approach ?
> 

@firesurfer have you taken a look at the pluginlib documentation?. I recommend doing so and do a proper fix. 

As we are talking about reconfiguring, it could be that the user wants to change the plugin type and it would be only possible at reconfigure instance, else he has to unload and load + configure the controller again

firesurfer:
@saikishor I just pushed the adaption I showed in my previous comment. 

I think this could be a proper solution as it makes sure that the `KinematicsInterface` is destroyed before we destroy and recreate the `ClassLoader`. Additionally as we also recreate the loader it is also possible to change the plugin type now. 

I did take a look at the ClassLoader headerfile but as far as I can tell there is no information on how to properly re-create a ClassLoader or about the destruction order

But what I could imagine what happens there is: 

When we destroy the class loader the linux dynamic library loader may unload/destroy the instance of the dynamic library https://linux.die.net/man/3/dlopen:
`The function dlclose() decrements the reference count on the dynamic library handle handle. If the reference count drops to zero and no other loaded libraries use symbols in it, then the dynamic library is unloaded. ` 
The kernel btw. does not have to unload the library. 

If we then destroy the interface handle it calls the destructor of the KinematicsInterface, which results in a segfault in case the library is already unloaded. I would guess that the timing between the calling dlclose and the library actually being unloaded may not be deterministic. 

But as I said this is just a guess.

bmagyar (approved):
I think this is an elegant solution to the problem you encountered

firesurfer:
The failing ci issues are as far as I can tell due to an unrelated test.

olivier-stasse (approved):
LGTM (after the dlopen discussions).

firesurfer:
@saikishor Is it fine for you to merge it  like this / does my explanation sound reasonable ?

saikishor (approved):

PR #1249, opened 2024-08-15
Add wrench offset for admittance controller

firesurfer:
This PR adds an additional WrenchStamped subscriber to the AdmittanceController in order to allow specifying a Wrench offset. 

The basic idea behind this PR is to simply add the offset to the measured force values which has the controller to produce an offset into the desired direction. I tested it on an UR16e where it seems to work quite well. 

Is this a desired feature? What do you think of it? Perhaps there are better ways to implement this feature in this controller.


admittance_controller/src/admittance_controller.cpp
```diff
@@ -27,6 +27,23 @@
...
+  res.force.y = a.force.y + b.force.y;
+  res.force.z = a.force.z + b.force.z;
+
+  res.torque.x = a.torque.x + b.torque.x;
+  res.torque.y = a.torque.y + b.torque.y;
+  res.torque.z = a.torque.z + b.torque.z;
+
+  return res;
+}
+
```
saikishor, line 46: Do we need additional subscriber to add these wrench offsets?. Does #1215 work for your case? Where the offsets are added at configure time and then it exposes state interfaces, these interfaces can be used by the admittance controller

firesurfer: The admittance controller reads the FT values directly via the corresponding state interfaces and not via a ros topic. I do not see how a change in the FTBroadcaster might help here.

Additionally adding an additional subscriber would allow to publish certain wrench profiles. I would guess the parameter system is way to slow for that.

saikishor: Got it. May be I misunderstood your description sorry. You mean to have this offsets in the wrench, so the admittance control can apply a certain force in that specific direction?

firesurfer: Exactly. Sorry if my description was misleading

admittance_controller/src/admittance_controller.cpp
```diff
@@ -27,6 +27,23 @@
 
 namespace admittance_controller
 {
+
+static geometry_msgs::msg::Wrench add_wrenches(
```
saikishor, line 31:
```suggestion
geometry_msgs::msg::Wrench add_wrenches(
```

I don't think static is necessary here

admittance_controller/src/admittance_controller.cpp
```diff
@@ -265,6 +283,20 @@ controller_interface::CallbackReturn AdmittanceController::on_configure(
...
+
+  input_wrench_command_subscriber_ =
+    get_node()->create_subscription<geometry_msgs::msg::WrenchStamped>(
+      "~/wrench_reference", rclcpp::SystemDefaultsQoS(),
+      [&](const geometry_msgs::msg::WrenchStamped & msg)
+      {
+        if (msg.header.frame_id != admittance_->parameters_.ft_sensor.frame.id)
+        {
+          RCLCPP_ERROR_STREAM(
+            get_node()->get_logger(), "Ignoring wrench reference as it is on the wrong frame");
```
saikishor, line 295:
```suggestion
            get_node()->get_logger(), "Ignoring wrench reference as it is on the wrong frame: " << msg.header.frame_id << ". Expected reference frame: << admittance_->parameters_.ft_sensor.frame_id);
```

admittance_controller/src/admittance_controller.cpp
```diff
@@ -265,6 +283,20 @@ controller_interface::CallbackReturn AdmittanceController::on_configure(
   input_joint_command_subscriber_ =
     get_node()->create_subscription<trajectory_msgs::msg::JointTrajectoryPoint>(
       "~/joint_references", rclcpp::SystemDefaultsQoS(), joint_command_callback);
+
+  input_wrench_command_subscriber_ =
+    get_node()->create_subscription<geometry_msgs::msg::WrenchStamped>(
+      "~/wrench_reference", rclcpp::SystemDefaultsQoS(),
+      [&](const geometry_msgs::msg::WrenchStamped & msg)
+      {
+        if (msg.header.frame_id != admittance_->parameters_.ft_sensor.frame.id)
```
saikishor, line 292:
```suggestion
        if (msg.header.frame_id != admittance_->parameters_.ft_sensor.frame.id && !msg.header.frame_id.empty())
```

saikishor (approved):
In general, the changes look good to me. I've left some minor comments. Please take a look

admittance_controller/src/admittance_controller.cpp
firesurfer: Depends on the desired visibility - I tend to make all freestanding functions static per default so that they are not visible to any other compilation units.

saikishor: If you don't make them static,  the modern compilers optimize them still, it's called return value optimization

saikishor:
@firesurfer can you remove the draft status on the title?

admittance_controller/src/admittance_controller.cpp
firesurfer: If we do not declare them static we basically tell the compiler that somewhere else in the code we may write a definition for that function and call it. Also from my point of view: If I do not declare a function in a header file I make it static as I do not want to expose it outside of the compilation unit.

firesurfer:
@saikishor I removed the draft status.

saikishor (approved):

firesurfer:
Any of the other reviewers willing to take a look at this ?

admittance_controller/src/admittance_controller.cpp
bmagyar: We've been using anonymous namespaces to achieve this which avoids mixing in the semantics of static functions/objects.

admittance_controller/src/admittance_controller.cpp
```diff
@@ -265,6 +283,20 @@ controller_interface::CallbackReturn AdmittanceController::on_configure(
...
+
+  input_wrench_command_subscriber_ =
+    get_node()->create_subscription<geometry_msgs::msg::WrenchStamped>(
+      "~/wrench_reference", rclcpp::SystemDefaultsQoS(),
+      [&](const geometry_msgs::msg::WrenchStamped & msg)
+      {
+        if (msg.header.frame_id != admittance_->parameters_.ft_sensor.frame.id && !msg.header.frame_id.empty())
+        {
+          RCLCPP_ERROR_STREAM(
+            get_node()->get_logger(), "Ignoring wrench reference as it is on the wrong frame: " << msg.header.frame_id << ". Expected reference frame: << admittance_->parameters_.ft_sensor.frame_id);
```
bmagyar, line 295:
```suggestion
            get_node()->get_logger(), "Ignoring wrench reference as it is on the wrong frame: " << msg.header.frame_id << ". Expected reference frame: " << admittance_->parameters_.ft_sensor.frame_id);
```
helps compiling

admittance_controller/src/admittance_controller.cpp
```diff
@@ -265,6 +283,20 @@ controller_interface::CallbackReturn AdmittanceController::on_configure(
...
+          return;
```
bmagyar, line 296: would it make sense to explicitly reset the last wrench to zero in this case?

bmagyar (changes requested):

bmagyar:
Do you have any ideas for a test for this by any chance?

firesurfer:
The way I tested it on hardware was: Load the controller and then command 1N into a single direction. One will then see a motion of the arm in that direction. I guess it would be possible to implement something similar in an unit test.

admittance_controller/src/admittance_controller.cpp
firesurfer: Well that's something which isn't easy to decide. 

I would argue: A wrong input shall not change the state of the system. Therefore we may not change the current offset.

firesurfer:
@destogl When browsing through the tests I found this:

```
    // TODO(destogl): comment in when using unified mode
    //     if (controller_->admittance_->unified_mode_) {
    //       force_command_publisher_->publish(force_msg);
    //     }
    //    pose_command_publisher_->publish(pose_msg);
```

Apparently there was some support for a force command planned but not implemented? Is this right?

admittance_controller/src/admittance_controller.cpp
```diff
@@ -265,6 +283,20 @@ controller_interface::CallbackReturn AdmittanceController::on_configure(
...
       "~/joint_references", rclcpp::SystemDefaultsQoS(), joint_command_callback);
+
+  input_wrench_command_subscriber_ =
+    get_node()->create_subscription<geometry_msgs::msg::WrenchStamped>(
+      "~/wrench_reference", rclcpp::SystemDefaultsQoS(),
+      [&](const geometry_msgs::msg::WrenchStamped & msg)
+      {
+        if (msg.header.frame_id != admittance_->parameters_.ft_sensor.frame.id && !msg.header.frame_id.empty())
+        {
+          RCLCPP_ERROR_STREAM(
```
destogl, line 294: Why don't we transform the wrench if it is in another frame to the sensor frame?

admittance_controller/src/admittance_controller.cpp
destogl: Agree with this logic

destogl (commented):
Generally, great!

What I would like to know is, do we want to enable this input even admittance controller is in a chain? As you did right now. I am more on a `yes` side, but would like to hear oppinions.

admittance_controller/src/admittance_controller.cpp
firesurfer: I had some very weird results in the past when transforming wrenches with TF (e.g. non zero torques afterwards). Additionally I don't think we have a tf buffer/listener in the controller at the moment and why add this additional overhead?

firesurfer:
So I just updated the code, that it compiles again in jazzy and rolling. 
How can we progress with this PR?

firesurfer:
@destogl and @bmagyar 
Just a friendly ping :)

christophfroehlich (changes requested):
I never used this controller so I don't comment on the changes itself but it seems to be valuable acc. to previous reviews.

But please:
- Update the [controller's documentation](https://github.com/ros-controls/ros2_controllers/blob/master/admittance_controller/doc/userdoc.rst)
- Add this new feature to the [release notes](https://github.com/ros-controls/ros2_controllers/blob/master/doc/release_notes.rst).

christophfroehlich (approved):

firesurfer:
So the only outstanding comment is the one from @destogl  about the force command transformation. I don't think this is a blocker at the moment as it could be changed rather easy afterwards.

PR #1251, opened 2024-08-17
Gpio command controller

Wiktor-99:
As I discussed with @christophfroehlich I've opened new pr with gpio controller.
 
This PR is a follow-up to the [thread](https://github.com/ros-controls/ros2_controllers/pull/342) and implements a controller to send commands to GPIO interfaces, allowing specific command interfaces for each GPIO.

I have made the following changes compared to the original PR:
- I utilized the 'DynamicJointState' message for both the command and state interfaces (this allows sending commands with specific GPIO values without having to worry about the order of ports in the command message).
- I added a parameters file, and the new input parameters file will look something like this:
```yml
gpios:
  - gpio1
  - gpio2
command_interfaces:
  - gpio1:
    - ports:
      - dig1
      - dig2
  - gpio2:
    - ports:
      - ana1
```
- I've done significant refactoring and have added many new UTs.


gpio_controllers/package.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>gpio_controllers</name>
+  <version>1.0.0</version>
```
olivier-stasse, line 5: Should the version number not be 4.12.1 as the rest of the repository ?

olivier-stasse (commented):
Thanks a lot. The unit tests looks pretty extensive.
It would be nice if a maintainer could allow the CI to run on your PR.

gpio_controllers/package.xml
Wiktor-99: Done

Wiktor-99:
There was some issue with includes I've fixed it so jobs can be run more time.

Wiktor-99:
I've fixed docks and pre-commit jobs other two jobs failed due to  pid controller and ackermann steering controller. 
I believe that now CI should be green.

Wiktor-99:
@christophfroehlich could you rerun the CI?

Wiktor-99:
> @christophfroehlich could you rerun the CI?

It appears that the CI is unstable and is failing on the diff drive and pid controller.

christophfroehlich:
> > @christophfroehlich could you rerun the CI?
> 
> It appears that the CI is unstable and is failing on the diff drive and pid controller.

Binary builds due to a recent API breaking change. Semi-binary seems to be happy.

Wiktor-99:
> > > @christophfroehlich could you rerun the CI?
> > 
> > 
> > It appears that the CI is unstable and is failing on the diff drive and pid controller.
> 
> Binary builds due to a recent API breaking change. Semi-binary seems to be happy.

Okay, so what steps should be taken to merge the pr?

christophfroehlich:
Someone has to find time to test and review it..

christophfroehlich (changes requested):
Sorry for the late feedback.

Great work, I successfully tested this controller, see the linked PR to the demos. I made some changes to the docs and minor changes to the code wrt. error handling and parameters. The only breaking changes: I renamed the `ports` parameter to `interfaces` to be aligned with the naming from the joint_state_broadcaster.

The only thing I'd like to discuss: Is it ok to assume that every command interface has a state interface with the same name? If this is not the case, the current implementation blows up:

`[ros2_control_node-1] [ERROR] [1730671224.262784257] [controller_manager]: Caught exception of type : St13runtime_error while claiming the state interfaces. Can't activate controller 'gpio_controller': State interface with key 'flange_vacuum/vacuum' does not exist
`

I suggest one of the options:
- make the state interfaces parameterizable, if parameter is empty then publish all (like joint_state_broadcaster does, but only for the configured gpios)
- just use state interfaces existing for the command ifs names, but continue gracefully
- add a validation check to have a more user-friendly error

Wiktor-99:
Thanks for feedback! 

Renaming `ports` to `interfaces` is reasonable.

I followed first options and implement parameterizable state interfaces. Currently you can configure command interfaces and state interfaces separately. When no state interfaces are specified then controller fetches them from urdf (only for configured gpios).  With such implementations you don't have to specify command or  state interfaces if you want to broadcast all states of configured gpios.

I've tested it with mentioned example and UTs and seems to work.

Wiktor-99:
I've fixed it. Now the CI no should be green.

christophfroehlich (approved):
Thanks a lot. I tested this again and updated the demo w.r.t. of the gpio_states topic.

Wiktor-99:
Great! I don't know code of conduct but if I can I could help with reviewing linked example.

gpio_controllers/doc/userdoc.rst
```diff
@@ -0,0 +1,55 @@
+.. _gpio_controllers_userdoc:
+
+gpio_controllers
+=====================
+
+This is a collection of controllers for hardware interfaces of type GPIO (``<gpio>`` tag in the URDF).
```
saikishor, line 6:
```suggestion
A controller for hardware interfaces of type GPIO (``<gpio>`` tag in the URDF) to be able to command the configured GPIOs and as well to publish their state
```

saikishor (changes requested):
Just few minor suggested changes from my side

gpio_controllers/doc/userdoc.rst
christophfroehlich: I think this was OK: gpio_controllers follow the structure of other collection packages we have (effort_controller), where only a single controller is implemented yet (gpio_controllers/gpio_command_controller, effort_controllers/JointGroupEffortController)

saikishor: Gotcha! 👍🏾

christophfroehlich:
@saikishor raised a legitimate question: Should we allow zero command interfaces for this controller (what happens actually then?)? or should we set the limit size>0 here, and create a gpio_controllers/GpioStateBroadcaster for broadcasting states only?

christophfroehlich:
> Great! I don't know code of conduct but if I can I could help with reviewing linked example.

Your review is very welcome!

Wiktor-99:
When we don't configure any command interface then controller will just behave as broadcaster and will publish state of configured gpios (explicitly configured or taken from urdf). Probably we  should't create `commands` topic it might fixed.

If we want to separate to controller/broadcaster we could fallback to initial version and use same interface name for commands and states and then add broadcaster.

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -0,0 +1,412 @@
...
+  }
+
+  store_command_interface_types();
+  store_state_interface_types();
+
+  if (command_interface_types_.empty() && state_interface_types_.empty())
+  {
+    RCLCPP_ERROR(get_node()->get_logger(), "No command or state interfaces are configured");
+    return CallbackReturn::ERROR;
+  }
```
saikishor, line 82: If we are having both empty, we are not publishing all the interfaces from the URDF now right?

Wiktor-99: No, we do. The state_interface_types variable stores data from the configuration file, and if state interfaces are empty, it tries to fetch it from urdf. We'll enter this branch when no state and command interfaces are set in config file, and no state interface has been found in urdf.

saikishor: Understood!

gpio_controllers/doc/userdoc.rst
```diff
@@ -0,0 +1,58 @@
...
+Parameters
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+This controller uses the `generate_parameter_library <https://github.com/PickNikRobotics/generate_parameter_library>`_ to handle its parameters. The parameter `definition file located in the src folder <https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/gpio_controllers_/src/gpio_command_controller_parameters.yaml>`_ contains descriptions for all the parameters used by the controller.
+
+.. generate_parameter_library_details::
+  ../src/gpio_command_controller_parameters.yaml
+
+The controller expects at least one GPIO interface and the corresponding command interface names or state interface.
+Command and state interfaces are optional.
```
saikishor, line 27:
```suggestion
The controller expects at least one GPIO interface and the corresponding command interface names or state interface. However, these Command and State interfaces are optional. The controller behaves as a broadcaster when no Command Interface is present, thereby publishing the configured GPIO state interfaces if set, else the one present in the URDF.
```

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+  gpios_command_subscriber_ = get_node()->create_subscription<CmdType>(
+    "~/commands", rclcpp::SystemDefaultsQoS(),
+    [this](const CmdType::SharedPtr msg) { rt_command_ptr_.writeFromNonRT(msg); });
```
saikishor, line 86:
```suggestion
  if(!command_interface_types_.empty())
  {
    gpios_command_subscriber_ = get_node()->create_subscription<CmdType>(
      "~/commands", rclcpp::SystemDefaultsQoS(),
      [this](const CmdType::SharedPtr msg) { rt_command_ptr_.writeFromNonRT(msg); });
  }
```

saikishor (changes requested):

christophfroehlich:
One last comment @saikishor brought up: Using [DynamicJointStates](https://github.com/ros-controls/control_msgs/blob/master/control_msgs/msg/DynamicJointState.msg) for the commands feels semantically more than wrong here. Should we add a copy and name it more appropriate? `DynamicResourceValues`or something better. Even for the broadcaster it is questionable.

Wiktor-99:
Yes, I totally agree with you. Speaking of naming I would propose to add new message `DynamicComponentValues`, resource sounds like it could be acquired. 

I've posted pr in the control_msgs repo https://github.com/ros-controls/control_msgs/pull/155. When it merges, I'll align this pr.

saikishor:
@Wiktor-99 @christophfroehlich may be this PR is also missing dependency in the ros2_controllers metapkg

christophfroehlich:
> @Wiktor-99 @christophfroehlich may be this PR is also missing dependency in the ros2_controllers metapkg

you are right!

Wiktor-99:
Yep, it had been missing. I've aligned it to use DynamicInterfaceGroupValues and fixed dependency in the meta pkg.

gpio_controllers/doc/userdoc.rst
```diff
@@ -0,0 +1,57 @@
...
+Description of controller's interfaces
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- ``/<controller_name>/gpio_states`` [``control_msgs/msg/DynamicJointState``]: Publishes all state interfaces of the given GPIO interfaces.
+- ``/<controller_name>/commands`` [``control_msgs/msg/DynamicJointState``]:  A subscriber for configured command interfaces.
+
+
+Parameters
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+This controller uses the `generate_parameter_library <https://github.com/PickNikRobotics/generate_parameter_library>`_ to handle its parameters. The parameter `definition file located in the src folder <https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/gpio_controllers_/src/gpio_command_controller_parameters.yaml>`_ contains descriptions for all the parameters used by the controller.
```
christophfroehlich, line 21:
```suggestion
This controller uses the `generate_parameter_library <https://github.com/PickNikRobotics/generate_parameter_library>`_ to handle its parameters. The parameter `definition file located in the src folder <https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/gpio_controllers/src/gpio_command_controller_parameters.yaml>`_ contains descriptions for all the parameters used by the controller.
```

gpio_controllers/doc/userdoc.rst
```diff
@@ -0,0 +1,57 @@
...
+
+.. generate_parameter_library_details::
+  ../src/gpio_command_controller_parameters.yaml
+
+The controller expects at least one GPIO interface and the corresponding command interface names or state interface. However, these Command and State interfaces are optional. The controller behaves as a broadcaster when no Command Interface is present, thereby publishing the configured GPIO state interfaces if set, else the one present in the URDF.
+
+.. note::
+
+  When no state interface is provided in the param file, the controller will try to use state_interfaces from ros2_control's config placed in the URDF for configured gpio interfaces.
+  However, Command interfaces will not be configured based on the available URDF setup.
```
christophfroehlich, line 31:
```suggestion
  However, command interfaces will not be configured based on the available URDF setup.
```

christophfroehlich (changes requested):
I once again tested it successfully with the demo PR. One last request from my side: please add a new section to the [release notes](https://github.com/ros-controls/ros2_controllers/blob/master/doc/release_notes.rst).

christophfroehlich (approved):
Perfect, thanks!

gpio_controllers/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
+cmake_minimum_required(VERSION 3.8)
+project(gpio_controllers)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
```
saikishor, line 10:
```suggestion
  add_compile_options(-Wall -Wextra -Wpedantic -Werror=conversion -Werror=unused-but-set-variable -Werror=return-type -Werror=shadow -Werror=format -Werror=range-loop-construct)
```

Can we test with these parameters, if the code is compiling or not?
@christophfroehlich @Wiktor-99

saikishor: These are the standard flags we tend to have across all the packages

gpio_controllers/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
+cmake_minimum_required(VERSION 3.8)
+project(gpio_controllers)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic -Werror=conversion -Werror=unused-but-set-variable -Werror=return-type -Werror=shadow -Werror=format -Werror=range-loop-construct -Werror=missing-braces)
```
christophfroehlich, line 10:
```suggestion
  add_compile_options(-Wall -Wextra -Wpedantic -Werror=conversion -Werror=unused-but-set-variable
  -Werror=return-type -Werror=shadow -Werror=format  -Werror=range-loop-construct
  -Werror=missing-braces)
```

christophfroehlich:
@Wiktor-99 could you please review https://github.com/ros-controls/ros2_control_demos/pull/627 ?

Wiktor-99:
@christophfroehlich  Yep I'll do it, can't promise I'll do it today but tomorrow seems feasible.

PR #1256, opened 2024-08-20
Fixes tests to work with use_global_arguments NodeOptions parameter 

saikishor:
Needs: https://github.com/ros-controls/ros2_control/pull/1701

Fixes: https://github.com/ros-controls/ros2_controllers/issues/1252
Fixes: https://github.com/ros-controls/ros2_controllers/issues/1253
Fixes: https://github.com/ros-controls/ros2_controllers/issues/1254
Fixes: https://github.com/ros-controls/ros2_controllers/issues/1255

### DIFF
--- a/ackermann_steering_controller/CMakeLists.txt
+++ b/ackermann_steering_controller/CMakeLists.txt
@@ -58,10 +58,8 @@ if(BUILD_TESTING)
   find_package(ros2_control_test_assets REQUIRED)
 
 
-  add_rostest_with_parameters_gmock(test_load_ackermann_steering_controller
-    test/test_load_ackermann_steering_controller.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/test/ackermann_steering_controller_params.yaml
-  )
+  add_definitions(-DTEST_FILES_DIRECTORY="${CMAKE_CURRENT_SOURCE_DIR}/test")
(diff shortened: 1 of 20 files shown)


christophfroehlich (commented):
why don't we need the changes for admittance controller and range_sensor_broadcaster?
the tests succeed, which means that they load successfully without parameters? 
Shouldn't we check if they initialized properly?
```
1: [ERROR] [1724184278.166220317] [load_admittance_controller]: Exception thrown during init stage with message: parameter 'joints' is not initialized 
1: 
1: [ERROR] [1724184278.166266164] [test_controller_manager]: Could not initialize the controller named 'load_admittance_controller'
1: [       OK ] TestLoadAdmittanceController.load_controller (40 ms)
```

range_sensor_broadcaster initializes fine, I guess it doesn't need non-default parameters.

saikishor:
@christophfroehlich it's probable that they are returning successful even without parameters. That's the only explanation. I could try fixing them tomorrow.

christophfroehlich:

```
```

saikishor:
Ok. I'm outside right now. I'll try to do that sooner

saikishor:
@christophfroehlich applied the fix to both the controllers as well

christophfroehlich (approved):
LGTM

bmagyar (approved):
Thank you!

PR #1259, opened 2024-08-21
Fix bug for displaying all controllers

fmrico:
Hi,

When using the rqt_joint_trajectory_controller, not all the controllers are shown on the controller button. This is because the list of limits was not accumulated, and only the last one was considered. This PR fixes this, showing all the valid controllers.

I hope it helps
Francisco

PR #1264, opened 2024-08-22
Also test if python files were changed

christophfroehlich:
I realized that the CI jobs are not running in https://github.com/ros-controls/ros2_controllers/pull/1259

Not that we would have any tests for rqt_jtc, but just in case we get some.


bmagyar (approved):
wunderbar!

PR #1269, opened 2024-08-23
[Joint State Broadcaster] Move urdf::Model initialization from on_activate to on_configure to enhance real-time performance

TakashiSato:
This PR aims to enhance real-time performance during the activation of the Joint State Broadcaster, which was introduced in #1233.

The `urdf::Model` initialization ([here](https://github.com/ros-controls/ros2_controllers/blob/20f6f0bba9a73cb49b4547cc5316ee058a2298aa/joint_state_broadcaster/src/joint_state_broadcaster.cpp#L248-L249)), previously executed within the on_activate function, incurred a substantial amount of time, leading to significant jitter in the Controller Manager's real-time loop.

To address this, the initialization of the `urdf::Model` has been relocated from on_activate to on_configure, as it can be effectively performed beforehand. In my testing environment, this modification has resulted in a remarkable improvement, accelerating the on_activate process by approximately 40 times (reducing the time from 5ms to 0.12ms).

### DIFF
--- a/joint_state_broadcaster/include/joint_state_broadcaster/joint_state_broadcaster.hpp
+++ b/joint_state_broadcaster/include/joint_state_broadcaster/joint_state_broadcaster.hpp
@@ -27,6 +27,7 @@
 #include "joint_state_broadcaster_parameters.hpp"
 #include "realtime_tools/realtime_publisher.h"
 #include "sensor_msgs/msg/joint_state.hpp"
+#include "urdf/model.h"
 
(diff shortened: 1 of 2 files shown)


christophfroehlich (approved):
Makes sense for me, thanks!

saikishor (approved):
Sure. Makes sense.

Thank you for the PR

bmagyar (approved):
great catch, thank you!

Issue #1239, opened 2024-08-02
publisher_forward_position_controller uses deprecated parameter API


christophfroehlich:
Compile the demos and run `ros2 launch ros2_control_demo_example_1 test_forward_position_controller.launch.py`

> [publisher_forward_position_controller-1] /opt/ros/rolling/lib/python3.12/site-packages/rclpy/node.py:481: UserWarning: when declaring parameter named 'pos3', declaring a parameter only providing its name is deprecated. You have to either:
> [publisher_forward_position_controller-1]       - Pass a name and a default value different to "PARAMETER NOT SET" (and optionally a descriptor).
> [publisher_forward_position_controller-1]       - Pass a name and a parameter type.
> [publisher_forward_position_controller-1]       - Pass a name and a descriptor with `dynamic_typing=True

christophfroehlich:
fyi: The publisher_joint_trajectory_controller doesn't throw warnings

PR #1280, opened 2024-09-01
Fix deprecation warning in paramater declaration

kumar-sanjeeev:
Updated the paramater declaration in `publisher_forward_position_controller` to resolve a deprecation warning when running the demo using  `ros2 launch ros2_control_demo_example_1 test_forward_position_controller.launch.py`.

Changes Made:
- Replaced `self.declare_parameter(name)` with `self.declare_parameter(name, descriptor=ParameterDescriptor(dynamic_typing=True))`

Fixes: #1239

### DIFF
--- a/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
+++ b/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
@@ -17,6 +17,7 @@
 
 import rclpy
 from rclpy.node import Node
+from rcl_interfaces.msg import ParameterDescriptor
 
 from std_msgs.msg import Float64MultiArray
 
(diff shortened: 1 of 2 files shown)


christophfroehlich (commented):
Thanks for your contribution. 
One question: Why shouldn't we specify the type of being a `PARAMETER_STRING` instead of activating `dynamic_typing`?

kumar-sanjeeev:
Hi @christophfroehlich 

Yes, that is another way to suppress the warnings. I opted for `dynamic_typing` because  same is implemented in the `publisher_joint_trajectory_controller`.

I also tried specifying the type, and it is doing the job: `self.declare_parameter(name, rclpy.Parameter.Type.STRING)`.

destogl:
> As you suggested, I tried using the type during the declaration of parameters, and it works as well: `self.declare_parameter(name, rclpy.Parameter.Type.DOUBLE_ARRAY)`

Let's use this exact approach. As rather to use `explicit` instead of `implicit`

bmagyar (approved):

christophfroehlich (approved):
LGTM!

saikishor (approved):
LGTM

PR #1285, opened 2024-09-10
fix(steering-odometry): handle infinite turning radius properly

reinzor:


steering_controllers_library/src/steering_odometry.cpp
```diff
@@ -133,11 +133,17 @@ double SteeringOdometry::get_linear_velocity_double_traction_axle(
   const double steer_pos)
 {
   double turning_radius = wheelbase_ / std::tan(steer_pos);
+  double vel_wheel_r = right_traction_wheel_vel * wheel_radius_;
+  double vel_wheel_l = left_traction_wheel_vel * wheel_radius_;
```
saikishor, line 137:
```suggestion
  const double vel_wheel_r = right_traction_wheel_vel * wheel_radius_;
  const double vel_wheel_l = left_traction_wheel_vel * wheel_radius_;
```

steering_controllers_library/src/steering_odometry.cpp
```diff
@@ -133,11 +133,17 @@ double SteeringOdometry::get_linear_velocity_double_traction_axle(
...
+
+  if (std::isinf(turning_radius))
```
saikishor, line 139: Will we need to handle cases of NaN as well here?

steering_controllers_library/src/steering_odometry.cpp
```diff
@@ -133,11 +133,17 @@ double SteeringOdometry::get_linear_velocity_double_traction_axle(
...
+    return (vel_wheel_r + vel_wheel_l) * 0.5;
+  }
+
   // overdetermined, we take the average
-  double vel_r = right_traction_wheel_vel * wheel_radius_ * turning_radius /
-                 (turning_radius + wheel_track_ * 0.5);
-  double vel_l = left_traction_wheel_vel * wheel_radius_ * turning_radius /
-                 (turning_radius - wheel_track_ * 0.5);
+  double vel_r = vel_wheel_r * turning_radius / (turning_radius + wheel_track_ * 0.5);
+  double vel_l = vel_wheel_l * turning_radius / (turning_radius - wheel_track_ * 0.5);
```
saikishor, line 146:
```suggestion
  const double vel_r = vel_wheel_r * turning_radius / (turning_radius + wheel_track_ * 0.5);
  const double vel_l = vel_wheel_l * turning_radius / (turning_radius - wheel_track_ * 0.5);
```

saikishor (changes requested):

steering_controllers_library/src/steering_odometry.cpp
reinzor: I am not sure; there are already input checks here: https://github.com/ros-controls/ros2_controllers/blob/c08bdab049cc090df0643d3399bf53862f4571be/tricycle_steering_controller/src/tricycle_steering_controller.cpp#L68 . I can update the PR if desired but what should be the behavior when the input is not finite, raise an exception?

saikishor: If the checks are already done. It is fine. Thank you!

bmagyar (approved):
Thank you!

PR #1288, opened 2024-09-12
fix(steering-odometry): convert twist to steering angle

reinzor:
The logic in this function was incorrect, if we divide by `0.`, the input of the `atan` goes to `inf` which should result in `PI/2` instead of `0.`. What should be handled instead are `nan` values as an input of the `atan` function that can happen if we input `0./0.`.


christophfroehlich (approved):
This makes sense, thanks!

bmagyar (approved):
also, correct math is always correct

unless you do some crazy physics but oh well... :D 

thank you!

PR #1289, opened 2024-09-13
fix(timeout): do not reset steer wheels to 0. on timeout

reinzor:
Instead of setting the incoming twist command velocity to zero, we should only set the wheel commands to zero without altering the steering angles in order to avoid in-place steering of the wheels.

### DIFF
--- a/steering_controllers_library/test/test_steering_controllers_library.cpp
+++ b/steering_controllers_library/test/test_steering_controllers_library.cpp
@@ -133,8 +133,8 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
   {
     EXPECT_TRUE(std::isnan(interface));
   }
-  EXPECT_TRUE(std::isnan((*(controller_->input_ref_.readFromNonRT()))->twist.linear.x));
-  EXPECT_TRUE(std::isnan((*(controller_->input_ref_.readFromNonRT()))->twist.angular.z));
+  ASSERT_EQ((*(controller_->input_ref_.readFromRT()))->twist.linear.x, TEST_LINEAR_VELOCITY_X);
+  ASSERT_EQ((*(controller_->input_ref_.readFromNonRT()))->twist.angular.z, TEST_ANGULAR_VELOCITY_Z);
 
   EXPECT_TRUE(std::isnan(controller_->reference_interfaces_[0]));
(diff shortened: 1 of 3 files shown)


christophfroehlich (changes requested):
This makes sense to me, thanks for the contribution.
Could you please have a look on the failing tests and change the assertions accordingly?
Please add also a short description in the [release_notes](https://github.com/ros-controls/ros2_controllers/blob/master/doc/release_notes.rst).

bmagyar (approved):
thank you!

saikishor (approved):

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -133,19 +133,22 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
   {
     EXPECT_TRUE(std::isnan(interface));
   }
-  EXPECT_TRUE(std::isnan((*(controller_->input_ref_.readFromNonRT()))->twist.linear.x));
-  EXPECT_TRUE(std::isnan((*(controller_->input_ref_.readFromNonRT()))->twist.angular.z));
+  ASSERT_EQ((*(controller_->input_ref_.readFromRT()))->twist.linear.x, TEST_LINEAR_VELOCITY_X);
```
christophfroehlich, line 136:
```suggestion
  ASSERT_EQ((*(controller_->input_ref_.readFromNonRT()))->twist.linear.x, TEST_LINEAR_VELOCITY_X);
```

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -174,19 +177,22 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
...
```
christophfroehlich, line 180:
```suggestion
  ASSERT_EQ((*(controller_->input_ref_.readFromNonRT()))->twist.linear.x, TEST_LINEAR_VELOCITY_X);
```

christophfroehlich (changes requested):
Thanks for the follow-up. Just two minor comments

christophfroehlich (approved):

christophfroehlich:
@mergifyio backport humble

PR #1306, opened 2024-10-07
[JTC] Sample at t + dT instead of the beginning of the control cycle

fmauch:
As mentioned in https://github.com/ros-controls/ros2_controllers/pull/1191#issuecomment-2284668456 and discussed in #697 the JTC currently samples the trajectory at the time given to the `update(time, period)` method. However, this actually is the beginning of the control cycle and it would make more sense to sample the setpoint for the end of the control cycle, resulting in `time  + controller_update_period`. This PR implements that.

I obviously had to update a couple of tests on the way, since with changing the sampling point, the resulting command can be significantly different.


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -194,12 +196,15 @@ controller_interface::return_type JointTrajectoryController::update(
         traj_external_point_ptr_->set_point_before_trajectory_msg(
           time, state_current_, joints_angle_wraparound_);
       }
+      traj_external_point_ptr_->sample(
+        // Sample once at the beginning to establish the start time if none was given
+        time, interpolation_method_, state_desired_, start_segment_itr, end_segment_itr);
```
saikishor, line 201:
```suggestion
      // Sample once at the beginning to establish the start time if none was given
      traj_external_point_ptr_->sample(
        time, interpolation_method_, state_desired_, start_segment_itr, end_segment_itr);
```

Just to avoid comment in the middle of the function arguments

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -194,12 +196,15 @@ controller_interface::return_type JointTrajectoryController::update(
...
     }
 
-    // find segment for current timestamp
-    TrajectoryPointConstIter start_segment_itr, end_segment_itr;
+    // Sample setpoint for next control cycle
     const bool valid_point = traj_external_point_ptr_->sample(
-      time, interpolation_method_, state_desired_, start_segment_itr, end_segment_itr);
+      time + update_period_, interpolation_method_, state_desired_, start_segment_itr,
+      end_segment_itr);
```
saikishor, line 207: Above you are doing the initial sampling and immediately overriding with this future one. Is this intended?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -279,7 +284,7 @@ controller_interface::return_type JointTrajectoryController::update(
             tmp_command_[i] = (state_desired_.velocities[i] * ff_velocity_scale_[i]) +
                               pids_[i]->computeCommand(
                                 state_error_.positions[i], state_error_.velocities[i],
-                                (uint64_t)period.nanoseconds());
+                                (uint64_t)update_period_.nanoseconds());
```
saikishor, line 287: shouldn't this use the period itself?
I think the `period` is more correct than the `update_period_`

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -961,6 +966,15 @@ controller_interface::CallbackReturn JointTrajectoryController::on_activate(
...
 
+  {
+    if (get_update_rate() == 0)
+    {
+      throw std::runtime_error("Controller's update rate is set to 0. This should not happen!");
+    }
+    update_period_ =
+      rclcpp::Duration(0.0, static_cast<uint32_t>(1.0e9 / static_cast<double>(get_update_rate())));
+  }
+
```
saikishor, line 977: Maybe this can go inside the `on_configure`? What do you think? @fmauch

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -644,7 +646,7 @@ TEST_P(TrajectoryControllerTestParameterized, position_error_not_angle_wraparoun
   size_t n_joints = joint_names_.size();
 
   // send msg
-  constexpr auto FIRST_POINT_TIME = std::chrono::milliseconds(250);
+  constexpr auto FIRST_POINT_TIME = std::chrono::milliseconds(300);
```
saikishor, line 649: If the control period is 100 milliseconds, shouldn't it be 350?
Am i missing something here?

saikishor (commented):
Look good. I've left some questions 
I'll continue reviewing after we discuss this part.

Thanks for your great work @fmauch

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: Yes. The sample call above is - as the comment says - only to establish the **actual** start time of the trajectory. However, also in the first control cycle when we started the trajectory we want to sample at $t+\Delta t$ to generate the commands.

This should be probably refactored in the trajectory representation that the start time would not implicitly be specified by the first sample, but I thought this might not be the right PR to address that...

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: It would usually not make that big of a difference, but isn't the goal here, as well, to generate a command to reach `state_desired` at the end of the current control cycle (which will be `update_period_` in the future).

joint_trajectory_controller/src/joint_trajectory_controller.cpp
fmauch: That seems reasonable, yes. I will change that.

joint_trajectory_controller/test/test_trajectory_controller.cpp
fmauch: Why? Would it be desired to set a target time between two control cycles?

With a control period of 100 ms we would call `update` at $t_0 + 100ms, t_0 + 200ms, t_0+300ms$... So, in the update call at $t=t_0+200ms$ it should generate the command to reach the setpoint which is what we wait for below.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
christophfroehlich: Actually, I agree with Sai here: This is not related to the trajectory sampling but only for calculating the feedback control law (in fact, only for integrating the error for the integral part of the command).

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -1578,9 +1600,20 @@ TEST_P(TrajectoryControllerTestParameterized, test_ignore_partial_old_trajectory
...
+
+  if (traj_controller_->has_position_command_interface())
+  {
+    //  Check that we reached end of points_old[0] trajectory
+    auto state_feedback = traj_controller_->get_state_feedback();
+    for (size_t i = 0; i < expected_actual.positions.size(); ++i)
+    {
+      EXPECT_NEAR(expected_actual.positions[i], state_feedback.positions[i], 0.01);
+    }
+  }
```
christophfroehlich, line 1616: Is there a reason that you don't call waitAndCompareState() here? why is now `traj_controller_->has_position_command_interface()` necessary, but wasn't before?

fmauch: If you look at the implementation of `waitAndCompareState`, you'll see that it is almost equivalent:

```c++
    traj_controller_->wait_for_trajectory(executor);
    auto end_time = updateControllerAsync(controller_wait_time, start_time);

    // get states from class variables
    auto state_feedback = traj_controller_->get_state_feedback();
    auto state_reference = traj_controller_->get_state_reference();

    for (size_t i = 0; i < expected_actual.positions.size(); ++i)
    {
      SCOPED_TRACE("Joint " + std::to_string(i));
      // TODO(anyone): add checking for velocities and accelerations
      if (traj_controller_->has_position_command_interface())
      {
        EXPECT_NEAR(expected_actual.positions[i], state_feedback.positions[i], allowed_delta);
      }
    }

    for (size_t i = 0; i < expected_desired.positions.size(); ++i)
    {
      SCOPED_TRACE("Joint " + std::to_string(i));
      // TODO(anyone): add checking for velocities and accelerations
      if (traj_controller_->has_position_command_interface())
      {
        EXPECT_NEAR(expected_desired.positions[i], state_reference.positions[i], allowed_delta);
      }
    }
```

* I preferred giving a specific update rate to the `updateControllerAsync` function that matches the actual update rate.
* The `waitAndCompareState` function only checks positions, hence I also only checked positions. So, it wasn't necessary before, because it was *inside* `waitAndCompareState`.
* More importantly, `waitAndCompareState` checks for the actual position and the generated position command. However, due to spline interpolation we don't necessarily know the generated command. Assuming `expected_desired = expected_actual;` is simply wrong as long as the robot is in motion (which it is, continuing to the second waypoint). It did work, more or less by accident, since the robot is initialized at `{1.2, 2.1, 3.1}` and moves to `{2.0, 3.0, 4.0}` in the first segment. Hence, the trajectory "arrives" at the setpoint quite slowly. However, in the next segment (which has the same time), it is supposed to go to `{4.0, 5.0, 6.0}`, so it has to move much faster in that segment. Before my change, sampling was basically moved back one cycle, so the desired state was 2.0 for joint1 which matches the desired state and the actual state was something right before that, which was in the range of the 0.1 tolerance (which is probably the reason why it was so large). However, after the changes from this PR, the **actual** state will be at 2.0 at this time and the **desired** will be on the way towards 4.0 already. (In fact, it will be at 2.3599999999999999). If we move sampling back by one control cycle to `rclcpp::Duration(delay) - rclcpp::Duration::from_seconds(0.1)` the desired state matches 2.0 exactly, while the actual state is 1.982, which falls into the 0.1 range. Since the comment to that check says `Check that we reached end of points_old[0]trajectory` I thought it would be better to check exactly that with a small tolerance instead of having to increase the tolerance to actually not fail a check that doesn't only check what we want to know at that point. I hope, that paragraph makes sense...

fmauch:
The changes from https://github.com/ros-controls/ros2_control/pull/1788 require revisiting the tests from this. I expect, that tests on the master branch will fail, as well, however, since I had to set correct rates all over the place within this PR.

joint_trajectory_controller/test/test_trajectory_controller.cpp
christophfroehlich: Ignore the question of `has_position_command_interface`, I have overlooked this in the implementation of `waitAndCompareState`.
The rest seems to be reasonable, thanks!

christophfroehlich (approved):

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: Thank you for the clarification

saikishor (approved):
LGTM

christophfroehlich (changes requested):
I just read once more the linked issue #697 and I think he is right in this point:

> However JTC in ros_controllers is comparing current state with future state (at T + dt), and this seems wrong. This will always lead to a tracking or goal error what ever perfect robot (mirrored cmd/state ifs) there is.

This is exactly what is done in this PR here?

christophfroehlich:
> In [0d00727](https://github.com/ros-controls/ros2_controllers/pull/1306/commits/0d00727ff1d3570ee6a67cb05df6953e07998e6e) I implemented using the trajectory sample point at time `T` as the `desired_state`. This is used for checking the tolerances, generating the feedback's desired state and error and to compute the PID value for the closed loop controller. I am unsure about the latter, as it was also discussed in #697.

I think this is a proper way in handling this, and it's a minimal change in the current behavior by now.
But I'm wondering if there is a problem with #1297: It may happen that the sampling at t_2 of the following update method will sample at a time before t_1+dt?

fmauch:
> > In [0d00727](https://github.com/ros-controls/ros2_controllers/pull/1306/commits/0d00727ff1d3570ee6a67cb05df6953e07998e6e) I implemented using the trajectory sample point at time `T` as the `desired_state`. This is used for checking the tolerances, generating the feedback's desired state and error and to compute the PID value for the closed loop controller. I am unsure about the latter, as it was also discussed in #697.
> 
> I think this is a proper way in handling this, and it's a minimal change in the current behavior by now. But I'm wondering if there is a problem with #1297: It may happen that the sampling at t_2 of the following update method will sample at a time before t_1+dt?

Yes, this may definitively happen, especially once we actually get to time-scaling as it is my overall target with this. I'm not entierly familiar with #1297, so I'll have to read into that.

fmauch:
> Yes, this may definitively happen, especially once we actually get to time-scaling as it is my overall target with this. I'm not entierly familiar with #1297, so I'll have to read into that.

Coming back to this: Yes, this is definitively a problem. One solution could be to make the index progression in the `sample` method optional such that it could only be moving forwards for the sampling at time T, but not at sampling for T+dT.

christophfroehlich (approved):
That's fine, I also thought about fixing it like this. ABI break should be fine for the trajectory class.

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -94,13 +96,16 @@ class Trajectory
    *      description above.
    * \param[out] end_segment_itr Iterator to the end segment for given \p sample_time. See
    *      description above.
+   * \param[in] progress_search_start_index If set to true, the next sample call will start
+   *      searching in the trajectory at the index of this call's result.
```
MarqRazz, line 100: This name makes me think the variable is an index. What do you think about renaming this to something like `search_monotonically_increasing`?

MarqRazz (approved):
Overall this looks good, just a few minor suggestions.

saikishor (approved):

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
bmagyar:
```suggestion
   * \param[in] search_monotonically_increasing If set to true, the next sample call will start
   *      searching in the trajectory at the index of this call's result.
```

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -69,8 +69,10 @@ class Trajectory
    * acceleration respectively. Deduction assumes that the provided velocity or acceleration have to
    * be reached at the time defined in the segment.
    *
-   * This function assumes that sampling is only done at monotonically increasing \p sample_time
-   * for any trajectory.
+   * This function by default assumes that sampling is only done at monotonically increasing \p
+   * sample_time for any trajectory. That means, it will only search for a point matching the sample
+   * time after the point it has been called before. If this isn't desired, set \p
+   * progress_search_start_index to false.
```
bmagyar, line 75:
```suggestion
   * search_monotonically_increasing to false.
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -91,7 +91,8 @@ bool Trajectory::sample(
   const rclcpp::Time & sample_time,
   const interpolation_methods::InterpolationMethod interpolation_method,
   trajectory_msgs::msg::JointTrajectoryPoint & output_state,
-  TrajectoryPointConstIter & start_segment_itr, TrajectoryPointConstIter & end_segment_itr)
+  TrajectoryPointConstIter & start_segment_itr, TrajectoryPointConstIter & end_segment_itr,
+  const bool progress_search_start_index)
```
bmagyar, line 95:
```suggestion
  const bool search_monotonically_increasing)
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -176,7 +177,10 @@ bool Trajectory::sample(
       }
       start_segment_itr = begin() + i;
       end_segment_itr = begin() + (i + 1);
-      last_sample_idx_ = i;
+      if (progress_search_start_index)
```
bmagyar, line 180:
```suggestion
      if (search_monotonically_increasing)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -194,13 +196,20 @@ controller_interface::return_type JointTrajectoryController::update(
         traj_external_point_ptr_->set_point_before_trajectory_msg(
           time, state_current_, joints_angle_wraparound_);
       }
+      // Sample once at the beginning to establish the start time if none was given
+      traj_external_point_ptr_->sample(
+        time, interpolation_method_, state_desired_, start_segment_itr, end_segment_itr);
```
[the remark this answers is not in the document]
fmauch: I think, you're right! With sampling at `time` for the state error, this is redundant.

bmagyar:
Thanks everyone!

bmagyar (approved):

PR #1311, opened 2024-10-09
Implement new PoseBroadcaster controller

RobertWilbrandt:
This PR implements a new broadcasting controller based on the `PoseSensor` semantic component introduced in https://github.com/ros-controls/ros2_control/pull/1775.

As described in the `PoseSensor` PR, my primary use-case for this is the publishing of tcp poses from robot arms. This can be more accurate than calculating forward kinematics, as robot calibrations etc. can be considered by the robot. Settings on the robot side concerning tool offsets etc. can also be included by this (at least for UR and kuka RSI). The ROS 1 `ur_robot_driver` used to publish this transform directly from within the hardware interface, but in ROS 2 there is not yet a standard way for providing this information. A PR for the UR ROS 2 driver already exists at https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1108.

This controller publishes poses directly as `geometry_msgs::msg::PoseStamped`s and as tf transforms (can be disabled).

### DIFF
--- a/pose_broadcaster/doc/userdoc.rst
+++ b/pose_broadcaster/doc/userdoc.rst
@@ -1,0 +1,27 @@
+:github_url: https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/pose_broadcaster/doc/userdoc.rst
+
+.. _pose_broadcaster_userdoc:
+
+Pose Broadcaster
+--------------------------------
+Broadcaster for poses measured by a robot or a sensor.
+The published message type is ``geometry_msgs/msg/PoseStamped``.
+
+The controller is a wrapper around the ``PoseSensor`` semantic component (see ``controller_interface`` package).
+
+Parameters
+^^^^^^^^^^^
+This controller uses the `generate_parameter_library <https://github.com/PickNikRobotics/generate_parameter_library>`_ to handle its parameters. The parameter `definition file located in the src folder <https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/pose_broadcaster/src/pose_broadcaster_parameters.yaml>`_ contains descriptions for all the parameters used by the controller.
+
+List of parameters
+=========================
+.. generate_parameter_library_details:: ../src/pose_broadcaster_parameters.yaml
+
+
+An example parameter file
+=========================
+
+An example parameter file for this controller can be found in `the test directory <https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/pose_broadcaster/test/pose_broadcaster_params.yaml>`_:
+
+.. literalinclude:: ../test/pose_broadcaster_params.yaml
+   :language: yaml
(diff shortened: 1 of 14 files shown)


pose_broadcaster/pose_broadcaster.xml
```diff
@@ -0,0 +1,9 @@
+<library path="pose_broadcaster">
+  <class name="pose_broadcaster/PoseBroadcaster"
+         type="pose_broadcaster::PoseBroadcaster"
+         base_class_type="controller_interface::ControllerInterface">
+    <description>
+      This controller publishes a cartesian state as a geometry_msgs/PoseStamped message.
```
fmauch, line 6: **Very** nitpicky, but AFAIK Cartesian is spelled with a capital C.
```suggestion
      This controller publishes a Cartesian state as a geometry_msgs/PoseStamped message.
```

pose_broadcaster/src/pose_broadcaster.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  if (realtime_tf_publisher_ && realtime_tf_publisher_->trylock())
+  {
+    bool do_publish = false;
+    // rlcpp::Time comparisons throw if clock types are not the same
+    if (tf_last_publish_time_.get_clock_type() != time.get_clock_type())
+    {
+      do_publish = true;
+    }
+    else if (!tf_publish_period_ || (tf_last_publish_time_ + *tf_publish_period_ < time))
```
fmauch, line 159: If we use ` < time`, the tf would only be published every second update if the controller rate is exactly the same as the tf publish rate, right?
```suggestion
    else if (!tf_publish_period_ || (tf_last_publish_time_ + *tf_publish_period_ <= time))
```

RobertWilbrandt: Good point, i didn't think of that

pose_broadcaster/doc/userdoc.rst
```diff
@@ -0,0 +1,27 @@
+:github_url: https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/pose_broadcaster/doc/userdoc.rst
+
+.. _pose_broadcaster_userdoc:
+
+Pose Broadcaster
+--------------------------------
+Broadcaster for poses measured by a robot or a sensor.
+The published message type is ``geometry_msgs/msg/PoseStamped``.
```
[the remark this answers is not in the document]
RobertWilbrandt: Good point, i updated this and the controller plugin file to mention that.

fmauch (approved):
Looks good to me! As @RobertWilbrandt wrote, we tested this in the UR driver already.

bmagyar:
Would you see use of this being a chainable controller at the same time?

RobertWilbrandt:
I don't have much experience with chaining controllers, but this sounds like a good extension. I can work on that and extend this PR.

fmauch:
@bmagyar This should probably be backported? The semantic component has already been backported.

christophfroehlich:
@mergifyio backport humble iron

PR #1314, opened 2024-10-13
steering_controllers_library: Add `reduce_wheel_speed_until_steering_reached` parameter

christophfroehlich:
Some features are missing in the library to feature-equally integrate the older tricycle_controller (#850).

`reduce_wheel_speed_until_steering_reached` is one step closer.

The old code
https://github.com/ros-controls/ros2_controllers/blob/97c1e2472f2bbc98a9cb86449b41611a3e298b30/tricycle_controller/src/tricycle_controller.cpp#L173-L189

I modified the function a bit to be continuously differentiable
image

This breaks ABI of the odometry class, but I don't think that lots of people use this directly without the derived controllers.


doc/release_notes.rst
```diff
@@ -57,6 +57,7 @@ steering_controllers_library
 * Changing default int values to double in steering controller's yaml file. The controllers should now initialize successfully without specifying these parameters (`#927 <https://github.com/ros-controls/ros2_controllers/pull/927>`_).
 * A fix for Ackermann steering odometry was added (`#921 <https://github.com/ros-controls/ros2_controllers/pull/921>`_).
 * Do not reset the steering wheels to ``0.0`` on timeout (`#1289 <https://github.com/ros-controls/ros2_controllers/pull/1289>`_).
+* New parameter ``reduce_wheel_speed_until_steering_reached`` was added. If set to true, then the wheel speed(s) is reduced until the steering angle has been reached. Only considered if ``open_loop = false`` (`#1314 <https://github.com/ros-controls/ros2_controllers/pull/1314>`_).
```
destogl, line 60: Do we need this "new", if yes for how long we keep it that way?

destogl (approved):
A small comment about docs, otherwise great!

I agree to break this on Humble

doc/release_notes.rst
christophfroehlich: This is part of the release notes, so I think that "new" is appropriate here?

bmagyar (approved):
Thank you!

PR #1325, opened 2024-10-29
Deprecating tf2 C Headers

CursedRock17:
Related to this [pull request](https://github.com/ros2/geometry2/pull/720) in `geometry2` in which we deprecated the `.h` style headers in favor of `.hpp`.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -26,7 +26,7 @@
 #include "hardware_interface/types/hardware_interface_type_values.hpp"
 #include "lifecycle_msgs/msg/state.hpp"
 #include "rclcpp/logging.hpp"
-#include "tf2/LinearMath/Quaternion.h"
+#include "tf2/LinearMath/Quaternion.hpp"
 
 namespace
 {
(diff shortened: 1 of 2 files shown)


saikishor (commented):
@CursedRock17 is this from rolling on? or does it affect Jazzy as well?

CursedRock17:
This is from rolling onward, there aren't any plans of back porting yet.

saikishor:
> This is from rolling onward, there aren't any plans of back porting yet.

Thank you!

We will have to wait for the original PR to be merged and then we can get back to this. 
In the meantime, Can you add conditioning in the code so that from rolling on it uses the `.hpp`, and in the prior versions it uses the `.h` version?

destogl (dismissed):
Thanks!

saikishor:
@CursedRock17 I changed this PR to draft for now. Once your original PR is merged, please mark it as ready

CursedRock17:
With the recent backports merged, this PR should be good for review.

christophfroehlich:
The backport to [jazzy](https://github.com/ros2/geometry2/commits/jazzy/) hasn't been released yet, and on rolling it seems that they are not yet available on testing repo although the PR on rosdistro was merged. Let's wait until then, then we could merge because the binary-main jobs are failing already because we did the same with realtime_tools headers.

bmagyar (approved):

PR #1331, opened 2024-10-30
[JSB] Fix the behaviour of publishing unavailable state interfaces when they are previously available

saikishor:
Hello!

While reviewing the [Example 13](https://github.com/ros-controls/ros2_control_demos/pull/417), I came across this situation described in the comment : https://github.com/ros-controls/ros2_control_demos/pull/417#discussion_r1823443110 

After debugging, the resource manager properly sets the state interfaces. However, some internal variables are not properly reset, and this results in the publishing of previously available interfaces.  

The proposed fix fixes the issue of properly publishing the state of the known state interfaces alone


bmagyar (approved):
oh mamma mia! that is a good catch!

christophfroehlich (approved):
I tested this right now with example_13: I could reproduce the issue and can confirm that this PR solves it.

PR #1359, opened 2024-11-07
Iron to use kimatics interface from Humble

bmagyar:


christophfroehlich:
@bmagyar or any chance to backport this https://github.com/ros-controls/ros2_controllers/pull/1354?

saikishor (commented):
May be it is better to branch of from last released version of ros2-master, or we can work on a minor changes to the kinematics_interface that way it works on both Iron and the latest distros

christophfroehlich (approved):
Whatever you decide, please change this on the master branch and backport to humble+iron to keep stuff in sync.
fyi: This has to be changed then as well
https://github.com/ros-controls/ros2_control_ci/blob/master/ros_controls.iron.repos

bmagyar:
You guys are correct, we should backport the recent changes

PR #1360, opened 2024-11-07
[JTC] Fix the JTC length_error exceptions in the tests

saikishor:
Fix the exception in JTC tests from time to time 

````
     [INFO] [1730930292.973865185] [test_joint_trajectory_controller]: Using 'splines' interpolation method.
    [INFO] [1730930292.974352669] [test_joint_trajectory_controller]: Action status changes will be monitored at 20.00 Hz.
    terminate called after throwing an instance of 'std::length_error'
      what():  basic_string::_M_create
````

https://github.com/ros-controls/ros2_controllers/actions/runs/11712509515/job/32623343057?pr=1297

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
@@ -26,6 +26,7 @@
 #include "hardware_interface/types/hardware_interface_type_values.hpp"
 #include "joint_trajectory_controller/joint_trajectory_controller.hpp"
 #include "joint_trajectory_controller/tolerances.hpp"
+#include "lifecycle_msgs/msg/state.hpp"
 #include "ros2_control_test_assets/descriptions.hpp"
 
 namespace
--- a/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
@@ -392,7 +392,7 @@ public:
     return traj_controller_->get_node()->activate();
   }
 
-  rclcpp_lifecycle::State DeactivateTrajectoryController()
+  void DeactivateTrajectoryController()
   {
     if (traj_controller_)
     {
@@ -400,12 +400,9 @@ public:
         traj_controller_->get_lifecycle_state().id() ==
         lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
       {
-        const auto new_state = traj_controller_->get_node()->deactivate();
-        EXPECT_EQ(new_state.id(), lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
-        return new_state;
+        traj_controller_->get_node()->deactivate();
       }
     }
-    return traj_controller_->get_lifecycle_state();
   }
 
   static void TearDownTestCase() { rclcpp::shutdown(); }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: joint_trajectory_controller/test/test_trajectory_controller_utils.hpp [hunk 1]
```diff
@@ -391,6 +392,22 @@ class TrajectoryControllerTest : public ::testing::Test
...
+        traj_controller_->get_lifecycle_state().id() ==
+        lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+      {
+        const auto new_state = traj_controller_->get_node()->deactivate();
+        EXPECT_EQ(new_state.id(), lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
+        return new_state;
+      }
+    }
+    return traj_controller_->get_lifecycle_state();
+  }
```
bmagyar, line 409: do we really need to return anything here?

```suggestion
    if (traj_controller_)
    {
      if (
        traj_controller_->get_lifecycle_state().id() ==
        lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
      {
        traj_controller_->get_node()->deactivate();
      }
    }
  }
```

saikishor: You are right. I added this before adding that EXPECT_EQ statement

saikishor: Bence can we leave the `EXPECT_EQ` part? or you prefer not to have it?

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
```diff
@@ -391,6 +392,22 @@ class TrajectoryControllerTest : public ::testing::Test
     return traj_controller_->get_node()->activate();
   }
 
+  rclcpp_lifecycle::State DeactivateTrajectoryController()
```
saikishor, line 395:
```suggestion
  void DeactivateTrajectoryController()
```

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
bmagyar: We could leave it, good point!

saikishor:
> Thanks for the fix :) should we backport this?

May be yes!

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
saikishor: ok

saikishor:
```suggestion
    if (traj_controller_)
    {
      if (
        traj_controller_->get_lifecycle_state().id() ==
        lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
      {
        const auto new_state = traj_controller_->get_node()->deactivate();
        EXPECT_EQ(new_state.id(), lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
      }
    }
  }
```

christophfroehlich:
Do we know why this is suddenly an issue? what has changed?

saikishor:
> Do we know why this is suddenly an issue? what has changed?

We are just exiting the test in active state and the LoanedInterfaces are still available and it is throwing this exception upon their destruction(this happens only time to time). I think it should happen in other distros as well, but I don't know why it is not happening.

bmagyar (approved):

PR #1385, opened 2024-11-23
Fix RealtimeBox API changes

christophfroehlich:
https://github.com/ros-controls/realtime_tools/issues/200

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -111,27 +111,27 @@ controller_interface::return_type DiffDriveController::update(
     return controller_interface::return_type::OK;
   }
 
-  std::shared_ptr<TwistStamped> last_command_msg;
-  received_velocity_msg_ptr_.try_get([&last_command_msg](const std::shared_ptr<TwistStamped> & msg)
-                                     { last_command_msg = msg; });
+  // if the mutex is unable to lock, last_command_msg_ won't be updated
+  received_velocity_msg_ptr_.try_get([this](const std::shared_ptr<TwistStamped> & msg)
+                                     { last_command_msg_ = msg; });
 
-  if (last_command_msg == nullptr)
+  if (last_command_msg_ == nullptr)
   {
     RCLCPP_WARN(logger, "Velocity message received was a nullptr.");
     return controller_interface::return_type::ERROR;
   }
 
-  const auto age_of_last_command = time - last_command_msg->header.stamp;
+  const auto age_of_last_command = time - last_command_msg_->header.stamp;
   // Brake if cmd_vel has timeout, override the stored command
   if (age_of_last_command > cmd_vel_timeout_)
   {
-    last_command_msg->twist.linear.x = 0.0;
-    last_command_msg->twist.angular.z = 0.0;
+    last_command_msg_->twist.linear.x = 0.0;
+    last_command_msg_->twist.angular.z = 0.0;
   }
 
   // command may be limited further by SpeedLimit,
   // without affecting the stored twist command
-  TwistStamped command = *last_command_msg;
+  TwistStamped command = *last_command_msg_;
   double & linear_command = command.twist.linear.x;
   double & angular_command = command.twist.angular.z;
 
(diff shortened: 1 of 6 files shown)


bmagyar (changes requested):
This PR swaps out 2 controllers' main input to use the stamped version without any deprecation. There is very little actual alignment with the realtime_tools API that I can see. Would it be possible to separate these onto 2 PRs?

christophfroehlich:
No, that is not true: There is no change! I was just puzzled by this alias and renamed it:
`using Twist = geometry_msgs::msg::TwistStamped;`
I can move that to a follow-up PR if you like to.

bmagyar:
Yes please and I stand corrected 🫡

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -503,7 +504,8 @@ bool DiffDriveController::reset()
   subscriber_is_active_ = false;
   velocity_command_subscriber_.reset();
 
-  received_velocity_msg_ptr_.set(nullptr);
+  received_velocity_msg_ptr_.set([](std::shared_ptr<TwistStamped> & stored_value)
+                                 { stored_value = nullptr; });
```
christophfroehlich, line 508: With latest fixes to realtime_tools this could stay like
```suggestion
  received_velocity_msg_ptr_.set(nullptr);
```
as the reviewers prefer.

tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -433,7 +434,8 @@ bool TricycleController::reset()
   subscriber_is_active_ = false;
   velocity_command_subscriber_.reset();
 
-  received_velocity_msg_ptr_.set(nullptr);
+  received_velocity_msg_ptr_.set([](std::shared_ptr<TwistStamped> & stored_value)
+                                 { stored_value = nullptr; });
```
christophfroehlich, line 438: With latest fixes to realtime_tools this could stay like
```suggestion
  received_velocity_msg_ptr_.set(nullptr);
```
as the reviewers prefer.

bmagyar (approved):
thank you!itemize

saikishor (commented):
I'm curious about the effect of the `try_get` we might have

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -112,7 +112,8 @@ controller_interface::return_type DiffDriveController::update(
   }
 
   std::shared_ptr<TwistStamped> last_command_msg;
-  received_velocity_msg_ptr_.get(last_command_msg);
+  received_velocity_msg_ptr_.try_get([&last_command_msg](const std::shared_ptr<TwistStamped> & msg)
+                                     { last_command_msg = msg; });
```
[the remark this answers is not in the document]
saikishor: Sounds like a nice plan!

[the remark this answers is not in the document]
christophfroehlich: done

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -326,8 +327,8 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
   }
 
   const TwistStamped empty_twist;
```
saikishor, line 329:
```suggestion
  last_command_msg_ = std::make_shared<TwistStamped>();
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -326,8 +327,8 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
...
-  received_velocity_msg_ptr_.set(std::make_shared<TwistStamped>(empty_twist));
-
+  received_velocity_msg_ptr_.set([empty_twist](std::shared_ptr<TwistStamped> & stored_value)
+                                 { stored_value = std::make_shared<TwistStamped>(empty_twist); });
```
saikishor, line 331:
```suggestion
                                 { stored_value = last_command_msg_; });
```

saikishor (commented):
May be it is better to set the `last_command_msg_` to empty in the configure. Same goes for the test changes

christophfroehlich:
> May be it is better to set the `last_command_msg_` to empty in the configure. Same goes for the test changes

We don't use try_get in the tests?

saikishor:
> We don't use try_get in the tests?

Sure it is not needed

saikishor (approved):
Looks spot on!

Thanks a lot for fixing things :)

christophfroehlich:
So we don't need to backport this after the branch-off of realtime_tools right?

Issue #271, opened 2021-11-18
[DiffDriveController] use "period" argument in the update method insead of calculating "dt"


destogl:
> we should have `period` available in the `update` method

_Originally posted by @destogl in https://github.com/ros-controls/ros2_controllers/pull/260#discussion_r752228051_

christophfroehlich:
Consider also the discussion in 
https://github.com/ros-controls/ros2_controllers/pull/1481

christophfroehlich:
> @christophfroehlich can you review the odometry implementation in #1535? If it looks good, I can make a PR making similar modifications to the ``diff_drive_controller``.

_Originally posted by @Amronos in https://github.com/ros-controls/ros2_controllers/issues/1394#issuecomment-2726314076_

Amronos:
Please assign this to me. I will start working on it in about a week.

PR #1394, opened 2024-11-27
Improve diff_drive_controller odometry velocity and timestep handling

josephduchesne:
This PR  updates diff_drive odometry class to
- consistantly use velocity where it claims to (the odometry class no longer arbitrarily swaps between position delta and velocity without comments, despite parameter and/or method names claim to take velocity).
- use a consistant timestep (dt) in every method that makes use of it
- rely on diff_drive_controller to decide which timestep source to use (avoiding existing and new cases where diff_drive_controller uses nominal timestep, and odometry uses 

This was requested in #271, although I took a more expansive approach since the odometry class is was filled with bug-landmines the way it was written.

- The result is that the class is very slightly less efficient (but still quite trivial) due to passing around a extra double.

- The other result is that it is no longer possible for velocity based odometry updates calculated incorrectly when the timestep fails to keep up.

To prevent position based updates from now suffering the inverse of the velocity timestep/dt mismatch, the measured delta is used exclusively for position based updates (retaining the existing behavior), while the configured timestep is used for open and closed loop velocity updates. This decision is made by diff_drive_controller, and out of scope of the odometry class.

I also added a 1s throttled warning if there is a major mismatch between measured and configured timestep (+/-20%) since this would cause a slight accumulation of integration errors on curved paths (I believe), depending on velocity, curvature and mismatch size. This could probably be expanded to a more general diagnostics WARN output, since controller/odometry updates failing to keep up can be serious.

Similar requested period/measured period mismatches may be a nuisance for open loop (which is velocity based) and closed loop velocity based updates, but the impact is likely negligible (in this class).

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -141,9 +141,10 @@ controller_interface::return_type DiffDriveController::update(
   const double left_wheel_radius = params_.left_wheel_radius_multiplier * params_.wheel_radius;
   const double right_wheel_radius = params_.right_wheel_radius_multiplier * params_.wheel_radius;
 
+
   if (params_.open_loop)
   {
-    odometry_.updateOpenLoop(linear_command, angular_command, time);
+    odometry_.updateFromVelocityOpenLoop(linear_command, angular_command, time, period.seconds());
   }
   else
   {
@@ -171,13 +172,28 @@ controller_interface::return_type DiffDriveController::update(
 
     if (params_.position_feedback)
     {
-      odometry_.update(left_feedback_mean, right_feedback_mean, time);
+      // Warn if actual timestep doesn't match period, 
+      // since 
+      const double measured_period = (
+        get_node()->get_clock()->now() - previous_update_timestamp_).seconds();
+      
+      // +/- 20% of nominal loop time as a starting point to warn wheel encoder
+      // users that the real update rate doesn't match the requested update period
+      if ( std::fabs(period.seconds()-measured_period) > period.seconds()/5.0 ) {
+        RCLCPP_WARN_THROTTLE(
+          get_node()->get_logger(), *(get_node()->get_clock()), 1000,
+          "Integrating odometry at measured delta of %.3fs, but desired update period is %.3fs",
+          measured_period, period.seconds());
+      }
+
+      // update odometry 
+      odometry_.updateFromPosition(left_feedback_mean, right_feedback_mean, time, measured_period);
     }
     else
     {
       odometry_.updateFromVelocity(
-        left_feedback_mean * left_wheel_radius * period.seconds(),
-        right_feedback_mean * right_wheel_radius * period.seconds(), time);
+        left_feedback_mean * left_wheel_radius,
+        right_feedback_mean * right_wheel_radius, time, period.seconds());
     }
   }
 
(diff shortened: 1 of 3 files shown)


diff_drive_controller/src/odometry.cpp
```diff
@@ -59,52 +58,49 @@ bool Odometry::update(double left_pos, double right_pos, const rclcpp::Time & ti
...
   // Compute linear and angular diff:
   const double linear = (left_vel + right_vel) * 0.5;
   // Now there is a bug about scout angular velocity
   const double angular = (right_vel - left_vel) / wheel_separation_;
 
   // Integrate odometry:
-  integrateExact(linear, angular);
+  integrateExact(linear, angular, dt);
 
   timestamp_ = time;
```
christophfroehlich, line 83: timestamp_ class variable could be removed now

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -171,13 +172,28 @@ controller_interface::return_type DiffDriveController::update(
 
     if (params_.position_feedback)
     {
-      odometry_.update(left_feedback_mean, right_feedback_mean, time);
+      // Warn if actual timestep doesn't match period, 
+      // since 
+      const double measured_period = (
```
christophfroehlich, line 177: Why should the measured period be different from the period argument of the update method? This is calculated from the controller_manager, we can assume that this is always valid?

christophfroehlich (changes requested):
Thanks for picking this old issue, I agree this should be more explicit.

But I have comments in the code here, I'm not sure if this is really an issue what you tried to fix.

In general: 

- Please check the pre-commit failures
- We should [deprecate](https://en.cppreference.com/w/cpp/language/attributes/deprecated) the old methods but keep them for a while
- If we are breaking API here, we should rename all odometry class methods to have snake case which is the standard of ros2_control (and ROS 2 in general afaik)

diff_drive_controller/src/odometry.cpp
josephduchesne: good catch

diff_drive_controller/src/diff_drive_controller.cpp
josephduchesne: I was thinking on this some more myself and came to the same conclusion.

I'm hoping to rework this PR sometime next week focusing on the code quality/cleanup aspects suggested here and remove the inconsistent time handling (which as you suggest is probably a distinction without a difference).

Recently I've switched to using the RT_Linux kernel and set up permissions for real-time and found a fairly dramatic difference in perceived responsiveness with it enabled in the diff drive controller. Something isn't handling time/dates/update timesteps properly and I think I incorrectly identified this as the culprit. It still might be, but I should have started off with unit tests (or at least minimum steps to reproduce) rather than complicating an otherwise straightforward code cleanup.

josephduchesne:
> In general:
> 
> * Please check the pre-commit failures
> * We should [deprecate](https://en.cppreference.com/w/cpp/language/attributes/deprecated) the old methods but keep them for a while
> * If we are breaking API here, we should rename all odometry class methods to have snake case which is the standard of ros2_control (and ROS 2 in general afaik)

Thanks, this is my first PR for the repo and I understand the value of consistency for a big open source project like this. I'll try to get things better aligned/more narrowly focused and either update the PR branch, or close this if I won't have the time in the next two weeks.

christophfroehlich:
@josephduchesne are you still working on this?

Amronos:
@christophfroehlich can you review the odometry implementation in #1535? If it looks good, I can make a PR making similar modifications to the ``diff_drive_controller``.

PR #1395, opened 2024-11-27
Jtc sum periods

fmauch:
Sample trajectory based on the sum of periods instead of the absolute time.

This is basically the second part of #1191. In the long run I want to be able to scale the period with a scaling factor in order to  artificially slow down / speed up progression in the trajectory. This PR is there to separate the necessary time counting change from the scaling functionality to make things more contained and easier to review.

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_actions.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_actions.cpp
@@ -83,12 +83,16 @@ protected:
       {
         // controller hardware cycle update loop
         auto clock = rclcpp::Clock(RCL_STEADY_TIME);
-        auto start_time = clock.now();
+        auto now_time = clock.now();
+        auto start_time = now_time;
+        auto last_time = start_time;
         rclcpp::Duration wait = rclcpp::Duration::from_seconds(2.0);
         auto end_time = start_time + wait;
         while (clock.now() < end_time)
         {
-          traj_controller_->update(clock.now(), clock.now() - start_time);
+          now_time = clock.now();
+          traj_controller_->update(now_time, now_time - last_time);
+          last_time = now_time;
         }
       });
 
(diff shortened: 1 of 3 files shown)


christophfroehlich (approved):
LGTM

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -83,12 +83,16 @@ class TestTrajectoryActions : public TrajectoryControllerTest
       {
         // controller hardware cycle update loop
         auto clock = rclcpp::Clock(RCL_STEADY_TIME);
-        auto start_time = clock.now();
+        auto now_time = clock.now();
+        auto start_time = now_time;
+        auto last_time = start_time;
         rclcpp::Duration wait = rclcpp::Duration::from_seconds(2.0);
         auto end_time = start_time + wait;
```
saikishor, line 90:
```suggestion
        auto last_time = now_time;
        rclcpp::Duration wait = rclcpp::Duration::from_seconds(2.0);
        auto end_time = last_time + wait;
```

saikishor (approved):
Looks good to me. 

Doing this way, we are kinda doing it open loop with the time. Am I right?

fmauch:
> Looks good to me.
> 
> Doing this way, we are kinda doing it open loop with the time. Am I right?

Yes, that is true. That requires correct period times without missing or skipping a period. Which is why we recently found issues with that and fixed them. From what I can draw from the design concepts this is an assumption we can make here, though.

saikishor:
> Yes, that is true. That requires correct period times without missing or skipping a period. Which is why we recently found issues with that and fixed them. From what I can draw from the design concepts this is an assumption we can make here, though.

Got it. Thanks for the explanation.

fmauch:
Semi-binary builds should now also succeed, now that https://github.com/ros-controls/ros2_control/pull/1898 is merged. I cannot restart the checks, though.

PR #1400, opened 2024-11-30
Don't call shutdown() after an exception

### LINKED ISSUE #1383: how to gracefully terminate a python node
I'm a little bit puzzled about the fact that the code in the finally branch has zero coverage. maybe rlcpy.ok() is always false here. Should the destroy_node() be called even if the context is not valid any more? @saikishor any idea?

image

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_controllers/issues/1369#issuecomment-2483034214_

This does not happen always, but if it enters there is a different error
```
2024-11-22T12:11:30.3702266Z [publisher_joint_trajectory_controller-2] Keyboard interrupt received. Shutting down node.
2024-11-22T12:11:30.3703121Z [publisher_joint_trajectory_controller-2] Traceback (most recent call last):
2024-11-22T12:11:30.3704880Z [publisher_joint_trajectory_controller-2]   File "/home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/install/ros2_controllers_test_nodes/lib/ros2_controllers_test_nodes/publisher_joint_trajectory_controller", line 33, in <module>
2024-11-22T12:11:30.3706968Z [publisher_joint_trajectory_controller-2]     sys.exit(load_entry_point('ros2-controllers-test-nodes==4.16.0', 'console_scripts', 'publisher_joint_trajectory_controller')())
2024-11-22T12:11:30.3708246Z [publisher_joint_trajectory_controller-2]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-11-22T12:11:30.3710170Z [publisher_joint_trajectory_controller-2]   File "/home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/install/ros2_controllers_test_nodes/lib/python3.12/site-packages/ros2_controllers_test_nodes/publisher_joint_trajectory_controller.py", line 196, in main
2024-11-22T12:11:30.3711758Z [publisher_joint_trajectory_controller-2]     rclpy.shutdown()
2024-11-22T12:11:30.3712778Z [publisher_joint_trajectory_controller-2]   File "/opt/ros/rolling/lib/python3.12/site-packages/rclpy/__init__.py", line 170, in shutdown
2024-11-22T12:11:30.3713790Z [publisher_joint_trajectory_controller-2]     _shutdown(context=context)
2024-11-22T12:11:30.9113697Z [publisher_joint_trajectory_controller-2]   File "/opt/ros/rolling/lib/python3.12/site-packages/rclpy/utilities.py", line 82, in shutdown
2024-11-22T12:11:30.9115644Z [publisher_joint_trajectory_controller-2]     context.shutdown()
2024-11-22T12:11:30.9117325Z [publisher_joint_trajectory_controller-2]   File "/opt/ros/rolling/lib/python3.12/site-packages/rclpy/context.py", line 179, in shutdown
2024-11-22T12:11:30.9118481Z [publisher_joint_trajectory_controller-2]     self._cleanup()
2024-11-22T12:11:30.9119840Z [publisher_joint_trajectory_controller-2]   File "/opt/ros/rolling/lib/python3.12/site-packages/rclpy/context.py", line 170, in _cleanup
2024-11-22T12:11:30.9121292Z [publisher_joint_trajectory_controller-2]     self.__context.shutdown()
2024-11-22T12:11:30.9122665Z [publisher_joint_trajectory_controller-2] rclpy._rclpy_pybind11.RCLError: failed to shutdown: rcl_shutdown already called on the given context, at ./src/rcl/init.c:290
2024-11-22T12:11:30.9125963Z [ERROR] [publisher_joint_trajectory_controller-2]: process has died [pid 30093, exit code 1, cmd '/home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/install/ros2_controllers_test_nodes/lib/ros2_controllers_test_nodes/publisher_joint_trajectory_controller --ros-args --params-file /home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/install/ros2_controllers_test_nodes/share/ros2_controllers_test_nodes/test/
```
https://github.com/ros-controls/ros2_controllers/actions/runs/11972094365/job/33378268325?pr=1314

christophfroehlich:
Closes #1383 acc. to the hint given by @firesurfer

### DIFF
--- a/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_joint_trajectory_controller.py
+++ b/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_joint_trajectory_controller.py
@@ -186,13 +186,10 @@ def main(args=None):
 
     try:
         rclpy.spin(publisher_joint_trajectory)
-    except KeyboardInterrupt:
+    except (KeyboardInterrupt, rclpy.executors.ExternalShutdownException):
         print("Keyboard interrupt received. Shutting down node.")
     except Exception as e:
         print(f"Unhandled exception: {e}")
-    finally:
-        if rclpy.ok():
-            publisher_joint_trajectory.destroy_node()
 
 
 if __name__ == "__main__":
--- a/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_joint_trajectory_controller.py
+++ b/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_joint_trajectory_controller.py
@@ -193,7 +193,6 @@ def main(args=None):
     finally:
         if rclpy.ok():
             publisher_joint_trajectory.destroy_node()
-            rclpy.shutdown()
 
 
 if __name__ == "__main__":
--- a/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
+++ b/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
@@ -77,7 +77,6 @@ def main(args=None):
     finally:
         if rclpy.ok():
             publisher_forward_position.destroy_node()
-            rclpy.shutdown()
 
 
 if __name__ == "__main__":
--- a/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
+++ b/ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_forward_position_controller.py
@@ -70,13 +70,10 @@ def main(args=None):
 
     try:
         rclpy.spin(publisher_forward_position)
-    except KeyboardInterrupt:
+    except (KeyboardInterrupt, rclpy.executors.ExternalShutdownException):
         print("Keyboard interrupt received. Shutting down node.")
     except Exception as e:
         print(f"Unhandled exception: {e}")
-    finally:
-        if rclpy.ok():
-            publisher_forward_position.destroy_node()
 
 
 if __name__ == "__main__":


saikishor (commented):
Isn't finally also applied even for the case without exception?

christophfroehlich:
> Isn't finally also applied even for the case without exception?

yes, but spin() never will exit without an exception, am I right? I'm not sure if destroy_node() is necessary at all: The examples on docs.ros.org don't do this.

firesurfer:
Without having the definite answer - my understanding is that you can not exit 
`rclpy.spin` without an exception:

a) Something throws an exception during the spin (e.g. one of your callbacks)
b) Any signal sent to the process also either generates a KeyboardInterrupt or rclpy.executors.ExternalShutdownException

#### But when would you like to explicitly destroy a node? 

Behind the node there is always the `Context`.   `rclpy.init` implicitly create a global context. In this context you could create and destroy as many nodes as you want as long as the context is valid (which is not the case when we exit with an exception due to a signal - at least that is what I saw so far). But you would have to manage spinning the nodes manually. 

The reason why there are methods for destroying publisher/subscription etc. and nodes is that the default garbage collection mechanism of python doesn't work with those. There is always a corresponding native object in the rcl layer and therefore it is necessary to manually release resources. (I've seen that there is some work done in rclpy to implement those as `ContextManagers` in order to handle it in a better way) 

Btw. if you need ROS communication after leaving spin due to an exception it is possible to create a new context and perform any communication inside that.

christophfroehlich:
Thanks for the background info. Would you say that the current state is fine then? I mean, this is a test node and not productive code, but still like to understand how to properly design this ;)

ros2_controllers_test_nodes/ros2_controllers_test_nodes/publisher_joint_trajectory_controller.py
```diff
@@ -193,7 +193,6 @@ def main(args=None):
     finally:
         if rclpy.ok():
             publisher_joint_trajectory.destroy_node()
```
firesurfer, line 195: This should also not be necessary.

firesurfer:
> Thanks for the background info. Would you say that the current state is fine then? I mean, this is a test node and not productive code, but still like to understand how to properly design this ;)

See my comment.
At least that's my understanding. 

I also think the rclpy documentation is not very clear about this - in addition there are also some things that not work the way they should (see the on_shutdown hook for example)

saikishor (approved):
Sounds fine. I hope it works the same back in Humble too

Zainskl:
Hello, @christophfroehlich if we want to do proper cleanup on SIGINT or SIGTERM
then
What should be our main block?

Is it good? **For humble?**

     
```
def cleanup(node):
    for i in range(20):
        time.sleep(1)
        node.get_logger().info(f"Sleep {i}")

def main(args=None):
    rclpy.init(args=args)

    node = myNode()

    try:
        rclpy.spin(node)
    except Exception as e:
        print(f"An error occurred: {e}")
    finally:
        cleanup(node)
        node.destroy_node()
        if rclpy.ok():
            rclpy.shutdown()
```

Or **Do i Need to use rclpy.context.Context.on_shutdown(somecallback)??**

I mean when OS gets shutdown call or Sigint or sigterm?
Then my cleanup(node) function must run before shutdown or reboot of os
please can you set this main block as per ros2 needs?

christophfroehlich:
@ZainSk-Reblade I really don't know, sorry.

PR #1403, opened 2024-12-01
Fix GPIO controller onn humble

Wiktor-99:
Compared to the rolling version, I have made the following changes:
* Fixed compilation issues
* Removed fetching GPIO state interfaces from URDF
* Aligned unit tests (UTs)
* Updated documentation


christophfroehlich (approved):
Thanks a lot, changes look good!

PR #1406, opened 2024-12-02
Use the .hpp headers from `realtime_tools` package

saikishor:
@christophfroehlich mentioned about the update here : https://github.com/ros-controls/realtime_tools/pull/206#issuecomment-2503276471

This PR addressed the same


christophfroehlich (approved):
Although this is not your first issue ;)

bmagyar (approved):
oi!

PR #1422, opened 2024-12-10
[CI] Add downstream jobs

christophfroehlich:
- run them only in PRs
- build+test downstream ros-controls packages
- build-only some downstream 3rd party packages. any more we should add?

### DIFF
--- a/.github/workflows/rolling-semi-binary-downstream-build.yml
+++ b/.github/workflows/rolling-semi-binary-downstream-build.yml
@@ -29,10 +29,9 @@ jobs:
       fail-fast: false
       matrix:
         ROS_DISTRO: [rolling]
-        ROS_REPO: [testing]
     with:
       ros_distro: ${{ matrix.ROS_DISTRO }}
-      ros_repo: ${{ matrix.ROS_REPO }}
+      ros_repo: testing
       ref_for_scheduled_build: master
       not_test_build: true
       upstream_workspace: ros2_controllers.${{ matrix.ROS_DISTRO }}.repos
--- a/.github/workflows/rolling-semi-binary-downstream-build.yml
+++ b/.github/workflows/rolling-semi-binary-downstream-build.yml
@@ -34,5 +34,26 @@ jobs:
       ros_repo: testing
       ref_for_scheduled_build: master
       upstream_workspace: ros2_controllers.${{ matrix.ROS_DISTRO }}.repos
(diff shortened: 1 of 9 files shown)


ros_controls.jazzy.repos
```diff
@@ -0,0 +1,9 @@
+repositories:
+  ros-controls/gz_ros2_control:
+    type: git
+    url: https://github.com/ros-controls/gz_ros2_control.git
+    version: jazzy
```
saikishor, line 5: Do we have tests over gz_ros2_control that use our ros2_controllers?

.github/workflows/rolling-semi-binary-downstream-build.yml
```diff
@@ -0,0 +1,54 @@
...
+        ROS_DISTRO: [rolling]
+        ROS_REPO: [testing]
+    with:
+      ros_distro: ${{ matrix.ROS_DISTRO }}
+      ros_repo: ${{ matrix.ROS_REPO }}
+      ref_for_scheduled_build: master
+      not_test_build: true
+      upstream_workspace: ros2_controllers.${{ matrix.ROS_DISTRO }}.repos
+      downstream_workspace: ros_controls.${{ matrix.ROS_DISTRO }}.repos
+      not_test_downstream: true
```
saikishor, line 54: Same question here as well : https://github.com/ros-controls/ros2_control/pull/1929#discussion_r1879001130

ros_controls.jazzy.repos
christophfroehlich: https://github.com/ros-controls/gz_ros2_control/blob/9ead4869949530722259898b192251706e083332/gz_ros2_control_tests/package.xml#L21-L22

.github/workflows/rolling-semi-binary-downstream-build.yml
christophfroehlich: you mean due to the matrix strategy? this makes it easier to add a new distro before we branch for it. I created already distinct files for jazzy, but kept this for the future

saikishor: Thanks

ros_controls.jazzy.repos
saikishor: Got it. Thanks

saikishor (approved):
LGTM

fmauch (approved):
Thanks for setting this up!

PR #1424, opened 2024-12-13
Add an error msg if empty message is received

christophfroehlich:
There existed one branch where a goal is rejected but no error message is given.


saikishor (dismissed):
LGTM

bmagyar (approved):

PR #1434, opened 2024-12-19
Fix bug in admittance controller access to reference interfaces

marip8:
This PR fixes an issue in the admittance controller where it always tries to access both the position and velocity reference interfaces even if one of them is not provided. Currently, if the `chainable_command_interfaces` parameter only includes position or velocity, the admittance controller will complain with the following message:

```
Export of the state or reference interfaces failed with following error: The internal storage for reference values 'reference_interfaces_' variable has size '6', but it is expected to have the size '12' equal to the number of exported reference interfaces. Please correct and recompile the controller with name 'admittance_controller' and try again.
```

To fix this issue, the controller should only access the reference interface(s) that match its chainable command interface(s)


bmagyar (approved):
The controllers I'm the least familiar with is probably this one but these changes look generally good to me. It will clash with the API updates for get/set but that shouldn't be your concern on this PR.

Thank you!

christophfroehlich (changes requested):
First, could you please fix the pre-commit issues? Install pre-commit in your workspace for the next time, then run `pre-commit run --all`.

And do you mind adding a test to see if valid combinations of chainable_command_interfaces don't throw errors?

christophfroehlich:
Thanks for your fixes, splitted this PR in #1196 and #1532

Issue #1442, opened 2024-12-22
Cleanup all the deprecations warning from the recent ros2_control variant changes


saikishor:
## Background

Recently after the new variants feature in Handles (https://github.com/ros-controls/ros2_control/pull/1688) We have a bunch of deprecation warnings that needs to be cleanup. This could be a nice first issue.

## Instructions

The best approach to this is to create the temporary variables where ever needed and then if the `set_value` or `get_value` returns `False`. Then, better to print a warning and simply return OK. This way, if we have some failures in getting data, the user will be able to know and can introspect properly. 

Moreover, If everything is running synchronously. it should work as expected without any failures.

For example:

For the following lines of the code in the diff_drive_controller:
https://github.com/ros-controls/ros2_controllers/blob/81c0d41e4feebfa7e7c81b01d0ac1101eaf10e5d/diff_drive_controller/src/diff_drive_controller.cpp#L268-L272

The expected code should look like:

```
  for (size_t index = 0; index < static_cast<size_t>(wheels_per_side_); ++index)
  {
    if (!registered_left_wheel_handles_[index].velocity.get().set_value(velocity_left) ||
    !registered_right_wheel_handles_[index].velocity.get().set_value(velocity_right))
    {
      RCLCPP_WARN(get_node()->get_logger(), "Error while setting the velocity_left : %f "
        "and velocity_right : %f to the command handles", velocity_left, velocity_right);
      return controller_interface::return_type::OK;
    }
  }
```

For the following `get_value` code in the diff_drive_controller:
https://github.com/ros-controls/ros2_controllers/blob/81c0d41e4feebfa7e7c81b01d0ac1101eaf10e5d/diff_drive_controller/src/diff_drive_controller.cpp#L162-L164

The expected outcome would be:

```
bool left_success = false;
const double left_feedback = [&]() {
    double temp;
    left_success = registered_left_wheel_handles_[index].feedback.get().get_value(temp);
    return temp;
}();

bool right_success = false;
const double right_feedback = [&]() {
    double temp;
    right_success = registered_right_wheel_handles_[index].feedback.get().get_value(temp);
    return temp;
}();

// Now you can use left_success and right_success to check if get_value was successful
if (!left_success || !right_success) {
    RCLCPP_WARN(get_node()->get_logger(), "Error while fetching information from the state interfaces");
    return controller_interface::return_type::OK;
}
```

or
```

double left_feedback, right_feedback; 
[...]

PR #1443, opened 2024-12-22
Draft: Fix deprecation warnings in controllers from recent ros2_control variant changes.

kumar-sanjeeev:
- Updated `get_value` and `set_value` in `diff_drive_controller` to fix deprecation warnings.

Fixes https://github.com/ros-controls/ros2_controllers/issues/1442

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -159,9 +159,28 @@ controller_interface::return_type DiffDriveController::update(
     double right_feedback_mean = 0.0;
     for (size_t index = 0; index < static_cast<size_t>(wheels_per_side_); ++index)
     {
-      const double left_feedback = registered_left_wheel_handles_[index].feedback.get().get_value();
-      const double right_feedback =
-        registered_right_wheel_handles_[index].feedback.get().get_value();
+      bool left_success = false;
+      const double left_feedback = [&]()
+      {
+        double temp{};
+        left_success = registered_left_wheel_handles_[index].feedback.get().get_value(temp);
+        return temp;
+      }();
+
+      bool right_success = false;
+      const double right_feedback = [&]()
+      {
+        double temp{};
+        right_success = registered_right_wheel_handles_[index].feedback.get().get_value(temp);
+        return temp;
+      }();
+
+      if (!left_success || !right_success)
+      {
+        RCLCPP_WARN(
+          get_node()->get_logger(), "Error while fetching information from the state interfaces");
+        return controller_interface::return_type::OK;
+      }
 
       if (std::isnan(left_feedback) || std::isnan(right_feedback))
       {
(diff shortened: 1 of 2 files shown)


saikishor:
@kumar-sanjeeev I'm discussing with @christophfroehlich on the best way to address these get_value methods. We will get back to you soon. So, you can apply the same to other controllers

kumar-sanjeeev:
> @kumar-sanjeeev I'm discussing with @christophfroehlich on the best way to address these get_value methods. We will get back to you soon. So, you can apply the same to other controllers

Okay, understood. I noticed that we also need to modify the test cases to suppress the warnings there as well.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -159,9 +159,28 @@ controller_interface::return_type DiffDriveController::update(
     double right_feedback_mean = 0.0;
     for (size_t index = 0; index < static_cast<size_t>(wheels_per_side_); ++index)
     {
-      const double left_feedback = registered_left_wheel_handles_[index].feedback.get().get_value();
-      const double right_feedback =
-        registered_right_wheel_handles_[index].feedback.get().get_value();
+      bool left_success = false;
```
christophfroehlich, line 162: shall we simplify this and use just a single bool variable if we don't use the information of left/right failure?
Please also add a comment here, that we only need the lambda functions here to be able to mark the double left/right_feedback variable as const.

kumar-sanjeeev:
```cpp
  bool success = true;
    // lambda function only to mark `double left_feedback` as `const`.
    const double left_feedback = [&]()
    {
      double temp{};
      success = registered_left_wheel_handles_[index].feedback.get().get_value(temp);
      return temp;
    }();

    // lambda function only to mark `double right_feedback` as `const`.
    const double right_feedback = [&]()
    {
      double temp{};
      success = registered_right_wheel_handles_[index].feedback.get().get_value(temp);
      return temp;
    }();

    if (!success)
    {
      RCLCPP_WARN(
        get_node()->get_logger(),
        "Error while fetching information from the state interfaces for %s wheel at index [%zu]",
        (index % 2 == 0) ? "left" : "right", index);
      return controller_interface::return_type::OK;
    }

```

@christophfroehlich  Is this make sense ?

kumar-sanjeeev: while working on another controller, I noticed that I often need to implement a similar setup for `get_value()`across multiple places, wouldn't it be a good idea to create a separate reusable function for this? We could then import it wherever needed, which might also be helpful in test files.

what do you think?

bmagyar (changes requested):
We need to discuss API further. I really don't like what this summary tells us:
`+44 -7` to achieve the exact same thing we have now. We need to add this to the framework before it blows up the codebase.

PR #1451, opened 2024-12-29
Remove visibility macros

bmagyar:


parallel_gripper_controller/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.16)
-project(parallel_gripper_controller LANGUAGES CXX)
+project(parallel_gripper_controller)
```
saikishor, line 2: Here it is missing

```
# using this instead of visibility macros
# S1 from https://github.com/ros-controls/ros2_controllers/issues/1053
set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```

saikishor (changes requested):
LGTM. Just a minor comment. 
parallel_gripper_controller is missing the `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)`

parallel_gripper_controller/CMakeLists.txt
bmagyar: There's also a comment a line below this that we don't support these controllers on windows or mac

saikishor: Ahh crap. I totally missed it. Thanks

saikishor (approved):

PR #1454, opened 2024-12-29
Fix open-loop odometry in case of ref timeout

christophfroehlich:
In case of ref timeout the open-loop odometry was wrong: It used the last valid command input instead.


steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -378,16 +378,16 @@ controller_interface::return_type SteeringControllersLibrary::update_and_write_c
...
     const auto timeout =
       age_of_last_command > ref_timeout_ && ref_timeout_ != rclcpp::Duration::from_seconds(0);
 
+    // store (for open loop odometry) and set commands
+    last_linear_velocity_ = timeout ? 0. : reference_interfaces_[0];
+    last_angular_velocity_ = timeout ? 0. : reference_interfaces_[1];
+
     auto [traction_commands, steering_commands] = odometry_.get_commands(
-      last_linear_velocity_, last_angular_velocity_, params_.open_loop,
+      reference_interfaces_[0], reference_interfaces_[1], params_.open_loop,
```
saikishor, line 390: Can't we leave the `last_linear_velocity_` and `last_angular_velocity_` here?

christophfroehlich: no, because it will be overriden with zero in case of timeout -> steering angle is zero -> it will move what have been fixed with #1289

saikishor (approved):

steering_controllers_library/src/steering_controllers_library.cpp
saikishor: Got it. We will need to add tests to this part of the code. This could be a good first or second issue as well.

christophfroehlich: I opened an issue

saikishor: Thanks

bmagyar (approved):

destogl (approved):
Looks good. Thanks.

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -378,16 +378,16 @@ controller_interface::return_type SteeringControllersLibrary::update_and_write_c
...
-    last_linear_velocity_ = reference_interfaces_[0];
-    last_angular_velocity_ = reference_interfaces_[1];
-
     const auto age_of_last_command = time - (*(input_ref_.readFromRT()))->header.stamp;
     const auto timeout =
       age_of_last_command > ref_timeout_ && ref_timeout_ != rclcpp::Duration::from_seconds(0);
 
+    // store (for open loop odometry) and set commands
+    last_linear_velocity_ = timeout ? 0. : reference_interfaces_[0];
+    last_angular_velocity_ = timeout ? 0. : reference_interfaces_[1];
```
destogl, line 387:
```suggestion
    last_linear_velocity_ = timeout ? 0.0 : reference_interfaces_[0];
    last_angular_velocity_ = timeout ? 0.0 : reference_interfaces_[1];
```

destogl (dismissed):

saikishor (approved):

PR #1470, opened 2025-01-03
Fix LifecycleNode in tests

christophfroehlich:
Since https://github.com/ros2/rclcpp/pull/2562 we get lots of warnings in the test logs like
```
[WARN] [1735845337.264023881] [rclcpp_lifecycle]: LifecycleNode is not shut down: Node still in state(1) in destructor
```

Triggering the shutdown transition before calling the destructor fixes this.

### DIFF
--- a/ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
+++ b/ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
@@ -128,7 +128,14 @@ public:
 
   void TearDown()
   {
-    controller_->get_node()->shutdown();
+    try
+    {
+      controller_->get_node()->shutdown();
+    }
+    catch (...)
+    {
+      // ignore case where node is not initialized
+    }
     controller_.reset(nullptr);  // this calls the dtor, but does not call shutdown transition
   }
 
--- a/ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
+++ b/ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
@@ -126,9 +126,13 @@ public:
       "/test_ackermann_steering_controller/reference", rclcpp::SystemDefaultsQoS());
   }
 
-  static void TearDownTestCase() {}
+  void TearDown()
+  {
+    controller_->get_node()->shutdown();
+    controller_.reset(nullptr);  // this calls the dtor, but does not call shutdown transition
+  }
 
-  void TearDown() { controller_.reset(nullptr); }
+  static void TearDownTestCase() {}
 
 protected:
   void SetUpController(const std::string controller_name = "test_ackermann_steering_controller")
--- a/admittance_controller/test/test_admittance_controller.hpp
+++ b/admittance_controller/test/test_admittance_controller.hpp
@@ -133,7 +133,14 @@ public:
 
   void TearDown()
   {
-    controller_->get_node()->shutdown();
+    try
+    {
+      controller_->get_node()->shutdown();
+    }
+    catch (...)
+    {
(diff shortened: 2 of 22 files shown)


saikishor (commented):
The changes look good to me. The range_sensor_broadcaster tests seem to fail. Can you take a look?

```
    [ RUN      ] RangeSensorBroadcasterTest.Initialize_RangeBroadcaster_Exception
    unknown file: Failure
    C++ exception with description "Lifecycle node hasn't been initialized yet!" thrown in TearDown().
    
    [  FAILED  ] RangeSensorBroadcasterTest.Initialize_RangeBroadcaster_Exception (0 ms)
```

ackermann_steering_controller/test/test_ackermann_steering_controller.hpp
```diff
@@ -126,9 +126,20 @@ class AckermannSteeringControllerFixture : public ::testing::Test
...
+  void TearDown()
+  {
+    try
+    {
+      controller_->get_node()->shutdown();
+    }
+    catch (...)
+    {
+      // ignore case where node is not initialized
+    }
```
saikishor, line 138:
```suggestion
    if(controller_->get_node().get_state() != lifecycle_msgs::msg::State::PRIMARY_STATE_UNKNOWN)
    {
      controller_->get_node()->shutdown();
    }
```

I think it is better to have it like this no?. trying and catching everything might hide some issues we might add in future. What's your opinion on this @christophfroehlich ?. If you think it's fine, I'm okay with it

christophfroehlich: The following code
```c++
  if(range_broadcaster_->get_node()->get_current_state().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_UNKNOWN)
  {
    range_broadcaster_->get_node()->shutdown();
  }
```

still throws
```

2: [ RUN      ] RangeSensorBroadcasterTest.Initialize_RangeBroadcaster_Exception
2: unknown file: Failure
2: C++ exception with description "Lifecycle node hasn't been initialized yet!" thrown in TearDown().
```

here

https://github.com/ros-controls/ros2_control/blob/d8a21c4e5effa2ac3b75a3dffbc6b48471e18a1f/controller_interface/src/controller_interface_base.cpp#L205-L212

I don't see any other possibility with the controller_interface API.

christophfroehlich: we could only catch `std::runtime_error` and rethrow others. 
Or we change the upstream API to check if the node is initialized or add a shutdown method, which itself checks if the node is initialized.

saikishor: Ahh I didn't know that this exception is in the controller interface base. If this is the case, we can add an upstream API to check if the controller is initialized or may be in the destructor of the base class, call shutdown if the node is valid and also the state makes sense. What do you think?

christophfroehlich: we never trigger the shutdown transition in the controller_manager, maybe this is a bug we should solve?
Currently we only have virtual default destructors of the `ControllerInterfaceBase` and `ControllerInterface` class, where would you add that?

saikishor: I would say in the destructor of the controller_interface_base. As that's the common one for both normal and chained controllers

christophfroehlich:
As suggested by @saikishor we should fix that upstream -> https://github.com/ros-controls/ros2_control/pull/1979

PR #1477, opened 2025-01-09
 Remove empty on_shutdown() callbacks

Juliaj:
Address https://github.com/ros-controls/ros2_controllers/issues/1474


saikishor (changes requested):
@Juliaj can you also remove all other empty callbacks? Not only that of on_shutdown?. I believe we need this general cleanup

christophfroehlich (approved):
Thanks for taking this one! We can remove the other callbacks also in a different PR if you like.

saikishor (approved):
If @christophfroehlich is fine with it. I'm fine with it ;)

PR #1478, opened 2025-01-09
Fix SpeedLimiter Constructor regression

saikishor:
After the https://github.com/ros-controls/ros2_controllers/pull/1315 and recent tag 4.18.0

I observed a regression in a controller that the SpeedLimiter is offering 2 constructors with default argument and this creates the following error: `error: call of overloaded 'SpeedLimiter()' is ambiguous`

I think it makes sense to have the new constructor more explicit without any default arguments. Moreover, the construction `limiter_angular_(std::numeric_limits<double>::quiet_NaN())` still doesn't help as it is still implicitly casting things to the deprecated constructor

`diff_drive_controller.cpp:50:1: warning: ‘diff_drive_controller::SpeedLimiter::SpeedLimiter(bool, bool, bool, double, double, double, double, double, double)’ is deprecated [-Wdeprecated-declarations]`

I decided to fix this also in the same PR by changing them to unique_ptr. I don't think it should be a problem as the controller is used as a plugin.

### DIFF
--- a/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
+++ b/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
@@ -123,8 +123,8 @@ protected:
   std::queue<TwistStamped> previous_commands_;  // last two commands
 
   // speed limiters
-  SpeedLimiter limiter_linear_;
-  SpeedLimiter limiter_angular_;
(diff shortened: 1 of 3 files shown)


christophfroehlich (approved):
Makes sense, thanks for fixing!

I could have used the unique_ptr from the beginning, you are right about the ABI.

bmagyar (approved):

PR #1479, opened 2025-01-09
[pose_broadcaster] Check for valid pose before attempting to publish a tf for it

urfeex:
Deactivate the controller otherwise

When using the pose_broadcaster e.g. with mock hardware, the pose state interfaces might not contain any value (NaN), if there are no initial values inside the URDF. However, it might not make sense to set an initial value as this would be plain wrong information. Imagine a robot-calculated forward kinematics that generates that pose. Initializing this to any fixed value will just be wrong. IMHO it would be better to explicitly state that there is no valid information available.

Currently, attempting to publish a tf with no pose information will result in spamming the shell with
```
TF_NAN_INPUT: Ignoring transform for child_frame_id "tool0_controller" from authority "Authority undetectable" because of a nan value in the transform (nan nan nan) (nan nan nan nan)
```

In this PR I added a check for a valid pose before attempting to publish a tf and return an error in case of an invalid pose.

### DIFF
--- a/pose_broadcaster/src/pose_broadcaster.cpp
+++ b/pose_broadcaster/src/pose_broadcaster.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #include "pose_broadcaster/pose_broadcaster.hpp"
+#include <cmath>
 
 namespace
 {
@@ -23,6 +24,17 @@ constexpr auto DEFAULT_TF_TOPIC = "/tf";
 
 namespace pose_broadcaster
 {
+
+bool is_pose_valid(const geometry_msgs::msg::Pose & pose)
+{
+  return !std::isnan(pose.position.x) && !std::isnan(pose.position.y) &&
+         !std::isnan(pose.position.z) && !std::isnan(pose.orientation.x) &&
+         !std::isnan(pose.orientation.y) && !std::isnan(pose.orientation.z) &&
+         !std::isnan(pose.orientation.w) &&
+         sqrt(
+           pow(pose.orientation.x, 2) + pow(pose.orientation.y, 2) + pow(pose.orientation.z, 2) +
+           pow(pose.orientation.w, 2)) > 0.0;
+}
 
 controller_interface::InterfaceConfiguration PoseBroadcaster::command_interface_configuration()
   const
@@ -148,40 +160,49 @@ controller_interface::return_type PoseBroadcaster::update(
     realtime_publisher_->unlockAndPublish();
   }
 
-  if (realtime_tf_publisher_ && realtime_tf_publisher_->trylock())
+  if (realtime_tf_publisher_)
   {
-    bool do_publish = false;
-    // rlcpp::Time comparisons throw if clock types are not the same
-    if (tf_last_publish_time_.get_clock_type() != time.get_clock_type())
-    {
-      do_publish = true;
-    }
-    else if (!tf_publish_period_ || (tf_last_publish_time_ + *tf_publish_period_ <= time))
-    {
-      do_publish = true;
-    }
-
-    if (do_publish)
-    {
-      auto & tf_transform = realtime_tf_publisher_->msg_.transforms[0];
-      tf_transform.header.stamp = time;
-
-      tf_transform.transform.translation.x = pose.position.x;
-      tf_transform.transform.translation.y = pose.position.y;
-      tf_transform.transform.translation.z = pose.position.z;
-
-      tf_transform.transform.rotation.x = pose.orientation.x;
-      tf_transform.transform.rotation.y = pose.orientation.y;
-      tf_transform.transform.rotation.z = pose.orientation.z;
-      tf_transform.transform.rotation.w = pose.orientation.w;
-
-      realtime_tf_publisher_->unlockAndPublish();
-
-      tf_last_publish_time_ = time;
-    }
-    else
+    if (!is_pose_valid(pose))
     {
       realtime_tf_publisher_->unlock();
+      return controller_interface::return_type::ERROR;
+    }
+
+    if (realtime_tf_publisher_->trylock())
+    {
+      bool do_publish = false;
+      // rlcpp::Time comparisons throw if clock types are not the same
+      if (tf_last_publish_time_.get_clock_type() != time.get_clock_type())
+      {
+        do_publish = true;
+      }
+      else if (!tf_publish_period_ || (tf_last_publish_time_ + *tf_publish_period_ <= time))
+      {
+        do_publish = true;
+      }
+
+      if (do_publish)
+      {
+        auto & tf_transform = realtime_tf_publisher_->msg_.transforms[0];
+        tf_transform.header.stamp = time;
+
+        tf_transform.transform.translation.x = pose.position.x;
+        tf_transform.transform.translation.y = pose.position.y;
+        tf_transform.transform.translation.z = pose.position.z;
+
+        tf_transform.transform.rotation.x = pose.orientation.x;
+        tf_transform.transform.rotation.y = pose.orientation.y;
+        tf_transform.transform.rotation.z = pose.orientation.z;
+        tf_transform.transform.rotation.w = pose.orientation.w;
+
+        realtime_tf_publisher_->unlockAndPublish();
+
+        tf_last_publish_time_ = time;
+      }
+      else
+      {
+        realtime_tf_publisher_->unlock();
+      }
     }
   }
 
--- a/pose_broadcaster/src/pose_broadcaster.cpp
+++ b/pose_broadcaster/src/pose_broadcaster.cpp
@@ -28,9 +28,15 @@ namespace pose_broadcaster
 
 bool is_pose_valid(const geometry_msgs::msg::Pose & pose)
 {
-         std::abs((pose.orientation.x * pose.orientation.x + pose.orientation.y * pose.orientation.y +
-          pose.orientation.z * pose.orientation.z + pose.orientation.w * pose.orientation.w) -
-          1.0) <= 10e-3;
+  return std::isfinite(pose.position.x) && std::isfinite(pose.position.y) &&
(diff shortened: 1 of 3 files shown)


pose_broadcaster/src/pose_broadcaster.cpp
```diff
@@ -24,6 +25,17 @@ constexpr auto DEFAULT_TF_TOPIC = "/tf";
 namespace pose_broadcaster
 {
 
+bool is_pose_valid(const geometry_msgs::msg::Pose & pose)
+{
+  return !std::isnan(pose.position.x) && !std::isnan(pose.position.y) &&
```
christophfroehlich, line 30: should we use `std::is_finite` instead?

christophfroehlich (commented):
Is there a use-case where hardware component just needs more time at startup for example, and the broadcaster should just not start publishing instead of being deactivated immediately?

urfeex:
> Is there a use-case where hardware component just needs more time at startup for example, and the broadcaster should just not start publishing instead of being deactivated immediately?

I've been asking myself the same question. The follow-up question would be what to do then in terms of feedback. Would a throttled logging be an option? It's in the `update` method, after all.

pose_broadcaster/src/pose_broadcaster.cpp
```diff
@@ -148,40 +160,53 @@ controller_interface::return_type PoseBroadcaster::update(
...
-    }
-    else if (!tf_publish_period_ || (tf_last_publish_time_ + *tf_publish_period_ <= time))
-    {
-      do_publish = true;
+      RCLCPP_ERROR(
+        get_node()->get_logger(), "Invalid pose [%f, %f, %f], [%f, %f, %f, %f]", pose.position.x,
+        pose.position.y, pose.position.z, pose.orientation.x, pose.orientation.y,
+        pose.orientation.z, pose.orientation.w);
+      realtime_tf_publisher_->unlock();
+      return controller_interface::return_type::ERROR;
```
saikishor, line 172: Better to return OK, if you want to skip the cycles, or return ERROR if you want to disable the controller.

What's your option here @urfeex ?

pose_broadcaster/src/pose_broadcaster.cpp
```diff
@@ -148,40 +160,53 @@ controller_interface::return_type PoseBroadcaster::update(
     realtime_publisher_->unlockAndPublish();
   }
 
-  if (realtime_tf_publisher_ && realtime_tf_publisher_->trylock())
+  if (realtime_tf_publisher_)
   {
-    bool do_publish = false;
-    // rlcpp::Time comparisons throw if clock types are not the same
-    if (tf_last_publish_time_.get_clock_type() != time.get_clock_type())
+    if (!is_pose_valid(pose))
```
saikishor, line 165: May be this check can be done outside the realtime_tf_publisher condition?. That way we don't disturb the previous code

saikishor (changes requested):

pose_broadcaster/src/pose_broadcaster.cpp
urfeex: That's basically the question raised earlier, as well. If a throttled output would be fine, I would suggest going for that.

saikishor: Ah sorry. I didn't look at previous comments. I did fast review from mobile. My apologies

pose_broadcaster/src/pose_broadcaster.cpp
```diff
@@ -24,6 +25,17 @@ constexpr auto DEFAULT_TF_TOPIC = "/tf";
...
 
+bool is_pose_valid(const geometry_msgs::msg::Pose & pose)
+{
+  return !std::isnan(pose.position.x) && !std::isnan(pose.position.y) &&
+         !std::isnan(pose.position.z) && !std::isnan(pose.orientation.x) &&
+         !std::isnan(pose.orientation.y) && !std::isnan(pose.orientation.z) &&
+         !std::isnan(pose.orientation.w) &&
+         sqrt(
+           pow(pose.orientation.x, 2) + pow(pose.orientation.y, 2) + pow(pose.orientation.z, 2) +
+           pow(pose.orientation.w, 2)) > 0.0;
```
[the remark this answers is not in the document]
saikishor: I believe it is. It is strictly necessary to have it convertible to euler angles.

Most of the methods they normalise it and it might give you wrong results

pose_broadcaster/test/test_pose_broadcaster.hpp
```diff
@@ -77,7 +78,10 @@ void PoseBroadcasterTest::subscribe_and_get_message(const std::string & topic, T
   constexpr size_t max_sub_check_loop_count = 5;
   for (size_t i = 0; !received_msg; ++i)
   {
-    ASSERT_LT(i, max_sub_check_loop_count);
+    if (i >= max_sub_check_loop_count)
+    {
+      throw std::runtime_error("Failed to receive message on topic: " + topic);
+    }
```
saikishor, line 84:
```suggestion
    ASSERT_LT(i, max_sub_check_loop_count) << "Failed to receive message on topic: " << topic;
```

Will this work?

Is there any specific reason for throwing?

fmauch: Yes, because this is expected to fail on an illegal quaternion. Throwing an exception here allows using an `EXPECT_THROW` statement around this function from the test.

pose_broadcaster/test/test_pose_broadcaster.cpp
```diff
@@ -186,6 +189,63 @@ TEST_F(PoseBroadcasterTest, PublishSuccess)
...
+  EXPECT_THROW(subscribe_and_get_message("/tf", tf_msg), std::runtime_error);
+  // Verify that no tf message was sent
+  ASSERT_EQ(tf_msg.transforms.size(), 0lu);
+
+  // Set valid position but invalid quaternion
+  ASSERT_TRUE(pose_position_x_.set_value(0.0));
+  ASSERT_TRUE(pose_orientation_x_.set_value(0.0));
+  ASSERT_TRUE(pose_orientation_y_.set_value(0.0));
+  ASSERT_TRUE(pose_orientation_z_.set_value(0.0));
+  ASSERT_TRUE(pose_orientation_w_.set_value(0.0));
```
saikishor, line 246:
```suggestion
```

Is a test missing here? os is there a reason for this? if not let's remove it

pose_broadcaster/test/test_pose_broadcaster.cpp
```diff
@@ -186,6 +189,63 @@ TEST_F(PoseBroadcasterTest, PublishSuccess)
...
+  ASSERT_TRUE(pose_position_x_.set_value(std::numeric_limits<double>::quiet_NaN()));
+
+  // Subscribe to pose topic
+  geometry_msgs::msg::PoseStamped pose_msg;
+  pose_msg.pose.position.x = 0.0;
+  pose_msg.pose.position.y = 0.0;
+  pose_msg.pose.position.z = 0.0;
+  pose_msg.pose.orientation.x = 0.0;
+  pose_msg.pose.orientation.y = 0.0;
+  pose_msg.pose.orientation.z = 0.0;
```
saikishor, line 221:
```suggestion
```
I believe the default value is zero.
If you think it is better to have this, its  fine for me

saikishor (changes requested):

pose_broadcaster/test/test_pose_broadcaster.hpp
saikishor: Got it

pose_broadcaster/test/test_pose_broadcaster.cpp
```diff
@@ -13,8 +13,11 @@
 // limitations under the License.
 #include "test_pose_broadcaster.hpp"
 
+#include <cmath>
+#include <limits>
 #include <utility>
 #include <vector>
+#include "gtest/gtest.h"
```
christophfroehlich, line 20: don't we use gmock here?

pose_broadcaster/src/pose_broadcaster.cpp
```diff
@@ -24,6 +26,18 @@ constexpr auto DEFAULT_TF_TOPIC = "/tf";
...
+bool is_pose_valid(const geometry_msgs::msg::Pose & pose)
+{
+  return std::isfinite(pose.position.x) && std::isfinite(pose.position.y) &&
+         std::isfinite(pose.position.z) && std::isfinite(pose.orientation.x) &&
+         std::isfinite(pose.orientation.y) && std::isfinite(pose.orientation.z) &&
+         std::isfinite(pose.orientation.w) &&
+
+         (pose.orientation.x * pose.orientation.x + pose.orientation.y * pose.orientation.y +
+          pose.orientation.z * pose.orientation.z + pose.orientation.w * pose.orientation.w -
+          1.0) <= 10e-3;
```
saikishor, line 38:
```suggestion
         std::abs((pose.orientation.x * pose.orientation.x + pose.orientation.y * pose.orientation.y +
          pose.orientation.z * pose.orientation.z + pose.orientation.w * pose.orientation.w) -
          1.0) <= 10e-3;
```

The absolute different is important here

saikishor (changes requested):

pose_broadcaster/test/test_pose_broadcaster.cpp
urfeex: Indeed I forgot checking for no tf again.

pose_broadcaster/test/test_pose_broadcaster.cpp
urfeex: I think my initial thought was to check the result for zeros, but that doesn't really make sense, I think it is fine to delete it.

saikishor (approved):
LGTM. Thank you for addressing all the comments

christophfroehlich (approved):
LGTM!

PR #1485, opened 2025-01-10
Make diff_drive_controller a ChainableControllerInterface

### LINKED ISSUE #1457: Convert the diff_drive_controller into a ChainableControllerInterface in the master branch
I am currently working on designing a differential drive robot with advanced navigation and balancing capabilities using ROS2 Jazzy. I would like to configure my controllers in a chained configuration similar to the one described in the [Controller Chaining Tutorial](https://control.ros.org/rolling/doc/ros2_control/controller_manager/doc/controller_chaining.html). However, in the ros2_controllers master branch it looks like the diff_drive_controller is still a standard ControllerInterface, not a ChainableControllerInterface, so I cannot put controllers that precede the diff_drive_controller.

I propose to convert the diff_drive_controller into a ChainableControllerInterface, and push these changes to the master branch.
I noticed that there is a chainable-diff-drive branch from last year that converts the diff_drive_controller into a ChainableControllerInterface as I desire. The problem is that this branch is many commits behind master at this point, and lacks functionality such as using TwistStamped instead of Twists for the velocity_command subscriber, among others. I would like to see the chainable-diff-drive branch merged with master (or, since this would be a complex merge, simply start from scratch using chainable-diff-drive as a guide).

I have not contributed to ros2_controllers before, but would be interested in taking this on as a challenge. My questions are: 
Why was this merge not made before? What are the blockers or nuances that I should watch out for?

arthurlovekin:
Using [dea3d3c](https://github.com/ros-controls/ros2_controllers/commit/dea3d3c027ef051a6a159264749bacab7065ecee) as a reference, I made the diff_drive_controller into a ChainableControllerInterface. I used a RealtimeBuffer to handle the incoming velocity command messages, and I also added lifecycle checks and support for the speed_limiter that were not included in [dea3d3c](https://github.com/ros-controls/ros2_controllers/commit/dea3d3c027ef051a6a159264749bacab7065ecee). I also added some tests that run the chainable controller in unchained mode, and I intend to test it in chained mode in one of my current projects (though I couldn't readily figure out how to make this into a test).


christophfroehlich:
Please always install pre-commit to avoid failing tests ;) otherwise, run it manually with `pre-commit run --all`

regarding tests: have a look at the mecanum controller, it is chainable anf there are exactly those tests already available

arthurlovekin:
I've fixed my formatting by running `pre-commit run --all-files` and added a test for chained mode, using the mecanum controller as a reference.

arthurlovekin:
To be honest I did not know the full implications of switching from a RealtimeBox to RealtimeBuffer, but I chose the change because every other example of chainable interface (eg. mecanum, passthrough, chainable-diff-drive) used it. I figured it probably had a reason for being used, but at the very least using the RealtimeBuffer would provide consistency across the chainable controllers. It seemed like the RealtimeBuffer was designed specifically for use-cases like the chainable interface, where the reference_interfaces could either be safely updated at a realtime rate or more slowly through the publisher, whereas the RealtimeBox would simply make a best-effort attempt at getting and setting the variables (which may in fact be sufficient for this use case) but didn't explicitly account for realtime vs non-realtime. Let me know if you would like me to proceed with changing RealtimeBuffer back to RealtimeBox, and I will do it.

I agree that there should be a check for properly exported reference interfaces. I will imitate the mecanum drive [when_controller_configured_expect_properly_exported_interfaces](https://github.com/ros-controls/ros2_controllers/blob/master/mecanum_drive_controller/test/test_mecanum_drive_controller.cpp) test. I will try to do that later today.

arthurlovekin:
I just added a test for export_reference_interfaces().

christophfroehlich (approved):
OK I see your point regarding the RT buffer (I also think that the box was not the right thing to use). Thanks for adding the additional test.

I tested this version with the demos in the non-chained mode, and it still works like expected.

For future PRs: It is not desired to squash the commits (we do that at the time of the merge in the base branch), it just takes more time to review because I can't see now what has changed since the time of my last review.

arthurlovekin:
Apologies, in the future I will not squash commits while the PR is in progress. The changes I made were to add the test `TEST_F(TestDiffDriveController, reference_interfaces_are_properly_exported)`, and I renamed `TEST_F(TestDiffDriveController, configure_succeeds_when_wheels_are_specified)` to `TEST_F(
  TestDiffDriveController,
  command_and_state_interface_configuration_succeeds_when_wheels_are_specified)` just to make it more clear that that test evaluates the state_interface_configuration() and command_interface_configuration() functions. The "Compare" button next to my force-push above shows these changes alone.

christophfroehlich:
> The "Compare" button next to my force-push above shows these changes alone.

you are right, I haven't realized that button before. You never stop learning ;)

arthurlovekin:
Is there a way that I can fix the Rolling ABI compatibility check? It seems to be broken because of the change from ControllerInterface to ChainableControllerInterface and I see no way around that. 

Side question: I see in the documentation that there is a Working Group meeting every second Wednesday, but I cannot find the link to join. Would it be possible for me to attend?

olivier-stasse:
> Side question: I see in the documentation that there is a Working Group meeting every second Wednesday, but I cannot find the link to join. Would it be possible for me to attend?

Dear @arthurlovekin the announcement for tomorrow in ROS discourse is here: https://discourse.ros.org/t/ros2-control-working-group-meeting-15th-january-2025/41528

To joint I copied paste the invitation on the announcement:
"You can join the meeting through [google groups](https://groups.google.com/g/ros-control-working-group-invites) or consult the calendar for the Google Meet link (note that the calendar may use a different time zone than yours)."

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -118,31 +137,61 @@ controller_interface::return_type DiffDriveController::update(
...
   // Brake if cmd_vel has timeout, override the stored command
   if (age_of_last_command > cmd_vel_timeout_)
   {
-    last_command_msg_->twist.linear.x = 0.0;
-    last_command_msg_->twist.angular.z = 0.0;
+    reference_interfaces_[0] = 0.0;
+    reference_interfaces_[1] = 0.0;
+  }
+  else if (
+    !std::isnan(command_msg_ptr->twist.linear.x) && !std::isnan(command_msg_ptr->twist.angular.z))
```
saikishor, line 156:
```suggestion
    std::isfinite(command_msg_ptr->twist.linear.x) && std::isfinite(command_msg_ptr->twist.angular.z))
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -419,8 +476,24 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
...
+      }
+      else
+      {
+        RCLCPP_WARN(
+          get_node()->get_logger(),
+          "Ignoring the received message (timestamp %.10f) because it is older than "
+          "the current time by %.10f seconds, which exceeds the allowed timeout (%.4f)",
+          rclcpp::Time(msg->header.stamp).seconds(), current_time_diff.seconds(),
+          cmd_vel_timeout_.seconds());
+      }
```
saikishor, line 496: This logic won't work if the header contains zeros
`if ((msg->header.stamp.sec == 0) && (msg->header.stamp.nanosec == 0))`

You should modify the logic to be able to work with it

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -498,6 +571,9 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
 controller_interface::CallbackReturn DiffDriveController::on_activate(
   const rclcpp_lifecycle::State &)
 {
+  // Set default value in command
+  reset_controller_reference_msg(*(received_velocity_msg_ptr_.readFromRT()), get_node());
```
saikishor, line 575: Personally, I don't like the idea of changing information of what is inside the buffer. The buffers should only be used to get but not modify it simultaneously.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -643,9 +720,35 @@ controller_interface::CallbackReturn DiffDriveController::configure_side(
 
   return controller_interface::CallbackReturn::SUCCESS;
 }
+
+bool DiffDriveController::on_set_chained_mode(bool chained_mode)
+{
+  // Always accept switch to/from chained mode
+  return true || chained_mode;
+}
```
saikishor, line 728:
```suggestion
bool DiffDriveController::on_set_chained_mode(bool /*chained_mode*/)
{
  // Always accept switch to/from chained mode
  return true;
}
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -36,6 +36,25 @@ constexpr auto DEFAULT_ODOMETRY_TOPIC = "~/odom";
...
+  const std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node)
+{
+  msg->header.stamp = node->now();
+  msg->twist.linear.x = std::numeric_limits<double>::quiet_NaN();
+  msg->twist.linear.y = std::numeric_limits<double>::quiet_NaN();
+  msg->twist.linear.z = std::numeric_limits<double>::quiet_NaN();
+  msg->twist.angular.x = std::numeric_limits<double>::quiet_NaN();
+  msg->twist.angular.y = std::numeric_limits<double>::quiet_NaN();
+  msg->twist.angular.z = std::numeric_limits<double>::quiet_NaN();
+}
```
saikishor, line 54: Do we need to set the the velocities to NaN especially?, why not simply zero? This way you can just overload with the default constructor one.

If there is a specific reason, do you mind explaining here?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -394,12 +448,15 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
...
-  last_command_msg_ = std::make_shared<TwistStamped>();
-  received_velocity_msg_ptr_.set([this](std::shared_ptr<TwistStamped> & stored_value)
-                                 { stored_value = last_command_msg_; });
-  // Fill last two commands with default constructed commands
-  previous_commands_.emplace(*last_command_msg_);
-  previous_commands_.emplace(*last_command_msg_);
+  const int nr_ref_itfs = 2;
+  reference_interfaces_.resize(nr_ref_itfs, std::numeric_limits<double>::quiet_NaN());
+  previous_two_commands_.push({{0.0, 0.0}});  // needs zeros (not NaN) to catch early accelerations
+  previous_two_commands_.push({{0.0, 0.0}});
```
saikishor, line 454: I've a question, do we need to also reset this to zeros in the on_activate? 
I'm not sure what would happen if you are deactivating and activating the controller and it might have the previous values still right?. May be I'm wrong here

saikishor (changes requested):
Thanks for the nice contribution. I've some questions around the code that I've reviewed. Thank you

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -118,31 +137,61 @@ controller_interface::return_type DiffDriveController::update(
...
+  auto logger = get_node()->get_logger();
+  if (get_lifecycle_state().id() == State::PRIMARY_STATE_INACTIVE)
+  {
+    if (!is_halted)
+    {
+      halt();
+      is_halted = true;
+    }
+    return controller_interface::return_type::OK;
   }
```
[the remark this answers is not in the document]
arthurlovekin: This was in the old update() function and I kept it in because without it I failed tests (eg. "cleanup" in test_diff_drive_controller.cpp) that were calling update after deactivation. Whether or not to keep it depends on your philosophy: how much do you trust the external controller manager to not call update after deactivating the node?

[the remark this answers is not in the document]
saikishor: It has to work that way AFAIK
https://github.com/ros-controls/ros2_control/blob/00c7088394773a61a8e6a9daf614591e31a2b55d/controller_manager/src/controller_manager.cpp#L2438

It first checks if it is active and then runs the update callback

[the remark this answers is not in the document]
saikishor: If there are tests linked to it. Maybe we should handle this in a different PR. What's your opinion @ros-controls/ros2-maintainers ?

diff_drive_controller/src/diff_drive_controller.cpp
arthurlovekin: I copied this from last year's chainable-diff-drive branch. I think there is probably a better way to do this (using the default constructor), but I need to think it through. What we want to capture is that there is a difference between reading a 0.0 (which would have been intentionally published to stop the robot) vs reading a NaN (which indicates that the controller simply hasn't received any command). We need to put that value (Nan or 0) into the received_velocity_msg_ptr_ RealtimeBuffer on initialization because we don't want to be pulling undefined values from the buffer.

diff_drive_controller/src/diff_drive_controller.cpp
arthurlovekin: Agreed. I'm trying to think of a good way to rework the reset_controller_reference_msg. In any case, I think any resetting should be done in deactivate().

diff_drive_controller/src/diff_drive_controller.cpp
arthurlovekin: I moved the reset_controller_reference_msg function into a private reset_buffers() member function. reset_buffers() is called in reset(), which is used by on_configure() and on_cleanup(). on_deactivate() just uses reset_buffers().

diff_drive_controller/src/diff_drive_controller.cpp
arthurlovekin: I removed the reset of the queue from on_activate() and moved it to the on_deactivate()

christophfroehlich:
> Is there a way that I can fix the Rolling ABI compatibility check? It seems to be broken because of the change from ControllerInterface to ChainableControllerInterface and I see no way around that.

This change truly is ABI breaking, but this is no show-stopper for this PR. Furthermore, for controllers this is not really an issue because we load them as plugins dynamically.

christophfroehlich (changes requested):
Can you please fix the merge conflict after #1478?

diff_drive_controller/src/diff_drive_controller.cpp
arthurlovekin: I am not sure what you mean, perhaps I am misunderstanding. I just added a test demonstrating that it works the way I expect it to. Let me know if my expectation is not correct.

arthurlovekin:
@christophfroehlich the merge conflict should be resolved

arthurlovekin:
One thing that occurred to me as I was writing the reset_buffer() code: If you want to switch from one controller to another, is it desirable for this controller to halt() on_deactivate()? Perhaps I misunderstand the controller_manager because it seems like you would want to be able to switch without stopping the motors. If you switch into the diff_drive_controller from another one it also seems like there could be issues with the acceleration limiter, since from the diff_drive_controller's perspective its previous_two_commands have been set to zero, but you may be switching at a time when the other controller is commanding a high value. (but if you don't set the previous_two_commands to zero then a single starting command at the beginning will be able to bypass the acceleration limits, which can only be applied if multiple commands have been sent).

christophfroehlich (changes requested):
Please fix the pre-commit (cpplint) error. (consider installing pre-commit to avoid this the next time)

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -703,9 +703,10 @@ controller_interface::CallbackReturn DiffDriveController::configure_side(
   return controller_interface::CallbackReturn::SUCCESS;
 }
 
-bool DiffDriveController::on_set_chained_mode(bool /*chained_mode*/)
+bool DiffDriveController::on_set_chained_mode(bool chained_mode)
 {
   // Always accept switch to/from chained mode
+  (void)chained_mode;
```
christophfroehlich, line 709: I know I proposed this, but I'm not sure if we should add this here :see_no_evil:

christophfroehlich: maybe this is why the weird
`return true || chained_mode;`
is there

saikishor: Semantically it doesn't make sense as if you use true || something. It will never even use it. 
Anyway, if that's the only way. Let's use it

christophfroehlich: I had a chat with my AI colleague.
It offered me the solution with (void), but also this one
```
bool DiffDriveController::on_set_chained_mode([[maybe_unused]] bool)
{
  // Always accept switch to/from chained mode
  return true;
}
```
all three work, but let's use the one with || and maybe change all chainable controllers at once if we want to change that.

christophfroehlich (dismissed):

arthurlovekin:
@saikishor can you let me know your thoughts on my responses to your comments?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -412,8 +441,24 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
           "time, this message will only be shown once");
         msg->header.stamp = get_node()->get_clock()->now();
```
saikishor, line 442:
```suggestion
        msg->header.stamp = get_node()->now();
```

Just to be consistent with your change

diff_drive_controller/src/diff_drive_controller.cpp
saikishor: Looks good to me.

saikishor (approved):
It looks pretty good to me now.

I've one question. I see that you are resetting the buffer in the on_configure, and inside we are setting a message with NaN's, so does that mean upon activation until the `cmd_timeout_` and if no message is received within this timeout, it will keep printing `Command message contains NaNs. Not updating reference interfaces.`?

arthurlovekin:
> I've one question. I see that you are resetting the buffer in the on_configure, and inside we are setting a message with NaN's, so does that mean upon activation until the `cmd_timeout_` and if no message is received within this timeout, it will keep printing `Command message contains NaNs. Not updating reference interfaces.`?

You are correct. For me, it was giving a warning about four times on startup. This is kind of annoying so I switched it to RCLCPP_WARN_SKIPFIRST_THROTTLE so that it will skip the first warning, then wait the cmd_vel_timeout_ until logging again (by which time it won't be hitting this warning). However, if you send it a NaN later, then it should give you a warning.

diff_drive_controller/src/diff_drive_controller.cpp
saikishor: ok

diff_drive_controller/src/diff_drive_controller.cpp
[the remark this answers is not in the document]
christophfroehlich: do we have a decision here?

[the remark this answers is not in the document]
saikishor: I would say, let's handle this in a different PR. it would be nice to open an issue to take a look again at this logic.

Thanks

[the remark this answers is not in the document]
christophfroehlich: done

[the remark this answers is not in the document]
saikishor: Awesome. Let's get this merged ;)

PR #1488, opened 2025-01-13
Remove empty callbacks

Juliaj:
Follow up of https://github.com/ros-controls/ros2_controllers/pull/1477#pullrequestreview-2539320377. This is to remove empty callback, e.g. `on_cleanup`, `on_deactivate` from code base.

A noob question, based on [node life_cycle doc ](https://design.ros2.org/articles/node_lifecycle.html), there are predefined roles for these call backs. Why were these coded as empty ? Was it for readability or a place holder for further development ?


christophfroehlich:
> A noob question, based on [node life_cycle doc ](https://design.ros2.org/articles/node_lifecycle.html), there are predefined roles for these call backs. Why were these coded as empty ? Was it for readability or a place holder for further development ?

These are virtual methods defined in the [LifecycleNodeInterface](https://github.com/ros2/rclcpp/blob/rolling/rclcpp_lifecycle/src/node_interfaces/lifecycle_node_interface.cpp), and don't need to be overridden if there is no action to be done at the respective transition. Why they are here? I guess just because of copying a template around.

christophfroehlich (approved):
Thanks for the cleanup.

saikishor (approved):
LGTM

bmagyar (approved):
thank you!

PR #1502, opened 2025-01-21
Improve antiwindup description

ViktorCVS:
Enhance the antiwindup description to accurately reflect its functionality in control_toolbox code.

[control_toolbox/src/pid.cpp::170:173](https://github.com/ros-controls/control_toolbox/blob/ros2-master/src/pid.cpp#L170C1-L173C4)
``` 
 if (!gains.antiwindup_) {
    // Limit i_term so that the limit is meaningful in the output
    i_term = std::clamp(i_term, gains.i_min_, gains.i_max_);
  }
```

### DIFF
--- a/pid_controller/src/pid_controller.yaml
+++ b/pid_controller/src/pid_controller.yaml
@@ -63,17 +63,20 @@ pid_controller:
       antiwindup: {
         type: bool,
         default_value: false,
-        description: "Antiwindup functionality."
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
Thanks! We can change that once again in case we have an upstream change to control_toolbox, see https://github.com/ros-controls/control_toolbox/issues/276

PR #1507, opened 2025-01-28
Reset PID controllers on activation and add `save_i_term` parameter

christophfroehlich:
Since https://github.com/ros-controls/control_toolbox/pull/180 there is a parameter to control the behavior of retention of the integral term after reset. 

Actually, we only have reset in on_configure here. I added the reset to on_activate() and documented the parameter in the GPL yaml. With the current default value of `save_i_term` there should be no behavior change.


Juliaj (approved):
LGTM. You really need to squint hard to read this as "save_i_term" :grin:.

saikishor (approved):

saikishor (approved):
Amazing. Looks perfect

PR #1513, opened 2025-01-31
Fix handles for following controllers of diff_drive_controller

christophfroehlich:
As reported in https://github.com/ros-controls/ros2_control_demos/pull/710, the pid_controller activation failed because it checked for prefix_name and interface_name separately, expecting prefix_name `pid_controller_left_wheel_joint/left_wheel_joint` with interface `velocity`, but actually it is prefix_name `pid_controller_left_wheel_joint` with interface_name `left_wheel_joint/velocity`.

Let's just check the full name.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -673,28 +673,37 @@ controller_interface::CallbackReturn DiffDriveController::configure_side(
     const auto state_handle = std::find_if(
       state_interfaces_.cbegin(), state_interfaces_.cend(),
       [&wheel_name, &interface_name](const auto & interface)
-      {
-        return interface.get_prefix_name() == wheel_name &&
-               interface.get_interface_name() == interface_name;
-      });
+      { return interface.get_name() == wheel_name + "/" + interface_name; });
 
     if (state_handle == state_interfaces_.cend())
     {
-      RCLCPP_ERROR(logger, "Unable to obtain joint state handle for %s", wheel_name.c_str());
+      auto full_name = wheel_name + "/" + interface_name;
+      RCLCPP_ERROR(logger, "Unable to obtain joint state handle for %s", full_name.c_str());
+      RCLCPP_ERROR(logger, "Available state interfaces:");
+      for (const auto & interface : state_interfaces_)
+      {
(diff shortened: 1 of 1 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -673,28 +673,37 @@ controller_interface::CallbackReturn DiffDriveController::configure_side(
...
-      RCLCPP_ERROR(logger, "Unable to obtain joint state handle for %s", wheel_name.c_str());
+      auto full_name = wheel_name + "/" + interface_name;
+      RCLCPP_ERROR(logger, "Unable to obtain joint state handle for %s", full_name.c_str());
+      RCLCPP_ERROR(logger, "Available state interfaces:");
+      for (const auto & interface : state_interfaces_)
+      {
+        RCLCPP_ERROR(
+          logger, " - %s: %s", interface.get_prefix_name().c_str(),
+          interface.get_interface_name().c_str());
+      }
```
Juliaj, line 688: I tested the fix above and it works with example_16. However, I wonder these debug message could lead to confusion if the developer is unaware about the chain created under the hood.  For example_16, prefix: interface pair is `pid_controller_left_wheel_joint` : `left_wheel_joint/velocity`.  

A naive alternative could be: 
- preserve the prefix and interface as in the current definition
- introduce a mode (e.g. chained_right) with additional params and logic to handle the use case.

Juliaj: Thought about this a bit more. Please ignore my comments on `alternative fix` above because I believe it is supported by default. This is illustrated by demo [example_12](https://github.com/ros-controls/ros2_control_demos/blob/master/example_12/bringup/config/rrbot_chained_controllers.yaml#L29-L35) where the `forward_position_controller` controls  "joints" exposed by the second level controller. These (virtual ?) joints are configured in a hierarchical pattern. 
```
# Third-level controllers
forward_position_controller:
  ros__parameters:
    type: forward_command_controller/ForwardCommandController
    joints:
      - position_controller/joint1_position_controller/joint1
      - position_controller/joint2_position_controller/joint2
    interface_name: position
```

Juliaj (approved):
LGTM. This was tested as part of demo example_16.

PR #1517, opened 2025-02-03
[JTC]: Abort goal on deactivate

urfeex:
Currently, when deactivating the JTC once a trajectory is being executed leads to the following problems:

- The trajectory doesn't stop
- The trajectory doesn't end in a result, as it doesn't get stopped.
- The controller's shell gets flooded with `[ros2_control_node-1] [WARN] [1737719032.541174555] [realtime_tools]: goal_handle attempted invalid transition from state EXECUTING with event CANCELED, at ./src/rcl_action/goal_handle.c:95` (See https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/1250)

I narrowed that down to the `RealtimeServerGoalHandle` calling [`gh_->canceled`](https://github.com/ros-controls/realtime_tools/blob/6b98fc112df38ec1e138063d60787ef6f9445d78/include/realtime_tools/realtime_server_goal_handle.hpp#L156) which is supposed to be called only, once the goal is already **cancelling** according to its [documentation](https://docs.ros2.org/latest/api/rclcpp_action/classrclcpp__action_1_1ServerGoalHandle.html#a146e57094758d960c84851d7118e9549). The transition doesn't succeed and hence that message gets repeated in every cycle of the non-realtime-thread, and the action never terminates.

I've added a test for that event in this PR and a possible solution: Calling `setAborted` instead of `setCanceled` in the JTC when it is being deactivated does work. I also think, this is a valid feedback to the user, as the goal isn't preempted by the user but it is aborted due to the controller going down.

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_actions.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_actions.cpp
@@ -1014,7 +1014,8 @@ TEST_P(TestTrajectoryActionsTestParameterized, deactivate_controller_aborts_acti
   controller_thread.join();
 
   EXPECT_TRUE(gh_future.get());
-  EXPECT_EQ(rclcpp_action::ResultCode::CANCELED, common_resultcode_);
+  EXPECT_EQ(rclcpp_action::ResultCode::ABORTED, common_resultcode_);
+
   auto state = traj_controller_->get_state_reference();
 
   // run an update
--- a/joint_trajectory_controller/test/test_trajectory_actions.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_actions.cpp
@@ -969,6 +969,116 @@ TEST_P(TestTrajectoryActionsTestParameterized, test_allow_nonzero_velocity_at_tr
   EXPECT_TRUE(gh_future.get());
 }
 
+TEST_P(TestTrajectoryActionsTestParameterized, deactivate_controller_aborts_action)
+{
+  // deactivate velocity tolerance
+  std::vector<rclcpp::Parameter> params = {
+    rclcpp::Parameter("constraints.stopped_velocity_tolerance", 0.0)};
+  SetUpExecutor(params, false, 1.0, 0.0);
+
+  // We use our own hardware thread here, as we want to make sure the controller is deactivated
+  auto controller_thread = std::thread(
+    [&]()
+    {
+      // controller hardware cycle update loop
+      auto clock = rclcpp::Clock(RCL_STEADY_TIME);
+      auto now_time = clock.now();
+      auto last_time = now_time;
+      rclcpp::Duration wait = rclcpp::Duration::from_seconds(0.5);
+      auto end_time = last_time + wait;
+      while (clock.now() < end_time)
+      {
+        now_time = clock.now();
+        traj_controller_->update(now_time, now_time - last_time);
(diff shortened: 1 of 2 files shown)


joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -969,6 +969,117 @@ TEST_P(TestTrajectoryActionsTestParameterized, test_allow_nonzero_velocity_at_tr
...
+      auto now_time = clock.now();
+      auto last_time = now_time;
+      rclcpp::Duration wait = rclcpp::Duration::from_seconds(0.5);
+      auto end_time = last_time + wait;
+      while (clock.now() < end_time)
+      {
+        now_time = clock.now();
+        traj_controller_->update(now_time, now_time - last_time);
+        last_time = now_time;
+      }
```
christophfroehlich, line 994: Shouldn't we use a fixed period instead of a while loop?
I see that we have that in `SetUpControllerHardware()`, but is this a good idea?

christophfroehlich (changes requested):
Thanks for this fix. 
But the test is failing in every CI job, does it succeed locally?

`setCanceled()` is also called in `preempt_active_goal()`, should we also change it there?

joint_trajectory_controller/test/test_trajectory_actions.cpp
urfeex: I agree, a fixed period would be more appropriate.

urfeex:
> Thanks for this fix. But the test is failing in every CI job, does it succeed locally?

Yes, they were succeeding locally, I will have a look.

> `setCanceled()` is also called in `preempt_active_goal()`, should we also change it there?

That seems reasonable, I will check that.

urfeex:
> `setCanceled()` is also called in `preempt_active_goal()`, should we also change it there?

Things seem to be a bit more complicated there.

From my understanding, we call `preempt_active_goal()` when [a new goal is accepted](https://github.com/ros-controls/ros2_controllers/blob/694d6f184b91d7d4c6f52052e298aa106a219253/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1175). Inside there, we use `setCanceled()` to set request the target state canceled. This would be handled by the next timer callback for `runNonRealtime()` but this is never happening, as in the same action callback we [reset the controller's timer](https://github.com/ros-controls/ros2_controllers/blob/694d6f184b91d7d4c6f52052e298aa106a219253/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1195).

If I add a manual call to `runNonRealtime()` as part of `preempt_active_goal()`, I get the expected output.

Trajectory replacement seems to be working fine and as expected, however the overwritten trajectory doesn't return until the overwriting trajectory finished. Then, it reports `CANCELED`. I noticed, that the `succeed` event is the first from the `runNonRealtime()` function that gets triggered, which is why this probably triggers the old action being canceled in the  underlying (non-realtime) action-server.

For the time being I would suggest to restrict this PR to the `deactivate` problem, as this is related to https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/1250 and we create an issue for the replacement separately.

saikishor (approved):
LGTM as well.
Thanks for adding a proper test :)

PR #1525, opened 2025-02-04
[JTC] Rename open_loop_control parameter and slightly change its scope

### LINKED ISSUE #1505: [JTC] Parameter open_loop_control should be the default
I am doing trajectory tracking with a Kinova Arm and I send a new trajectory every 0.1s. This only works with `open_loop_control` set to `true` because otherwise the current state is taken as the start for the interpolation. This is weird to me as it is always desired to update from the current desired state and not from the actual state as you are updating the desired trajectory and not the actual trajectory. The controller is also still doing closed loop control with respect to the desired state right? 

Can you explain the thought behind updating from the actual state and why this is useful? Maybe I am mistaken. 

I think the `open_loop_control` parameter should be `true` by default and maybe renamed as there still is a closed control loop if you set it to true. 

In case I am mistaken, feel free to clear things up :) Thanks for your work.

Thieso:
Resolves #1505


joint_trajectory_controller/doc/parameters_context.yaml
```diff
@@ -16,3 +14,6 @@ gains: |
   i.e., the shortest rotation to the target position is the desired motion.
   Otherwise  :math:`e = s_d - s` is used, with the desired position :math:`s_d` and the measured
   position :math:`s` from the state interface.
+
+  If you want to turn of the PIDs (open loop control), set the gains to zero.
```
christophfroehlich, line 18:
```suggestion
  If you want to turn off the PIDs (open loop control), set the feedback gains to zero.
```
`ff_velocity_scale` has to be set to 1.0 in this case, we probably should change the default value.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -44,15 +44,15 @@ joint_trajectory_controller:
     default_value: false,
     description: "Allow joint goals defining trajectory for only some joints.",
   }
-  open_loop_control: {
+  interpolate_from_desired_state: {
     type: bool,
-    default_value: false,
-    description: "Use controller in open-loop control mode
+    default_value: true,
```
christophfroehlich, line 49: we should leave this default to false, otherwise we change the behavior of existing setups
```suggestion
    default_value: false,
```

christophfroehlich (changes requested):
Thanks for your PR. However, we must not change the default behavior of existing setups. Your change would go into the stable jazzy release as well.

1. please keep the old `open_loop` parameter and add a deprecation warning if it is set to the non-default value (`true`)
2. if `open_loop` is set to true, we still should deactivate the `use_closed_loop_pid_adapter_`.
3. Mark deprecated code clearly, so we can remove that in the release after next.
4. Please add a migration note and release note [here](https://github.com/ros-controls/ros2_controllers/tree/master/doc).

Thieso:
Thanks for the comments. This should be better now.

christophfroehlich (approved):
Thanks for the follow-up, but please use pre-commit the next time (fixed that for you).

christophfroehlich (approved):

christophfroehlich:
@thieso as you are currently working on the open-loop mode, could you have a look on #780 please and approve it if it makes sense for you? Then, please change also all occurances of `open-loop` in the code comments in this PR here ;)

Thieso:
> @Thieso as you are currently working on the open-loop mode, could you have a look on #780 please and approve it if it makes sense for you? Then, please change also all occurances of `open-loop` in the code comments in this PR here ;)

Hi, I took a look and tested it on my example and it actually fixes another bug which I was not yet able to find. So it looks good to me but what exactly do you mean by replacing `open-loop` occurances?

christophfroehlich:
can you leave a review in the other pr please? 
i just mean that in the comments of some tests there is written `open-loop`, and after your changes this does not make sense any more

bmagyar (approved):

saikishor (approved):
LGTM

christophfroehlich (approved):

christophfroehlich:
@Thieso as this is merged, could you please open the follow-up PR removing the deprecated code? We then can merge that as soon as we branch for jazzy.

PR #1527, opened 2025-02-05
Use global cmake macros and fix gcc-10 build

christophfroehlich:
The change is motivated from the fact that we added a compiler option, which was introduced with gcc-11. But, debian11 (the humble tier 3), has only gcc-10 per default. (We don't have CI jobs for compatibility of the rolling version on humble debian/rhel.)

```
#24 258.0 --- stderr: joint_state_broadcaster
#24 258.0 cc1plus: error: '-Werror=range-loop-construct': no option '-Wrange-loop-construct'
```

So I added a switch depending on the gcc version to add this option or not. To make the options simpler to handle in the future, I propose to add a global cmake macro instead.


mathias-luedtke:
@christophfroehlich: Careful, AFAIK bloom does not "bundle" files outside of the package folders.

christophfroehlich:
oh, thanks for the hint. This means that the packages won't be built on the build farm right?

christophfroehlich:
it's not the tier 3 I'm concerned about, I have a personal interest due to one of my projects at work ;) but could solve it differently, sure thing. 
and the cmake file was just an idea in making the settings global, independent of the actual issue.

ros2_controllers-not-released.rolling.repos
```diff
@@ -4,3 +4,7 @@ repositories:
 #     type: git
 #     url: git@github.com:<some_github_namespace>/<some_ros_package>.git
 #     version: master
+  ros2_control_cmake:
+    type: git
+    url: https://github.com/ros-controls/ros2_control_cmake.git
+    version: cmake
```
christophfroehlich, line 10:
```suggestion
    version: master
```

ros2_controllers.rolling.repos
```diff
@@ -19,3 +19,7 @@ repositories:
     type: git
     url: https://github.com/ros-controls/kinematics_interface.git
     version: master
+  ros2_control_cmake:
+    type: git
+    url: https://github.com/ros-controls/ros2_control_cmake.git
+    version: cmake
```
christophfroehlich, line 25:
```suggestion
    version: master
```

ros2_controllers-not-released.jazzy.repos
```diff
@@ -4,3 +4,7 @@ repositories:
 #     type: git
 #     url: git@github.com:<some_github_namespace>/<some_ros_package>.git
 #     version: master
+  ros2_control_cmake:
+    type: git
+    url: https://github.com/ros-controls/ros2_control_cmake.git
+    version: cmake
```
christophfroehlich, line 10:
```suggestion
    version: master
```

saikishor (approved):
Looks great

PR #1528, opened 2025-02-05
Fix the exported interface naming in the chainable controllers

saikishor:
Fix the exported interfaces naming from the chainable controllers to have a proper standard interface name so that they can chain easily with other controllers without major changes.

Needs: https://github.com/ros-controls/ros2_control/pull/2038


christophfroehlich (dismissed):
Changes LGTM, and tested successfully with https://github.com/ros-controls/ros2_control_demos/pull/710

Juliaj (approved):
LGTM.

christophfroehlich (approved):
Great!

PR #1529, opened 2025-02-06
Fix reference in chained diff drive controller

tpoignonec:
Hi,

There seems to be an issue with the new chained diff drive implementation from PR #1485.

I initially had an issue when chaining this (`malloc(): unaligned tcache chunk detected` after loading the upstream controller). However, while hunting down the error, I found that the cleanup phase is a more easily replicable scenario to get the same problem.
The bug can be reproduced following [these steps](https://github.com/tpoignonec/test_chained_diff_drive_controller )), but it boils down to dereferencing of the `reference_interfaces_` data in `DiffDriveController::reset_buffers()`: 

https://github.com/ros-controls/ros2_controllers/blob/5486fcd2a651e0f6ca9a1827d5d2871917b99c2c/diff_drive_controller/src/diff_drive_controller.cpp#L619-L627

This method is called at different points in the controller: `on_configure()`, `on_error()`, and `on_cleanup`. Each time, the previously exported reference interface is invalidated.

The `reset_buffer` method should not assign a new vector, at it invalidates the references. 
As far as I can see, there are only two options:

```cpp
// Keeps size, just updates values
reference_interfaces_.assign(
  reference_interfaces_.size(),

// or just use std::fill
std::fill(
  reference_interfaces_.begin(),
  reference_interfaces_.end(),
```

In the PR, I went with the `std::fill(...)` option, not sure if there is any argument for one against the other...

Additionally, I moved the resize to the `on_configure()` method to make sure the memory is allocated right away. Otherwise, some tests break (`reference_interfaces_` not allocated at first use).
Are there any guidelines about that?
That seems to be the way it is done in ros2_control_demos ([example_12](https://github.com/ros-controls/ros2_control_demos/blob/332ede0ee44f9c3382666df91a0b7d49a368652f/example_12/controllers/src/passthrough_controller.cpp#L82-L86)), but not in the steering_controllers_library.cpp:

https://github.com/ros-controls/ros2_controllers/blob/5486fcd2a651e0f6ca9a1827d5d2871917b99c2c/steering_controllers_library/src/steering_controllers_library.cpp#L311-L328

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -618,7 +618,9 @@ bool DiffDriveController::reset()
 
 void DiffDriveController::reset_buffers()
 {
-  reference_interfaces_ = std::vector<double>(2, std::numeric_limits<double>::quiet_NaN());
+  std::fill(
+    reference_interfaces_.begin(), reference_interfaces_.end(),
+    std::numeric_limits<double>::quiet_NaN());
   // Empty out the old queue. Fill with zeros (not NaN) to catch early accelerations.
   std::queue<std::array<double, 2>> empty;
   std::swap(previous_two_commands_, empty);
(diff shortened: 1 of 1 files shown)


saikishor (approved):
Thanks for fixing it. 
LGTM

christophfroehlich (approved):
Thanks, LGTM!

PR #1534, opened 2025-02-11
Cleanup wrong lifecycle transitions in tests and unnecessary checks

### LINKED ISSUE #405: Tests that call on_configure without a successful on_init
In writing PRs for the parameter change I realized that many tests depend on being able to call on_configure before ever calling on_init or when they have called on_init and it has failed. Here is an example:

https://github.com/ros-controls/ros2_controllers/blob/4138b734c3c1389fd715408ead3d24485d3e4c25/joint_state_broadcaster/test/test_joint_state_broadcaster.cpp#L156

### LINKED ISSUE #417: Cleanup tests for controllers to not call `on_configure` if `on_init` fails
It guards against on_configure being called and resulting in a segfault in the case where on_init failed and param_listener_ is not yet valid.

_Originally posted by @tylerjw in https://github.com/ros-controls/ros2_controllers/pull/395#discussion_r947001737_

Then also cleanup the code from guards.

### LINKED ISSUE #1501: diff_drive_controller: lifecycle checks in update method
This was in the old update() function and I kept it in because without it I failed tests (eg. "cleanup" in test_diff_drive_controller.cpp) that were calling update after deactivation. Whether or not to keep it depends on your philosophy: how much do you trust the external controller manager to not call update after deactivating the node?

_Originally posted by @arthurlovekin in https://github.com/ros-controls/ros2_controllers/pull/1485#discussion_r1915271862_

christophfroehlich:
- We can assume that only valid transitions are performed (on_configure only if on_init succeeded)
- CM calls the update methods of controllers only in active state, so we don't have to check that again, and we also should not do that in the tests.

Closes #417 and closes #405 and closes #1501

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -101,15 +101,6 @@ controller_interface::return_type DiffDriveController::update_reference_from_sub
   const rclcpp::Time & time, const rclcpp::Duration & /*period*/)
 {
   auto logger = get_node()->get_logger();
-  if (get_lifecycle_state().id() == State::PRIMARY_STATE_INACTIVE)
-  {
-    if (!is_halted)
-    {
-      halt();
-      is_halted = true;
-    }
-    return controller_interface::return_type::OK;
-  }
 
   const std::shared_ptr<TwistStamped> command_msg_ptr = *(received_velocity_msg_ptr_.readFromRT());
 
@@ -149,15 +140,6 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
   const rclcpp::Time & time, const rclcpp::Duration & period)
 {
   auto logger = get_node()->get_logger();
-  if (get_lifecycle_state().id() == State::PRIMARY_STATE_INACTIVE)
-  {
-    if (!is_halted)
-    {
-      halt();
-      is_halted = true;
-    }
-    return controller_interface::return_type::OK;
-  }
 
   // command may be limited further by SpeedLimit,
   // without affecting the stored twist command
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -544,7 +544,6 @@ controller_interface::CallbackReturn DiffDriveController::on_activate(
     return controller_interface::CallbackReturn::ERROR;
   }
 
-  is_halted = false;
   subscriber_is_active_ = true;
 
   RCLCPP_DEBUG(get_node()->get_logger(), "Subscriber and publisher are now active.");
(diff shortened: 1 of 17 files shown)


saikishor (dismissed):

bmagyar (approved):

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -101,15 +101,6 @@ controller_interface::return_type DiffDriveController::update_reference_from_sub
   const rclcpp::Time & time, const rclcpp::Duration & /*period*/)
 {
   auto logger = get_node()->get_logger();
-  if (get_lifecycle_state().id() == State::PRIMARY_STATE_INACTIVE)
-  {
-    if (!is_halted)
```
saikishor, line 106: Do we also need to remove the is_halted variable, if it is a member variable?

tricycle_controller/src/tricycle_controller.cpp
```diff
@@ -392,6 +382,11 @@ CallbackReturn TricycleController::on_activate(const rclcpp_lifecycle::State &)
 CallbackReturn TricycleController::on_deactivate(const rclcpp_lifecycle::State &)
 {
   subscriber_is_active_ = false;
+  if (!is_halted)
+  {
+    halt();
+    is_halted = true;
+  }
```
saikishor, line 389: @christophfroehlich why do we need this here?

diff_drive_controller/src/diff_drive_controller.cpp
christophfroehlich: we use it in on_deactivate() transition

tricycle_controller/src/tricycle_controller.cpp
christophfroehlich: to stop the vehicle in on_deactivate transition

diff_drive_controller/src/diff_drive_controller.cpp
saikishor: Can't we simply call halt in the deactivate method always, instead of using this is_halted variable?

What do you think?

I can recheck the code later and get back to you

tricycle_controller/src/tricycle_controller.cpp
saikishor: Same as above, can't we simply call halt without any check in the deactivate method?

diff_drive_controller/src/diff_drive_controller.cpp
christophfroehlich: I think we could, sure.

saikishor (approved):
Thanks for cleaning things.
Looks very good to me. 

Great job 👏🏾

PR #1535, opened 2025-02-12
Add omni_wheel_drive_controller

Amronos:
This is a controller for mobile robots with omnidirectional drive. It supports using 3 or more omni wheels spaced at an equal angle from each other in a circular formation.

I have added many tests similar to other controllers and updated the docs.
Please review this controller and tell me if you have any doubts regarding its working.

I will be updating [mobile_robot_kinematics.rst](https://github.com/ros-controls/ros2_controllers/blob/master/doc/mobile_robot_kinematics.rst) for the kinematics of this controller and for that wanted to ask what software has been used to make the diagrams in that doc. :smile:

PR #1551, opened 2025-02-20
Add omni drive robot to mobile_robot_kinematics

Amronos:
…sing Omni Wheels

Updated ``mobile_robot_kinematics.rst`` as asked in [#1535](https://github.com/ros-controls/ros2_controllers/pull/1535#pullrequestreview-2629212240).


christophfroehlich (changes requested):
Thanks. I just found another thing: $$R$$ should be measured to the center of the wheel, right? compare to the wheeltrack of the DiffDrive sketch, maybe you can add black dimension lines (and leave the green ones only for velocity, angular rate).

Amronos:
>  R should be measured to the center of the wheel, right?

Nope! It's from the center of the robot to the point where the wheel touches the robot as shown by the green arrow.

I have represented the velocities of the robot in blue and things related to measurements in green.

christophfroehlich (changes requested):
> > R should be measured to the center of the wheel, right?
> 
> Nope! It's from the center of the robot to the point where the wheel touches the robot as shown by the green arrow.

Nope, according to the literature it is the center of the wheel, not the innermost point. See this screenshot of the ModernRobotics book, which is linked on the page.
image

> I have represented the velocities of the robot in blue and things related to measurements in green.

What about this style? It has similar style as the other sketches.

omni_wheel_omnidirectional_drive
[omni_wheel_omnidirectional_drive.zip](https://github.com/user-attachments/files/18894230/omni_wheel_omnidirectional_drive.zip)

Amronos:
> Nope, according to the literature it is the center of the wheel, not the innermost point. See this screenshot of the ModernRobotics book, which is linked on the page.


Have a look at this screenshot from the same book, look at the symbol d (it is the radius). The screenshot that you shared just represents how a mecanum wheel works.

> What about this style? It has similar style as the other sketches.

I am okay with it, though I liked mine more.

christophfroehlich:
> Have a look at this screenshot from the same book, look at the symbol d (it is the radius). The screenshot that you shared just represents how a mecanum wheel works.

The arrows in Fig 13.5 are not very technical but only stylistic. Insert the respective values of $$\gamma_i,\beta_i,x_i,y_i$$ in (13.8) and you will get (13.9), where $$(x_i,y_i)$$ is the center of the wheel in $`\{b \}`$.

To simplify this: if velocities are zero we have $$\omega_1=\omega_2=\omega_3=- \frac{R \omega_{b,z}}{r}$$. This will not change with the "width" of the wheel (as we actually assume zero width), as it would be the case with your definition.

> > What about this style? It has similar style as the other sketches.
> 
> I am okay with it, though I liked mine a bit more. Please commit whatever you think would be the best and merge the PR, I won't be able to make any commits for a few days.

I don't have a preference in either one, but I like it to be consistent with the others.

christophfroehlich (approved):

Amronos:
@christophfroehlich from my testing in both simulation and real life of the controller, the radius when measured from the robot's center to the innermost point of the wheel yields the correct linear and angular velocities of the robot. But when taking the radius as from the robot's center to the wheel's center the linear and angular velocities of the robot are completely different from the commanded velocities.
I am not too sure on why this is the case, but I think we should measure the radius from the robot's center to the wheel's innermost point.

Amronos:
@christophfroehlich 
https://journal.umy.ac.id/index.php/jrc/article/view/10664/5908 suggests the following:
- $R$ is the distance from the wheel to the center of the robot and r is the radius of the omni wheel.
-
image

-
image

https://www.researchgate.net/publication/324929202_Motion_Improvement_of_Four-Wheeled_Omnidirectional_Mobile_Robots_for_Indoor_Terrain/fulltext/5aebadf2458515f59981e3b1/Motion-Improvement-of-Four-Wheeled-Omnidirectional-Mobile-Robots-for-Indoor-Terrain.pdf suggests the following:
-  $R$ is the radius of the robot from the center to the wheel.
- 
image

- 
image

And from my testing in both simulation and real life:
- When $R$ is the distance from the wheel to the center of the robot, the robot accurately moves at the commanded velocities.
- When $R$ is the distance from the center of the wheel to the center of the robot, the robot's movement is very different from the commanded velocities.

So I think we should consider $R$ as the distance from the point where the wheel touches the robot to the robot's center.

christophfroehlich:
But then, the same argument should hold for simple diffdrive robots, where the wheel track is _the distance between the centerline of two wheels_ ([wikipedia](https://en.wikipedia.org/wiki/Axle_track))? or acc. to modern robotics:
image

christophfroehlich:
and continuing throwing publications around (these are from a high-ranked robotics journal, with the distance always counted from the center of the wheel):

- https://ieeexplore.ieee.org/document/5764857
- https://ieeexplore.ieee.org/document/4768681

christophfroehlich:
No worries, I am always open for a constructive and respectful discussion. I'm glad that I could help!

Amronos:
@christophfroehlich After further testing it seems like taking the radius from the robot center to the outermost point of the wheel gives the most accuracy possible, and this time I made sure that I was not repeating the same mistakes in my testing. I am sorry!

PR #1554, opened 2025-02-23
Add GPSBroadcaster

Wiktor-99:
As in the title. I'm open to any suggestions.


gps_sensor_broadcaster/src/gps_sensor_broadcaster.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+  return callback_return_type::SUCCESS;
+}
+catch (const std::exception & e)
+{
+  fprintf(
+    stderr, "Exception thrown during publisher creation at configure stage with message: %s \n",
+    e.what());
+  return callback_return_type::ERROR;
+}
```
saikishor, line 91:
```suggestion
callback_return_type GPSSensorBroadcaster::setup_publisher()
{
try
{
  sensor_state_publisher_ = get_node()->create_publisher<sensor_msgs::msg::NavSatFix>(
    "~/gps/fix", rclcpp::SystemDefaultsQoS());
  realtime_publisher_ = std::make_unique<StatePublisher>(sensor_state_publisher_);
  realtime_publisher_->lock();
  realtime_publisher_->msg_.header.frame_id = params_.frame_id;
  setup_covariance();
  realtime_publisher_->unlock();

  return callback_return_type::SUCCESS;
}
catch (const std::exception & e)
{
  fprintf(
    stderr, "Exception thrown during publisher creation at configure stage with message: %s \n",
    e.what());
  return callback_return_type::ERROR;
}
}
```

Missing enclosing parenthesis

Timple:
Conceptual question:

We also have [battery_state_broadcaster](https://docs.ros.org/en/jazzy/p/battery_state_broadcaster/). Now we get a `gps_broadcaster`.

Now I can imagine we get a lot of future PR's creating packages for all items in this list: [sensor_msgs](https://docs.ros2.org/foxy/api/sensor_msgs/index-msg.html).

Would it make sense to template this such that we can simply send any message from sensor_msgs?

Wiktor-99:
@Timple Perhaps you're right; it could be a template class. However, since all sensors would use a different set of state interfaces, we would probably need to create a bunch of new semantic components (which can be done).

Also, it looks like this decision was already settled. We've got pose/IMU/range broadcasters, and templating them would affect existing configs.

gps_sensor_broadcaster/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
+add_compile_options(-Wall -Wextra -Wpedantic -Werror=conversion -Werror=unused-but-set-variable
+                    -Werror=return-type -Werror=shadow -Werror=format -Werror=range-loop-construct
+                    -Werror=missing-braces)
+endif()
+
+# using this instead of visibility macros
+# S1 from https://github.com/ros-controls/ros2_controllers/issues/1053
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
christophfroehlich, line 12: See https://github.com/ros-controls/ros2_controllers/pull/1527, needs also the dependency in the package.xml of course.
```suggestion
find_package(ros2_control_cmake REQUIRED)
set_compiler_options()
export_windows_symbols()
```

gps_sensor_broadcaster/include/gps_sensor_broadcaster/gps_sensor_broadcaster.hpp
```diff
@@ -0,0 +1,69 @@
...
+  callback_return_type on_init() override;
+  callback_return_type on_configure(const rclcpp_lifecycle::State & previous_state) override;
+  callback_return_type on_activate(const rclcpp_lifecycle::State & previous_state) override;
+  callback_return_type on_deactivate(const rclcpp_lifecycle::State & previous_state) override;
+  controller_interface::return_type update(const rclcpp::Time &, const rclcpp::Duration &) override;
+
+protected:
+  using GPSSensorOption = semantic_components::GPSSensorOption;
+  using GPSSensorVariant = std::variant<
+    std::monostate, semantic_components::GPSSensor<GPSSensorOption::WithCovariance>,
```
christophfroehlich, line 52: why do we need the `std::monostate` here? When will the variant be used before we call `on_configure`? can't we just remove it, and get along without the visitors?

christophfroehlich (changes requested):
Thanks for the follow-up to the semantic component. But I have some comments

btw: To fix the docs workflow, please add the userdoc.rst [here](https://github.com/ros-controls/ros2_controllers/blob/master/doc/controllers_index.rst#broadcasters).

gps_sensor_broadcaster/include/gps_sensor_broadcaster/gps_sensor_broadcaster.hpp
Wiktor-99: There's no issue with calling, there is no default ctor so we cannot avoid this. In other broadcasters we use pointers so we don't run into this scenario.

gps_sensor_broadcaster/CMakeLists.txt
Wiktor-99: Done

gps_sensor_broadcaster/include/gps_sensor_broadcaster/gps_sensor_broadcaster.hpp
christophfroehlich: Got it, thanks!

christophfroehlich (approved):
LGTM

saikishor (approved):

PR #1555, opened 2025-02-23
Cleanup dependencies (add `angles`, rm `filters`)

saikishor:
Might fix the failing job: https://build.ros2.org/job/Hbin_uJ64__admittance_controller__ubuntu_jammy_amd64__binary/155/consoleFull#console-section-15

It could be that one of the dependencies of the package removed this dependency and that's why we are noticing it now

### DIFF
--- a/admittance_controller/package.xml
+++ b/admittance_controller/package.xml
@@ -27,20 +27,19 @@
   <depend>control_msgs</depend>
   <depend>control_toolbox</depend>
   <depend>controller_interface</depend>
-  <depend>kinematics_interface</depend>
-  <depend>filters</depend>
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
Thanks. This was silently included with JTC before https://github.com/ros-controls/ros2_controllers/pull/1532
I rechecked the deps and removed the `filters` dependency, we only use a header file of `control_toolbox`.

saikishor:
> Thanks. This was silently included with JTC before #1532 I rechecked the deps and removed the `filters` dependency, we only use a header file of `control_toolbox`.

LGTM

PR #1560, opened 2025-02-28
Fix multi-time trajectory controller failing activate

bijoua29:
Fix multi-time trajectory controller failing activate when feedback is not available
- fix by delaying initialization to update() if necessary


saikishor (commented):
Closed by the author :)

PR #1564, opened 2025-03-02
fixed on_set_chained_mode implementations to avoid cpplint warnings

bhagyeshagresar:
Fixed warnings generated by https://github.com/cpplint/cpplint/issues/131 on some of the controllers that implement the chained mode.


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -689,8 +689,9 @@ controller_interface::CallbackReturn DiffDriveController::configure_side(
 
 bool DiffDriveController::on_set_chained_mode(bool chained_mode)
 {
-  // Always accept switch to/from chained mode (without linting type-cast error)
-  return true || chained_mode;
+  // Fix to adhere to CppLint standards
```
bmagyar, line 692: A motor reference isn't really adhering to cppkint standards ;)

bmagyar: You can achieve the same by commenting out the name of the argument. It tricks the parser and also the compiler. We already use that approach in the codebase, just look for "/*" in ros2_control

bmagyar (changes requested):
Thank you for the contribution! Great that you found out this. Please take a look at the solution I recommended.

diff_drive_controller/src/diff_drive_controller.cpp
christophfroehlich: No, I don't think so. See https://github.com/ros-controls/ros2_controllers/issues/1491 and the linked cpplint issue.

saikishor (approved):

PR #1565, opened 2025-03-02
Use new handles API in diff_drive_controller

saikishor:
This PR applies the new API changes in the handles of ros2_control in the diff_drive_controller

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -167,9 +167,19 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
     double right_feedback_mean = 0.0;
     for (size_t index = 0; index < static_cast<size_t>(wheels_per_side_); ++index)
     {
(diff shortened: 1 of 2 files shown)


christophfroehlich (approved):
Great!

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -167,9 +167,19 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
+      {
+        RCLCPP_DEBUG(logger, "Unable to retrieve the data from the left or right wheels feedback!");
+        return controller_interface::return_type::OK;
+      }
+
+      const double left_feedback = left_feedback_op.value();
+      const double right_feedback = right_feedback_op.value();
 
       if (std::isnan(left_feedback) || std::isnan(right_feedback))
       {
```
Amronos, line 185: Should this be removed now?

saikishor: Nope, this should stay as the handle may still have NaNs that might be set by hardware component

Amronos (approved):

PR #1568, opened 2025-03-04
[ForwardCommandController] Fix the duplicate command interface types when reconfiguring the controller

saikishor:
if we reconfigure the controller now, it simply appends and creates duplicates. Found the issue with : https://github.com/ros-controls/ros2_control/pull/2090


christophfroehlich (approved):
lgtm

PR #1570, opened 2025-03-04
Fix the `ActivateWithWrongJointsNamesFails` test

christophfroehlich:
It came up with the fix introduced with #1568


christophfroehlich:
@mergifyio backport mergify/bp/humble/pr-1568

saikishor (approved):

saikishor:
@christophfroehlich thanks for taking care of the fix

PR #1573, opened 2025-03-05
[mecanum_drive_controller] Fix Odometry Initialization 

Juliaj:
This PR addresses an issue where the odometry object wasn't initialized properly in the mecanum_drive_controller. The odometry_.init() call was missing, which the `base_frame_offset_` could contain random values, leading to incorrect calculations in the transformation matrix. Specifically, this affects the line:

```
tf2::Vector3 linear_transformation_from_center_2_base =
    angular_transformation_from_center_2_base *
    tf2::Vector3(-base_frame_offset_[0], -base_frame_offset_[1], 0.0);
```
and causes `velocity_in_base_frame_linear_x` and `velocity_in_base_frame_linear_y` to grow unbound.  A test case was added to guard against this issue.


christophfroehlich (changes requested):
I see your point of the missing init. But the test is succeeding also without the init call. Can you verify that?

Juliaj:
> I see your point of the missing init. But the test is succeeding also without the init call. Can you verify that? 

It is because that issue is related to accessing uninitialized memory. Even though it is fairly repro,  it is not 100% reproducible.  I  added explicit steps to verify the base_frame_offset is set properly.

christophfroehlich (approved):
Thanks!

saikishor (approved):
LGTM

PR #1575, opened 2025-03-06
[mecanum_drive_controller] Fix `mergify/bp/humble/pr 512`

soham2560:
## Brief
In this PR
- Completed Humble backport of #1376 

Note: Refer [this comment](https://github.com/ros-controls/ros2_controllers/pull/1376#issuecomment-2702966012) for exact motivation

### DIFF
--- a/mecanum_drive_controller/CMakeLists.txt
+++ b/mecanum_drive_controller/CMakeLists.txt
@@ -2,9 +2,7 @@ cmake_minimum_required(VERSION 3.16)
 project(mecanum_drive_controller LANGUAGES CXX)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
-  add_compile_options(-Wall -Wextra -Wpedantic -Werror=conversion -Werror=unused-but-set-variable
-  -Werror=return-type -Werror=shadow -Werror=format -Werror=range-loop-construct
-  -Werror=missing-braces)
+  add_compile_options(-Wall -Wextra -Wpedantic -Wconversion)
 endif()
 
 # find dependencies
--- a/mecanum_drive_controller/include/mecanum_drive_controller/mecanum_drive_controller.hpp
+++ b/mecanum_drive_controller/include/mecanum_drive_controller/mecanum_drive_controller.hpp
@@ -26,7 +26,7 @@
 #include "controller_interface/chainable_controller_interface.hpp"
 #include "mecanum_drive_controller/odometry.hpp"
 #include "mecanum_drive_controller/visibility_control.h"
-#include "mecanum_drive_controller_parameters.hpp"
+#include <mecanum_drive_controller/mecanum_drive_controller_parameters.hpp>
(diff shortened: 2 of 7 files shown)


soham2560:
I do have some questions about this implementation though
- the odom transforms seems to be publishing to a non-standard topic `~/tf_odometry` 
https://github.com/ros-controls/ros2_controllers/blob/11ed083dbc1126d8db03d90c32395ee18e6e4be4/mecanum_drive_controller/src/mecanum_drive_controller.cpp#L165
- there is no unstamped cmd vel topic, I believe we need that for humble? (also the stamped cmd vel topic is `~/reference` by default)

let me know if anything is to be done about this

christophfroehlich:
> I do have some questions about this implementation though
> 
>     * the odom transforms seems to be publishing to a non-standard topic `~/tf_odometry`
>       https://github.com/ros-controls/ros2_controllers/blob/11ed083dbc1126d8db03d90c32395ee18e6e4be4/mecanum_drive_controller/src/mecanum_drive_controller.cpp#L165
> 
>     * there is no unstamped cmd vel topic, I believe we need that for humble? (also the stamped cmd vel topic is `~/reference` by default)
> 
> 
> let me know if anything is to be done about this

The non-default topics are used by other controllers here to, one has to remap them if to be used in the "standard" way.
About unstamped cmd_vel, you are maybe right. Have a look how the switch is implemented in [diff_drive_controller](https://github.com/ros-controls/ros2_controllers/blob/5de2dfa9a4ffbd64ae6522f8da0d4663e8bb542c/diff_drive_controller/src/diff_drive_controller.cpp#L343-L386) for example.

soham2560:
@christophfroehlich I have added stamped vel support (refer [97e8ba9](https://github.com/ros-controls/ros2_controllers/pull/1575/commits/97e8ba9f96e2940d0ae47b5f2d89b03b5b9e6a43)), can you have a look?

christophfroehlich (approved):
Thanks for the changes. I tested the package locally, it seems to be fine

christophfroehlich:
@soham2560 please install pre-commit and run the checks the next time before opening a PR ;)

soham2560:
> @soham2560 please install pre-commit and run the checks the next time before opening a PR ;)

oh I completely missed that, had run it but it skipped al the files since they were already commited, just checked, should've used `--all-files` flag in that scenario, thanks for the pointer will keep it in mind

soham2560:
>The non-default topics are used by other controllers here to, one has to remap them if to be used in the "standard" way.

@christophfroehlich apologies for commenting here, but is remapping possible on humble if the `spawner` does not support the `--ros-args` arg like in https://github.com/ros-controls/ros2_control/pull/1713

saikishor:
> 
> @christophfroehlich apologies for commenting here, but is remapping possible on humble if the `spawner` does not support the `--ros-args` arg like in https://github.com/ros-controls/ros2_control/pull/1713


You can remap by adding the remapping arg to ros2_control_node.

PR #1578, opened 2025-03-10
Update documentation of rqt_joint_trajectory_controller

AdityaPawar162:
This PR updates the documentation for the rqt_joint_trajectory_controller, as discussed while addressing [this issue](https://github.com/ros-controls/ros2_control_demos/pull/734).


christophfroehlich (changes requested):
image
Thanks for the update, but please fix the pre-commit issue (always install pre-commit in the repositories, and let them check before you commit)

AdityaPawar162:
Hi @christophfroehlich,
Thanks for the feedback! I've run pre-commit run --all-files, and I believe it should work now.

PR #1584, opened 2025-03-11
[JSB] cleanup the activation error message

### LINKED ISSUE #1580: Joint_state_broadcaster throws error when no states exist, even if none are "requested"
By default, joint_state_broadcaster publishes all available states when none are specified in its parameters. If there are no states at all (none present in the urdf), it throws a confusing error message `[joint_state_broadcaster]: None of requested interfaces exist. Controller will not run`. This message implies that there are joint states that are supposed to exist but don't, when in reality there are simply no states to publish

This behavior can be undesirable for example, when a project is in development it can often make sense to first implement open loop control to verify the hardware is working, before moving on to closed loop control. 

I propose replacing this behavior with a simple warning indicating that no states exist to publish

saikishor:
Fixes #1580 

I also cleaned up the other message, as this will never happen, and it would fail much earlier to this activate callback

https://github.com/ros-controls/ros2_control/blob/cc3491c9935b908cba2135fa12f629c7e5063e58/controller_manager/src/controller_manager.cpp#L1916-L1942

### DIFF
--- a/joint_state_broadcaster/src/joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/src/joint_state_broadcaster.cpp
@@ -179,22 +179,13 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_activate(
   if (!init_joint_data())
   {
     RCLCPP_ERROR(
-      get_node()->get_logger(), "None of requested interfaces exist. Controller will not run.");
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
LGTM

Issue #1574, opened 2025-03-06
[CI] `test_force_torque_sensor_broadcaster` regularily times out


christophfroehlich:
```
test_results/force_torque_sensor_broadcaster/test_force_torque_sensor_broadcaster.gtest.xml
2025-03-06T02:46:41.8447062Z 2: [==========] Running 13 tests from 1 test suite.
2025-03-06T02:46:41.8447487Z 2: [----------] Global test environment set-up.
2025-03-06T02:46:41.8447818Z 2: [----------] 13 tests from ForceTorqueSensorBroadcasterTest
2025-03-06T02:46:41.8448266Z 2: [ RUN      ] ForceTorqueSensorBroadcasterTest.SensorName_InterfaceNames_NotSet
2025-03-06T02:46:41.8449075Z 2: [ERROR] [1741229141.866531381] [test_force_torque_sensor_broadcaster]: 'sensor_name' or at least one 'interface_names.[force|torque].[x|y|z]' parameter has to be specified.
2025-03-06T02:46:41.8449907Z 2: [       OK ] ForceTorqueSensorBroadcasterTest.SensorName_InterfaceNames_NotSet (12 ms)
2025-03-06T02:46:41.8450451Z 2: [ RUN      ] ForceTorqueSensorBroadcasterTest.SensorName_InterfaceNames_Set
2025-03-06T02:46:41.8451315Z 2: [ERROR] [1741229141.872119148] [test_force_torque_sensor_broadcaster]: both 'sensor_name' and 'interface_names.[force|torque].[x|y|z]' parameters can not be specified together.
2025-03-06T02:46:41.8452133Z 2: [       OK ] ForceTorqueSensorBroadcasterTest.SensorName_InterfaceNames_Set (5 ms)
2025-03-06T02:46:41.8452710Z 2: [ RUN      ] ForceTorqueSensorBroadcasterTest.SensorName_IsEmpty_InterfaceNames_NotSet
2025-03-06T02:46:41.8453534Z 2: [ERROR] [1741229141.876958103] [test_force_torque_sensor_broadcaster]: 'sensor_name' or at least one 'interface_names.[force|torque].[x|y|z]' parameter has to be specified.
2025-03-06T02:46:41.8454478Z 2: [       OK ] ForceTorqueSensorBroadcasterTest.SensorName_IsEmpty_InterfaceNames_NotSet (5 ms)
[...]

Issue #1559, opened 2025-02-28
[CI] `ForceTorqueSensorBroadcasterTest.InterfaceNames_Configure_Success` sometimes times out


christophfroehlich:
```
2025-02-28T02:40:40.1906356Z 2: Test command: /usr/bin/python3 "-u" "/opt/ros/humble/share/ament_cmake_test/cmake/run_test.py" "/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/build/force_torque_sensor_broadcaster/test_results/force_torque_sensor_broadcaster/test_force_torque_sensor_broadcaster.gtest.xml" "--package-name" "force_torque_sensor_broadcaster" "--output-file" "/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/build/force_torque_sensor_broadcaster/test_results/force_torque_sensor_broadcaster/test_force_torque_sensor_broadcaster.txt" "--command" "/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/build/force_torque_sensor_broadcaster/test_force_torque_sensor_broadcaster" "--ros-args" "--params-file" "/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/src/ros-controls/ros2_controllers/force_torque_sensor_broadcaster/test/force_torque_sensor_broadcaster_params.yaml" "--" "--gtest_output=xml:/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/build/force_torque_sensor_broadcaster/test_results/force_torque_sensor_broadcaster/test_force_torque_sensor_broadcaster.gtest.xml"
2025-02-28T02:40:40.1910196Z 2: Test timeout computed to be: 60
2025-02-28T02:40:40.1910766Z 2: -- run_test.py: invoking following command in '/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/build/force_torque_sensor_broadcaster':
[...]

PR #1586, opened 2025-03-12
[CI] Time out from test_force_torque_sensor_broadcaster 

Juliaj:
This change attempts to debug and reproduce CI failures: https://github.com/ros-controls/ros2_controllers/issues/1574 and https://github.com/ros-controls/ros2_controllers/issues/1559.  

In both issues, it appears that the timeout was caused by the test "InterfaceNames_Configure_Success", which primarily tests the `on_configure` function. No hypothesis has been established so far.
- The code in "InterfaceNames_Configure_Success" is straightforward.
- I’ve reviewed the `on_configure` function in `ForceTorqueSensorBroadcaster` and the changes introduced around the time of the failure. Other than [a change](https://github.com/ros-controls/realtime_tools/pull/280) made in realtime_tools regarding `RTPublisher`, no other changes were found that could impact on_configure.


saikishor (commented):
I've a feeling that it happens mostly or only in Humble. Right now the Rolling compatibility build of Humble is also failed in this PR

https://github.com/ros-controls/ros2_controllers/actions/runs/13800804076/job/38602688612#step:5:1

Juliaj:
> I've a feeling that it happens mostly or only in Humble. Right now the Rolling compatibility build of Humble is also failed in this PR

Setup a docker container with Humble, no repro ...

Juliaj:
@christophfroehlich, I only hit the failure once (as Sai pointed out above) and haven't been able to reproduce it in subsequent attempts. Is there anything else I should try to reproduce this issue?

There are failures in the Coverage build and Rolling Binary Build. The test failure in the coverage build appears to be intermittent. The failures in the rolling binary build seem to be related to a dependency issue. Please let me know if these need further investigation.

@saikishor, the controller loading tests reported warnings related to introspection. I'm unable to determine if this is expected behavior.
```
 1: [ RUN      ] TestLoadAdmittanceController.load_controller
  1: [INFO] [1741819586.401847780] [test_controller_manager]: Using Steady (Monotonic) clock for triggering controller manager cycles.
  1: [INFO] [1741819586.433478203] [test_controller_manager]: Loading hardware 'TestActuatorHardware' 
[...]
  1: [WARN] [1741819586.435269427] [pal_statistics]: Unable to register entity state_interface.joint1/position in ros2_control, as the registry is not found. Try initializing it!
  1: [WARN] [1741819586.435294563] [pal_statistics]: Unable to register entity state_interface.joint1/velocity in ros2_control, as the registry is not found. Try initializing it!
```

saikishor:
> @saikishor, the controller loading tests reported warnings related to introspection. I'm unable to determine if this is expected behavior.
> ```
>  1: [ RUN      ] TestLoadAdmittanceController.load_controller
>   1: [INFO] [1741819586.401847780] [test_controller_manager]: Using Steady (Monotonic) clock for triggering controller manager cycles.
>   1: [INFO] [1741819586.433478203] [test_controller_manager]: Loading hardware 'TestActuatorHardware' 
> [...]
>   1: [WARN] [1741819586.435269427] [pal_statistics]: Unable to register entity state_interface.joint1/position in ros2_control, as the registry is not found. Try initializing it!
>   1: [WARN] [1741819586.435294563] [pal_statistics]: Unable to register entity state_interface.joint1/velocity in ros2_control, as the registry is not found. Try initializing it!
> ```

This needs to be handled at the pal_statistics level. I'll take care of it. 

I've retriggered the CI, let's see if we can reproduce it.

christophfroehlich:
We had some similar timeouts in the control_toolbox repo, fixed with https://github.com/ros-controls/control_toolbox/pull/261
I haven't checked yet, but does this apply here maybe too?

Juliaj:
> We had some similar timeouts in the control_toolbox repo, fixed with [ros-controls/control_toolbox#261](https://github.com/ros-controls/control_toolbox/pull/261) I haven't checked yet, but does this apply here maybe too?

@christophfroehlich , thank you for pointing that out. It’s hard to tell whether this is similar to the one you mentioned above. To reduce the chance, I’ve updated the test teardown logic with more cleanup. I also implemented a timeout mechanism to prevent tests from hanging indefinitely. Could you please review these changes and let me know your thoughts?

saikishor:
@Juliaj I believe we also have a default timeout of 60 secs for the gtest. At least that's what I thought, isn't it the case @christophfroehlich ?

christophfroehlich (approved):
Let's give it a try

christophfroehlich (commented):
For reference: We already added an auto-shutdown in the destructor
https://github.com/ros-controls/ros2_control/blob/95188b92c9bf628e72dc90a671e1f6c00f0b7611/controller_interface/src/controller_interface_base.cpp#L26-L40

At assigning a new unique_ptr with `fts_broadcaster_ = std::make_unique<FriendForceTorqueSensorBroadcaster>();` it should call this transition already.

saikishor (approved):

christophfroehlich:
Ah, I saw now that we haven't backported https://github.com/ros-controls/ros2_control/pull/1979. This might be a reason why it fails more often with the humble version.

force_torque_sensor_broadcaster/test/test_force_torque_sensor_broadcaster.cpp
```diff
@@ -40,15 +40,43 @@ constexpr auto NODE_ERROR = controller_interface::CallbackReturn::ERROR;
...
+      lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED)
+    {
+      if (
+        fts_broadcaster_->get_lifecycle_state().id() ==
+        lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+      {
+        ASSERT_EQ(fts_broadcaster_->on_deactivate(rclcpp_lifecycle::State()), NODE_SUCCESS);
+      }
+      // Clean up the broadcaster
+      ASSERT_EQ(fts_broadcaster_->on_cleanup(rclcpp_lifecycle::State()), NODE_SUCCESS);
```
christophfroehlich, line 73: Shouldn't we have called the lifecycle node's transition calls instead of the callbacks? The lifecycle node won't have the correct state afterwards.

saikishor: You are right!

christophfroehlich: @Juliaj could you fix this in a new PR please?

PR #1587, opened 2025-03-12
Use constructor parameters instead of initializer list

fmauch:
I stumbled upon a two memory issues inside the JTC tests / the framework. This PR is about one of them, I'll open an issue for the other one.

Using the initializer list leads to vectors with two entries [3, NaN] and we get a heap overflow error in the tests as we access 3 elements (one for each joint).


christophfroehlich (approved):
Thanks!

Issue #1588, opened 2025-03-12
heap-use-after-free thrown in tests


fmauch:
**Describe the bug**
Running the JTC tests with ASAN enabled raises a heap-use-after-free error. I stumbled upon this while writing tests for #1191 where I get crashes on **some** configurations. Tracing things down I realized that there are memory issues reported also on the master branch and I would like to get them fixed, first.

```
=================================================================
==197664==ERROR: AddressSanitizer: heap-use-after-free on address 0x5160000bbed8 at pc 0x79ea5f6fb42e bp 0x7ffc58fb3210 sp 0x7ffc58fb29b8
READ of size 15 at 0x5160000bbed8 thread T0
    #0 0x79ea5f6fb42d in memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115
    #1 0x79ea5eef3a85  (/opt/ros/jazzy/lib/librclcpp.so+0xf3a85) (BuildId: a6e4def86fc9f569833617ff077616d6d3344142)
    #2 0x79ea5ef2c0be in rclcpp::get_logger(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (/opt/ros/jazzy/lib/librclcpp.so+0x12c0be) (BuildId: a6e4def86fc9f569833617ff077616d6d3344142)
    #3 0x5965fd023844 in hardware_interface::LoanedStateInterface::~LoanedStateInterface() /home/feex/checkout/ros2_control_dev/colcon_ws/src/ros2_control/hardware_interface/include/hardware_interface/loaned_state_interface.hpp:61
    #4 0x79ea5f5c2713 in void std::_Destroy<hardware_interface::LoanedStateInterface>(hardware_interface::LoanedStateInterface*) /usr/include/c++/13/bits/stl_construct.h:151
    #5 0x79ea5f5c2713 in void std::_Destroy_aux<false>::__destroy<hardware_interface::LoanedStateInterface*>(hardware_interface::LoanedStateInterface*, hardware_interface::LoanedStateInterface*) /usr/include/c++/13/bits/stl_construct.h:163
    #6 0x79ea5f5c2713 in void std::_Destroy<hardware_interface::LoanedStateInterface*>(hardware_interface::LoanedStateInterface*, hardware_interface::LoanedStateInterface*) /usr/include/c++/13/bits/stl_construct.h:196
[...]

PR #1596, opened 2025-03-18
[jtc tests] avoid dangling ref of command / state interfaces

fmauch:
This should avoid the problem mentioned in #1588. This makes the JTC tests run with ASAN without issues, but I haven't tested all the other packages.

### DIFF
--- a/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
@@ -259,6 +259,12 @@ public:
       controller_name_ + "/joint_trajectory", rclcpp::SystemDefaultsQoS());
   }
 
+  void TearDown() override
+  {
+    DeactivateTrajectoryController();
+    traj_controller_.reset();
+  }
+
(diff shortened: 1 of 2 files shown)


saikishor (dismissed):
LGTM. Thanks

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
christophfroehlich, line 1: shouldn't this be added to TrajectoryControllerTestParameterized TearDown method, too?

christophfroehlich (commented):
Thanks for digging into this. Could you please open a good-second issue for checking that with the other packages? would be great to also add a step-by-step guide there use ASAN (like you did in the other issue)

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
fmauch: It doesn't seem to be strictly necessary, but it might be a good call.

saikishor (approved):

Issue #1456, opened 2024-12-29
Extend tests for odometry of steering controllers library


christophfroehlich:
We had an issue (fixed with #1454) where the odometry was wrong in case of a reference timeout.

There should be a test for this behavior, maybe in the test [test_both_update_methods_for_ref_timeout](https://github.com/ros-controls/ros2_controllers/blob/f23b7d833a41bff6e2125ff5243e0ed95133e6d8/steering_controllers_library/test/test_steering_controllers_library.cpp#L82)

AdityaPawar162:
Hi @christophfroehlich ,
I would like to work on this issue.

AdityaPawar162:
Hi @christophfroehlich,
From what i understood, I need to extend the existing the test for odometry behaviour during reference timeout. To do that i need to,
1. Track odometry values before and after the timeout.
2. Verify that position values remain stable and don't get incorrectly updated.
3. Check that velocity values in odometry reset to zero.

Am i heading in the right direction?

PR #1166, opened 2024-06-08
Remove front_steering from steering library

qinqon:
To Accommodate controllers that are not only steering at front or rear
this change remove the `front_steering` variable from
steering_controller_library, as a byproduct of that the notion of
front or rear wheel radius is also removed from dependant controllers
and the library has know "traction_joints_names" and
"steering_joints_names" instead of "front_wheels_names" and
"rear_wheels_names".

Depends on:
- https://github.com/ros-controls/ros2_controllers/pull/1150

PR #1606, opened 2025-03-24
Extend odometry tests for steering controllers

AdityaPawar162:
This PR addresses issue #1456 by:
1.Tracking odometry values before and after the timeout.
2.Ensuring position values remain stable.
3.Verifying that linear and angular velocities reset to zero.

### DIFF
--- a/steering_controllers_library/test/test_steering_controllers_library.cpp
+++ b/steering_controllers_library/test/test_steering_controllers_library.cpp
@@ -123,12 +123,26 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
   // case 1 position_feedback = false
   controller_->params_.position_feedback = false;
 
+  // pre odometry values
+  auto pre_odom_x_1 = controller_->odometry_.get_x();
+  auto pre_odom_y_1 = controller_->odometry_.get_y();
+
   // age_of_last_command > ref_timeout_
   ASSERT_FALSE(age_of_last_command <= controller_->ref_timeout_);
   ASSERT_EQ((*(controller_->input_ref_.readFromRT()))->twist.linear.x, TEST_LINEAR_VELOCITY_X);
   ASSERT_EQ(
     controller_->update(controller_->get_node()->now(), rclcpp::Duration::from_seconds(0.01)),
     controller_interface::return_type::OK);
+
+  // post odometry values
+  auto post_odom_x_1 = controller_->odometry_.get_x();
+  auto post_odom_y_1 = controller_->odometry_.get_y();
+
+  const double position_tolerance = 1e-5;
+
+  // Position should remain stable
+  EXPECT_NEAR(pre_odom_x_1, post_odom_x_1, position_tolerance);
+  EXPECT_NEAR(pre_odom_y_1, post_odom_y_1, position_tolerance);
 
   EXPECT_TRUE(std::isnan(controller_->reference_interfaces_[0]));
   EXPECT_TRUE(std::isnan(controller_->reference_interfaces_[1]));
@@ -153,8 +167,16 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
   EXPECT_NEAR(controller_->command_interfaces_[2].get_value(), 0.575875, 1e-6);
   EXPECT_NEAR(controller_->command_interfaces_[3].get_value(), 0.575875, 1e-6);
 
+  // Linear and angular velocity should be reset to zero
+  EXPECT_DOUBLE_EQ(controller_->odometry_.get_linear(), 0.0);
+  EXPECT_DOUBLE_EQ(controller_->odometry_.get_angular(), 0.0);
+
   // case 2 position_feedback = true
   controller_->params_.position_feedback = true;
+
+  // pre odometry values
+  auto pre_odom_x_2 = controller_->odometry_.get_x();
+  auto pre_odom_y_2 = controller_->odometry_.get_y();
 
   // adjusting to achieve age_of_last_command > ref_timeout
   msg->header.stamp = controller_->get_node()->now() - controller_->ref_timeout_ -
@@ -167,6 +189,14 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
   msg->twist.angular.z = TEST_ANGULAR_VELOCITY_Z;
   controller_->input_ref_.writeFromNonRT(msg);
 
+  // post odometry values
+  auto post_odom_x_2 = controller_->odometry_.get_x();
+  auto post_odom_y_2 = controller_->odometry_.get_y();
+
+  // Position should remain stable
+  EXPECT_NEAR(pre_odom_x_2, post_odom_x_2, position_tolerance);
+  EXPECT_NEAR(pre_odom_y_2, post_odom_y_2, position_tolerance);
+
   // age_of_last_command > ref_timeout_
   ASSERT_FALSE(age_of_last_command <= controller_->ref_timeout_);
   ASSERT_EQ((*(controller_->input_ref_.readFromRT()))->twist.linear.x, TEST_LINEAR_VELOCITY_X);
@@ -196,6 +226,10 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
   // Steer angles should not reset
   EXPECT_NEAR(controller_->command_interfaces_[2].get_value(), 0.575875, 1e-6);
   EXPECT_NEAR(controller_->command_interfaces_[3].get_value(), 0.575875, 1e-6);
+
+  // Linear and angular velocity should be reset to zero
+  EXPECT_DOUBLE_EQ(controller_->odometry_.get_linear(), 0.0);
+  EXPECT_DOUBLE_EQ(controller_->odometry_.get_angular(), 0.0);
 }
 
 int main(int argc, char ** argv)


steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -167,6 +189,14 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
...
   controller_->input_ref_.writeFromNonRT(msg);
 
+  // post odometry values
+  auto post_odom_x_2 = controller_->odometry_.get_x();
+  auto post_odom_y_2 = controller_->odometry_.get_y();
+
+  // Position should remain stable
+  EXPECT_NEAR(pre_odom_x_2, post_odom_x_2, position_tolerance);
+  EXPECT_NEAR(pre_odom_y_2, post_odom_y_2, position_tolerance);
+
```
christophfroehlich, line 199: this should go after the update() call?

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -123,13 +123,27 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
...
 
+  // post odometry values
+  auto post_odom_x_1 = controller_->odometry_.get_x();
+  auto post_odom_y_1 = controller_->odometry_.get_y();
+
+  const double position_tolerance = 1e-5;
+
+  // Position should remain stable
+  EXPECT_NEAR(pre_odom_x_1, post_odom_x_1, position_tolerance);
+  EXPECT_NEAR(pre_odom_y_1, post_odom_y_1, position_tolerance);
```
christophfroehlich, line 145: we should also test for the heading using `odometry_.get_heading()`

christophfroehlich (changes requested):
Thanks for giving a try! But apparently, this does not test the original issue. What I tried: Run colcon build and test with this version, and then with
`git revert 2375aca`
It succeeds in both cases.

AdityaPawar162:
Hi @christophfroehlich,
I have updated the function as per suggested changes,I have also added a check for the `last_linear_velocity_` and `last_angular_velocity_`.

christophfroehlich:
tests are failing now

AdityaPawar162:
> tests are failing now

Sorry , for the incorrect push.

AdityaPawar162:
Hi @christophfroehlich ,
I have pushed the correct changes, please do suggest any changes to be made.

christophfroehlich (changes requested):
Thanks, the tests for the last*velocity now fails without the bugfix.

I had a look now why the other test does not fail: 
1.) The test is in open_loop=false mode
2.) The `update_odometry` call in `update_and_write_commands` is overriden in the `TestableSteeringControllersLibrary` and does not update the odometry class. 


We should add a new testfile with a different yaml (see add_rostest_with_parameters_gmock in the CMakeLists file) and add something like this in the TestableSteeringControllersLibrary (from bicycle_steering_controller)
```cpp
  if (params_.open_loop)
  {
    odometry_.update_open_loop(last_linear_velocity_, last_angular_velocity_, period.seconds());
  }
```

AdityaPawar162:
> Thanks, the tests for the last*velocity now fails without the bugfix.
> 
> I had a look now why the other test does not fail: 1.) The test is in open_loop=false mode 2.) The `update_odometry` call in `update_and_write_commands` is overriden in the `TestableSteeringControllersLibrary` and does not update the odometry class.
> 
> We should add a new testfile with a different yaml (see add_rostest_with_parameters_gmock in the CMakeLists file) and add something like this in the `TestableSteeringControllersLibrary::update_odometry` (from bicycle_steering_controller)
> 
> ```c++
>   if (params_.open_loop)
>   {
>     odometry_.update_open_loop(last_linear_velocity_, last_angular_velocity_, period.seconds());
>   }
> ```

ok

AdityaPawar162:
Hi @christophfroehlich,
As per your suggestion, I created a new YAML file with `open_loop` set to true and added a separate test file named `test_steering_controllers_open_library`. I also modified the `TestableSteeringControllersLibrary::update_geometry` function. However, I'm unsure why the builds are failing.

steering_controllers_library/CMakeLists.txt
```diff
@@ -74,6 +74,20 @@ if(BUILD_TESTING)
...
+  target_link_libraries(test_steering_controllers_open_library steering_controllers_library)
+  ament_target_dependencies(
+    test_steering_controllers_open_library
+    controller_interface
+    hardware_interface
+  )
+
+  ament_add_gmock(test_steering_odometry_open_loop_timeout test/test_steering_odometry.cpp)
+  target_link_libraries(test_steering_odometry_open_loop_timeout steering_controllers_library)
+
```
christophfroehlich, line 90:
```suggestion
```
why is this here?

steering_controllers_library/test/steering_controllers_library_open_loop_timeout_params.yaml
```diff
@@ -0,0 +1,17 @@
+test_steering_controllers_library_open_loop_timeout:
```
christophfroehlich, line 1:
```suggestion
test_steering_controllers_library:
```
this has to match the node name, which is hardcoded [here](https://github.com/ros-controls/ros2_controllers/blob/3850a08d73424963831a92bda2a9c332d8f0aa6f/steering_controllers_library/test/test_steering_controllers_library.hpp#L155)

christophfroehlich (changes requested):
See my other comments. With these changes, the test runs but fails with

```
3: /workspaces/ros2_rolling_ws/src/ros2_controllers/steering_controllers_library/test/test_steering_controllers_library.cpp:173: Failure
3: The difference between controller_->command_interfaces_[2].get_value() and 0.575875 is 0.071721078654024661, which exceeds 1e-6, where
3: controller_->command_interfaces_[2].get_value() evaluates to 0.50415392134597536,
3: 0.575875 evaluates to 0.57587500000000003, and
3: 1e-6 evaluates to 9.9999999999999995e-07.
3: 
3: /workspaces/ros2_rolling_ws/src/ros2_controllers/steering_controllers_library/test/test_steering_controllers_library.cpp:174: Failure
3: The difference between controller_->command_interfaces_[3].get_value() and 0.575875 is 0.092000372941643649, which exceeds 1e-6, where
3: controller_->command_interfaces_[3].get_value() evaluates to 0.66787537294164367,
3: 0.575875 evaluates to 0.57587500000000003, and
3: 1e-6 evaluates to 9.9999999999999995e-07.
3: 
3: /workspaces/ros2_rolling_ws/src/ros2_controllers/steering_controllers_library/test/test_steering_controllers_library.cpp:239: Failure
3: The difference between controller_->command_interfaces_[2].get_value() and 0.575875 is 0.071721078654024661, which exceeds 1e-6, where
3: controller_->command_interfaces_[2].get_value() evaluates to 0.50415392134597536,
3: 0.575875 evaluates to 0.57587500000000003, and
3: 1e-6 evaluates to 9.9999999999999995e-07.
3: 
3: /workspaces/ros2_rolling_ws/src/ros2_controllers/steering_controllers_library/test/test_steering_controllers_library.cpp:240: Failure
3: The difference between controller_->command_interfaces_[3].get_value() and 0.575875 is 0.092000372941643649, which exceeds 1e-6, where
3: controller_->command_interfaces_[3].get_value() evaluates to 0.66787537294164367,
3: 0.575875 evaluates to 0.57587500000000003, and
3: 1e-6 evaluates to 9.9999999999999995e-07.
```

I'm not sure why these are different in open_loop mode, but haven't investigated further now

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -123,13 +123,33 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
...
 
+  // pre odometry values
+  auto pre_odom_x_1 = controller_->odometry_.get_x();
+  auto pre_odom_y_1 = controller_->odometry_.get_y();
+  auto pre_odom_heading_1 = controller_->odometry_.get_heading();
+
   // age_of_last_command > ref_timeout_
   ASSERT_FALSE(age_of_last_command <= controller_->ref_timeout_);
   ASSERT_EQ((*(controller_->input_ref_.readFromRT()))->twist.linear.x, TEST_LINEAR_VELOCITY_X);
   ASSERT_EQ(
```
christophfroehlich, line 134: Another thing I realized: We have to call update() twice in the test, because the update_odometry() will be called in the beginning, but the last*velocity will be updated later. Calling update twice here, and we get also errors as (with the reverted fix)
```
3: /workspaces/ros2_rolling_ws/src/ros2_controllers/steering_controllers_library/test/test_steering_controllers_library.cpp:151: Failure
3: The difference between pre_odom_heading_1 and post_odom_heading_1 is 0.0030000000000000001, which exceeds position_tolerance, where
3: pre_odom_heading_1 evaluates to 0,
3: post_odom_heading_1 evaluates to 0.0030000000000000001, and
3: position_tolerance evaluates to 1.0000000000000001e-05.
```

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -174,6 +203,19 @@ TEST_F(SteeringControllersLibraryTest, test_both_update_methods_for_ref_timeout)
     controller_->update(controller_->get_node()->now(), rclcpp::Duration::from_seconds(0.01)),
```
christophfroehlich, line 203: see above

steering_controllers_library/CMakeLists.txt
AdityaPawar162: Oh, I thought I need to add another gmock test target.

christophfroehlich: No, because this test file does not use the parameters at all?

steering_controllers_library/test/test_steering_controllers_library.cpp
AdityaPawar162: Called update function twice in the next commit.

steering_controllers_library/test/test_steering_controllers_library.cpp
AdityaPawar162: Called update function twice in the next commit.

AdityaPawar162:
Hi @christophfroehlich,
The build is failing because it is not able to find ros2_control_cmake 
Starting >>> realtime_tools
Starting >>> control_msgs
Starting >>> ros2_control_test_assets
Starting >>> controller_manager_msgs
--- stderr: realtime_tools
CMake Error at CMakeLists.txt:4 (find_package):
  By not providing "Findros2_control_cmake.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "ros2_control_cmake", but CMake did not find one.

  Could not find a package configuration file provided by
  "ros2_control_cmake" with any of the following names:

    ros2_control_cmakeConfig.cmake
    ros2_control_cmake-config.cmake

  Add the installation prefix of "ros2_control_cmake" to CMAKE_PREFIX_PATH or
  set "ros2_control_cmake_DIR" to a directory containing one of the above
  files.  If "ros2_control_cmake" provides a separate development package or
  SDK, be sure it has been installed.


---
Failed   <<< realtime_tools [0.74s, exited with code 1]
Aborted  <<< ros2_control_test_assets [3.00s]
Aborted  <<< controller_manager_msgs [6.3[9](https://github.com/ros-controls/ros2_controllers/actions/runs/14100888744/job/39496745904?pr=1606#step:7:10)s]
Aborted  <<< control_msgs [7.15s]

Summary: 0 packages finished [7.53s]
  1 package failed: realtime_tools
  3 packages aborted: control_msgs controller_manager_msgs ros2_control_test_assets
  1 package had stderr output: realtime_tools
  32 packages not processed
   
I believe there shouldn't be any installation errors in GitHub Actions, right?

AdityaPawar162:
Hi @christophfroehlich,
I have made suggested changes, please do take a look.

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -339,6 +339,15 @@ controller_interface::CallbackReturn SteeringControllersLibrary::on_activate(
   // Set default value in command
   reset_controller_reference_msg(*(input_ref_.readFromRT()), get_node());
 
+  // Set the initial steering angle values
+  size_t steering_count_idx =
+    params_.front_steering ? params_.front_wheels_names.size() : params_.rear_wheels_names.size();
+
+  for (size_t i = 0; i < steering_count_idx; i++)
+  {
+    command_interfaces_[steering_count_idx + i].set_value(0.575875);
```
christophfroehlich, line 348: ?

AdityaPawar162: The test clearly expects the steering angles (command_interfaces_[2] and command_interfaces_[3]) to remain at 0.575875 during timeouts but somehow they were not so i directly set the steering angles to the expected value in the controller activation method (on_activate).

AdityaPawar162: I have added a timeout check in the `update_and_write_commands` function, ensuring that commands are updated only when the timeout is false.

christophfroehlich: I don't think that this is the right approach. This value should be the steering angle for the requested twist, which should remain unchanged after timeout.

But something is strange, both steering angles should not be the same as this test uses ackermann config
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L125

christophfroehlich:
While checking the test class, I see that there are some unused/duplicate constants.

Could you clean them up as well please?
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L57-L61
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L292-L295
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L300-L309
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L311-L315
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L324

steering_controllers_library/src/steering_controllers_library.cpp
christophfroehlich: This answer refers to the line in the code of the thread.

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -386,32 +395,52 @@ controller_interface::return_type SteeringControllersLibrary::update_and_write_c
     last_linear_velocity_ = timeout ? 0.0 : reference_interfaces_[0];
     last_angular_velocity_ = timeout ? 0.0 : reference_interfaces_[1];
 
-    auto [traction_commands, steering_commands] = odometry_.get_commands(
-      reference_interfaces_[0], reference_interfaces_[1], params_.open_loop,
-      params_.reduce_wheel_speed_until_steering_reached);
-
-    if (params_.front_steering)
+    // calculate new commands if not in timeout
+    if (!timeout)
```
christophfroehlich, line 399: I don't see the necessity of this change. 
In case of timeout it SHOULD set the appropriate commands see L397, L401 or L409/413, respectively.

AdityaPawar162: Actually, I specially wanted to update traction_commands and steering_commands when the timeout is false, so I thought there is a need to add this    
     `auto [traction_commands, steering_commands] = odometry_.get_commands(
      reference_interfaces_[0], reference_interfaces_[1], params_.open_loop,
      params_.reduce_wheel_speed_until_steering_reached);
    ` inside the flag.

AdityaPawar162:
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L57-L61
This is getting used in the function `configure_odometry`
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L121-L128

christophfroehlich:
> This is getting used in the function `configure_odometry`

only FRONT_WHEELS_RADIUS_, WHEELBASE_, REAR_WHEEL_TRACK_

AdityaPawar162:
Hi @christophfroehlich,
The `joint_command_values` are hardcoded to 2.2 and 4.4, I think so because of this the steering angle is not getting updated ?
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L318

AdityaPawar162:
> Hi @christophfroehlich, The `joint_command_values` are hardcoded to 2.2 and 4.4, I think so because of this the steering angle is not getting updated ?
> 
> https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/steering_controllers_library/test/test_steering_controllers_library.hpp#L318

Hi @christophfroehlich,
Please clearify my doubt.

christophfroehlich:
@AdityaPawar162 I try to catch up and respond to your comments within the next days, could you please fix the conflicts with the master with a merge? (which come from #1166)

christophfroehlich:
@AdityaPawar162 are you still working on this?

Issue #1409, opened 2024-12-05
High CPU load due to joint state broadcaster


firesurfer:
**Describe the bug**

I am currently looking into the issue why our ros2control node is consuming 100% cpu load. 
For testing purposes I disabled all loaded controllers one by one. As expected the cpu drops a bit for each controller. But for the joint state broadcaster this was rather extreme. I see a change in cpu load for node of ~15-20%. 
This is not the case for other controllers that publish messages at the same rate of 500Hz. 

We have a rather large amount of joints in our system: 34. 

**Expected behavior**

I would expect that such a simple controller shouldn't consume that much cpu power. 

**Environment (please complete the following information):**
 - OS:  ubuntu 24.04
 - Version Jazzy - 4.16.0

**Additional context**

When taking a look at the code I would guess at the moment three places that could result in a higher cpu load:

1. there are rather many operations on the `map_interface_to_joint_state_` -> but unordered_map should be fast enough
2. mutex operation in the RT publisher
3. Reading the values from the state interfaces -> Not sure if there is any synchronization happening ?

christophfroehlich:
Strange. To narrow that down, I'd comment the different publishers in the update() method and see if the publishers or the `state_interface.get_value()` is causing this behavior..

saikishor:
I highly doubt that it is caused due to the handles. In a synchronous system, mutexes shouldn't do anything. So, as @christophfroehlich suggested try to check with the publishers and see.

jordan-palacios:
We are currently having issues with this broadcaster too. In our case I think simple due to the sheer amount of state interfaces our setup has: ~550. 

After a brief look, the broadcaster could use some optimization. There a lot of map lookups on the `update()` function that could be cached. I'll prepare a PR.

saikishor:
> We are currently having issues with this broadcaster too. In our case I think simple due to the sheer amount of state interfaces our setup has: ~550. 
> 
> After a brief look, the broadcaster could use some optimization. There a lot of map lookups on the `update()` function that could be cached. I'll prepare a PR.

Thank you @jordan-palacios

jordan-palacios:
Performance of the current version in `master`.

```
[INFO] [1744704812.699545539] [joint_state_broadcaster]: Average update time: 240.174700 us
[INFO] [1744704812.699598892] [joint_state_broadcaster]: Variance: 231730.945780 us
--
[...]

PR #1609, opened 2025-03-25
[JSB] skip publish on no subscription

saikishor:
might help with #1409 
@firesurfer can you check if this helps?


firesurfer:
@saikishor I don't think this will help with the other issue as there are subscribers on the joint state topic - I also explained in the other issue why I think this happens in our setup.

Also a small comment from my side on this patch. I am not sure how ROS2 actually determines the amount of subscribers of a topic. I could imagine that this might be a rather expensive operation. It might be cheaper just to publish the data.

saikishor:
> @saikishor I don't think this will help with the other issue as there are subscribers on the joint state topic - I also explained in the other issue why I think this happens in our setup.

I thought the overhead might come from the dynamic_joint_states. 

> Also a small comment from my side on this patch. I am not sure how ROS2 actually determines the amount of subscribers of a topic. I could imagine that this might be a rather expensive operation. It might be cheaper just to publish the data. 

I don't think it's expensive operation, I have used in RT loops and didn't find any major jitters

bmagyar (approved):

firesurfer:
> I thought the overhead might come from the dynamic_joint_states.

I see. I can hopefully give it a try next week if our setup is free.

joint_state_broadcaster/src/joint_state_broadcaster.cpp
```diff
@@ -351,7 +351,9 @@ controller_interface::return_type JointStateBroadcaster::update(
       state_interface.get_value());
   }
 
-  if (realtime_joint_state_publisher_ && realtime_joint_state_publisher_->trylock())
+  if (
+    realtime_joint_state_publisher_ && joint_state_publisher_->get_subscription_count() > 0u &&
```
mhubii, line 355: looks good. Are the `rmw` and implementations thereof doing this implicitly already?

saikishor: I believe yes

bmagyar:
So we wait for @firesurfer 's feedback?

mhubii:
If it's not broken, don't fix it. Might make sense to benchmark this before modifying such a sensitive line of code based on empirical insights

joint_state_broadcaster/src/joint_state_broadcaster.cpp
mhubii: meaning subscription count is polled outside the RT loop where it is definitely safe, right? To my best understanding, the `publish` operation just appends a buffer

mhubii:
wasn't the initial issue closed?

saikishor:
> If it's not broken, don't fix it. Might make sense to benchmark this before modifying such a sensitive line of code based on empirical insights

@mhubii Will you be able to test it on your setup?

mhubii:
essentially run a mock JSB with and without modification, spawn subscriptions + monitor CPU load each across all available middlewares?

bmagyar:
Change of heart, I am with @mhubii on this. Instead of benchmarking though I think we should look closer at the JSB with a profiler first. The changes suggested on this PR do not fix the underlying issue, only avoid them when there aren't subscribers which will lead to a massive performance change when there *is* a subscriber. Let's fix the root cause instead.

bmagyar (changes requested):
I suggest profiling first.

PR #1622, opened 2025-04-03
[CI] Revert test changes to debug the timeout from test_force_torque_sensor_broadcaster 

Juliaj:
This is to address https://github.com/ros-controls/ros2_controllers/pull/1586#discussion_r2015022950

@christophfroehlich, instead of switching the callback to lifecycle `deactivate`, `cleanup` etc, I removed the code change in test tear down so that it utilizes the auto-shutdown in the destructor.  Let me know if you'd like me to keep the current code and make the fix.

I was able to reproduce the timeout locally from `ForceTorqueSensorBroadcasterTest.InterfaceNames_Configure_Success`.  Issue was reported as https://github.com/ros-controls/realtime_tools/issues/307. Once we understand a bit more cause, I'll explore additional fix if needed.


christophfroehlich:
@mergifyio backport mergify/bp/humble/pr-1586

christophfroehlich (approved):
Thanks for the follow-up

PR #1625, opened 2025-04-04
fix/use-gmock-in-tests

yassine-cherni:
Replacing gmock by gtest [gtest https://github.com/ros-controls/ros2_controllers/issues/1621[](url)


admittance_controller/test/test_admittance_controller.hpp
```diff
@@ -398,7 +398,8 @@ class AdmittanceControllerTest : public ::testing::Test
   rclcpp::Node::SharedPtr test_broadcaster_node_;
 };
 
-// From the tutorial: https://www.sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest
+// From the tutorial: https://www.sandordargo.com/blog/2022/03/02/mocking-non-virtual-and-free-functions
```
christophfroehlich, line 401: why changing this?

christophfroehlich (changes requested):
Thanks, just one comment

admittance_controller/test/test_admittance_controller.hpp
yassine-cherni: I through it will be helpfull to change it for the global library tutorial.

christophfroehlich: but the new link does not cover the parameterized tests?

saikishor: There is a similar link also in the JTC and well as in admittance in line 436. Why only change here?

saikishor (commented):
Apart from that comment, rest LGTM

admittance_controller/test/test_admittance_controller.hpp
yassine-cherni: No, it's for gmock
I'll return the last one if you want, thanks.

admittance_controller/test/test_admittance_controller.hpp
christophfroehlich, line 1: ?

yassine-cherni: Fixed, Thanks a lot 💖

admittance_controller/test/test_admittance_controller.hpp
```diff
@@ -456,4 +456,4 @@ class AdmittanceControllerTestParameterizedInvalidParameters
   }
 };
 
-#endif  // TEST_ADMITTANCE_CONTROLLER_HPP_
+#endif  // TEST_ADMITTANCE_CONTROLLER_HPP_
```
saikishor:
```suggestion
#endif  // TEST_ADMITTANCE_CONTROLLER_HPP_

```

saikishor (approved):

christophfroehlich (approved):

PR #1630, opened 2025-04-07
Improve steering controller tests

### LINKED ISSUE #930: Improve Steering Controller Tests
> > Yes, you're right, there are a lot of things open to test -- but right now I just wanted to provide a quick fix for the Ackermann odometry, only rudimentary tested. Maybe we should open a new issue, tagged "help wanted" for proper testing of the steering lib -- starting with tests for the integrator functions, etc. But maybe first we should check for a redesign cf. #692
> 
> ok, I'll create a new issue then.
> 
> > BTW, I just added this file 'test_steering_odometry.cpp', with some hard-coded tests. Is that the way to go, or do we want to have a more advanced testing style?
> 
> Should the expectations of the FW kinematics be true for all the different kinematic configurations? Then we could change this to a parameterized test as we have with JTC
> https://github.com/ros-controls/ros2_controllers/blob/1d0d7531ef8d24ea52e761a872b8bf277652dd33/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp#L695-L710
> 
> https://github.com/ros-controls/ros2_controllers/blob/1d0d7531ef8d24ea52e761a872b8bf277652dd33/joint_trajectory_controller/test/test_trajectory_controller.cpp#L1840-L1849

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_controllers/issues/921#issuecomment-1851853293_

Vedant87:
Closes #930

- Converted the steering odometry tests to parameterized tests.
- Added a new file for defining the parameterized test classes.
- Introduced 2 parameterized test classes: 1) For regular test cases and 2) For steering limited test cases.


christophfroehlich:
Thanks for your contribution, I try to come back with a review soon. Could you please review #1166 in the meantime? This should get merged soon.

christophfroehlich (commented):
I had a look now, but I don't understand the approach you have used here. For me it seems that in the end, we have the exact same tests like before, but added 650 lines of code? What is the purpose of using parameterized tests, but have a condition in every test to skip if it does not match the single test which was here before? Please explain, maybe I just don't see it.

PR #1634, opened 2025-04-09
Implemented a generic filters plugin 

ankurbodhe:
This PR migrates and enhances the chained_filter_controller originally implemented during the [ROSCon 2024 control workshop](https://github.com/ros-controls/roscon2024_control_workshop). The controller enables chaining and filtering of state interfaces through a sequence of filters.

**Changes introduced:**
1. Ported chained_filter_controller from workshop reference implementation.
2. Developed unit tests covering:
    - Initialization and lifecycle behavior.
    - Filter configuration and chaining.
    - Correct propagation of state values through the filter stack.


ankurbodhe:
@christophfroehlich I just made another commit to fix the issues in the workflow.. requesting you to approve to workflow

saikishor:
@ankurbodhe can you fix the pre-commit jobs?

Is this ready for review?

chained_filter_controller/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
+# cmake_minimum_required and project setup
+cmake_minimum_required(VERSION 3.8)
+project(chained_filter_controller)
+
+# Enable warnings
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
```
christophfroehlich, line 8: please have a look at the other packages, we use ros2_control_cmake instead

chained_filter_controller/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
...
+
+# Dependencies
+find_package(ament_cmake REQUIRED)
+find_package(controller_interface REQUIRED)
+find_package(filters REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
+find_package(hardware_interface REQUIRED)
+find_package(generate_parameter_library REQUIRED)
```
christophfroehlich, line 18: please have a look at the other packages to follow the same style (look for `THIS_PACKAGE_INCLUDE_DEPENDS`)

chained_filter_controller/config/chained_filter_parameters.yaml
christophfroehlich, line 1: we usually put this in the src folder

chained_filter_controller/package.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>chained_filter_controller</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
+  <maintainer email="ankur-u24@todo.todo">ankur-u24</maintainer>
+  <license>TODO: License declaration</license>
```
christophfroehlich, line 8: please copy maintainer, license from the other packages of this repo, and add you and Bence as authors
`  <author email="bence.magyar.robotics@gmail.com">Bence Magyar</author>`

you can copy the description of the plugin xml below.

chained_filter_controller/include/chained_filter_controller/chained_filter.hpp
christophfroehlich, line 1: please insert a copyright notice

chained_filter_controller/src/chained_filter.cpp
```diff
@@ -0,0 +1,107 @@
+// chained_filter.cpp (migrated from ROSCon 2024 workshop)
```
christophfroehlich, line 1:
```suggestion

```
please insert a copyright notice instead

chained_filter_controller/test/test_chained_filter.cpp
```diff
@@ -0,0 +1,93 @@
+#include <gtest/gtest.h>
```
christophfroehlich, line 1:
```suggestion
#include <gmock/gmock.h>
```
We are using gmock consistently, see https://github.com/ros-controls/ros2_controllers/pull/1625

chained_filter_controller/test/test_chained_filter.cpp
```diff
@@ -0,0 +1,93 @@
...
+    controller_->on_cleanup(rclcpp_lifecycle::State());
+  });
+}
+
+TEST_F(ChainedFilterControllerTest, ShutdownDoesNotCrash)
+{
+  EXPECT_NO_THROW({
+    controller_->on_shutdown(rclcpp_lifecycle::State());
+  });
+}
```
christophfroehlich, line 78: I don't think this is necessary to test. The controller manager assures that only the correct state transitions are called.

chained_filter_controller/test/test_chained_filter.cpp
```diff
@@ -0,0 +1,93 @@
...
+TEST_F(ChainedFilterControllerTest, ConfigureReturnsFailureWithNoParams)
+{
+  // Initialize the controller (should set up internal node)
+  ASSERT_EQ(
+    controller_->init(
+      "chained_filter",         // controller name
+      "test_ns",                // namespace
+      0,                        // intra-process (0 = disabled)
+      "test_node",              // RCL node name
+      rclcpp::NodeOptions()),   // node options
```
christophfroehlich, line 36: where do you have this from? This is completely wrong, see 
https://github.com/ros-controls/ros2_control/blob/c19997a8e614773cf69e5f3e6ffd7b511f2a4f5c/controller_interface/include/controller_interface/controller_interface_base.hpp#L154-L156

chained_filter_controller/test/test_chained_filter.cpp
christophfroehlich, line 1: Please add a test if a filter is loaded and used to process data.

christophfroehlich (changes requested):
Thank you for contribution!
In general, have a look at other packages in this repository to follow the (code) style, copyright claims etc.

Please address my comments and also fix the pre-commit failures as Sai already has mentioned.

christophfroehlich:
@ankurbodhe can I help you to finish this PR? We also have to apply the changes of #1697 to the new CMakeLists.

christophfroehlich:
@ankurbodhe please don't force-push to already reviewed PRs, it just makes subsequent reviews slow. There is no need for a clean history, as we squash PRs at the time of merging into the base branches.
Ping me again if it is ready for the next review round.

christophfroehlich (approved):
I needed something like this in a project of mine and finished this PR.

As an example, adding this to example_16 gives the wheel acceleration by a derivative filter
```yaml
left_wheel_filter:
  ros__parameters:
    type: chained_filter_controller/ChainedFilter
    filter_chain:
      filter1:
        name: filter1
        type: "filters/TransferFunctionFilterDouble"
        params:
          # Transfer function coefficients for a third-order state variable filter, giving the first derivative
          a: [1.0, -1.104, 0.406, -0.04979]
          b: [1.839, -1.163, -0.6767]
    input_interfaces: ["left_wheel_joint/velocity"]
    output_interfaces: ["left_wheel_joint/acceleration"]
```

image

chained_filter_controller/src/chained_filter.cpp
```diff
@@ -0,0 +1,167 @@
...
+  {
+    const auto sensor_op = state_interfaces_[i].get_optional();
+    if (!sensor_op.has_value())
+    {
+      RCLCPP_ERROR(
+        get_node()->get_logger(), "Failed to read sensor value from state interface '%s'.",
+        state_interfaces_[0].get_name().c_str());
+      continue;
+    }
+    if (!std::isnan(sensor_op.value()))
```
saikishor:
```suggestion
    if (std::isfinite(sensor_op.value()))
```

chained_filter_controller/src/chained_filter.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+  return controller_interface::return_type::OK;
+}
+
+rclcpp::NodeOptions ChainedFilter::define_custom_node_options() const
+{
+  return rclcpp::NodeOptions()
+    .allow_undeclared_parameters(true)
+    .automatically_declare_parameters_from_overrides(false);
+}
```
saikishor: Won't it work with the default settings?, Just curious

chained_filter_controller/src/chained_filter_parameters.yaml
```diff
@@ -0,0 +1,33 @@
...
+    }
+  filter_chain: {
+    type: none,
+    description: "Map of parameters that defines a filter chain, containing filterN as key. Valid for all interfaces, overrides <input_interfaces>.filter_chain if set.
+    The fields for each filter are:
+    type: The filter plugin to be loaded
+    name: Actual name semantically describing the filter, e.g., low_pass_filter
+    params: And underlying map of parameters needed for a specific filter, refer to the specific filter documentation."
+  }
+  __map_input_interfaces:
```
saikishor, line 26: One doubt, can the parameter have "/" in ROS 2?

saikishor: Ok. I see it here: https://github.com/ros-controls/ros2_controllers/blob/66a760038ca85e801a370017398a889e83b29d47/chained_filter_controller/test/config/test_multiple_chained_filter.yaml

chained_filter_controller/test/test_multiple_chained_filter.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+void MultipleChainedFilterTest::TearDown() { controller_.reset(nullptr); }
+
+void MultipleChainedFilterTest::SetUpController(
+  const std::string node_name, const std::vector<rclcpp::Parameter> & parameters)
+{
+  auto node_options = controller_->define_custom_node_options();
+  node_options.parameter_overrides(parameters);
+
```
saikishor, line 50: Don't we need to parse `multiple_chained_filter.yaml` this file here?

saikishor (commented):
Very impressive work 👏🏾 

Leaving some comments of my partial review

chained_filter_controller/src/chained_filter_parameters.yaml
christophfroehlich: Yess, I doubted it as well but it worked <3

chained_filter_controller/test/test_multiple_chained_filter.cpp
christophfroehlich: this is done via the cmake method
https://github.com/ros-controls/ros2_controllers/pull/1634/files#diff-a3a22369907540a2005b908a839ba9368165749c3c9c57d6f8480a562dc2aba5R70

chained_filter_controller/src/chained_filter.cpp
christophfroehlich: haven't tested it before, but it seems to work without

christophfroehlich (approved):

chained_filter_controller/include/chained_filter_controller/chained_filter.hpp
```diff
@@ -0,0 +1,63 @@
...
+    const rclcpp_lifecycle::State & previous_state) override;
+
+  controller_interface::CallbackReturn on_activate(
+    const rclcpp_lifecycle::State & previous_state) override;
+
+  controller_interface::return_type update_and_write_commands(
+    const rclcpp::Time & time, const rclcpp::Duration & period) override;
+
+protected:
+  std::vector<hardware_interface::StateInterface> on_export_state_interfaces() override;
```
saikishor, line 51: So, this controller right now only works as a filter for state Interfaces?

chained_filter_controller/src/chained_filter.cpp
```diff
@@ -0,0 +1,160 @@
...
+  return controller_interface::return_type::OK;
+}
+
+std::vector<hardware_interface::StateInterface> ChainedFilter::on_export_state_interfaces()
+{
+  std::vector<hardware_interface::StateInterface> state_interfaces;
+  for (size_t i = 0; i < params_.output_interfaces.size(); ++i)
+  {
+    state_interfaces.emplace_back(
+      get_node()->get_name(), params_.output_interfaces.at(i), &output_state_values_.at(i));
```
saikishor: One curious question here 

If you remember in the past, we said only the last part of the interface needs to be used in the interface part and the rest we use in the prefix.

For instance, if we use the state interface `joint1/position` and then when we filter the exposed interface will be with prefix: `ctrl_name` and the interface as `joint1/position` instead of `ctrl_name/joint1` as prefix and `position` as interface. 

Does it make sense to follow the same strategy here?

saikishor (approved):
Just two things. Rest looks spot on

chained_filter_controller/include/chained_filter_controller/chained_filter.hpp
christophfroehlich: I think so yes ;) Do you think this should be two different controllers, one for state and one for command interfaces? Or add another parameter list, and handle them in parallel?
If this does not need a change to the package or plugin name, I'd add a follow-up issue for that.

chained_filter_controller/src/chained_filter.cpp
christophfroehlich: Here we propose the list of output_interfaces, for example we can have input_interface `joint1/position` and output interface `joint1/velocity` for a derivative filter. This would mean that we have to split the string and add only `velocity` as interface. OK for this simple example, but this might be tricky for multiple filters in a chain.

If we care (what are the implications, if we don't?), we should change the parameters to have `interface_name` (like `joint1`, it is not only limited to joints right?), and `input_interface_type` (like position) and `output_interface_type` (like velocity)?

saikishor: Why not split the string and use the last one for the interface always and the rest for the prefix?

chained_filter_controller/include/chained_filter_controller/chained_filter.hpp
saikishor: Yes, we can create a follow-up issue for that. I think the same controller can be utilized for this by extending it with parameters etc

chained_filter_controller/src/chained_filter.cpp
christophfroehlich: which string, the string given in `output_interfaces`?

saikishor: Yesss

christophfroehlich: done

christophfroehlich (dismissed):

chained_filter_controller/src/chained_filter.cpp
```diff
@@ -0,0 +1,173 @@
...
+    auto pos = full_name.rfind('/');
+    const std::string iface_name =
+      (pos != std::string::npos) ? full_name.substr(pos + 1) : full_name;
+    const std::string iface_prefix =
+      (pos != std::string::npos) ? "/" + full_name.substr(0, pos) : "";
+    std::cout << "Exporting state interface with prefix: " << get_node()->get_name() + iface_prefix
+              << " and interface " << iface_name << std::endl;
+
+    state_interfaces.emplace_back(
+      get_node()->get_name() + iface_prefix, iface_name, &output_state_values_.at(i));
```
saikishor:
```suggestion
    const std::string exp_intf_prefix = get_node()->get_name() + iface_prefix;
    RCLCPP_INFO(get_logger(), "Exporting state interface with prefix: '%s' and interface: '%s'",  exp_intf_prefix.c_str(), iface_name.c_str());

    state_interfaces.emplace_back(
      exp_intf_prefix, iface_name, &output_state_values_.at(i));
```
Change to RCLCPP_INFO instead of std::cout.

If it's not a big requirement, let's leave it as it as, it's okay for me as well

saikishor (dismissed):

chained_filter_controller/src/chained_filter.cpp
```diff
@@ -0,0 +1,173 @@
...
+  {
+    // Split the output interface name by the last "/" if present
+    const std::string & full_name = params_.output_interfaces.at(i);
+    auto pos = full_name.rfind('/');
+    const std::string iface_name =
+      (pos != std::string::npos) ? full_name.substr(pos + 1) : full_name;
+    const std::string iface_prefix =
+      (pos != std::string::npos) ? "/" + full_name.substr(0, pos) : "";
+    std::cout << "Exporting state interface with prefix: " << get_node()->get_name() + iface_prefix
+              << " and interface " << iface_name << std::endl;
```
christophfroehlich:
```suggestion
```

chained_filter_controller/src/chained_filter.cpp
christophfroehlich: oh this was only for debugging. removed it instead

saikishor (approved):
LGTM

saikishor (approved):
Spot on!

PR #1636, opened 2025-04-11
Add `motion_primitives_forward_controller` for interfacing motion primitive messages with hardware interfaces

mathias31415:
This PR adds the `motion_primitives_forward_controller`, a controller for forwarding motion primitive commands to motion-primitive-capable hardware interfaces.  
It was developed alongside the [`motion_primitive_ur_driver`](https://github.com/StoglRobotics-forks/Universal_Robots_ROS2_Driver_MotionPrimitive) and is intended to support future hardware interfaces for additional robot manufacturers.  
The corresponding PR for integrating the `motion_primitive_ur_driver` into the official Universal Robots ROS 2 driver can be found here: [UniversalRobots/Universal_Robots_ROS2_Driver#1341](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1341).

The controller subscribes to `MotionPrimitive.msg` from the [`industrial_robot_motion_interfaces`](https://github.com/StoglRobotics-forks/industrial_robot_motion_interfaces/tree/helper-types) package. The `MotionPrimitive.msg` has been extended with additional helper types:
  - `STOP_MOTION`: Immediately interrupts execution and clears all queued primitives.
  - `MOTION_SEQUENCE_START` and `MOTION_SEQUENCE_END`: Define a sequence of primitives to be executed as a blended motion block.

The controller uses the following interfaces:

**Command Interfaces**
- `motion_type`: Type of motion primitive (e.g., LINEAR_JOINT, LINEAR_CARTESIAN, CIRCULAR_CARTESIAN, etc.)
- `q1` – `q6`: Target joint positions for joint-based motion
- `pos_x`, `pos_y`, `pos_z`: Target Cartesian position
- `pos_qx`, `pos_qy`, `pos_qz`, `pos_qw`: Orientation quaternion of the target pose
- `pos_via_x`, `pos_via_y`, `pos_via_z`: Intermediate via-point position for circular motion
- `pos_via_qx`, `pos_via_qy`, `pos_via_qz`, `pos_via_qw`: Orientation quaternion of via-point
- `blend_radius`: Blending radius for smooth transitions
- `velocity`: Desired motion velocity
- `acceleration`: Desired motion acceleration
- `move_time`: Optional duration for time-based execution

**State Interfaces**
- `execution_status`: Indicates the current execution state of the primitive.
- `ready_for_new_primitive`: Boolean flag indicating whether the interface is ready to receive a new motion primitive

I'd appreciate any feedback or suggestions – thanks in advance!


motion_primitives_forward_controller/include/motion_primitives_forward_controller/visibility_control.h
```diff
@@ -0,0 +1,54 @@
...
+#define MOTION_PRIMITIVES_CONTTROLLER_PKG__VISIBILITY_IMPORT
+#if __GNUC__ >= 4
+#define MOTION_PRIMITIVES_CONTTROLLER_PKG__VISIBILITY_PUBLIC __attribute__((visibility("default")))
+#define MOTION_PRIMITIVES_CONTTROLLER_PKG__VISIBILITY_LOCAL __attribute__((visibility("hidden")))
+#else
+#define MOTION_PRIMITIVES_CONTTROLLER_PKG__VISIBILITY_PUBLIC
+#define MOTION_PRIMITIVES_CONTTROLLER_PKG__VISIBILITY_LOCAL
+#endif
+#define MOTION_PRIMITIVES_CONTTROLLER_PKG__VISIBILITY_PUBLIC_TYPE
+#endif
```
bmagyar, line 52: we don't do this sort of stuff anymore, please take a look at how other controllers  do it

motion_primitives_forward_controller/package.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>motion_primitives_forward_controller</name>
+  <version>0.0.0</version>
+  <description>Package to control robots using motion primitives like PTP, LIN and CIRC</description>
+  <maintainer email="todo@todo.de">todo</maintainer>
+  <author>Mathias Fuhrer</author>
+  <license>BSD-3-Clause</license>
```
bmagyar, line 9: It would be ideal to update this license to Apache 2 if possible.

motion_primitives_forward_controller/motion_primitives_forward_controller.rolling.repos
```diff
@@ -0,0 +1,6 @@
+repositories:
```
bmagyar, line 1: these repos files are not needed

motion_primitives_forward_controller/package.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>motion_primitives_forward_controller</name>
+  <version>0.0.0</version>
+  <description>Package to control robots using motion primitives like PTP, LIN and CIRC</description>
+  <maintainer email="todo@todo.de">todo</maintainer>
+  <author>Mathias Fuhrer</author>
```
bmagyar, line 8: email for you could be defined, the release version could be updated to match the rest of the repo and please copy the maintainer fields from other controllers in t his repo

motion_primitives_forward_controller/package.xml
```diff
@@ -0,0 +1,36 @@
...
+  <depend>industrial_robot_motion_interfaces</depend>
+
+  <depend>rclcpp</depend>
+
+  <depend>rclcpp_lifecycle</depend>
+
+  <depend>realtime_tools</depend>
+
+  <depend>std_srvs</depend>
+
```
bmagyar, line 32: that's a lot of whitespace that could be removed ;)

motion_primitives_forward_controller/LICENSE
```diff
@@ -0,0 +1,25 @@
...
+
+   * Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
+
+   * Neither the name of the copyright holder nor the names of its
+     contributors may be used to endorse or promote products derived from
+     this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
```
bmagyar, line 15: this file isn't needed

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+  # )
+
+  # add_rostest_with_parameters_gmock(test_motion_primitives_forward_controller_preceeding test/test_motion_primitives_forward_controller_preceeding.cpp ${CMAKE_CURRENT_SOURCE_DIR}/test/motion_primitives_forward_controller_preceeding_params.yaml)
+  # target_include_directories(test_motion_primitives_forward_controller_preceeding PRIVATE include)
+  # target_link_libraries(test_motion_primitives_forward_controller_preceeding motion_primitives_forward_controller)
+  # ament_target_dependencies(
+  #   test_motion_primitives_forward_controller_preceeding
+  #   controller_interface
+  #   hardware_interface
+  # )
```
bmagyar, line 99: are these expected to work? if not, could be removed. If yes, but not yet, we should put them in a follow-up PR and still remove them from this PR for now

motion_primitives_forward_controller/motion_primitives_forward_controller.xml
```diff
@@ -0,0 +1,28 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+
+Source of this file are templates in
+[RosTeamWorkspace](https://github.com/StoglRobotics/ros_team_workspace) repository.
+-->
```
bmagyar, line 19: there is absolutely nothing to copyright or license in this file

motion_primitives_forward_controller/src/motion_primitives_forward_controller.yaml
```diff
@@ -0,0 +1,50 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors: Mathias Fuhrer
+
+
```
bmagyar, line 17: don't think this warrants a license header

motion_primitives_forward_controller/test/motion_primitives_forward_controller_params.yaml
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# Source of this file are templates in
+# [RosTeamWorkspace](https://github.com/StoglRobotics/ros_team_workspace) repository.
+#
+
```
bmagyar, line 19: same here

motion_primitives_forward_controller/test/test_load_motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,44 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//
+// Source of this file are templates in
+// [RosTeamWorkspace](https://github.com/StoglRobotics/ros_team_workspace) repository.
+//
+
```
bmagyar, line 19: I'd suggest a general cleanup on all TEMPLATE license stuff here

motion_primitives_forward_controller/test/test_motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,277 @@
...
+// }
+
+// int main(int argc, char ** argv)
+// {
+//   ::testing::InitGoogleTest(&argc, argv);
+//   rclcpp::init(argc, argv);
+//   int result = RUN_ALL_TESTS();
+//   rclcpp::shutdown();
+//   return result;
+// }
```
bmagyar, line 277: 2 problems:
* it's all commented out :D 
* it's always better to use gmock, not gtest. You get much better prints.

.gitignore
```diff
@@ -0,0 +1 @@
+**/COLCON_IGNORE
```
bmagyar, line 1: I'm afraid it won't be this PR pioneering on the .gitignore file ;) please keep this to your local workspace

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+  motion_primitives_forward_controller
+  RUNTIME DESTINATION bin
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+)
+
+# install(
+#   DIRECTORY include/
+#   DESTINATION include/${PROJECT_NAME}
+# )
```
bmagyar, line 65: boop

motion_primitives_forward_controller/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Architecture Overview
+The following diagram shows the architecture for a UR robot.
+For this setup, the [`motion_primitive_ur_driver`](https://github.com/StoglRobotics-forks/Universal_Robots_ROS2_Driver_MotionPrimitive) is used.
+
+![UR Robot Architecture](doc/ros2_control_motion_primitives_ur_whiteBackground.drawio.png)
+
+
+1. **Command Reception**  
+   Python scripts can publish motion primitives to the `~/reference` topic. These messages are received by the `reference_callback()` method in the controller.
```
bmagyar, line 29: Do they have to be Python scripts? CLI, C++ or Rust are disqualified? ;)

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_primitives_forward_controller.hpp
```diff
@@ -0,0 +1,112 @@
...
+  using ControllerStatePublisher = realtime_tools::RealtimePublisher<ControllerStateMsg>;
+
+  rclcpp::Publisher<ControllerStateMsg>::SharedPtr s_publisher_;
+  std::unique_ptr<ControllerStatePublisher> state_publisher_;
+
+private:
+  // callback for topic interface
+  MOTION_PRIMITIVES_CONTTROLLER_PKG__VISIBILITY_LOCAL
+  void reference_callback(const std::shared_ptr<ControllerReferenceMsg> msg); // callback for reference message
+  // std::atomic<bool> new_msg_available_ = false; // flag to indicate if new message is available
```
bmagyar, line 98: boop

bmagyar (changes requested):
There is a lot of stuff coming from templates in this PR and you either used some AI tool for coding or you are very good at writing AI-style comments.

Please try removing all commented code or fix them and remove all comments that add nothing to the code itself (which in this code is almost all comments...)

motion_primitives_forward_controller/README.md
```diff
@@ -0,0 +1,48 @@
+motion_primitives_forward_controller
+==========================================
+
+Package to control robots using motion primitives like LINEAR_JOINT (PTP/ MOVEJ), LINEAR_CARTESIAN (LIN/ MOVEL) and CIRCULAR_CARTESIAN (CIRC/ MOVEC)
+
+![Licence](https://img.shields.io/badge/License-BSD-3-Clause-blue.svg)
```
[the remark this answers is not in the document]
mathias31415: fixed it:)

motion_primitives_forward_controller/README.md
mathias31415: Obviously, it doesn't have to be a Python script. A Python script is just what I provided in the [`motion_primitive_ur_driver`](https://github.com/StoglRobotics-forks/Universal_Robots_ROS2_Driver_MotionPrimitive/blob/main/ur_robot_driver/examples/send_dummy_motion_primitives.py) . I clarified that in the README ;)

mathias31415:
Thanks @bmagyar for your review! I've addressed and resolved all the comments. Looking forward to any further feedback!

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+  rclcpp
+  rclcpp_lifecycle
+  realtime_tools
+  std_srvs
+)
+
+find_package(ament_cmake REQUIRED)
+find_package(generate_parameter_library REQUIRED)
+find_package(ament_cmake_gmock REQUIRED)
+find_package(controller_manager REQUIRED)
```
fmauch, line 22: IMHO test dependencies should be find_packaged inside the `if(BUILD_TESTING)` block below.

Also, the `package.xml` misses some `<test_depend>` entries, e.g. `controller_manager`.

fmauch (changes requested):
This is an incomplete early review. I just noticed that I needed to do this to get it built locally.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,485 @@
...
+    return controller_interface::CallbackReturn::ERROR;
+  }
+
+  // topics QoS
+  auto subscribers_qos = rclcpp::SystemDefaultsQoS();
+  subscribers_qos.keep_last(1);
+  subscribers_qos.best_effort();
+
+  // Reference Subscriber
+  ref_subscriber_ = get_node()->create_subscription<ControllerReferenceMsg>(
```
fmauch, line 106: I honestly think that the interface should be an action rather than a topic. (or, provide both, as the JTC does). That would also remove the need of the helpers you introduced in the interfaces package.

Furthermore, I think it makes sense to always provide a sequence as an interface. The sequence can, obviously, also have size 1. 

We have prepared the [`ExecuteMotion` action](https://github.com/UniversalRobots/industrial_robot_motion_interfaces/blob/main/action/ExecuteMotion.action) for that already.

motion_primitives_forward_controller/CMakeLists.txt
mathias31415: Fixed it, thanks for the hint

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
mathias31415: @fmauch I changed it to the action you recommended. Looking forward to your feedback.

mathias31415:
It would also be great if the actions and messages from [`industrial_robot_motion_interfaces`](https://github.com/UniversalRobots/industrial_robot_motion_interfaces) could be merged into the appropriate ROS 2 repositories if possible @fmauch / @urfeex :)

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+add_library(
+  motion_primitives_forward_controller
+  SHARED
+  src/motion_primitives_forward_controller.cpp
+)
+target_include_directories(motion_primitives_forward_controller PUBLIC
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>")
+target_link_libraries(motion_primitives_forward_controller motion_primitives_forward_controller_parameters)
+ament_target_dependencies(motion_primitives_forward_controller
```
fmauch: `ament_target_dependencies` is deprecated and was even temporarily removed from Rolling. See https://github.com/ament/ament_cmake/issues/584#issuecomment-2931061571 for details.

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+  TARGETS
+    motion_primitives_forward_controller
+  RUNTIME DESTINATION bin
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+)
+
+install(
+  DIRECTORY include/
+  DESTINATION include
```
fmauch:
```suggestion
  DESTINATION include/${PROJECT_NAME}
```

See https://colcon.readthedocs.io/en/released/user/overriding-packages.html#install-headers-to-a-unique-include-directory

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+  find_package(hardware_interface REQUIRED)
+  find_package(ros2_control_test_assets REQUIRED)
+
+  ament_add_gmock(
+    test_load_motion_primitives_forward_controller
+    test/test_load_motion_primitives_forward_controller.cpp
+  )
+  target_include_directories(
+    test_load_motion_primitives_forward_controller PRIVATE include
+  )
```
fmauch: Wouldn't it make more sense to add `motion_primitives_forward_controller` to the target_link_libraries` instead of manually adding the include dir?

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+  ament_target_dependencies(
```
fmauch: `target_link_libraries`

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+    test/test_motion_primitives_forward_controller.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/test/motion_primitives_forward_controller_params.yaml
+  )
+  target_include_directories(
+    test_motion_primitives_forward_controller PRIVATE include
+  )
+  target_link_libraries(
+    test_motion_primitives_forward_controller motion_primitives_forward_controller
+  )
+  ament_target_dependencies(
```
fmauch: `target_link_libraries`

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+    test/test_motion_primitives_forward_controller_preceeding.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/test/motion_primitives_forward_controller_preceeding_params.yaml
+  )
+  target_include_directories(
+    test_motion_primitives_forward_controller_preceeding PRIVATE include
+  )
+  target_link_libraries(
+    test_motion_primitives_forward_controller_preceeding motion_primitives_forward_controller
+  )
+  ament_target_dependencies(
```
fmauch: `target_link_libraries`

motion_primitives_forward_controller/README.md
```diff
@@ -0,0 +1,38 @@
+motion_primitives_forward_controller
+==========================================
+
+Package to control robots using motion primitives like LINEAR_JOINT (PTP/ MOVEJ), LINEAR_CARTESIAN (LIN/ MOVEL) and CIRCULAR_CARTESIAN (CIRC/ MOVEC)
+
+[![Licence](https://img.shields.io/badge/License-Apache-2.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+
+
+This project provides an interface for sending motion primitives to an industrial robot controller using the `ExecuteMotion.action` action from the [industrial_robot_motion_interfaces](https://github.com/b-robotized-forks/industrial_robot_motion_interfaces/tree/helper-types) package. The controller receives the primitives via the action interface and forwards them through command interfaces to the robot-specific hardware interface. Currently, hardware interfaces for [Universal Robots](https://github.com/b-robotized-forks/Universal_Robots_ROS2_Driver_MotionPrimitive) and [KUKA Robots](https://github.com/b-robotized-forks/kuka_experimental/tree/motion_primitive_kuka_driver) are implemented.
```
fmauch: Let's try to reference the final package destinations here.

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_primitives_forward_controller.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include "industrial_robot_motion_interfaces/msg/motion_primitive.hpp"
+#include "rclcpp_action/rclcpp_action.hpp"
+
+namespace motion_primitives_forward_controller
+{
+// name constants for state interfaces
+static constexpr size_t STATE_MY_ITFS = 0;
+
+// name constants for command interfaces
+static constexpr size_t CMD_MY_ITFS = 0;
```
fmauch: That's certainly a way to handle this. In my opinion keeping a reference in the controller rather than an index is more readable and less error-prune.

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_type.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef MOTION_PRIMITIVES_FORWARD_CONTROLLER__MOTION_TYPE_HPP_
+#define MOTION_PRIMITIVES_FORWARD_CONTROLLER__MOTION_TYPE_HPP_
+
+namespace MotionType
+{
+// Motion Primitives
+static constexpr uint8_t LINEAR_JOINT = 10;  // changed to 10 because 0 is default value
+static constexpr uint8_t LINEAR_CARTESIAN = 50;
+static constexpr uint8_t CIRCULAR_CARTESIAN = 51;
```
fmauch: Wouldn't it be better to re-use the constants defined in https://github.com/UniversalRobots/industrial_robot_motion_interfaces/blob/main/msg/MotionPrimitive.msg?

motion_primitives_forward_controller/include/motion_primitives_forward_controller/ready_for_new_primitive.hpp
fmauch, line 1: How about defining those in an enum inside the controller's namespace rather than creating a new namespace? in a new file?

motion_primitives_forward_controller/include/motion_primitives_forward_controller/validate_motion_primitives_forward_controller_parameters.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace parameter_traits
+{
+Result forbidden_interface_name_prefix(rclcpp::Parameter const & parameter)
+{
+  auto const & interface_name = parameter.as_string();
+
+  if (interface_name.rfind("blup_", 0) == 0)
+  {
+    return ERROR("'interface_name' parameter can not start with 'blup_'");
```
fmauch: That seems overly specific. Is that validation function really useful or was it only added from a template?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.yaml
```diff
@@ -0,0 +1,33 @@
...
+  state_interfaces: {
+    type: string_array,
+    default_value: [],
+    description: "Names of the state interfaces used by the controller.",
+    read_only: true,
+    validation: {
+      unique<>: null,
+      not_empty<>: null,
+    }
+  }
```
fmauch: Wouldn't it be simpler to just name the command_interfaces and state_interfaces without the `name/` prefix? I don't really see the benefits of doing that. It removes one small configuration duplication with the cost more complexity that seems not to be justified in that case. (My opinion)

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+  RMW_QOS_POLICY_HISTORY_KEEP_ALL,
+  1,  // message queue depth
+  RMW_QOS_POLICY_RELIABILITY_RELIABLE,
+  RMW_QOS_POLICY_DURABILITY_VOLATILE,
+  RMW_QOS_DEADLINE_DEFAULT,
+  RMW_QOS_LIFESPAN_DEFAULT,
+  RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT,
+  RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT,
+  false};
+}  // namespace
```
fmauch: This appears to be unused.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+  try
+  {
+    param_listener_ =
+      std::make_shared<motion_primitives_forward_controller::ParamListener>(get_node());
+  }
+  catch (const std::exception & e)
+  {
+    fprintf(stderr, "Exception thrown during controller's init with message: %s \n", e.what());
+    return controller_interface::CallbackReturn::ERROR;
+  }
```
fmauch: Why is that in a try-catch block?

And: Maybe use ROS logging here? Maybe the message should also be more specific?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+namespace motion_primitives_forward_controller
+{
+MotionPrimitivesForwardController::MotionPrimitivesForwardController()
+: controller_interface::ControllerInterface()
+{
+}
+
+controller_interface::CallbackReturn MotionPrimitivesForwardController::on_init()
+{
+  RCLCPP_INFO(get_node()->get_logger(), "Initializing Motion Primitives Forward Controller");
```
fmauch: Bear in mind that this controller will be part of a potentially large set of controllers to be started. There's a lot of output going on already and the controller manager will also say which controller it is loading. So maybe make that a debug output?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+    return controller_interface::CallbackReturn::ERROR;
+  }
+
+  // Check if there are exactly 25 command interfaces
+  if (params_.command_interfaces.size() != 25)
+  {  // motion_type + 6 joints + 2*7 positions + blend_radius + velocity + acceleration + move_time
+    RCLCPP_ERROR(
+      get_node()->get_logger(), "Error: Exactly 25 command interfaces must be provided!");
+    return controller_interface::CallbackReturn::ERROR;
+  }
```
fmauch: Do they all have to provide an interface for position, velocity and acceleration?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+    state_interfaces_config.names.push_back(params_.name + "/" + interface_name);
+  }
+  return state_interfaces_config;
+}
+
+controller_interface::CallbackReturn MotionPrimitivesForwardController::on_activate(
+  const rclcpp_lifecycle::State & /*previous_state*/)
+{
+  reset_command_interfaces();
+  RCLCPP_INFO(get_node()->get_logger(), "Controller activated");
```
fmauch: Same comment about logging.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+  return controller_interface::CallbackReturn::SUCCESS;
+}
+
+controller_interface::CallbackReturn MotionPrimitivesForwardController::on_deactivate(
+  const rclcpp_lifecycle::State & /*previous_state*/)
+{
+  reset_command_interfaces();
+  RCLCPP_INFO(get_node()->get_logger(), "Controller deactivated");
```
fmauch: Same comment about logging.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+rclcpp_action::CancelResponse MotionPrimitivesForwardController::goal_cancelled_callback(
+  const std::shared_ptr<rclcpp_action::ServerGoalHandle<ExecuteMotion>>)
+{
+  RCLCPP_INFO(get_node()->get_logger(), "Canceling goal");
+
+  std::lock_guard<std::mutex> guard(command_mutex_);
+  reset_command_interfaces();
+  robot_stopped_ = true;
+  // send stop command immediately to the hw-interface
+  (void)command_interfaces_[0].set_value(static_cast<double>(MotionType::STOP_MOTION));
```
fmauch: Instead of writing to the command interfaces directly, you could write to a buffer that is then being read in the `update` method. An `std::atomic<bool> cancel_requested_;` as a member should do.

With the mutex you effectively delay the control thread massively.

All of the state transitions can be done in the control thread (inside the `update` method). This callback can return before cancellation has actually finished. It's only necessary to trigger the stop and report back that the stop has been triggered.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+  robot_stopped_ = false;
+
+  RCLCPP_INFO(get_node()->get_logger(), "Robot stopped, ready for new motion primitives.");
+  return rclcpp_action::CancelResponse::ACCEPT;
+}
+
+void MotionPrimitivesForwardController::goal_accepted_callback(
+  const std::shared_ptr<rclcpp_action::ServerGoalHandle<ExecuteMotion>> goal_handle)
+{
+  std::thread{std::bind(&MotionPrimitivesForwardController::execute_goal, this, goal_handle)}
```
fmauch: Starting this in a thread will bring you into all kinds of thread-related trouble. I'm not sure whether that's necessarily worth the pain.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+    .detach();
+}
+
+void MotionPrimitivesForwardController::execute_goal(
+  const std::shared_ptr<rclcpp_action::ServerGoalHandle<ExecuteMotion>> goal_handle)
+{
+  const auto & goal = goal_handle->get_goal();
+
+  if (robot_stopped_)
```
fmauch: This is missing thread synchronization. It should at least be a `std::atomic<bool>`.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+}
+
+rclcpp_action::CancelResponse MotionPrimitivesForwardController::goal_cancelled_callback(
+  const std::shared_ptr<rclcpp_action::ServerGoalHandle<ExecuteMotion>>)
+{
+  RCLCPP_INFO(get_node()->get_logger(), "Canceling goal");
+
+  std::lock_guard<std::mutex> guard(command_mutex_);
+  reset_command_interfaces();
+  robot_stopped_ = true;
```
fmauch: nitpicking: Isn't that variable encoding "robot stop requested" rather than "robot stopped"?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+      &MotionPrimitivesForwardController::goal_cancelled_callback, this, std::placeholders::_1),
+    std::bind(
+      &MotionPrimitivesForwardController::goal_accepted_callback, this, std::placeholders::_1));
+
+  queue_size_ = params_.queue_size;
+  if (queue_size_ == 0)
+  {
+    RCLCPP_ERROR(get_node()->get_logger(), "Error: Queue size must be greater than 0!");
+    return controller_interface::CallbackReturn::ERROR;
+  }
```
fmauch: Why would it be necessary to restrict the queue size?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+  if (moprim_queue_.size() + required_queue_space > queue_size_)
+  {
+    RCLCPP_ERROR(
+      get_node()->get_logger(),
+      "Received more motion primitives than queue can store. "
+      "Current queue size: %zu, required space: %zu. "
+      "Please increase the queue size in the controller parameters.",
+      moprim_queue_.size(), required_queue_space);
+    return;
+  }
```
fmauch: As mentioned above, I don't quite see the necessity of restricting this.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+    std::shared_ptr<MotionPrimitiveMsg> start_marker = std::make_shared<MotionPrimitiveMsg>();
+    start_marker->type = MotionType::MOTION_SEQUENCE_START;
+    moprim_queue_.push(start_marker);
+  }
+
+  // Add motion primitives to the queue
+  for (const auto & primitive : primitives)
+  {
+    moprim_queue_.push(std::make_shared<MotionPrimitiveMsg>(primitive));
+  }
```
fmauch: You could wrap that into a lambda and then only have one conditional block:

```c++
auto add_motions = [this](primitives){
  for (const auto & primitive : primitives)
  {
    moprim_queue_.push(std::make_shared<MotionPrimitiveMsg>(primitive));
  }
};

if (primitives.size() > 1)
{
  std::shared_ptr<MotionPrimitiveMsg> start_marker = std::make_shared<MotionPrimitiveMsg>();
  start_marker->type = MotionType::MOTION_SEQUENCE_START;
  moprim_queue_.push(start_marker);

  add_motions(primitives);

  std::shared_ptr<MotionPrimitiveMsg> end_marker = std::make_shared<MotionPrimitiveMsg>();
  end_marker->type = MotionType::MOTION_SEQUENCE_END;
  moprim_queue_.push(end_marker);
}
else
{
  add_motions(primitives);
}
```

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+
+  RCLCPP_INFO(
+    get_node()->get_logger(), "Accepted goal with %zu motion primitives%s.", primitives.size(),
+    add_sequence_wrappers ? " (wrapped in sequence markers)" : "");
+
+  ExecuteMotion::Result result;
+  rclcpp::Rate rate(50);
+  while (rclcpp::ok())
+  {
+    auto execution_status = static_cast<uint8_t>(std::round(state_interfaces_[0].get_value()));
```
fmauch: You are basically accessing a shared resource (`state_interfaces_` gets set inside the control thread) without thread synchronization. One more point against using the execute thread.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+  auto opt_value_ready = state_interfaces_[1].get_optional();
+  if (!opt_value_ready.has_value())
+  {
+    RCLCPP_ERROR(
+      get_node()->get_logger(), "State interface 1 (ready_for_new_primitive) returned no value.");
+    return controller_interface::return_type::ERROR;
+  }
+  uint8_t ready_for_new_primitive = static_cast<int8_t>(std::round(opt_value_ready.value()));
+
+  if (!moprim_queue_.empty())  // check if new command is available
```
fmauch: Access to shared resource without thread synchronization. I am not sure whether `empty()` is a thread-safe call, but I wouldn't expect it.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,554 @@
...
+  {
+    switch (ready_for_new_primitive)
+    {
+      case ReadyForNewPrimitive::NOT_READY:
+      {
+        return controller_interface::return_type::OK;
+      }
+      case ReadyForNewPrimitive::READY:
+      {
+        if (!set_command_interfaces())
```
fmauch: This is where you potentially block the control thread as this will wait for aquiring the lock for accessing the queue.

fmauch (changes requested):
Great work so far!

In my opinion this lacks a bit of a threading concept as the RT thread must not wait under any circumstances which is not given by this implementation as far as I can see.

motion_primitives_forward_controller/doc/Moprim_Controller_ExecuteMotion_Action.drawio.png
fmauch, line 1: I guess, the German came in with the "Python S**k**ript" ;-)

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
mathias31415: Some of the controllers use a try-catch block here.
For example:
https://github.com/ros-controls/ros2_controllers/blob/2800f21c048dcd91973a1c9452dba8a16d5fde69/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L53 
```
controller_interface::CallbackReturn JointTrajectoryController::on_init()
{
  try
  {
    // Create the parameter listener and get the parameters
    param_listener_ = std::make_shared<ParamListener>(get_node());
    params_ = param_listener_->get_params();
  }
  catch (const std::exception & e)
  {
    fprintf(stderr, "Exception thrown during init stage with message: %s \n", e.what());
    return CallbackReturn::ERROR;
  }
```

https://github.com/ros-controls/ros2_controllers/blob/2800f21c048dcd91973a1c9452dba8a16d5fde69/admittance_controller/src/admittance_controller.cpp#L71
```
controller_interface::CallbackReturn AdmittanceController::on_init()
{
  // initialize controller config
  try
  {
    parameter_handler_ = std::make_shared<admittance_controller::ParamListener>(get_node());
    admittance_ = std::make_unique<admittance_controller::AdmittanceRule>(parameter_handler_);
  }
  catch (const std::exception & e)
  {
    RCLCPP_ERROR(
      get_node()->get_logger(), "Exception thrown during init stage with message: %s \n", e.what());
    return controller_interface::CallbackReturn::ERROR;
  }
```

Do you recommend removing the try-catch block and just keeping:
`param_listener_ =
      std::make_shared<motion_primitives_forward_controller::ParamListener>(get_node());`

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_type.hpp
mathias31415: I’ve now changed it to an enum in the motion_primitives_forward_controller namespace, similar to `ExecutionState` and `ReadyForNewPrimitive`.
If I reuse the constants defined in `industrial_robot_motion_interfaces/msg/MotionPrimitive.msg`, I’m missing the stop parameters `STOP_MOTION` and `RESET_STOP`, as well as the sequence parameters `MOTION_SEQUENCE_START` and `MOTION_SEQUENCE_END`.
What is your recommendation for this?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
mathias31415: I have removed the execution thread. The stop commands and primitives are now all set from within the update() function. I also removed the mutex.
If I understand correctly, calling `if (!set_command_interfaces())` should no longer be an issue, right?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
mathias31415: What do you mean by "all"?

1 motion_type
6 joints         // joint positions for PTP
2 * 7 positions   // (x,y,z,qx,qy,qz,qw) For LIN 1 * 7. For  CIRC 2 * 7 (via and goal pose)
1 blend_radius
1 velocity      // one joint or cartesian speed value depending on the motion_type
1 acceleration  //                  "               acceleration                       "
1 move_time    // as an alternative for vel + acc
= 25

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
mathias31415: I introduced a `std::atomic<bool>` called `moprim_queue_write_enabled_`. In `goal_received_callback()`, moprim_queue_write_enabled_ is checked to determine if the queue is enabled for writing, otherwise the goal is rejected. In `goal_accepted_callback()`, all primitives are first written into the queue and then `moprim_queue_write_enabled_` is set to `false`, allowing the queue to be read in the `update()` function. The flag is only set back to `true` once the queue is empty.

This means that the primitives are only forwarded to the hardware interface once all primitives of the action goal have been written to the queue. I think this approach is acceptable.

Do you have a better suggestion for making `moprim_queue_` thread-safe without using a mutex?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
fmauch: I was not aware of [this](https://github.com/PickNikRobotics/generate_parameter_library/blob/ba34f27c9eccfc9d55aa47edce909ae0a1305df9/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header#L94), so catching that might be a good option, indeed.

Using ROS logging might still be beneficial from my point of view.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
fmauch: I mean: Does every hardware interface using this controller have to provide interfaces for position, velocity and acceleration?

But I think, I forgot the fact, that we are talking about https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1341/files#diff-93110686145e59c58f86e3a61c7e8fdb0c9c66b8e6160918d72d09119498e0f9R302-R331 and not the standard joint interfaces.

All good then, I guess.

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_primitives_forward_controller.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+protected:
+  std::shared_ptr<motion_primitives_forward_controller::ParamListener> param_listener_;
+  motion_primitives_forward_controller::Params params_;
+
+  using MotionPrimitive = industrial_robot_motion_interfaces::msg::MotionPrimitive;
+  std::queue<std::shared_ptr<MotionPrimitive>> moprim_queue_;
+
+  using ExecuteMotion = industrial_robot_motion_interfaces::action::ExecuteMotion;
+  rclcpp_action::Server<ExecuteMotion>::SharedPtr action_server_;
```
fmauch: May I suggest using the [`realtime_tools::RealtimeServerGoalHandle`](https://github.com/ros-controls/realtime_tools/blob/master/realtime_tools/include/realtime_tools/realtime_server_goal_handle.hpp) for this? This way you will gain
- Proper thread synchronization between the accepted callback and the update method
- non-blocking calls to set the result in the update method.

I am not sure how severe things are with the latter point, but it seems like the right choice nevertheless. The JTC and also our passthrough trajectory controller use that.

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_primitives_forward_controller.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+  controller_interface::return_type update(
+    const rclcpp::Time & time, const rclcpp::Duration & period) override;
+
+protected:
+  std::shared_ptr<motion_primitives_forward_controller::ParamListener> param_listener_;
+  motion_primitives_forward_controller::Params params_;
+
+  using MotionPrimitive = industrial_robot_motion_interfaces::msg::MotionPrimitive;
+  std::queue<std::shared_ptr<MotionPrimitive>> moprim_queue_;
```
fmauch: With that queue you are creating a couple of potential issues:

- pushing and popping from this in the real-time thread will create new (shared pointer) objects and therefore allocate memory in the real-time thread. One way to circumvent this would be to use an `std::vector<MotionPrimitive>` container and instead of popping items increase an index variable to get **references** to the respective data stored at the index.
- You'll need to do thread synchronization as you did with the `moprim_queue_write_enabled_` member. That definitively does work that way, but seems a bit fragile because it doesn't really **protect** the queue as a mutex would.

You could potentially use the [`realtime_tools::LockFreeQueue`](https://github.com/ros-controls/realtime_tools/blob/master/realtime_tools/include/realtime_tools/lock_free_queue.hpp) to avoid both problems.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
fmauch: See my comment about the LockFreeQueue below.

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_primitives_forward_controller.hpp
mathias31415: I changed `moprim_queue_` to use `realtime_tools::LockFreeSPSCQueue`. From what I understand, the queue now requires a compile-time fixed capacity instead of a dynamic size.

urfeex:
@mathias31415 I've added a PR to control_msgs with the motion primitives in https://github.com/ros-controls/control_msgs/pull/228. Would you like to use those for your PR?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -145,9 +144,10 @@ controller_interface::return_type MotionPrimitivesForwardController::update(
...
     // send stop command immediately to the hw-interface
     (void)command_interfaces_[0].set_value(static_cast<double>(MotionType::STOP_MOTION));
-    while (!moprim_queue_.empty())
-    {  // clear the queue
-      moprim_queue_.pop();
+    std::shared_ptr<MotionPrimitive> primitive;
+    while (moprim_queue_.pop(primitive))
+    {
+      // clear the queue
     }
```
urfeex: You could use `get_latest` to achieve the same.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -145,9 +144,10 @@ controller_interface::return_type MotionPrimitivesForwardController::update(
     reset_command_interfaces();
     // send stop command immediately to the hw-interface
     (void)command_interfaces_[0].set_value(static_cast<double>(MotionType::STOP_MOTION));
-    while (!moprim_queue_.empty())
-    {  // clear the queue
-      moprim_queue_.pop();
+    std::shared_ptr<MotionPrimitive> primitive;
```
urfeex: Maybe preallocate that outside of the update method?

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -308,13 +305,10 @@ void MotionPrimitivesForwardController::reset_command_interfaces()
 bool MotionPrimitivesForwardController::set_command_interfaces()
 {
   // Get the oldest message from the queue
-  std::shared_ptr<MotionPrimitive> current_moprim = moprim_queue_.front();
-  moprim_queue_.pop();
-
-  // Check if the message is valid
-  if (!current_moprim)
+  std::shared_ptr<MotionPrimitive> current_moprim;
```
urfeex: Use the same preallocated one as mentioned before.

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_primitives_forward_controller.hpp
```diff
@@ -91,7 +91,7 @@ class MotionPrimitivesForwardController : public controller_interface::Controlle
   motion_primitives_forward_controller::Params params_;
 
   using MotionPrimitive = industrial_robot_motion_interfaces::msg::MotionPrimitive;
-  std::queue<std::shared_ptr<MotionPrimitive>> moprim_queue_;
+  realtime_tools::LockFreeSPSCQueue<std::shared_ptr<MotionPrimitive>, 128> moprim_queue_;
```
urfeex: A capacity of 128 seems to be a bit low. Also, it should be checked when accepting a trajectory.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -479,27 +465,36 @@ void MotionPrimitivesForwardController::goal_accepted_callback(
   {
     for (const auto & primitive : motion_primitives)
     {
-      moprim_queue_.push(std::make_shared<MotionPrimitive>(primitive));
+      if (!moprim_queue_.push(std::make_shared<MotionPrimitive>(primitive)))
```
urfeex: What is the reason for storing the primitives in shared pointers? Wouldn't it be simpler just to push the primitives into the queue directly?

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_primitives_forward_controller.hpp
mathias31415: I added the check. What capacity do you recommend?

urfeex: I guess, having at least 1024 or even 2048 might be good. When planning motions with short segments you might end up with that number quite quickly. I think, going with 1024 for the beginning might be good. We can see how it's going to be used, then and then increase it if needed.

motion_primitives_forward_controller/include/motion_primitives_forward_controller/motion_type.hpp
mathias31415: I’ve adjusted it again now. MotionType is now taken from the msg definition, and only the MotionHelperTypes are defined in the enum within the controller. In my opinion, this should make the whole thing a bit less error-prone. What’s your opinion on this approach?

urfeex: That looks great!

motion_primitives_forward_controller/package.xml
```diff
@@ -0,0 +1,43 @@
...
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <build_depend>generate_parameter_library</build_depend>
+
+  <depend>control_msgs</depend>
+  <depend>controller_interface</depend>
+  <depend>hardware_interface</depend>
+  <depend>pluginlib</depend>
+  <depend>industrial_robot_motion_interfaces</depend>
```
fmauch: This needs to get removed (and all places where this was used replaced by https://github.com/ros-controls/control_msgs/pull/228)

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,515 @@
...
+{
+  controller_interface::InterfaceConfiguration command_interfaces_config;
+  command_interfaces_config.type = controller_interface::interface_configuration_type::INDIVIDUAL;
+
+  command_interfaces_config.names.reserve(params_.command_interfaces.size());
+
+  // Iterate over all command interfaces from the config yaml file
+  for (const auto & interface_name : params_.command_interfaces)
+  {
+    command_interfaces_config.names.push_back("motion_primitive/" + interface_name);
```
urfeex: Currently, all 25 command interfaces are given to the controller and then added with the `motion_primitive` prefix. The actual names of the interfaces don't matter as far as I can see from your implementation, they could also be `i0` to `i24`. The prefix, however (the gpio group name), has to be exactly `motion_primitive`. That's not necessarily the case. As [in your implementation for UR](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1341/files#diff-93110686145e59c58f86e3a61c7e8fdb0c9c66b8e6160918d72d09119498e0f9R309), the group name is `${tf_prefix}motion_primitive` which would mean that this controller would not work if a `tf_prefix` is set. I would suggest to either

- pass full interface names to the controller (such as `$(var tf_prefix)motion_primitive/q1`) or
- pass both, the group name and each interface name as separate parameters or
- Have the interface names fixed and only pass the group name

mathias31415: I adjusted it so that the full name is passed to the controller ($(var tf_prefix)motion_primitive/q1).

ros2_controllers.rolling.repos
```diff
@@ -9,8 +9,10 @@ repositories:
     version: master
   control_msgs:
     type: git
-    url: https://github.com/ros-controls/control_msgs.git
-    version: master
+    url: https://github.com/urfeex/control_msgs.git
+    version: motion_primitives
+    # url: https://github.com/ros-controls/control_msgs.git
+    # version: master
```
```suggestion
    url: https://github.com/ros-controls/control_msgs.git
    version: master
```

motion_primitives_forward_controller/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+If multiple motion primitives are passed to the controller via the action, the controller forwards them to the hardware interface as a sequence. To do this, it first sends `MOTION_SEQUENCE_START`, followed by each individual primitive, and finally `MOTION_SEQUENCE_END`. All primitives between these two markers will be executed as a single, continuous sequence. This allows seamless transitions (blending) between primitives.
+
+The action interface also allows stopping the current execution of motion primitives. When a stop request is received, the controller sends `STOP_MOTION` to the hardware interface, which then halts the robot's movement. Once the controller receives confirmation that the robot has stopped, it sends `RESET_STOP` to the hardware interface. After that, new commands can be sent.
+
+![Action Image](doc/Moprim_Controller_ExecuteMotion_Action.drawio.png)
+
+This can be done, for example, via a Python script as demonstrated in the [`example python script`](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/examples/send_dummy_motion_primitives_hka_ur10e.py) in the `Universal_Robots_ROS2_Driver` package.
```
fmauch:
```suggestion
This can be done, for example, via a Python script as demonstrated in the [`example python script`](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/examples/ssend_dummy_motion_primitives_ur10e.py) in the `Universal_Robots_ROS2_Driver` package.
```

motion_primitives_forward_controller/test/test_motion_primitives_forward_controller_preceeding.cpp
fmauch, line 1: Am I missing anything or is this test exactly contained in `motion_primitives_forward_controller/test/test_motion_primitives_forward_controller.cpp`? From my understanding the `proceeding` test could be deleted.

fmauch (commented):
This looks good to me now. Before finally approving this, I would prefer getting https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1341 to it's final shape in case we do find something we need to adapt here. We are pretty close to getting that done, though, as well in my opinion.

motion_primitives_forward_controller/src/motion_primitives_forward_controller.cpp
```diff
@@ -0,0 +1,538 @@
...
+    }
+  }
+  return true;
+}
+
+rclcpp_action::GoalResponse MotionPrimitivesForwardController::goal_received_callback(
+  const rclcpp_action::GoalUUID &, std::shared_ptr<const ExecuteMotionAction::Goal> goal)
+{
+  RCLCPP_INFO(get_node()->get_logger(), "Received new action goal");
+
```
fmauch, line 372: While testing it occurred to me that trajectories were accepted even though the controller was not running. I think it would be an improvement to reject trajectories when the controller isn't active.

```suggestion

  // Precondition: Running controller
  if (get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE) {
    RCLCPP_ERROR(get_node()->get_logger(), "Can't accept new trajectories. Controller is not running.");
    return rclcpp_action::GoalResponse::REJECT;
  }

```

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
+cmake_minimum_required(VERSION 3.8)
+project(motion_primitives_forward_controller)
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
+  add_compile_options(-Wall -Wextra -Werror=conversion -Werror=unused-but-set-variable -Werror=return-type -Werror=shadow)
+endif()
```
christophfroehlich:
```suggestion
find_package(ros2_control_cmake REQUIRED)
set_compiler_options()
export_windows_symbols()
```
to be consistent with the other packages in ros2_control

motion_primitives_forward_controller/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
+cmake_minimum_required(VERSION 3.8)
```
christophfroehlich:
```suggestion
cmake_minimum_required(VERSION 3.10)
```

<3.10 is deprecated

motion_primitives_forward_controller/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Demo-Video with KR3
+[![KUKA demo video](doc/moprim_forward_controller_kuka_demo_thumbnail.png)](https://youtu.be/_BWCO36j9bg)
+
+
+# TODOs/ improvements
+- Use references for command and state interfaces to improve code readability and less error-prone.
+- Extend the tests
+   - Test for a sequence of multiple primitives
+   - Test for canceling movement
```
christophfroehlich:
```suggestion
```

I'd suggest open an issue with a description of the necessary tasks instead. Or rephrase the section as "current limitations".

motion_primitives_forward_controller/README.md
```diff
@@ -0,0 +1,44 @@
+motion_primitive_controllers
+==========================================
+
+Package to control robots using motion primitives like LINEAR_JOINT (PTP/ MOVEJ), LINEAR_CARTESIAN (LIN/ MOVEL) and CIRCULAR_CARTESIAN (CIRC/ MOVEC)
+
+[![Licence](https://img.shields.io/badge/License-Apache-2.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+
```
christophfroehlich:
```suggestion
```

No need to specify this here, as it is global in the whole repo.

motion_primitives_forward_controller/README.md
christophfroehlich, line 1: can you please link the parameter yaml like we have in other controllers?
https://github.com/ros-controls/ros2_controllers/blob/a01fd0bdf99d443c2d8a4ed9a306d27aa5510e52/omni_wheel_drive_controller/doc/userdoc.rst?plain=1#L68-L78

christophfroehlich, line 1: From this description only, it is not clear for me which interfaces a custom hardware needs to export. Also the parameter yaml is not specifc here (e.g. which `command_interfaces` are allowed/needed?)

Or are only the two mentioned drivers supported (then this should be more explicit).

motion_primitives_forward_controller/README.md
mathias31415: I have removed it and will open an issue once the PR has been merged.

motion_primitives_forward_controller/README.md
mathias31415: I added a description for the needed command and state interfaces. Let me know if it’s clarified now.

mathias31415:
Thanks @christophfroehlich for your review. I addressed all your comments. Looking forward to your feedback.

motion_primitives_forward_controller/README.md
christophfroehlich: If the listed interfaces are _required_, why does one has to list them as ROS parameter?

mathias31415: Do you recommend hardcoding them in the controller instead?
@urfeex / @fmauch what do you think about that?

christophfroehlich: If they naming is fixed, then why make it configurable as parameter? But as I said, I haven't look into details of the code if there would be a justification for it.

mathias31415: Yes, you’re actually right. It would also be sufficient to only pass the tf_prefix as a parameter and implement the rest (motion_primitive/<Interface-Name>) fixed in the controller.

I already had a short discussion with Felix about passing the interface names [here](https://github.com/ros-controls/ros2_controllers/pull/1636#discussion_r2213128713).

If you prefer, I can implement the interface names fixed in the controller.

christophfroehlich: Yes please, this seems to be more practical and understandable in my opinion.

mathias31415: I adjusted it so that only tf_prefix is passed as a parameter.

motion_primitives_forward_controller/README.md
```diff
@@ -0,0 +1,68 @@
...
+![Action Image](doc/Moprim_Controller_ExecuteMotion_Action.drawio.png)
+
+This can be done, for example, via a Python script as demonstrated in the [`example python script`](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_robot_driver/examples/send_dummy_motion_primitives_ur10e.py) in the `Universal_Robots_ROS2_Driver` package.
+
+## Command and State Interfaces
+To transmit the motion primitives, the following command and state interfaces are required:
+
+### Command Interfaces
+These interfaces are used to send motion primitive data to the hardware interface:
+- `motion_type`: Type of motion primitive (LINEAR_JOINT, LINEAR_CARTESIAN, CIRCULAR_CARTESIAN)
```
christophfroehlich, line 27: The actual interface is 
`tf_prefix_ + "motion_primitive/motion_type"`
Please add the prefix somehow here (is the same for all command + state interfaces).

mathias31415:
> Oh, I haven't realized earlier that you have written the documentation in a README.md file. This won't be published on control.ros.org per default. You can either convert it to an rst document and add it to the [index](https://github.com/ros-controls/ros2_controllers/blob/master/doc/controllers_index.rst#controllers-for-manipulators-and-other-robots) here, or you import it to a rst wrapper (like [here](https://github.com/ros-controls/control_toolbox/blob/ros2-master/doc/index.rst?plain=1#L10-L11)) and add this rst file again to the index.
> 
> Apart from that LGTM.

@christophfroehlich Could you help me properly include the RST file? I’m getting the doc8 error: No module named 'myst_parser'. What am I missing?

mathias31415:
> > Oh, I haven't realized earlier that you have written the documentation in a README.md file. This won't be published on control.ros.org per default. You can either convert it to an rst document and add it to the [index](https://github.com/ros-controls/ros2_controllers/blob/master/doc/controllers_index.rst#controllers-for-manipulators-and-other-robots) here, or you import it to a rst wrapper (like [here](https://github.com/ros-controls/control_toolbox/blob/ros2-master/doc/index.rst?plain=1#L10-L11)) and add this rst file again to the index.
> > Apart from that LGTM.
> 
> @christophfroehlich Could you help me properly include the RST file? I’m getting the doc8 error: No module named 'myst_parser'. What am I missing?

@christophfroehlich i just saw, that you [added `'--ignore-path=doc/index.rst'` in control_toolbox](https://github.com/ros-controls/control_toolbox/commit/516eccdf217608e0ac45a9ccf4eb48be812b655b#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R117)

christophfroehlich:
@mergifyio backport jazzy

PR #1639, opened 2025-04-16
[CI]  test_force_torque_sensor_broadcaster regularily times out

Juliaj:
This adds debug logs to investigate https://github.com/ros-controls/ros2_controllers/issues/1574. We still saw timeout time to time in CI.


saikishor (changes requested):
Can you add 2 more log info statements at L79 and L84

Just after the semantic component as well as before the publishers

Rest looks spot on

saikishor (approved):
Thank you @Juliaj

christophfroehlich (approved):
Thanks!

PR #1654, opened 2025-04-22
Branch for Jazzy

christophfroehlich:
and 

- some cleanup in the from the reusable workflows from ros2_control_ci
- adding badges for build.ros2.org


christophfroehlich:
already done with #1646

PR #1655, opened 2025-04-23
Fix preceeding->preceding typos

christophfroehlich:
As @juliaj mentioned in https://github.com/ros-controls/ros2_controllers/pull/1166#discussion_r2055262690

I fixed that in a separate PR to avoid polluting the other one ;)


Juliaj:
That's lighting speed!

PR #1666, opened 2025-04-29
Reset the PIDs states when controller is at rest

KmakD:
This feature can help when it is impossible to send commands to hardware but the controller may receive state values.

To justify that let me give an example. You have a robot that has a hardware e-stop. When it is in e-stop state the controller receives a false reading (e.g. someone manually moves the robot). This increases the integral part of the PID, which sets the output to some fixed value. If the robot is smart it won't let you reset the e-stop when it receives non zero command, if not it will move unexpectedly as soon as e-stop is reset. This functionality will reset the PID if the controller is at rest (error and input is 0) for a specified time preventing that.

- [X] Fork the repository.
- [ ] Send a pull request, answering any default questions in the pull request interface.

### DIFF
--- a/pid_controller/src/pid_controller.cpp
+++ b/pid_controller/src/pid_controller.cpp
@@ -496,7 +496,7 @@ controller_interface::return_type PidController::update_and_write_commands(
   {
     for (size_t i = 0; i < measured_state_values_.size(); ++i)
     {
-      measured_state_values_[i] = state_interfaces_[i].get_value();
+      measured_state_values_[i] = state_interfaces_[i].get_optional().value();
     }
   }
 
@@ -533,6 +533,23 @@ controller_interface::return_type PidController::update_and_write_commands(
       }
 
       double error = reference_interfaces_[i] - measured_state_values_[i];
+
+      const auto zero_threshold = params_.gains.dof_names_map[params_.dof_names[i]].zero_threshold;
+      if (std::abs(reference_interfaces_[i]) < zero_threshold && std::abs(error) < zero_threshold)
+      {
+        reset_pid_time_ += period.seconds();
+        if (
+          reset_pid_time_ >
+          params_.gains.dof_names_map[params_.dof_names[i]].zero_threshold_duration)
+        {
+          pids_[i]->reset();
+        }
+      }
+      else
+      {
+        reset_pid_time_ = 0.0;
+      }
+
       if (params_.gains.dof_names_map[params_.dof_names[i]].angle_wraparound)
       {
         // for continuous angles the error is normalized between -pi<error<pi
@@ -601,7 +618,7 @@ controller_interface::return_type PidController::update_and_write_commands(
       state_publisher_->msg_.dof_states[i].time_step = period.seconds();
       // Command can store the old calculated values. This should be obvious because at least one
       // another value is NaN.
-      state_publisher_->msg_.dof_states[i].output = command_interfaces_[i].get_value();
+      state_publisher_->msg_.dof_states[i].output = command_interfaces_[i].get_optional().value();
     }
     state_publisher_->unlockAndPublish();
   }
--- a/pid_controller/src/pid_controller.cpp
+++ b/pid_controller/src/pid_controller.cpp
@@ -496,7 +496,7 @@ controller_interface::return_type PidController::update_and_write_commands(
   {
     for (size_t i = 0; i < measured_state_values_.size(); ++i)
     {
-      measured_state_values_[i] = state_interfaces_[i].get_optional().value();
+      measured_state_values_[i] = state_interfaces_[i].get_value();
     }
   }
 
@@ -618,7 +618,7 @@ controller_interface::return_type PidController::update_and_write_commands(
       state_publisher_->msg_.dof_states[i].time_step = period.seconds();
       // Command can store the old calculated values. This should be obvious because at least one
       // another value is NaN.
-      state_publisher_->msg_.dof_states[i].output = command_interfaces_[i].get_optional().value();
+      state_publisher_->msg_.dof_states[i].output = command_interfaces_[i].get_value();
     }
     state_publisher_->unlockAndPublish();
   }
--- a/pid_controller/include/pid_controller/pid_controller.hpp
+++ b/pid_controller/include/pid_controller/pid_controller.hpp
@@ -86,6 +86,8 @@ protected:
   // Feed-forward velocity weight factor when calculating closed loop pid adapter's command
   std::vector<double> feedforward_gain_;
 
+  double reset_pid_time_;
+
   // Command subscribers and Controller State publisher
   rclcpp::Subscription<ControllerReferenceMsg>::SharedPtr ref_subscriber_ = nullptr;
   realtime_tools::RealtimeBuffer<std::shared_ptr<ControllerReferenceMsg>> input_ref_;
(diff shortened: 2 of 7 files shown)


christophfroehlich:
in your example, when should the controller be reset? if the robot is moved manually, the error won't be zero in the end right?

and in the case where the input is zero, and with some integral action the error also converges to zero. If the controller will be reset, the robot will jump/sag?

bmagyar:
@christophfroehlich @KmakD shouldn't we consider this in the `joint_trajectory_controller` too?

christophfroehlich:
I have objections in this approach, see my unanswered question

KmakD:
Sorry for late respone, I had a few days off.

> in your example, when should the controller be reset?

The robot should be reset only when the input is zero and the error is zero.

> if the robot is moved manually, the error won't be zero in the end right?

Assuming velocity control, when the robot is no longer moved, the velocity feedback becomes zero and the error will be also zero.

> and in the case where the input is zero, and with some integral action the error also converges to zero. If the controller will be reset, the robot will jump/sag?

I'm not sure if I fully understans this, If controller is reset when the input and error are 0, the output will also be 0, so there is no way it will jump/sag

pid_controller/src/pid_controller.cpp
```diff
@@ -533,6 +533,23 @@ controller_interface::return_type PidController::update_and_write_commands(
       }
 
       double error = reference_interfaces_[i] - measured_state_values_[i];
+
+      const auto zero_threshold = params_.gains.dof_names_map[params_.dof_names[i]].zero_threshold;
+      if (std::abs(reference_interfaces_[i]) < zero_threshold && std::abs(error) < zero_threshold)
```
saikishor, line 538: I agree with @christophfroehlich 

Zero itself is a valid reference. IMHO using that to reset the PIDs is not a smart decision

KmakD: I agree, using reference values is not a perfect solution, but I can't think of a better way to detect this. 

There was once an idea to add emergency-stop feature in ros2_control that can be set in hardware component and propagated to all controllers (as a flag if I remember correctly). That would be perfect in this case. What happened to that idea?

Juliaj: is “Configuration/Movement/Safety-critical" interfaces definition ( https://github.com/ros-controls/roadmap/pull/51) relevant for this scenario ?

KmakD: If I understand this correctly this will only allow to configure command interfaces depending if hardware component is active/inactive. In controller we would like to be able to read hardware interface state and act if it was deactivated/halted, or have some global flag that indicates hardware e-stop was triggerd

christophfroehlich:
> > and in the case where the input is zero, and with some integral action the error also converges to zero. If the controller will be reset, the robot will jump/sag?
> 
> I'm not sure if I fully understans this, If controller is reset when the input and error are 0, the output will also be 0, so there is no way it will jump/sag

Let's think of an effort-controlled robot arm, where the reference is velocity. steady-state for zero velocity input (and error) will be a non-zero effort for gravity compensation. 

I understand that your approach might make sense for some robots, but not in general. This means that your addition should be optional, and not default on. setting the default value of `zero_threshold_duration` to `.inf` should do this. 

> To justify that let me give an example. You have a robot that has a hardware e-stop. When it is in e-stop state the controller receives a false reading (e.g. someone manually moves the robot).

Shouldn't your hardware component go in inactive state if the e-stop is hit?

emergency-stop feature is discussed again and again, but no one is working on it right now afaik.

KmakD:
> I understand that your approach might make sense for some robots, but not in general. This means that your addition should be optional, and not default on. setting the default value of `zero_threshold_duration` to `.inf` should do this.

It's already optional, setting `zero_threshold` to 0.0 (the default value) disables it.

> Shouldn't your hardware component go in inactive state if the e-stop is hit?

It should, but right now both state and command interfaces will be available for controller so it will keep making computations even if the hardware is inactive.

christophfroehlich:
> > I understand that your approach might make sense for some robots, but not in general. This means that your addition should be optional, and not default on. setting the default value of `zero_threshold_duration` to `.inf` should do this.
> 
> It's already optional, setting `zero_threshold` to 0.0 (the default value) disables it.

Ok, this is implicit included here what I haven't seen in first place: 
`if (std::abs(reference_interfaces_[i]) < zero_threshold && std::abs(error) < zero_threshold)`
As comparisons agains zero might fail, could you please add an additional check against eps, like we do [here](https://github.com/ros-controls/control_toolbox/blob/b2b3425d07d85f5941f6f002e1d066f259327c62/control_toolbox/include/control_toolbox/pid.hpp#L199-L202)?

> > Shouldn't your hardware component go in inactive state if the e-stop is hit?
> 
> It should, but right now both state and command interfaces will be available for controller so it will keep making computations even if the hardware is inactive.

This will change soon, see https://github.com/ros-controls/ros2_control/pull/2334

saikishor:
Hello @KmakD 

I'm still not sure if this makes sense. What happens if the reference for my arm is zero, then when the PID is tracking it for the zero threshold duration and then it resets and then the arm might fall right? . How do you foresee to correct this issue?. Moreover, when the zero threshold is a parameter, and how it is integrated currently, if I set it to 1, then any value less than 1 will have this effect.

saikishor:
Can a service to reset the PIDs solve your issue?

> To justify that let me give an example. You have a robot that has a hardware e-stop. When it is in e-stop state the controller receives a false reading (e.g. someone manually moves the robot). This increases the integral part of the PID, which sets the output to some fixed value. If the robot is smart it won't let you reset the e-stop when it receives non zero command, if not it will move unexpectedly as soon as e-stop is reset. This functionality will reset the PID if the controller is at rest (error and input is 0) for a specified time preventing that.

As per your description, I personally don't think this is a good approach. Maybe what you need to check is irrespective of the command you are reproducing, the error keeps increasing or your error is huge and not converging

KmakD:
Hi @saikishor 

This is the solution I suggested when it was impossible for controller to tell if hardware can be operated or not. As I mentioned, I know it is not perfect.
With https://github.com/ros-controls/ros2_control/pull/2334 however, this may not be needed anymore. I'll test that when I find some time and come back to report on this.

saikishor:
> Hi @saikishor 
> 
> This is the solution I suggested when it was impossible for controller to tell if hardware can be operated or not. As I mentioned, I know it is not perfect.
> With https://github.com/ros-controls/ros2_control/pull/2334 however, this may not be needed anymore. I'll test that when I find some time and come back to report on this.

Thank you!

KmakD:
I did some tests and found a few issues:

- The controller is only deactivated if the hardware component returns `return_type::DEACTIVATE` in write method. If deactivating hw component using CLI the controller stays active.
- This probably depends on above, but It is possible to activate the controller even if the hardware component is still in INACTIVE state.
- Only the controller that has claimed hardware interfaces is being deactivated. Other controllers that are chained are still active. Not sure if this should work that way. If you try to deactivate the controller that is chained using CLI it will throw an error.
- Right now to resolve the inactive state you have to activate the hardware component and then activate the controller. It would be nice if this could be automated. For example when calling activate on the controller also try to activate hardware component it depends on. Or allow auto reactivation of controller if hardware becomes active again. Or my favorite one but it would require some work: allow to define a "system group" that defines all dependent hardware components and controllers. Then you could call activate on a group and manager would attempt to activate all hw components and controllers. 

Should I open an issue for this?

saikishor:
@KmakD makes sense. I agree that the behaviour should be as you said the controllers should also go to inactive. Can you also share the logic of the throw when deactivating chaining controllers in that issue?

Thank you!

christophfroehlich:
I think we can close this PR. @KmakD If you find new issues as discussed above, please open a dedicated issue for it.

PR #1677, opened 2025-05-11
JTC add more options for tolerance configuration

### LINKED ISSUE #1180: [JTC] Add more options for tolerance configuration
> From the node parameter we cannot set velocity or acceleration tolerances (except for a single `stopped_velocity_tolerance` for the goal tolerances of all joints). Should we add them as parameters as well, to have the same structure like the action message?

This would be nice to have in the follow-up PR.

_Originally posted by @destogl in https://github.com/ros-controls/ros2_controllers/pull/716#pullrequestreview-2136399188_

Vedant87:
Closes #1180 

Added new Parameters velocity_tolerance and acceleration_tolerance, to the Joint Trajectory Controller.


joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
christophfroehlich, line 1: I think the initial idea was to have them per-joint, as it is used in the action 
https://github.com/ros-controls/control_msgs/blob/master/control_msgs/action/FollowJointTrajectory.action
and defined here
https://github.com/ros-controls/control_msgs/blob/master/control_msgs/msg/JointTolerance.msg

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -55,6 +55,10 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
     // Create the parameter listener and get the parameters
     param_listener_ = std::make_shared<ParamListener>(get_node());
     params_ = param_listener_->get_params();
+    // Read new velocity and acceleration tolerances
+    velocity_tolerance_ = get_node()->get_parameter("constraints.velocity_tolerance").as_double();
+    acceleration_tolerance_ =
+      get_node()->get_parameter("constraints.acceleration_tolerance").as_double();
```
christophfroehlich, line 61: you don't need to use get_parameter() method. Have a look at the structure params_ , generate_parameter_library automatically fills this.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
christophfroehlich: I suggest adding the parameters `constraints.__map_joints.trajectory_velocity`, `constraints.__map_joints.goal_velocity` etc. We cannot easily rename the old parameters, as it would change existing code. If necessary, we have to deprecate the old parameters and add something like
constraints.__map_joints.path.position
constraints.__map_joints.path.velocity
constraints.__map_joints.path.acceleration

christophfroehlich (changes requested):
Thanks for tackling this issue. 

This PR is not functional yet, right? However, I added some early review comments.

Vedant87:
> Thanks for tackling this issue. 
> 
> 
> 
> This PR is not functional yet, right? However, I added some early review comments.

Yes this PR is still a work in progress, have been busy with other work hence haven't got enough time to complete this.

Issue #1676, opened 2025-05-09
Update mecanum_drive_controller with common controller features


KmakD:
It appears that the mecanum_drive_controller is lacking some features that are present in other controllers and seem to be important or useful. Compared to the diff_drive_controller, there are several features that could be implemented in the mecanum_drive_controller:

- Velocity limits (`linear.x`, `linear.y`, `angular.z`)
- `cmd_vel_timeout`: Timeout in seconds, after which input command on cmd_vel topic is considered staled.
- tf frame prefix parameters:
  - `tf_frame_prefix_enable`:  Enables or disables appending tf_prefix to tf frame id's.
  - `tf_frame_prefix`: (optional) Prefix to be appended to the tf frames, will be added to odom_id and base_frame_id before publishing. If the parameter is empty, controller's namespace will be used.
- Wheel radius multiplier - Correction factor when radius of wheel differs from the nominal value in `wheel_radius` parameter.
- `publish_rate`: Publishing rate (Hz) of the odometry and TF messages.

All of these features are already implemented in the diff_drive_controller, so it seems the task would be to port them into the mecanum_drive_controller.

christophfroehlich:
regarding the velocity limits, I planned to create a chainable TwistLimiter which can be used with all mobile robot controllers, instead of duplicating the code in all of them.

Regarding the other points, contributions are welcome!

KmakD:
> wheel radius: why one would need that, instead of changing the original parameter?
It's a thing in diff drive controller also separated to left and right wheels. I guess it makes sense there because you can have inflated tires and you may want to correct nominal radius value depending on pressure in the wheels. For mecanum, it looks like it would be useless.

Do you have any idea, when can we expect that TwistLimiter?

KmakD:
I figured that `cmd_vel_timeout` in mecanum coresponds to `reference_timeout`. This made me notice another inconsistency between different controllers regarding topics. For diff and tricycle topic are:
`~/cmd_vel`
`~/odom`
`/tf`

But for mecanum and controllers that use steering_controllers_library it is:
`~/reference`
`~/odometry`
`~/tf_odometry`

Do you plan to unify that at some point? Personally, I find `~/tf_odometry` a bit strange, knowing that some standard tools like Rviz require a global `/tf` topic. In most cases this topic would need to be remapped to be usefull anyway.

Juliaj:
[...]

PR #1680, opened 2025-05-14
Add tf_frame_prefix parameters to mecanum_drive_controller

KmakD:
Focuses one of the points described here: https://github.com/ros-controls/ros2_controllers/issues/1676

Pretty much copied and adapted form diff drive controller. I've tested it on real robot and it worked fine

### DIFF
--- a/mecanum_drive_controller/src/mecanum_drive_controller.cpp
+++ b/mecanum_drive_controller/src/mecanum_drive_controller.cpp
@@ -148,10 +148,31 @@ controller_interface::CallbackReturn MecanumDriveController::on_configure(
     return controller_interface::CallbackReturn::ERROR;
   }
 
+  // Append the tf prefix if there is one
+  std::string tf_prefix = "";
+  if (params_.tf_frame_prefix_enable)
+  {
+    tf_prefix = params_.tf_frame_prefix != "" ? params_.tf_frame_prefix
+                                              : std::string(get_node()->get_namespace());
+
+    // Make sure prefix does not start with '/' and always ends with '/'
+    if (tf_prefix.back() != '/')
+    {
+      tf_prefix = tf_prefix + "/";
+    }
+    if (tf_prefix.front() == '/')
+    {
+      tf_prefix.erase(0, 1);
+    }
+  }
+
+  const auto odom_frame_id = tf_prefix + params_.odom_frame_id;
+  const auto base_frame_id = tf_prefix + params_.base_frame_id;
+
   rt_odom_state_publisher_->lock();
   rt_odom_state_publisher_->msg_.header.stamp = get_node()->now();
-  rt_odom_state_publisher_->msg_.header.frame_id = params_.odom_frame_id;
-  rt_odom_state_publisher_->msg_.child_frame_id = params_.base_frame_id;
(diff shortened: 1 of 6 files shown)


bmagyar (approved):

mecanum_drive_controller/src/mecanum_drive_controller.cpp
```diff
@@ -148,10 +148,31 @@ controller_interface::CallbackReturn MecanumDriveController::on_configure(
...
+    // Make sure prefix does not start with '/' and always ends with '/'
+    if (tf_prefix.back() != '/')
+    {
+      tf_prefix = tf_prefix + "/";
+    }
+    if (tf_prefix.front() == '/')
+    {
+      tf_prefix.erase(0, 1);
+    }
+  }
```
christophfroehlich, line 167: should we move these lines to a shared place, so that we can include it in all mobile base controllers? Maybe control_toolbox would be the right place?

mecanum_drive_controller/src/mecanum_drive_controller.yaml
```diff
@@ -113,12 +113,21 @@ mecanum_drive_controller:
       read_only: false,
     }
 
+  tf_frame_prefix_enable: {
+    type: bool,
+    default_value: true,
+    description:  "Enables or disables appending tf_prefix to tf frame id's.",
```
christophfroehlich:
```suggestion
    description:  "Enables or disables appending ``tf_frame_prefix`` to tf frame id's. See its parameter description for more information.",
```

mecanum_drive_controller/test/test_mecanum_drive_controller.cpp
christophfroehlich, line 1: we could then also move this tests to control_toolbox

christophfroehlich (commented):
The changes are fine for me, I'm just opening the discussion if we should move this to a shared place?

mecanum_drive_controller/src/mecanum_drive_controller.cpp
bmagyar: That is a good idea but I think we could do that in a follow-up, and swap the code out from all controllers at once.

KmakD: I agree with @bmagyar, this is a good idea but probably should be done in different PR, as it affects multiple controllers

bmagyar: https://github.com/ros-controls/ros2_controllers/issues/1745

bmagyar (approved):

christophfroehlich:
@bmagyar @saikishor shall we backport this to jazzy/humble? Backports #1807 and #1806 fail otherwise.

saikishor:
> @bmagyar @saikishor shall we backport this to jazzy/humble? Backports #1807 and #1806 fail otherwise.

Sure, we can do it for Jazzy. If it's straight forward for Humble, let's do that as well.

christophfroehlich:
> If it's straight forward for Humble, let's do that as well.

No it's not, I closed the backport and removed the respective tests in https://github.com/ros-controls/ros2_controllers/pull/1806

saikishor:
Thank you @christophfroehlich

PR #1684, opened 2025-05-17
Remove deprecated parameters from steering_controllers_library

christophfroehlich:
Related to #1166 which was released to rolling 4.24.0, which got not synced already -> we should not merge/release this before the sync


bmagyar (approved):

saikishor (approved):
LGTM

PR #1614, opened 2025-03-28
Deprecate tf.publish_rate parameter for pose_broadcaster

Amronos:
First step towards #473.

I have deprecated the ``tf.publish_rate`` parameter and added necessary warnings.
I have added TODOs wherever needed for whenever Jazzy is branched off.

PR #1685, opened 2025-05-17
Remove deprecated parameter from pose_broadcaster

christophfroehlich:
Related to #1614 which was released to rolling 4.24.0, which got not synced already -> we should not merge/release this before the sync


Amronos (changes requested):
You didn't remove the member variables from the header file. Is that intentional?
Also, sorry for not getting to this earlier myself.

christophfroehlich:
No, I just missed that. Thanks for pointing out :)

Amronos (approved):

PR #1690, opened 2025-05-17
Use `warning`  attribute of RcutilsLogger

saikishor:
Might lead to similar errors like https://github.com/ros-controls/ros2_control/issues/2243

It looks like the warn method is deprecated long back and they have removed it recently in https://github.com/ros2/rclpy/pull/1456


christophfroehlich (approved):
Thank you!

PR #1709, opened 2025-05-27
Fix: Reset both sec and nanosec in time_from_start

arnavk23:
Fixes issue #1708 by resetting both seconds and nanoseconds in `time_from_start`.

Previously, only seconds were reset with `set__sec(0)`, but nanoseconds remained unchanged, causing incorrect time resets.

This patch sets both `sec` and `nanosec` to zero explicitly.


christophfroehlich (approved):
Thanks. I had a quick look to find a one-line assignment, but there is also none in [rclcpp::duration](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/duration.hpp).

PR #1721, opened 2025-05-30
Update realtime containers

christophfroehlich:
Apply changes from the guidelines added with https://github.com/ros-controls/realtime_tools/pull/347

- I changed all Boxes to use the messages itself, instead of a smart pointer to it. 
- Messages get pre-allocated and saved later for the case that any try_get fails in the RT thread
- I refactored the steering_controllers_lib a bit, to have the same timeout logic that the mecanum_controller has. This needed some updates of the tests, too.
- Apart from that, the behavior of all controllers should not change at all


mecanum_drive_controller/src/mecanum_drive_controller.cpp
```diff
@@ -303,8 +304,11 @@ bool MecanumDriveController::on_set_chained_mode(bool /*chained_mode*/) { return
 controller_interface::CallbackReturn MecanumDriveController::on_activate(
   const rclcpp_lifecycle::State & /*previous_state*/)
 {
-  // Set default value in command
-  reset_controller_reference_msg(*(input_ref_.readFromRT()), get_node());
+  // Try to set default value in command.
+  // If this fails, then another command will be received soon anyways.
+  ControllerReferenceMsg emtpy_msg;
+  reset_controller_reference_msg(emtpy_msg, get_node());
+  input_ref_.try_set(emtpy_msg);
```
saikishor, line 311:
```suggestion
  input_ref_.set(emtpy_msg);
```
Shouldn't it be a set in this case? As we are in the activate callback

saikishor: Yes, you are right. In this particular case, it is wiser to use a set, as you want the controller to start with a zero state. Moreover, it is hard to have it blocking, unless you keep spamming the topic even before the controller started

christophfroehlich: but if the lock fails here, wouldn't it be very likely that the recently received command can be read with the try_get at the beginning of the update method?

christophfroehlich: then this blocking set here has no use

admittance_controller/src/admittance_controller.cpp
```diff
@@ -25,6 +25,27 @@
...
+  msg.positions.clear();
+  msg.velocities.clear();
+}
+
+// called from RT control loop
+void reset_wrench_msg(
+  geometry_msgs::msg::WrenchStamped & msg,
+  const std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node)
+{
+  msg.header.stamp = node->now();
```
saikishor, line 43: Can't we leave it with zero time by default?

forward_command_controller/src/forward_controllers_base.cpp
```diff
@@ -58,7 +72,7 @@ controller_interface::CallbackReturn ForwardControllersBase::on_configure(
 
   joints_command_subscriber_ = get_node()->create_subscription<CmdType>(
     "~/commands", rclcpp::SystemDefaultsQoS(),
-    [this](const CmdType::SharedPtr msg) { rt_command_ptr_.writeFromNonRT(msg); });
+    [this](const CmdType::SharedPtr msg) { rt_command_.set(*msg); });
```
saikishor, line 75: May be we should check here with std::all_of that all of the data elements are finite values

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -33,6 +33,15 @@ void print_interface(const rclcpp::Logger & logger, const T & command_interfaces
   }
 }
 
+// called from RT control loop
+void reset_controller_reference_msg(
+  gpio_controllers::CmdType & msg, const std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node)
+{
+  msg.header.stamp = node->now();
```
saikishor, line 40: same here,c can't we have simply with zero timestamp?

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -31,16 +31,15 @@ using ControllerTwistReferenceMsg =
...
+  ControllerTwistReferenceMsg & msg, const std::shared_ptr<rclcpp_lifecycle::LifecycleNode> & node)
 {
-  msg->header.stamp = node->now();
-  msg->twist.linear.x = std::numeric_limits<double>::quiet_NaN();
-  msg->twist.linear.y = std::numeric_limits<double>::quiet_NaN();
-  msg->twist.linear.z = std::numeric_limits<double>::quiet_NaN();
-  msg->twist.angular.x = std::numeric_limits<double>::quiet_NaN();
-  msg->twist.angular.y = std::numeric_limits<double>::quiet_NaN();
-  msg->twist.angular.z = std::numeric_limits<double>::quiet_NaN();
+  msg.header.stamp = node->now();
```
saikishor, line 36: same  here. May be we can handle all of these in a future PR

saikishor (approved):
LGTM

forward_command_controller/src/forward_controllers_base.cpp
christophfroehlich: This would be a behavior change right? maybe we can do that in a follow-up PR?

steering_controllers_library/src/steering_controllers_library.cpp
christophfroehlich: ok, let's do that in a future PR because this very method was there already before.

PR #1746, opened 2025-06-07
Added frame_id to Joint State Broadcaster

delipl:
Added `frame_id` parameter to fill the `joint_states` message in the Joint State Broadcaster. Thanks to this parameter rviz2 allows  to visualize the effort of the joints.

image

### DIFF
--- a/joint_state_broadcaster/src/joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/src/joint_state_broadcaster.cpp
@@ -184,9 +184,7 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_configure(
   frame_id_ = params_.frame_id;
   if (frame_id_.empty())
   {
-    RCLCPP_WARN(
-      get_node()->get_logger(),
-      "Frame ID is not set.");
+    RCLCPP_WARN(get_node()->get_logger(), "Frame ID is not set.");
   }
 
   return CallbackReturn::SUCCESS;


saikishor (changes requested):
@delipl you have access to the URDF model in the joint state broadcaster. Why not use it to get its root link and publish by default, instead of setting a custom frame_id from the user?

AFAIK, the visualization is same irrespective of the type of the frame you choose. Am I right?

delipl:
> @delipl you have access to the URDF model in the joint state broadcaster. Why not use it to get its root link and publish by default, instead of setting a custom frame_id from the user?
> 
> AFAIK, the visualization is same irrespective of the type of the frame you choose. Am I right?

Good point. I can change the frame id to the root of the urdf.

Edit:
I added the parameter because I'm using humble. There is no model_.

saikishor:
> Good point. I can change the frame id to the root of the urdf.
> 
> Edit:
> I added the parameter because I'm using humble. There is no model_.

Sure. But for Jazzy upon we have a URDF Model. May be you can use these changes to open another PR for Humble, but for this one, please use the URDF root link

bmagyar:
@saikishor what if we merge this as-is as a first version, backport all the way to humble and then modify things to use the URDF instead? Is there any chance you'd want to *not* use the absolute root of the URDF?

saikishor:
> @saikishor what if we merge this as-is as a first version, backport all the way to humble and then modify things to use the URDF instead?

@bmagyar I checked locally, it is not directly backportable, so I opened a new PR : https://github.com/ros-controls/ros2_controllers/pull/1751 cherry-picking @delipl's changes and refining a bit.  

>  Is there any chance you'd want to _not_ use the absolute root of the URDF?

I don't think the RViZ2 visualization changes based on the frame_id. I think it only check that it should be a valid frame

bmagyar (approved):
Thank you!

zimmerman-pliant:
Is it right that this is not backported and released via Jazzy binaries yet?

Issue #1617, opened 2025-03-31
[JTC] Documentation about PID is wrong


christophfroehlich:
The [documentation](https://control.ros.org/rolling/doc/ros2_controllers/joint_trajectory_controller/doc/parameters.html) was not properly updated with https://github.com/ros-controls/ros2_controllers/pull/1200

Check
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/joint_trajectory_controller/doc/userdoc.rst?plain=1#L29-L35
and update 
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/joint_trajectory_controller/doc/parameters_context.yaml#L4-L11
accordingly.

PR #1748, opened 2025-06-07
[JTC Doc] Update PID documentation for effort related command interface support

Juliaj:
Address https://github.com/ros-controls/ros2_controllers/issues/1617

### DIFF
--- a/joint_trajectory_controller/doc/parameters_context.yaml
+++ b/joint_trajectory_controller/doc/parameters_context.yaml
@@ -2,22 +2,33 @@ constraints:
     Default values for tolerances if no explicit values are set in the ``JointTrajectory`` message.
 
 gains: |
-  The parameters are used to configure the PID loop for the ``velocity`` or ``effort`` only command interface.
+  The parameters are used to configure PID loops for the ``velocity`` or ``effort``-only command interfaces. This structure contains the controller gains for every joint with following the control laws
 
-  If ``velocity`` is the only command interface, PID controllers are used for every joint.
-  This structure contains the controller gains for every joint with the control law
+  * If ``velocity`` is the only command interface:    
 
-  .. math:: u = k_{ff} v_d + k_p e + k_i \sum e dt + k_d (v_d - v)
+    .. math::
+    
+      u = k_{ff} v_d + k_p e + k_i \sum e dt + k_d (v_d - v)
 
-  with the desired velocity :math:`v_d`, the measured velocity :math:`v`, the position error :math:`e` (definition see below),
+    with the desired velocity :math:`v_d`, the measured velocity :math:`v`, the position error :math:`e` (definition see below),
+    the controller period :math:`dt`, and the ``velocity`` manipulated variable (control variable) :math:`u`, respectively.
   the controller period :math:`dt`, and the ``velocity`` manipulated variable (control variable) :math:`u`, respectively.
 
-  If ``effort`` is the only command interface, PID controllers are used for every joint. Following control law is applied:
+  * If ``effort`` is the only command interface:
 
-  .. math:: u = (k_{ff} v_d + effort_d) + k_p e + k_i \sum e dt + k_d (v_d - v)
+    .. math::
+    
+      u = k_{ff} v_d + \delta_d + k_p e + k_i \sum e dt + k_d (v_d - v)
 
-  with the desired velocity :math:`v_d`, the desired effort :math:`effort_d` if it is provided in the trajectory or 0 if not, the measured velocity :math:`v`, the position error :math:`e` (definition see below),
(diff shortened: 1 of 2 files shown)


joint_trajectory_controller/doc/parameters_context.yaml
```diff
@@ -2,13 +2,22 @@ constraints:
...
 gains: |
-  If ``velocity`` is the only command interface for all joints or an ``effort`` command interface is configured, PID controllers are used for every joint.
+  The parameters are used to configure the PID loop for the ``velocity`` or ``effort`` only command interface.
+
+  If ``velocity`` is the only command interface, PID controllers are used for every joint.
   This structure contains the controller gains for every joint with the control law
 
   .. math:: u = k_{ff} v_d + k_p e + k_i \sum e dt + k_d (v_d - v)
 
   with the desired velocity :math:`v_d`, the measured velocity :math:`v`, the position error :math:`e` (definition see below),
```
christophfroehlich:
```suggestion
  The parameters are used to configure PID loops for the ``velocity`` or ``effort``-only command interfaces. This structure contains the controller gains for every joint with following the control laws

  * If ``velocity`` is the only command interface:    

    .. math::
    
      u = k_{ff} v_d + k_p e + k_i \sum e dt + k_d (v_d - v)

    with the desired velocity :math:`v_d`, the measured velocity :math:`v`, the position error :math:`e` (definition see below),
    the controller period :math:`dt`, and the ``velocity`` manipulated variable (control variable) :math:`u`, respectively.
```

joint_trajectory_controller/doc/parameters_context.yaml
```diff
@@ -2,13 +2,22 @@ constraints:
...
-  the controller period :math:`dt`, and the ``velocity`` or ``effort`` manipulated variable (control variable) :math:`u`, respectively.
+  the controller period :math:`dt`, and the ``velocity`` manipulated variable (control variable) :math:`u`, respectively.
+
+  If ``effort`` is the only command interface, PID controllers are used for every joint. Following control law is applied:
+
+  .. math:: u = (k_{ff} v_d + effort_d) + k_p e + k_i \sum e dt + k_d (v_d - v)
+
+  with the desired velocity :math:`v_d`, the desired effort :math:`effort_d` if it is provided in the trajectory or 0 if not, the measured velocity :math:`v`, the position error :math:`e` (definition see below),
+  the controller period :math:`dt`, and the ``effort`` manipulated variable (control variable) :math:`u`, respectively.
```
christophfroehlich:
```suggestion
  * If ``effort`` is the only command interface:

    .. math::
    
      u = k_{ff} v_d + \delta_d + k_p e + k_i \sum e dt + k_d (v_d - v)

    with the desired velocity :math:`v_d`, the desired effort :math:`\delta_d` if provided in the trajectory (or 0 otherwise), the measured velocity :math:`v`, the position error :math:`e` (definition see below),
    the controller period :math:`dt`, and the ``effort`` manipulated variable (control variable) :math:`u`, respectively.

  If the joint is of continuous type, the position error :math:`e = normalize(s_d - s)` is normalized between :math:`-\pi, \pi`,
  i.e., the shortest rotation to the target position is the desired motion.
  Otherwise  :math:`e = s_d - s` is used, with the desired position :math:`s_d` and the measured
  position :math:`s` from the state interface.

  If you want to turn off the PIDs (open loop control), set the feedback gains to zero and an appropriate feed-forward gain :math:`k_{ff}`
```

christophfroehlich (changes requested):
Thanks for picking this! I rendered this with Sphinx, but there were some display problems. I tried to fix them (naturally, I`d use $$\tau$$ for the effort variable, but this does not work with sphinx and I couldn't figure out why) and suggest to rephrase it a bit. What do you think?
image

christophfroehlich (dismissed):
Thanks!

joint_trajectory_controller/doc/parameters_context.yaml
```diff
@@ -2,17 +2,30 @@ constraints:
...
 
-  with the desired velocity :math:`v_d`, the measured velocity :math:`v`, the position error :math:`e` (definition see below),
-  the controller period :math:`dt`, and the ``velocity`` or ``effort`` manipulated variable (control variable) :math:`u`, respectively.
+    .. math::
+
+      u = k_{ff} v_d + k_p e + k_i \sum e dt + k_d (v_d - v)
+
+    with the desired velocity :math:`v_d`, the measured velocity :math:`v`, the position error :math:`e` (definition see below),
+    the controller period :math:`dt`, and the ``velocity`` manipulated variable (control variable) :math:`u`, respectively.
+  the controller period :math:`dt`, and the ``velocity`` manipulated variable (control variable) :math:`u`, respectively.
```
christophfroehlich:
```suggestion
```

PR #1720, opened 2025-05-30
JTC: Use std::atomic<bool>

christophfroehlich:
Apply changes from the guidelines added with https://github.com/ros-controls/realtime_tools/pull/347

PR #1749, opened 2025-06-09
Fix atomic variables in JTC

christophfroehlich:
Explicitly initialize them, should fix the error reported with https://github.com/ros-controls/ros2_controllers/pull/1720#issuecomment-2954973398

I don't know why the tests did not cover this?

Furthermore, I saw that `subscriber_is_active_` should be atomic too, as it is used in both threads.


bmagyar (approved):

bmagyar:
@lqiang2003cn  could you please confirm is this fixes the issue you are currently seeing?

PR #1772, opened 2025-06-25
Mecanum Drive: Populate the pose covariance matrix

hilary-luo:
During recent testing I noticed that the pose covariance matrix was not being populated in the mecanum drive controller. This PR is to populate the covariance matrix in the message using the coordinating parameters in the same way that the twist covariance matrix is populated.

Once this is merged I do want to get it backported, I'm not certain what the process is for that.


Juliaj:
The change looks good, please help add a test.

mecanum_drive_controller/src/mecanum_drive_controller.cpp
```diff
@@ -173,13 +173,14 @@ controller_interface::CallbackReturn MecanumDriveController::on_configure(
...
+  auto & pose_covariance = rt_odom_state_publisher_->msg_.pose.covariance;
+  auto & twist_covariance = rt_odom_state_publisher_->msg_.twist.covariance;
   constexpr size_t NUM_DIMENSIONS = 6;
   for (size_t index = 0; index < 6; ++index)
   {
     const size_t diagonal_index = NUM_DIMENSIONS * index + index;
-    covariance[diagonal_index] = params_.pose_covariance_diagonal[index];
-    covariance[diagonal_index] = params_.twist_covariance_diagonal[index];
+    pose_covariance[diagonal_index] = params_.pose_covariance_diagonal[index];
+    twist_covariance[diagonal_index] = params_.twist_covariance_diagonal[index];
```
moriarty, line 183: Line 182 immediately overwrote thing set on line 181, good catch.

destogl (dismissed):
This is good,but let's add test as @christophfroehlich mentioned.

moriarty:
@destogl is it enough to just modify one of the existing tests to catch this?

Eg: just update one of these two lines

https://github.com/ros-controls/ros2_controllers/blob/e1f5116ea8467c447c431cc21469d358b8aff8e2/mecanum_drive_controller/test/mecanum_drive_controller_params.yaml#L19-L20

https://github.com/ros-controls/ros2_controllers/blob/e1f5116ea8467c447c431cc21469d358b8aff8e2/mecanum_drive_controller/test/mecanum_drive_controller_params.yaml#L40-L41

christophfroehlich:
> @destogl is it enough to just modify one of the existing tests to catch this?

this is not enough because there is no test for checking the values ;)

This can be added to `when_update_is_called_expect_status_message` for example.

hilary-luo:
Please let me know if this requires any additional changes from me to get this merged.

PR #1781, opened 2025-06-30
add read_only parameters to diff_drive_controller

adhithiyan110:
This PR adds the read_only attribute to key parameters of diff_drive_controller.Related to #1696


diff_drive_controller/src/diff_drive_controller_parameter.yaml
```diff
@@ -35,94 +39,112 @@ diff_drive_controller:
...
     type: double_array,
     default_value: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
     description: "Odometry covariance for the encoder output of the robot for the speed. These values should be tuned to your robot's sample odometry data, but these values are a good place to start: ``[0.001, 0.001, 0.001, 0.001, 0.001, 0.01]``.",
+    read_only: true,
   }
   open_loop: {
     type: bool,
     default_value: false,
     description: "If set to true the odometry of the robot will be calculated from the commanded values and not from feedback.",
+    read_only: false,
```
saikishor: Do we want this to be false?. Does it make sense to change this online?

diff_drive_controller/src/diff_drive_controller_parameter.yaml
```diff
@@ -35,94 +39,112 @@ diff_drive_controller:
...
   }
   position_feedback: {
     type: bool,
     default_value: true,
     description: "Is there position feedback from hardware.",
+    read_only: false,
```
saikishor:
```suggestion
    read_only: true,
```

diff_drive_controller/src/diff_drive_controller_parameter.yaml
```diff
@@ -35,94 +39,112 @@ diff_drive_controller:
...
     type: int,
     default_value: 10,
     description: "Size of the rolling window for calculation of mean velocity use in odometry.",
+    read_only: true,
   }
   publish_rate: {
     type: double,
     default_value: 50.0, # Hz
     description: "Publishing rate (Hz) of the odometry and TF messages.",
+    read_only: false,
```
saikishor:
```suggestion
    read_only: true,
```
This should be constant too

saikishor:
```suggestion
    read_only: true,
```
This should be constant too

diff_drive_controller/src/diff_drive_controller_parameter.yaml
adhithiyan110: can i make this true?

saikishor: Sure

christophfroehlich (changes requested):
The parameters itself look good. But we don't update the parameters post on_configure. I propose changing this to a simple get_params() as on_configure is not in the RT thread 
https://github.com/ros-controls/ros2_controllers/blob/922e49b89f0903614f72e1093ac1060ab30012d7/diff_drive_controller/src/diff_drive_controller.cpp#L313-L317
but add this `try_update_params` to the update_reference_from_subscribers or update_and_write_commands method.

Please review then also if we need local copies of the params_ struct and either remove it or update them after try_update_params. (cmd_vel_timeout_, publish_limited_velocity_,..)

christophfroehlich:
@adhithiyan110 do you need any input for going on with this PR?

adhithiyan110:
@christophfroehlich can you please review changes

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -107,6 +107,37 @@ controller_interface::return_type DiffDriveController::update_reference_from_sub
...
+  publish_limited_velocity_ = params_.publish_limited_velocity;
+    // limit the publication on the topics /odom and /tf
+    publish_rate_ = params_.publish_rate;
+    publish_period_ = rclcpp::Duration::from_seconds(1.0 / publish_rate_);
+  const double wheel_separation = params_.wheel_separation_multiplier * params_.wheel_separation;
+  const double left_wheel_radius = params_.left_wheel_radius_multiplier * params_.wheel_radius;
+  const double right_wheel_radius = params_.right_wheel_radius_multiplier * params_.wheel_radius;
+
+  odometry_.setWheelParams(wheel_separation, left_wheel_radius, right_wheel_radius);
+  odometry_.setVelocityRollingWindowSize(static_cast<size_t>(params_.velocity_rolling_window_size));
```
christophfroehlich, line 126: these parameters are read-only, better leave them in on_configure

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -107,6 +107,37 @@ controller_interface::return_type DiffDriveController::update_reference_from_sub
...
+    params_.linear.x.min_velocity, params_.linear.x.max_velocity,
+    params_.linear.x.max_acceleration_reverse, params_.linear.x.max_acceleration,
+    params_.linear.x.max_deceleration, params_.linear.x.max_deceleration_reverse,
+    params_.linear.x.min_jerk, params_.linear.x.max_jerk);
+
+  limiter_angular_ = std::make_unique<SpeedLimiter>(
+    params_.angular.z.min_velocity, params_.angular.z.max_velocity,
+    params_.angular.z.max_acceleration_reverse, params_.angular.z.max_acceleration,
+    params_.angular.z.max_deceleration, params_.angular.z.max_deceleration_reverse,
+    params_.angular.z.min_jerk, params_.angular.z.max_jerk);
```
christophfroehlich, line 138: This method is part of the RT loop, we must not allocate memory here. Better leave the initial construction of the objects in the on_configure, but update the parameters (only if they changed, i.e., inside the if clause above).
To achieve this, we have to expose the `control_toolbox::RateLimiter::set_params` method in the `diff_drive_controller::SpeedLimiter` class.
https://github.com/ros-controls/control_toolbox/blob/c46e3c5b56d62c30e30e42d3658821fbf909dd3e/control_toolbox/include/control_toolbox/rate_limiter.hpp#L107-L132

Furthermore, we should change the rate_limiter class to throw the errors of the input checks before the member variables are set. So we can catch the errors in the `diff_drive_controller::SpeedLimiter::set_params` method, otherwise we have to reimplement the checks to avoid crashing the controller: https://github.com/ros-controls/control_toolbox/issues/437

christophfroehlich (changes requested):

christophfroehlich (changes requested):
And please install pre-commit, so that it automatically checks for the rules (see the failing pre-commit jobs).

christophfroehlich:
@adhithiyan110 do you need any input for going on with this PR?

adhithiyan110:
> @adhithiyan110 do you need any input for going on with this PR?

yes please, I'm a bit  stuck

christophfroehlich:
> > @adhithiyan110 do you need any input for going on with this PR?
> 
> yes please, I'm a bit stuck

Start by following my comments above, and reply there if something is not clear. (for pre-commit please see https://github.com/ros-controls/ros2_controllers?tab=contributing-ov-file#contributing-via-pull-requests)

christophfroehlich:
Got obsolete by #2074

PR #1783, opened 2025-06-30
Handle cases of state interfaces with different data types

saikishor:
Needs https://github.com/ros-controls/ros2_control/pull/2351

### DIFF
--- a/joint_state_broadcaster/src/joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/src/joint_state_broadcaster.cpp
@@ -201,6 +201,29 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_activate(
     return CallbackReturn::ERROR;
   }
 
+  // Erase the interfaces that are not castable to double
+  std::vector<hardware_interface::LoanedStateInterface> state_interfaces_to_keep;
+  std::for_each(
+    state_interfaces_.begin(), state_interfaces_.end(),
+    [&state_interfaces_to_keep](hardware_interface::LoanedStateInterface & si)
+    {
+      if (si.is_castable_to_double())
+      {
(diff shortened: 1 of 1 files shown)


christophfroehlich (dismissed):
LGTM, should we also add a test for that?

destogl (commented):
I was thinking to approach this in another way. I think we should have per default conversion between double and other types (if possible). This will help us with the reusability of HW interface and controllers. We had similar issues in CAN implementation and solved it with `memcpy`. I know this is not beautiful, but it is extremely useful. We should give warningonce per interface that we are handing the type using `double` but it is a compatible type and then just do it. Otherwise, I am afraid that we will end up in the endless `if - else if` loops and having many issues with the controllers — HW incompatibilities, even though they could work with the proposed type support.

joint_state_broadcaster/src/joint_state_broadcaster.cpp
```diff
@@ -201,6 +201,29 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_activate(
...
+      {
+        RCLCPP_ERROR(
+          rclcpp::get_logger("JointStateBroadcaster"),
+          "State interface '%s' is not castable to double. "
+          "It will not be used in JointStateBroadcaster.",
+          si.get_name().c_str());
+      }
+    });
+
+  state_interfaces_.clear();
```
Maverobot, line 224: This `clear()` function call seems unnecessary.

christophfroehlich:
We decided to not do this, but
- add a parameter back to humble to disable dynamic_joint_states
- deprecate dynamic_joint_states on rolling, with migration notes to gpio_command_controller, pal_statistics, and [interface_state_broadcaster](https://github.com/ros-controls/ros2_controllers/pull/2006)
- remove it in an upcoming release

PR #1791, opened 2025-07-03
fix rqt_joint_trajectory_controller namespace for robots with namespace

oscar-lima:
rqt_joint_trajectory_controller does not work if your robot has a namespace. This PR fixes that, to test do:

`ros2 run rqt_joint_trajectory_controller rqt_joint_trajectory_controller --ros-args -r __ns:=/my_robot_ns`

behavior: arm cannot be moved, because in fact not even the drop-down menu from the right combo box (jtc_combo) is filled with data.

I have tested these changes with robots with namespace and without namespace and works fine, so it does not break things for others.


oscar-lima:
Closing because PR has to be performed against master branch as mergify bot pointed out.

oscar-lima:
Opened a new PR that targets master branch instead of jazzy, see: https://github.com/ros-controls/ros2_controllers/pull/1792

PR #1792, opened 2025-07-03
fix rqt_joint_trajectory_controller for robots with namespace

oscar-lima:
rqt_joint_trajectory_controller does not work if your robot has a namespace. This PR fixes that, to test do:

`ros2 run rqt_joint_trajectory_controller rqt_joint_trajectory_controller --ros-args -r __ns:=/my_robot_ns`

behavior: arm cannot be moved, because in fact not even the drop-down menu from the right combo box (jtc_combo) is filled with data.

I have tested these changes with robots with namespace and without namespace and works fine, so it does not break things for others.

### DIFF
--- a/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
+++ b/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
@@ -464,7 +464,7 @@ def _jtc_joint_names(jtc_info):
 
     joint_names = []
     for interface in jtc_info.required_state_interfaces:
-        name = interface.split("/")[-2]
+        name = "/".join(interface.split("/")[:-1])
         if name not in joint_names:
             joint_names.append(name)
 


prototriangle:
Crazy coincidence, I was just coming here to submit a first time contribution to change exactly this!

This is what I was going to suggest:

```
diff --git a/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py b/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
index 12ac10c6..7049c2b8 100644
--- a/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
+++ b/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
@@ -464,7 +464,7 @@ def _jtc_joint_names(jtc_info):
 
     joint_names = []
     for interface in jtc_info.required_state_interfaces:
-        name = interface.split("/")[-2]
+        name = interface.rpartition("/")[0]
         if name not in joint_names:
             joint_names.append(name)
 ```

I'm not sure why the original code was selecting the penultimate "split" rather than selecting everything before the final /

christophfroehlich:
@prototriangle could you please approve this PR if it works for you?

prototriangle (approved):
Tested on humble with a robot with 4 namespaced controller_managers/controllers

Works for me.

bmagyar (approved):
Thank you guys!

Issue #1757, opened 2025-06-13
Proposal: Modular Reference Handling in ROS 2 Controllers


my-rice:
Hi, we are writing this issue as part of the [Automatic Control Group from the University of Salerno](https://github.com/unisa-acg). We have previously contributed to the ROS ecosystem, for example, through pull requests such as [here](https://github.com/ros-controls/kinematics_interface/pull/92) and [here](https://github.com/ros-controls/kinematics_interface/pull/93), as well as by deploying ROS 2 packages to the community like [here](https://index.ros.org/p/inverse_dynamics_solver/).

---
We are opening this issue to propose a new design approach for the control framework, specifically aimed at improving how controllers handle reference inputs.

**Is your feature request related to a problem? Please describe.**
After reviewing the current implementations, we observed that a significant portion of controller code is dedicated to managing reference acquisition, typically via subscribers or action servers. For example, [here](https://github.com/ros-controls/ros2_controllers/tree/master/admittance_controller) and [here](https://github.com/ros-controls/ros2_controllers/tree/master/joint_trajectory_controller).

**Describe the solution you'd like**
To improve modularity and separation of concerns, we propose a cleaner structure that decouples reference acquisition from control logic. The idea is to divide responsibilities into two distinct modules:

1. **Reference Generator** – This component is practically a controller, but its purpose is to receive and validate reference trajectories or setpoints from the ROS 2 network. It abstracts the communication interface and provides a clean output to downstream controllers, using [controller chaining](https://control.ros.org/rolling/doc/ros2_control/controller_manager/doc/controller_chaining.html).

2. **Controller** – This component is exclusively responsible for computing the control law. It receives input from the Reference Generator and remains agnostic to how references are produced or delivered.

This modular structure promotes **better organization and a clearer separation of responsibilities**. Moreover, Reference Generators **increase code reuse** across multiple controllers. The diagram below illustrates the proposed architecture:
![Image](https://github.com/user-attachments/assets/6f1a32b6-c277-475b-93ac-917fb669a292)
Importantly, **this proposal does not require any modifications to existing controllers**. The framework is fully backward-compatible, allowing existing controllers to continue functioning without changes. Developers creating new controllers may adopt this structure if it aligns with their needs.

**Describe alternatives you've considered**
The most straightforward alternative is to maintain the current design, leaving users responsible for implementing their own reference-handling logic, even when this functionality could be modularized and reused across controllers.

**Additional context**
[...]

Issue #1638, opened 2025-04-15
Ask about JTC and chained DDC


moveito:
Hello community!

I have created a controller that would concatenate the output of the JTC to the DDC.
Take a look at the link.
https://github.com/moveito/mom_diff_drive_controller/tree/main
MOM is an acronym for MObile_Manipulator.
The purpose of creating this is to control a robot that is driven differentially by Moveit2.

What I did was to DDC.
1. added on_export_state_interfaces()
2. added odom values to state_interfaces_values_.
3. changed the calculation method of linear_command and angular_command values

I checked the changes with my own robot.
Maybe it is working.
https://x.com/move_ito/status/1905974761924305273

If possible, it would be great if this change could be reflected in ros2_controllers.
I'm new to this kind of OSS activity, so I'd like to know.

1. Is it already implemented in other ways to begin with?
2. which is the best direction to reflect this change?
　2-1. reflect this change in DDC
　2-2. create another controller (e.g. Planar Base Controller (PBC)) and make it chain Joint Trajectory Controller (JTC) → PBC → Diff Drive Controller (DDC)

Thank you in advance.

christophfroehlich:
We may have overlooked your issue. DDC is now a chainable controller, but I think there is still something missing in between: If you plan a path in moveit, where is the position control law creating the twist command for DDC? This is not trivial with a nonholonomic mobile base.

moveito:
Thank you for your reply.

> If you plan a path in moveit, where is the position control law creating the twist command for DDC?  

In moveit2, we can define a planar joint and output x, y, and theta. In the section below, these x, y, and theta values are converted into translational velocity and rotational velocity. After that, just like in DDC, I think you should convert them into the rotational angular velocities of the left and right wheels.
https://github.com/moveito/mom_diff_drive_controller/blob/main/src/diff_drive_controller.cpp#L144-L152  
  
However, since this implementation cannot handle standard `cmd_vel` values, I’m considering creating a new controller (one that converts `x`, `y`, and `theta` to `cmd_vel`) or something similar.

christophfroehlich:
Do you get position of x,y,theta, or also it's derivative? Can you add the nonholonomic constraints (no sideward movement)?

moveito:
> Do you get position of x,y,theta, or also it's derivative?  

What is the point of this question?

> Can you add the nonholonomic constraints (no sideward movement)?  

[...]

PR #300, opened 2022-02-23
[WIP] Cartesian twist controller

livanov93:
PR info:
 - add twist controller (cartesian controller) using input type message [`geometry_msgs/TwistStamped`](https://docs.ros2.org/foxy/api/geometry_msgs/msg/TwistStamped.html)
    
TODOs:
-  add tests

PR #1801, opened 2025-07-05
Add cmd_vel_controller

Amronos:
This is a controller designed to filter command velocity. It subscribes to the ``/cmd_vel`` topic, filters the velocity, and then commands reference interfaces of Mobile Robot Controllers alongside publishing the filtered velocity on the ``/cmd_vel_out`` topic.

It includes/will include the following features:
- [x] Limiting of command velocity
- [ ] PID filtering of command velocity, using ``/odom`` for feedback.
- [ ] Some features of the [nav2_velocity_smoother](https://github.com/ros-navigation/navigation2/tree/main/nav2_velocity_smoother) (Future Work)

I am creating a draft PR for now, as I would like to get feedback on the controller even though it is not complete yet. There are no docs or tests yet. I will add them soon.

Related to #1676 and #1757.


christophfroehlich:
Initially I was thinking of making such a controller the default way in handling input processing in mobile platforms. But there are voice in not doing so for sake of simplicity for simple (beginner) setups. I'll discuss this in next week's PMC meeting, how to proceed with this PR and https://github.com/ros-controls/ros2_controllers/pull/300

christophfroehlich:
We discussed this in today's meeting:

- We should revive #300 and add it to the forward_command_controller package. Note that the twist_controller is for full 6DOF twist, while for mobile bases we normally have 2 or 3 DOFs only.
- Instead of adding complexity for users to use a controller chain to add twist input processing, better create a library (header only?) and reuse that in all mobile base controllers. As we now have the rate limiters in control_toolbox repo, maybe this library would fit there as well? Do you want to take this and make a proposal for the API of this lib?

Amronos:
Regarding creating a library instead of a new controller:
- I feel that if someone needs to use this controller, they are probably already using the ``pid_controller``, to use which you do need controller chaining.
- I also feel that adding this controller to all the mobile robot controllers will over-complicate them and make it harder for beginners to understand and implement a simple ros2_control setup.
- The maintenance cost will also probably increase as one will need to maintain all the mobile robot controllers and the library.

Amronos:
I also made this diagram some time back, to assist with how I was envisioning control for my robot and expecting the ``cmd_vel_controller`` to be used:
Control Diagram

I hope this is helpful in understanding why I want to keep the ``cmd_vel_controller`` as a separate controller.

moveito:
## Are this PR and issue #1638 actually solving different problems?

I'd like to raise a question about whether they should really be treated as the same problem.

As I understand it:

- **This PR** addresses filtering and limiting of `geometry_msgs/Twist` commands coming from navigation stacks like Nav2. The input is already a velocity in the right form — the goal is to smooth or constrain it before passing it to a mobile base controller.
- **Issue #1638** addresses translating JTC joint trajectory commands into linear/angular velocity references for DDC. The input is in a fundamentally different form (joint positions/velocities from a trajectory), and the core challenge is interface translation, not filtering.

Both sit "in front of DDC" in the controller chain, but their responsibilities feel distinct to me. One is about *signal conditioning*, the other is about *interface conversion*.

I may be missing context here — is the intention to solve both with a single unified mechanism? If so, I'd appreciate an explanation of how they fit together.

If they are indeed separate problems, would it make sense to track and resolve them independently, so that this PR can be evaluated on its own merits without waiting for the broader design question to be settled?

Happy to hear any thoughts on this.

Amronos:
The features described in #1638 aren't something I have ever thought about or planned to add to this controller.
If needed, this controller can also solve #1638.
IMO, though, it would be best to solve this directly in MoveIt.

moveito:
Thank you for your comment.  
Personally, I think it would be better to create a separate controller from cmd_vel_controller.

> it would be best to solve this directly in MoveIt.  

Since Moveit generates trajectories for the planner, I believe it would be better to implement this on the ROS 2 controller side.

@christophfroehlich 

1. Do you think we should create a separate controller?

2. Would it be better to convert the trajectory to `cmd_vel` on the Moveit side?

christophfroehlich:
My idea wasn't to include this in this controller, but add another controller in the same package.
But as you pointed out, this maybe should be done as post-trajectory-generation step as it can be calculated offline, there is no need to do this online in the controller chain?

PR #1814, opened 2025-07-11
Add filtering capability to ft_broadcaster

OscarMrZ:
Good morning!

This is a continuation of https://github.com/ros-controls/ros2_controllers/pull/559 as it was inactive. 

The changes are minimal:

1. Applied @saikishor suggestions
  - Added `sensor_filter_chain` to the parameters, which serves as documentation.
  -  I removed all the logic about the additional frames. I agree with @saikishor that transforming to different frames should be handled by the user in their application. With this, no tf calls inside the RT are needed.
  - No more publishing NaN if the update of the filter failed.
 
 2. Added some relevant tests
  IMHO, the previous tests were testing the business logic of the filters themselves, that is, checking that the actual computations were correct, rather than the actual new logic for adding filters. I added three tests for configuring, activating/deactivating, and updating with a configured filter chain. Just one consideration: I changed a little bit the logic of SetUpFTSBroadcaster a little bit to accept a node name. This is because I added some parameters in the YAML that are needed for the broadcaster with filters. I cannot do `set_parameter` because the filter chain expects the parameters to be overrides.
  
 Thanks a lot for your time!
 
 Óscar


OscarMrZ:
Also, an additional consideration about the logic:

Because of https://github.com/ros/filters/issues/89, there's not really a way to tell if a chain is actually configured. So now, even if the user doesn't provide a chain config, it will be configured with an empty filter list. At the end, this just means that the filtered message will always be published, with output = input. If https://github.com/ros/filters/pull/90 gets merged, the broadcaster will know no filter chain was configured and omit publishing the unnecessary message.

christophfroehlich (commented):
Thanks for picking this up.

> This is a continuation of #559 as it was inactive.

It was not inactive, but it was missing more reviews ;)

 >     2. Added some relevant tests
>        IMHO, the previous tests were testing the business logic of the filters themselves, that is, checking that the actual computations were correct, rather than the actual new logic for adding filters. I added three tests for configuring, activating/deactivating, and updating with a configured filter chain. Just one consideration: I changed a little bit the logic of SetUpFTSBroadcaster a little bit to accept a node name. This is because I added some parameters in the YAML that are needed for the broadcaster with filters. I cannot do `set_parameter` because the filter chain expects the parameters to be overrides.

But now there is no test if any filtering is performed, or not? E.g., not calling the filter-update will not be covered by the tests?

OscarMrZ:
Hello @christophfroehlich 

Thanks for your comment!

True, my bad! I hope it's ok opening this new MR.

I had actually run it with pre-commit, but somehow pushed with some rules failing. It should be fixed now. Sorry for that!

> But now there is no test if any filtering is performed, or not? E.g., not calling the filter-update will not be covered by the tests?

Yeah, you are right. I didn't know how to properly add it without including unnecessary logic checks for the underlying filters. Would you prefer re-adding a test with one of the simple filters or creating a dummy filter for testing? 

P.D.: Sorry for the force push!

christophfroehlich:
creating a dummy filter would be ok, but using an existing one is also fine as we have control_toolbox as dependency in this repo (but not in this package, right? this would speak for the dummy or one from ros/filters directly)

saikishor:
> creating a dummy filter would be ok, but using an existing one is also fine as we have control_toolbox as dependency in this repo (but not in this package, right? this would speak for the dummy or one from ros/filters directly)

Sounds good to me.

OscarMrZ:
Good morning!

Yeah, there's no dependency for `control_toolbox` in the ft_broadcaster, but I think it's perfectly fine to use a filter from filters as we are introducing the dependency for the chain.

I added `dummy_filter.cpp`, which performs the template specialization for the WrenchStamped type of the [increment filter](https://github.com/ros/filters/blob/ros2/include/filters/increment.hpp) update method. It will just add 1.0 to every field of the message.

I updated the tests accordingly and included an additional one to ensure the correct message is received. Everything seems to be working fine!

Thanks a lot!

christophfroehlich (changes requested):
This looks quite good already. Can you please update

- https://github.com/ros-controls/ros2_controllers/blob/master/force_torque_sensor_broadcaster/doc/userdoc.rst
- https://github.com/ros-controls/ros2_controllers/blob/master/doc/release_notes.rst

then this is good to be merged.

force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
```diff
@@ -87,12 +90,39 @@ controller_interface::CallbackReturn ForceTorqueSensorBroadcaster::on_configure(
...
+  }
+  catch (const std::exception & e)
+  {
+    fprintf(stderr, "Exception thrown during filter chain creation with message : %s \n", e.what());
+    return CallbackReturn::ERROR;
+  }
+
+  // As the sensor_filter_chain parameter is of type 'none', we cannot directly check if it is
+  // present. Even if the sensor_filter_chain parameter is not specified, the filter chain will be
+  // correctly configured with an empty list of filters (https://github.com/ros/filters/issues/89).
```
christophfroehlich: how to proceed here?

force_torque_sensor_broadcaster/CMakeLists.txt
```diff
@@ -69,7 +71,33 @@ if(BUILD_TESTING)
   target_include_directories(test_force_torque_sensor_broadcaster PRIVATE include)
   target_link_libraries(test_force_torque_sensor_broadcaster
     force_torque_sensor_broadcaster
+    force_torque_sensor_broadcaster_parameters
```
christophfroehlich:
```suggestion
```
do we have to link against the parameters here?

force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
OscarMrZ: I would say until the new version of filters with my MR gets synced into Humble, it has to stay like it is. After that, we can check the length of the chain and act accordingly.

OscarMrZ:
Good morning @christophfroehlich!

Sorry for the delay, things get slower in summer! :smile: 

Everything should be ready for merging. Please let me know if you think further changes may be needed.

Thanks a lot for your assistance!

force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
christophfroehlich: filters 2.2.2 is released, and only the sync to jazzy is missing. So we can update this PR already, and just hold an eventual backport to jazzy until jazzy got synced.

saikishor: Good job guys

christophfroehlich: Shouldn't we use `get_length()` now?

OscarMrZ: I just committed a quick change, so we can both check if the chain is properly configured and with some filters in it.

force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
```diff
@@ -104,12 +104,17 @@ controller_interface::CallbackReturn ForceTorqueSensorBroadcaster::on_configure(
...
   // present. Even if the sensor_filter_chain parameter is not specified, the filter chain will be
   // correctly configured with an empty list of filters (https://github.com/ros/filters/issues/89).
-  has_filter_chain_ = filter_chain_->configure(
+  bool filter_chain_configured = filter_chain_->configure(
     "sensor_filter_chain", get_node()->get_node_logging_interface(),
     get_node()->get_node_parameters_interface());
 
+  // Even on successful configure, if empty, the chain won't be used
+  size_t filter_chain_length = filter_chain_->get_length();
+  has_filter_chain_ = filter_chain_configured && filter_chain_length > 0;
```
christophfroehlich:
```suggestion
  has_filter_chain_ = filter_chain_configured && filter_chain_->get_length() > 0;
```
Is the get_length() well defined if the configure step failed?

OscarMrZ: Afaik yes! If configuration fails, the function simply returns false and `reference_pointers_` remains uninitialized, that is, with size = 0.

force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
```diff
@@ -87,12 +90,43 @@ controller_interface::CallbackReturn ForceTorqueSensorBroadcaster::on_configure(
...
+  }
+  catch (const std::exception & e)
+  {
+    fprintf(stderr, "Exception thrown during filter chain creation with message : %s \n", e.what());
+    return CallbackReturn::ERROR;
+  }
+
+  // As the sensor_filter_chain parameter is of type 'none', we cannot directly check if it is
+  // present. Even if the sensor_filter_chain parameter is not specified, the filter chain will be
+  // correctly configured with an empty list of filters (https://github.com/ros/filters/issues/89).
```
christophfroehlich:
```suggestion
  // correctly configured with an empty list of filters.
```
The issue is kind of resolved https://github.com/ros/filters/issues/89?

force_torque_sensor_broadcaster/src/force_torque_sensor_broadcaster.cpp
```diff
@@ -87,12 +90,43 @@ controller_interface::CallbackReturn ForceTorqueSensorBroadcaster::on_configure(
...
+  bool filter_chain_configured = filter_chain_->configure(
+    "sensor_filter_chain", get_node()->get_node_logging_interface(),
+    get_node()->get_node_parameters_interface());
+
+  // Even on successful configure, if empty, the chain won't be used
+  has_filter_chain_ = filter_chain_configured && filter_chain_->get_length() > 0;
+
+  RCLCPP_INFO_EXPRESSION(
+    get_node()->get_logger(), has_filter_chain_, "Filter active with %zu filters!",
+    filter_chain_length);
```
saikishor:
```suggestion
    filter_chain_->get_length());
```

christophfroehlich (dismissed):
Thanks for the follow-ups, LGTM!

christophfroehlich (approved):

Issue #1775, opened 2025-06-25
forward_command_controller: Check message for finite values


christophfroehlich:
> May be we should check here with std::all_of that all of the data elements are finite values

_Originally posted by @saikishor in https://github.com/ros-controls/ros2_controllers/pull/1721#discussion_r2164917620_

lfurno:
Hello @christophfroehlich. I would like to claim this issue as my first open source contribution :)

christophfroehlich:
Fixed with #1815

PR #1815, opened 2025-07-14
Reject non-finite values in forward controller subscriber callback (#…

lfurno:
This PR adds a check to the ForwardControllersBase subscriber callback to reject any non-finite command values and log a warning. It prevents NaNs or infinities from entering the real-time buffer. Related to issue [#1775](https://github.com/ros-controls/ros2_controllers/issues/1775).

### DIFF
--- a/forward_command_controller/src/forward_controllers_base.cpp
+++ b/forward_command_controller/src/forward_controllers_base.cpp
@@ -72,7 +72,21 @@ controller_interface::CallbackReturn ForwardControllersBase::on_configure(
 
   joints_command_subscriber_ = get_node()->create_subscription<CmdType>(
     "~/commands", rclcpp::SystemDefaultsQoS(),
-    [this](const CmdType::SharedPtr msg) { rt_command_.set(*msg); });
+    [this](const CmdType::SharedPtr msg)
+    {
+      const auto cmd = *msg;
+
+      if (!std::all_of(
+            cmd.data.cbegin(), cmd.data.cend(),
+            [](const auto & value) { return std::isfinite(value); }))
+      {
+        RCLCPP_WARN_THROTTLE(
+          get_node()->get_logger(), *(get_node()->get_clock()), 1000,
+          "Non-finite value received. Dropping message");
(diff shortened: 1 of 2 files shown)


lfurno:
@christophfroehlich Hello! Is this the behavior you envisaged for the check? 

If yes, I verified the new behavior using the existing ForwardCommandControllerTest.CommandCallbackTest. By inserting
std::numeric_limits<double>::infinity() into the published command, I confirmed that the warning is emitted and the real-time buffer is not updated.

Would you like me to add a dedicated unit test that explicitly checks that, when a non-finite value is received, the joint command handle remains unchanged?

christophfroehlich:
> @christophfroehlich Hello! Is this the behavior you envisaged for the check?

Yes, this is perfect. I'm just thinking if we should use a throttled warning instead?

> Would you like me to add a dedicated unit test that explicitly checks that, when a non-finite value is received, the joint command handle remains unchanged?

yes please.

christophfroehlich:
Please don't force-push if there is no need to, it just will make future review rounds more time consuming.

christophfroehlich (commented):
not sure what we should do if a single field is infinite, should we drop the whole message or only the infinite value?

saikishor:
> not sure what we should do if a single field is infinite, should we drop the whole message or only the infinite value?

I think it makes sense to drop the whole message. If needed, we can bring this up in PMC meeting

lfurno:
> Please don't force-push if there is no need to, it just will make future review rounds more time consuming.

Ok, I am sorry. The best practice would have been to add a new commit on top of the old one, instead of push forcing the modifications in the same commit?

christophfroehlich:
> > Please don't force-push if there is no need to, it just will make future review rounds more time consuming.
> 
> Ok, I am sorry. The best practice would have been to add a new commit on top of the old one, instead of push forcing the modifications in the same commit? 

exactly. we squash all commits when we merge the PRs anyways.

saikishor (approved):

PR #1816, opened 2025-07-15
Remove unneccessary usage of `get_ordered_interfaces`

### LINKED ISSUE #116: Update documentation for order of `command_interfaces_`
> @bmagyar This assumes that command_interfaces_ comes ordered as you requested it, but I did not find any documentation explicitly mentioning this, that's why I went with the `get_ordered_interfaces` thing.
> 
> we could just explicitely mention that the itnerfaces are provided in the order they were requested, that would simplify things.

_Originally posted by @v-lopez in https://github.com/ros-controls/ros2_controllers/pull/99#discussion_r532390821_

christophfroehlich:
Fixes #116

Admittance_controller: There were some copies of the interface names without ever using them again.
forward_command_controller:  I changed the joint and interface parameters to be read_only, then we don't need the ordering there at all. This might be a breaking change, but I don't think that dynamically changing that would have worked without notifying the CM?

### DIFF
--- a/forward_command_controller/test/test_forward_command_controller.hpp
+++ b/forward_command_controller/test/test_forward_command_controller.hpp
@@ -57,7 +57,7 @@ public:
   void SetUp();
   void TearDown();
 
-  void SetUpController();
+  void SetUpController(const std::vector<rclcpp::Parameter> & parameters = {});
   void SetUpHandles();
 
 protected:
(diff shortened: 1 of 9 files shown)


forward_command_controller/test/test_forward_command_controller.hpp
```diff
@@ -57,7 +57,7 @@ class ForwardCommandControllerTest : public ::testing::Test
   void SetUp();
   void TearDown();
 
-  void SetUpController();
+  void SetUpController(const std::vector<rclcpp::Parameter> & parameters = {});
   void SetUpHandles();
```
christophfroehlich:
```suggestion
```

saikishor (approved):

PR #1823, opened 2025-07-21
Apply API change of PidROS

christophfroehlich:
See https://github.com/ros-controls/control_toolbox/pull/431

Earlier, the state was published to global `/<controller_name>/<dof_name>/pid_state` instead. Should we better not change this?

### DIFF
--- a/ros2_controllers.kilted.repos
+++ b/ros2_controllers.kilted.repos
@@ -14,7 +14,7 @@ repositories:
   control_toolbox:
     type: git
     url: https://github.com/ros-controls/control_toolbox.git
-    version: ros2-master
+    version: pid/prefix
   kinematics_interface:
     type: git
     url: https://github.com/ros-controls/kinematics_interface.git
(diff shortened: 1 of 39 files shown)


ros2_controllers.kilted.repos
```diff
@@ -14,7 +14,7 @@ repositories:
   control_toolbox:
     type: git
     url: https://github.com/ros-controls/control_toolbox.git
-    version: ros2-master
+    version: pid/prefix
```
christophfroehlich:
```suggestion
    version: ros2-master
```

ros2_controllers.rolling.repos
```diff
@@ -14,7 +14,7 @@ repositories:
   control_toolbox:
     type: git
     url: https://github.com/ros-controls/control_toolbox.git
-    version: ros2-master
+    version: pid/prefix
```
```suggestion
    version: ros2-master
```

saikishor (approved):
The changes look good to me

PR #1824, opened 2025-07-21
Change the tests to work without deprecated PID settings

christophfroehlich:
as per title. I had to fix some test criteria as the integral action is not added at the first update call (integration is done after calculating the command).

needs https://github.com/ros-controls/control_toolbox/pull/432

### DIFF
--- a/ros2_controllers.kilted.repos
+++ b/ros2_controllers.kilted.repos
@@ -14,7 +14,7 @@ repositories:
   control_toolbox:
     type: git
     url: https://github.com/ros-controls/control_toolbox.git
-    version: fix/pid
+    version: ros2-master
   kinematics_interface:
     type: git
     url: https://github.com/ros-controls/kinematics_interface.git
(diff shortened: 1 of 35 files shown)


ros2_controllers.kilted.repos
```diff
@@ -14,7 +14,7 @@ repositories:
   control_toolbox:
     type: git
     url: https://github.com/ros-controls/control_toolbox.git
-    version: ros2-master
+    version: fix/pid
```
christophfroehlich:
```suggestion
    version: ros2-master
```

ros2_controllers.rolling.repos
```diff
@@ -14,7 +14,7 @@ repositories:
   control_toolbox:
     type: git
     url: https://github.com/ros-controls/control_toolbox.git
-    version: ros2-master
+    version: fix/pid
```
```suggestion
    version: ros2-master
```

bmagyar (approved):
thanks for doing the math :D

saikishor (approved):

PR #1829, opened 2025-07-23
Extend Tricycle Controller with “Safe” Velocity Clipping

ViktorHotsii:
Extend Tricycle Controller with “Safe” Velocity Clipping

What’s Changed:
- New Params
-- use_clipping (bool): enable/disable wheel‑level clipping
-- clipping_speed_limits & clipping_angle_limits (arrays): define Ws vs. steering‑angle envelopes

- Clipping Logic
 -- Two private methods clamp wheel angular speed and steering angle against the limits

- Zero Impact by Default

- Clipping disabled unless use_clipping is true

 - All original kinematics and limiters remain intact


christophfroehlich:
It might be a good idea to pitch your idea before working on it and opening a PR ;) Do you know that the clipping of commands is done already in the joint_limits plugins of the resource manager?

PR #1830, opened 2025-07-23
Update API for realtime publisher

christophfroehlich:
https://github.com/ros-controls/realtime_tools/pull/323

admittance controller didn't use the try_lock in the RT loop btw.


saikishor (approved):
Looks super good to me. We should definitely backport this

PR #1833, opened 2025-07-24
Add calibration possibility to IMU broadcaster

christophfroehlich:
Similar to multiplier+offset of the FTS broadcaster, I changed the IMU broadcaster to

- a chainable controller, exporting the fields of the message as state interfaces
- apply a calibration of the IMU, to be configured as a roll-pitch-yaw orientation from parameters

I used the method from [imu_pipeline](https://github.com/ros-perception/imu_pipeline/blob/jazzy/imu_transformer/include/imu_transformer/tf2_sensor_msgs.h), but copied it over to avoid pulling in another dependency. If you prefer, I could do and use a tf2 transform instead of Eigen quaternion to save the calibrated orientation.


imu_sensor_broadcaster/include/imu_sensor_broadcaster/imu_sensor_broadcaster.hpp
```diff
@@ -49,18 +51,25 @@ class IMUSensorBroadcaster : public controller_interface::ControllerInterface
...
 
+  controller_interface::return_type update_reference_from_subscribers(
+    const rclcpp::Time & time, const rclcpp::Duration & period) override;
+
+  std::vector<hardware_interface::StateInterface> on_export_state_interfaces() override;
+
 protected:
   std::shared_ptr<ParamListener> param_listener_;
   Params params_;
+  Eigen::Quaternion<double> r_;
```
saikishor:
```suggestion
  Eigen::Quaterniond r_;
```

There is already an internal typdef mapping that to d
https://eigen.tuxfamily.org/dox-devel/classEigen_1_1Quaternion.html

Not a big deal, if this is not applied.

imu_sensor_broadcaster/include/imu_sensor_broadcaster/imu_transform.hpp
```diff
@@ -0,0 +1,100 @@
...
+    Eigen::AngleAxisd(yaw, Eigen::Vector3d::UnitZ()) *
+    Eigen::AngleAxisd(pitch, Eigen::Vector3d::UnitY()) *
+    Eigen::AngleAxisd(roll, Eigen::Vector3d::UnitX()));
+}
+
+/**
+ * @brief Transforms a covariance array from one frame to another
+ */
+inline void transformCovariance(
+  const std::array<double, 9> & in, std::array<double, 9> & out, Eigen::Quaternion<double> r)
```
saikishor:
```suggestion
inline void transformCovariance(
  const std::array<double, 9> & in, const Eigen::Quaternion<double> & r, std::array<double, 9> & out)
```

Can we change the order and have `r` as const, so the first arguments are const and then comes the output. Does it make sense?

r is the offset or the transformation that we need to apply. Am i right?.

imu_sensor_broadcaster/include/imu_sensor_broadcaster/imu_transform.hpp
```diff
@@ -0,0 +1,100 @@
...
+  Eigen::Map<Eigen::Matrix<double, 3, 3, Eigen::RowMajor> > cov_out(out.data());
+  cov_out = r * cov_in * r.inverse();
+}
+
+/**
+ * @brief Transforms sensor_msgs::Imu data from one frame to another
+ */
+inline void doTransform(
+  const sensor_msgs::msg::Imu & imu_in, sensor_msgs::msg::Imu & imu_out,
+  const Eigen::Quaternion<double> r)
```
saikishor:
```suggestion
inline void doTransform(
  const sensor_msgs::msg::Imu & imu_in, const Eigen::Quaternion<double> &r, sensor_msgs::msg::Imu & imu_out)
```

It's just my habit to have const and inputs at one side of the functions, if we need to leave it how it is, that is totally okay with me

imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
```diff
@@ -43,7 +44,19 @@ controller_interface::CallbackReturn IMUSensorBroadcaster::on_init()
 controller_interface::CallbackReturn IMUSensorBroadcaster::on_configure(
   const rclcpp_lifecycle::State & /*previous_state*/)
 {
-  params_ = param_listener_->get_params();
+  try
+  {
+    params_ = param_listener_->get_params();
+    r_ =
+      quat_from_euler(params_.calibration.roll, params_.calibration.pitch, params_.calibration.yaw);
```
saikishor:
```suggestion
    r_ =
      quat_from_euler(params_.calibration.roll, params_.calibration.pitch, params_.calibration.yaw);
    r_.normalize();
```

I made it a habit to normalize the quaternions

imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
```diff
@@ -110,22 +120,75 @@ controller_interface::CallbackReturn IMUSensorBroadcaster::on_deactivate(
   return CallbackReturn::SUCCESS;
 }
 
-controller_interface::return_type IMUSensorBroadcaster::update(
+controller_interface::return_type IMUSensorBroadcaster::update_and_write_commands(
   const rclcpp::Time & time, const rclcpp::Duration & /*period*/)
 {
-  if (realtime_publisher_ && realtime_publisher_->trylock())
+  imu_sensor_->get_values_as_message(state_message_);
+  doTransform(state_message_, state_message_, r_);
```
saikishor:
```suggestion
  sensor_msgs::msg::Imu input_imu;
  imu_sensor_->get_values_as_message(input_imu);
  doTransform(input_imu, r_, state_message_);
```

Should we do it this way?. No extra memory is allocated this way?. I'm a bit skeptic about using same variable in input and output. It is just me XD

imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
```diff
@@ -110,22 +120,75 @@ controller_interface::CallbackReturn IMUSensorBroadcaster::on_deactivate(
...
+}
+
+std::vector<hardware_interface::StateInterface> IMUSensorBroadcaster::on_export_state_interfaces()
+{
+  std::vector<hardware_interface::StateInterface> exported_state_interfaces;
+
+  std::string export_prefix = get_node()->get_name();
+  if (!params_.sensor_name.empty())
+  {
+    // Update the prefix and get the proper force and torque names
```
saikishor:
```suggestion
    // Update the prefix and get the proper IMU sensor naming
```

saikishor (commented):
In general, looks great

Amazing job as always @christophfroehlich

imu_sensor_broadcaster/include/imu_sensor_broadcaster/imu_transform.hpp
christophfroehlich: As I mentioned, I basically copied it over from the other repo. I usually order it like (output, input) like memcpy for example, I can apply that.

imu_sensor_broadcaster/src/imu_sensor_broadcaster.cpp
christophfroehlich: I had to initialize input_imu to have the fixed frame_id. I'm not sure about memory allocation here, isn't the current code always allocating memory? Or do you expect the compiler to optimize that?

saikishor: Yes, I think the compiler will be able to optimize here

saikishor: image
It shouldn't be a problem as most of it is fixed size

PR #1850, opened 2025-08-03
Filling `index` field in feedback message of the action interface

GiuseppeMonetti95:
Hi everyone!

I took this task from the good-first-issue here: https://github.com/ros-controls/ros2_controllers/issues/1742

According to the new field description in the .action file, this looks like working for me.
The only doubt I have is about testing, I don't know if this change is worth a test case.

Looking forward to have reviews 😄

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -175,7 +175,7 @@ controller_interface::return_type JointTrajectoryController::update(
     }
   }
 
-  // don't update goal after we sampled the trajectory to avoid any racecondition
+  // don't update goal after we sampled the trajectory to avoid any race condition
   const auto active_goal = *rt_active_goal_.readFromRT();
 
   // Check if a new trajectory message has been received from Non-RT threads
@@ -235,6 +235,8 @@ controller_interface::return_type JointTrajectoryController::update(
     const bool valid_point = current_trajectory_->sample(
       traj_time_ + update_period_, interpolation_method_, command_next_, start_segment_itr,
       end_segment_itr, false);
+
+    const size_t next_point_index = std::distance(current_trajectory_->begin(), end_segment_itr);
 
     state_current_.time_from_start = time - current_trajectory_->time_from_start();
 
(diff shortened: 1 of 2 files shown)


christophfroehlich (changes requested):
Thanks a lot. Except for the conversion error below, this looks very good. Testing in existing testcases is fine.

```
   /home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/src/ros2_controllers/joint_trajectory_controller/src/joint_trajectory_controller.cpp:239:37: error: implicit conversion changes signedness: 'typename iterator_traits<__normal_iterator<const JointTrajectoryPoint_<allocator<void>> *, vector<JointTrajectoryPoint_<allocator<void>>>>>::difference_type' (aka 'long') to 'const size_t' (aka 'const unsigned long') [-Werror,-Wsign-conversion]
    239 |     const size_t next_point_index = std::distance(current_trajectory_->begin(), end_segment_itr);
        |                  ~~~~~~~~~~~~~~~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

GiuseppeMonetti95:
Implicit conversion problem has been fixed. However, I had a test failure that I was not capable to reproduce in the little time I had today: Even if the `time_from_start ` was less then the one of the next via point in the trajectory, in the feedback message there was already the next point.
I have a feeling that it may be due to path and/or goal tolerances. I will investigate more.
Does the problem description sound familiar? Do you have a guess such that I can find faster the problem?

christophfroehlich:
> I have a feeling that it may be due to path and/or goal tolerances. I will investigate more. 

I don't think that this relates, I can't imagine now why this should happen. 

but the failing jazzy test now shows the same?
```
   <<< failure message
    /home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/src/ros2_controllers/joint_trajectory_controller/test/test_trajectory_actions.cpp:395
    Expected equality of these values:
      static_cast<int32_t>(expected_index)
        Which is: 0
      feedback_msg->index
        Which is: 1
  >>>

```

GiuseppeMonetti95:
There was a logic error in my implementation, now it works!
I also made the conversion explicit.

christophfroehlich (approved):
Thanks, now LGTM

GiuseppeMonetti95:
> Thanks, now LGTM

Thanks! Do you understand why the pipelines are failing? It looks like the failure is in the `apt install ...` command. I don't see why the failure should be caused by my changes

christophfroehlich:
not related at all, moveit packages are not available

GiuseppeMonetti95:
Should I do something extra to get this merged?

saikishor (approved):

PR #1854, opened 2025-08-06
Update odometry implementation in diff_drive

### LINKED ISSUE #271: [DiffDriveController] use "period" argument in the update method insead of calculating "dt"

### LINKED ISSUE #357: Odometry position approximation: exact integration / rungeKutta2 integration / rungeKutta4 integration?
Line 150 of odometry.cpp gives the comment:
`/// Exact integration (should solve problems when angular is zero):`
However, it is executed when `!(fabs(angular) < 1e-6)` as it is in the else clause. Is this supposed to be the other way around? Or is the comment wrong?
And maybe this discussion is better another place, but when I see the RungeKutta2 function, I'm not much of a Mathematician, but it looks a lot different from what I can find online about RungeKutta2. Is this a work in progress? And maybe we can implement RungeKutta4 instead for higher accuracy over longer distances? I understand with a high sample rate this shouldn't have much effect, but when driving fast / sampling little it will.
Background information: As I'm looking for ways to convert recorded motor controller positions to XY location, in order to visualise this, I try to see how ROS handles this.

Amronos:
Fixes #271.
Fixes #357.

I have updated the odometry implementation of the ``diff_drive_controller`` to fix the above two issues alongside improving the code's readability. This implementation is similar to that of ``omni_wheel_drive_controller``.
I have created some new methods and deprecated the older ones.

Should I also make similar changes to other controllers? If yes, should I do that in a separate PR?


diff_drive_controller/include/diff_drive_controller/odometry.hpp
```diff
@@ -37,10 +37,26 @@ class Odometry
...
+  [[deprecated(
+    "Replaced by bool updateFromPos(const double & left_pos, const double & right_pos, const "
+    "rclcpp::Time & time).")]]
   bool update(double left_pos, double right_pos, const rclcpp::Time & time);
+  [[deprecated(
+    "Replaced by bool updateFromVel(const double & left_vel, const double & right_vel, const "
+    "rclcpp::Time & time).")]]
   bool updateFromVelocity(double left_vel, double right_vel, const rclcpp::Time & time);
+  [[deprecated(
+    "Replaced by bool updateOpenLoop(const double & linear_vel, const double & angular_vel, const "
```
christophfroehlich:
```suggestion
    "Replaced by bool tryUpdateOpenLoop(const double & linear_vel, const double & angular_vel, const "
```

diff_drive_controller/include/diff_drive_controller/odometry.hpp
```diff
@@ -37,10 +37,26 @@ class Odometry
...
+    "rclcpp::Time "
+    "& time).")]]
   void updateOpenLoop(double linear, double angular, const rclcpp::Time & time);
+
+  bool updateFromPos(const double & left_pos, const double & right_pos, const rclcpp::Time & time);
+  bool updateFromVel(const double & left_vel, const double & right_vel, const rclcpp::Time & time);
+  bool tryUpdateOpenLoop(
+    const double & linear_vel, const double & angular_vel, const rclcpp::Time & time);
```
christophfroehlich:
```suggestion
  [[deprecated]]
  void init(const rclcpp::Time & time);
  [[deprecated(
    "Replaced by bool updateFromPos(const double left_pos, const double right_pos, const "
    "rclcpp::Time & time).")]]
  bool update(double left_pos, double right_pos, const rclcpp::Time & time);
  [[deprecated(
    "Replaced by bool updateFromVel(const double left_vel, const double right_vel, const "
    "rclcpp::Time & time).")]]
  bool updateFromVelocity(double left_vel, double right_vel, const rclcpp::Time & time);
  [[deprecated(
    "Replaced by bool updateOpenLoop(const double linear_vel, const double angular_vel, const "
    "rclcpp::Time "
    "& time).")]]
  void updateOpenLoop(double linear, double angular, const rclcpp::Time & time);

  bool updateFromPos(const double left_pos, const double right_pos, const rclcpp::Time & time);
  bool updateFromVel(const double left_vel, const double right_vel, const rclcpp::Time & time);
  bool tryUpdateOpenLoop(
    const double linear_vel, const double angular_vel, const rclcpp::Time & time);
```
IMHO passing by value is preferred for trivial types?

diff_drive_controller/include/diff_drive_controller/odometry.hpp
```diff
@@ -60,8 +76,12 @@ class Odometry
   using RollingMeanAccumulator = rcppmath::RollingMeanAccumulator<double>;
 #endif
 
+  [[deprecated("Replaced by void integrate(const double & dx, const double & dheading).")]]
   void integrateRungeKutta2(double linear, double angular);
+  [[deprecated("Replaced by void integrate(const double & dx, const double & dheading).")]]
   void integrateExact(double linear, double angular);
+
+  void integrate(const double & dx, const double & dheading);
```
christophfroehlich:
```suggestion

  void integrate(const double dx, const double dheading);
```

diff_drive_controller/src/odometry.cpp
```diff
@@ -100,6 +125,30 @@ bool Odometry::updateFromVelocity(double left_vel, double right_vel, const rclcp
   return true;
 }
 
+bool Odometry::updateFromVel(
+  const double & left_vel, const double & right_vel, const rclcpp::Time & time)
+{
+  const double dt = time.seconds() - timestamp_.seconds();
```
christophfroehlich: this is exactly the pattern mentioned with #271 to be removed?

diff_drive_controller/src/odometry.cpp
```diff
@@ -112,6 +161,23 @@ void Odometry::updateOpenLoop(double linear, double angular, const rclcpp::Time
...
+  // Integrate odometry:
+  integrate(linear_vel * dt, angular_vel * dt);
+
+  timestamp_ = time;
+
+  // Save last linear and angular velocity:
+  linear_ = linear_vel;
+  angular_ = angular_vel;
+
+  return true;
```
christophfroehlich, line 167: why a return value, which is always true?

diff_drive_controller/src/odometry.cpp
```diff
@@ -161,6 +227,24 @@ void Odometry::integrateExact(double linear, double angular)
...
 }
 
+void Odometry::integrate(const double & dx, const double & dheading)
+{
+  if (fabs(dheading) < 1e-6)
+  {
+    // For very small dheading, approximate to linear motion
+    x_ = x_ + (dx * std::cos(heading_));
+    y_ = y_ + (dx * std::sin(heading_));
+    heading_ = heading_ + dheading;
```
christophfroehlich:
```suggestion
    x_ += (dx * std::cos(heading_));
    y_ += (dx * std::sin(heading_));
    heading_ += dheading;
```

diff_drive_controller/src/odometry.cpp
```diff
@@ -161,6 +227,24 @@ void Odometry::integrateExact(double linear, double angular)
...
+  }
+  else
+  {
+    const double heading_old = heading_;
+    heading_ = heading_ + dheading;
+    x_ = x_ + ((dx / dheading) * (std::sin(heading_) - std::sin(heading_old)));
+    y_ = y_ - (dx / dheading) * (std::cos(heading_) - std::cos(heading_old));
```
christophfroehlich:
```suggestion
    heading_ += dheading;
    x_ += ((dx / dheading) * (std::sin(heading_) - std::sin(heading_old)));
    y_ += -(dx / dheading) * (std::cos(heading_) - std::cos(heading_old));
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -200,63 +202,64 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
     {
-      odometry_.update(left_feedback_mean, right_feedback_mean, time);
+      odometry_updated = odometry_.updateFromPos(left_feedback_mean, right_feedback_mean, time);
     }
     else
     {
-      odometry_.updateFromVelocity(
-        left_feedback_mean * left_wheel_radius * period.seconds(),
-        right_feedback_mean * right_wheel_radius * period.seconds(), time);
+      odometry_updated = odometry_.updateFromVel(left_feedback_mean, right_feedback_mean, time);
```
christophfroehlich: where did the wheel_radius go?

christophfroehlich (changes requested):
Let's discuss my points first, and merge the outcome. If necessary, let's open a new PR afterwards for the other controllers.

diff_drive_controller/src/odometry.cpp
Amronos: Sorry, I should have explained this in the PR description.
The reasoning here is the same as that of this https://github.com/ros-controls/ros2_controllers/pull/260#discussion_r746362028.
Additionally, ``timestamp_`` records the last time an update was run; therefore, the correct value of ``dt`` can only be obtained from there.
It is used in this function despite ``(std::fabs(dt) < 1e-6)`` not being used, as it is also called from ``updateFromPos()`` which uses the ``timestamp_`` variable.

diff_drive_controller/src/odometry.cpp
Amronos: This is there to maintain consistency with other update methods, make it easier to add code to this function in the future that may return false, and make the ``diff_drive_controller.cpp`` code simpler.
``updateFromVel()`` also has this line for the same reason.
I can change everything back to the normal ``updateOpenLoop()`` function if needed, though.

diff_drive_controller/include/diff_drive_controller/odometry.hpp
Amronos: Fixed in commit [629cf5e](https://github.com/ros-controls/ros2_controllers/pull/1854/commits/629cf5efd393657112d120f8860a16ff0a8c3d65),

diff_drive_controller/src/diff_drive_controller.cpp
Amronos: My bad! This was meant to be multiplied in the ``updateFromVel()`` function. Fixed with [d387750](https://github.com/ros-controls/ros2_controllers/pull/1854/commits/d387750bff2973b112223ee21cf7b3335eeca5b5).

Amronos:
Sorry for the delay in fixing the failing checks here! I was busy shifting Linux distros and had tried to make the commits that caused the checks to fail through the web.

diff_drive_controller/src/odometry.cpp
```diff
@@ -73,6 +73,31 @@ bool Odometry::update(double left_pos, double right_pos, const rclcpp::Time & ti
...
+
+  // Update old position with current:
+  left_wheel_old_pos_ = left_pos;
+  right_wheel_old_pos_ = right_pos;
+
+  if (updateFromVel(left_vel, right_vel, time))
+  {
+    return true;
+  }
+  return false;
```
christophfroehlich:
```suggestion
  return updateFromVel(left_vel, right_vel, time);
```

diff_drive_controller/src/odometry.cpp
christophfroehlich: Sorry, I still don't understand. Why not using the period argument from the update method and pass it to the update method?

Amronos: - Basically due to the ``if (std::fabs(dt) < 1e-6) {return false;}`` check in ``updateFromPos()`` we need to maintain the ``timestamp_`` variable. 
- ``timestamp_`` is only updated when the check returns false, and we continue with the code. If ``timestamp_`` was updated every time, it would yield incorrect results, as we only update odometry when the code passes through that check.
- If ``period`` was directly passed, then it would yield an incorrect ``dt`` (``dt`` will now be replaced with ``period``), as the current way to calculate ``dt`` is through the ``timestamp_``, which is not updated every time the method is called.
- The ``updateFromVel()`` function can be changed to use ``period`` if the arguments of ``updateFromPos()`` remain the same, and we pass it ``dt`` from there when using position feedback, and when using velocity feedback, we pass ``period`` from the main update method. This can only be done, though, if we are sure that the user can't/won't switch between velocity and position feedback while the controller is running (#1781).

christophfroehlich: Thanks, I understand the motivation now.
1. Is the period argument ever that small so we need the check? (might come from ROS 1 days)
2. If so (maybe the first update call after activation or something like this. This can't happen on a regular basis), Is the error really that high if we use the period and still skip if it is close to zero?

christophfroehlich: https://github.com/ros-controls/ros2_control/issues/2336

Amronos: 1. I also thought about this, as the period argument would only be this small if the update rate is 10^6; I don't think people set the update rate that high, but I still kept the check. Because of https://github.com/ros-controls/ros2_control/issues/2336, ig we will either way need it.
2. Didn't think about it like that :). There would only be a minor 10^-6 or less seconds unaccounted for, which shouldn't have any effect on the odometry. That is assuming that the update rate would never be set that high to make this error accumulate.

So should I pass period to both the functions, in ``updateFromPos()`` keep the zero check, and stop the use of ``timestamp_`` in the new functions?

christophfroehlich: yes please do so :+1:

Amronos:
@christophfroehlich pre-commit isn't working for me for some reason, but besides that I have made the changed you requested. If possible, could you clone the pr, and push a fix for the pre-commit? Thanks!

diff_drive_controller/src/odometry.cpp
Amronos: Done in [77eade3](https://github.com/ros-controls/ros2_controllers/pull/1854/commits/77eade3a2f89d51392a52d7755dffb8ea1ac56cf).

Amronos:
Sorry for the delay on this! All requested changes have been made now, please review!

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -200,62 +203,65 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
+        odometry_message_.pose.pose.orientation.x = orientation.x();
+        odometry_message_.pose.pose.orientation.y = orientation.y();
+        odometry_message_.pose.pose.orientation.z = orientation.z();
+        odometry_message_.pose.pose.orientation.w = orientation.w();
+        odometry_message_.twist.twist.linear.x = odometry_.getLinear();
+        odometry_message_.twist.twist.angular.z = odometry_.getAngular();
+        realtime_odometry_publisher_->try_publish(odometry_message_);
+      }
+
+      if (params_.enable_odom_tf && realtime_odometry_transform_publisher_->trylock())
```
Amronos:
```suggestion
      if (params_.enable_odom_tf && realtime_odometry_transform_publisher_
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -200,62 +203,65 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
+        odometry_message_.pose.pose.orientation.x = orientation.x();
+        odometry_message_.pose.pose.orientation.y = orientation.y();
+        odometry_message_.pose.pose.orientation.z = orientation.z();
+        odometry_message_.pose.pose.orientation.w = orientation.w();
+        odometry_message_.twist.twist.linear.x = odometry_.getLinear();
+        odometry_message_.twist.twist.angular.z = odometry_.getAngular();
+        realtime_odometry_publisher_->try_publish(odometry_message_);
+      }
+
+      if (params_.enable_odom_tf && realtime_odometry_transform_publisher_
```
Amronos:
```suggestion
      if (params_.enable_odom_tf && realtime_odometry_transform_publisher_)
```

diff_drive_controller/src/odometry.cpp
```diff
@@ -161,6 +212,24 @@ void Odometry::integrateExact(double linear, double angular)
   }
 }
 
+void Odometry::integrate(const double dx, const double dheading)
```
christophfroehlich: I'd find it more meaningful to pass `dt` to the integration method as well, instead of delta values only?
Or we rename the method to `accumulate` or something.

diff_drive_controller/include/diff_drive_controller/odometry.hpp
```diff
@@ -37,10 +37,25 @@ class Odometry
...
   bool updateFromVelocity(double left_vel, double right_vel, const rclcpp::Time & time);
+  [[deprecated(
+    "Replaced by bool tryUpdateOpenLoop(const double linear_vel, const double angular_vel, const "
+    "rclcpp::Time "
+    "& time).")]]
   void updateOpenLoop(double linear, double angular, const rclcpp::Time & time);
+
+  bool updateFromPos(const double left_pos, const double right_pos, const double dt);
+  bool updateFromVel(const double left_vel, const double right_vel, const double dt);
+  bool tryUpdateOpenLoop(const double linear_vel, const double angular_vel, const double dt);
```
christophfroehlich: - no need for const qualifier for non-ref datatypes. 
- let's change the code style to snake case for new methods.

christophfroehlich (changes requested):
Thanks for your patience. This is almost good to be merged, two minor comments only

diff_drive_controller/include/diff_drive_controller/odometry.hpp
Amronos: Done in commit [6a6c38c](https://github.com/ros-controls/ros2_controllers/pull/1854/commits/6a6c38c98383646e5a3cac22007c78d152f0b46a).

diff_drive_controller/src/odometry.cpp
Amronos: Done in commit [6a6c38c](https://github.com/ros-controls/ros2_controllers/pull/1854/commits/6a6c38c98383646e5a3cac22007c78d152f0b46a).

christophfroehlich (approved):
Thanks for the latest cleanup! LGTM

diff_drive_controller/include/diff_drive_controller/odometry.hpp
```diff
@@ -37,10 +37,25 @@ class Odometry
 public:
   explicit Odometry(size_t velocity_rolling_window_size = 10);
 
+  [[deprecated]]
   void init(const rclcpp::Time & time);
+  [[deprecated(
+    "Replaced by bool update_from_pos(double left_pos, double right_pos, const "
+    "rclcpp::Time & time).")]]
```
Juliaj:
```suggestion
  [[deprecated(
    "Replaced by bool update_from_pos(double left_pos, double right_pos, double "
    "dt).")]]
```

diff_drive_controller/include/diff_drive_controller/odometry.hpp
```diff
@@ -37,10 +37,25 @@ class Odometry
...
   bool update(double left_pos, double right_pos, const rclcpp::Time & time);
+  [[deprecated(
+    "Replaced by bool update_from_vel(double left_vel, double right_vel, const "
+    "rclcpp::Time & time).")]]
```
Juliaj:
```suggestion
 [[deprecated(
    "Replaced by bool update_from_vel(double left_vel, double right_vel, double "
    "dt).")]]
```

diff_drive_controller/include/diff_drive_controller/odometry.hpp
```diff
@@ -37,10 +37,25 @@ class Odometry
...
   bool updateFromVelocity(double left_vel, double right_vel, const rclcpp::Time & time);
+  [[deprecated(
+    "Replaced by bool try_update_open_loop(double linear_vel, double angular_vel, const "
+    "rclcpp::Time "
+    "& time).")]]
```
Juliaj:
```suggestion
  [[deprecated(
       "Replaced by bool try_update_open_loop(double linear_vel, double angular_vel, double "
    "dt).")]]
```

diff_drive_controller/src/odometry.cpp
```diff
@@ -100,6 +119,26 @@ bool Odometry::updateFromVelocity(double left_vel, double right_vel, const rclcp
   return true;
 }
 
+bool Odometry::update_from_vel(double left_vel, double right_vel, double dt)
+{
+  // Compute linear and angular velocities of the robot:
```
Juliaj, line 128: Let's add this check back
```
  if (dt < 0.0001)
  {
    return false;  // Interval too small to integrate with
  }
```

diff_drive_controller/src/odometry.cpp
```diff
@@ -161,6 +212,26 @@ void Odometry::integrateExact(double linear, double angular)
   }
 }
 
+void Odometry::integrate(double linear_vel, double angular_vel, double dt)
+{
+  const double dx = linear_vel * dt;
```
Juliaj, line 226:
```suggestion
  // Skip integration for invalid time intervals
  if (std::fabs(dt) < 1e-6)
  {
    return;
  }
  const double dx = linear_vel * dt;
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -200,62 +203,65 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
-    if (previous_publish_timestamp_ + publish_period_ < time)
+    tf2::Quaternion orientation;
+    orientation.setRPY(0.0, 0.0, odometry_.getHeading());
+
+    bool should_publish = false;
+    try
     {
-      previous_publish_timestamp_ += publish_period_;
-      should_publish = true;
+      if (previous_publish_timestamp_ + publish_period_ < time)
```
Juliaj:
```suggestion
      if (previous_publish_timestamp_ + publish_period_ <= time)
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -200,62 +203,65 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
-      odometry_message_.pose.pose.orientation.y = orientation.y();
-      odometry_message_.pose.pose.orientation.z = orientation.z();
-      odometry_message_.pose.pose.orientation.w = orientation.w();
-      odometry_message_.twist.twist.linear.x = odometry_.getLinear();
-      odometry_message_.twist.twist.angular.z = odometry_.getAngular();
-      realtime_odometry_publisher_->try_publish(odometry_message_);
+      // Handle exceptions when the time source changes and initialize publish timestamp
+      previous_publish_timestamp_ = time;
+      should_publish = true;
     }
```
Juliaj, line 235: Is this block meant to catch the error of previous_publish_timestamp as unintialized ? if so, we should initialize it in on_configure. If not, this block may not be reachable.

Juliaj:
@Amronos, thanks for working on this PR. It would be great to add some tests for the odometry changes. If you can't accommodate this within this PR, please consider creating a GitHub issue so the community can pick it up.

diff_drive_controller/src/odometry.cpp
Amronos:
```suggestion
  if (std::fabs(dt) < 1e-6)
  {
    return false;  // Interval too small to integrate with
  }
  // Compute linear and angular velocities of the robot:
```

diff_drive_controller/src/diff_drive_controller.cpp
Amronos: This isn't really a part of the changes made in this PR, perhaps create a separate issue? A similar pattern exists in other controllers as well, so the change will need to be propagated to them as well.
If you create another issue, please also link [this](https://github.com/ros-controls/ros2_controllers/pull/1854#discussion_r2573152912) suggestion in it.

Amronos:
> @Amronos, thanks for working on this PR. It would be great to add some tests for the odometry changes. If you can't accommodate this within this PR, please consider creating a GitHub issue so the community can pick it up.

Thanks a lot for the review! I don't really have the time to work on tests for this PR rn, and so have gone ahead and created #2033.

diff_drive_controller/src/diff_drive_controller.cpp
christophfroehlich: You are right, this existed before and the diff is just not properly rendered by the UI. @Juliaj please open a (first) issue with this.

christophfroehlich (approved):

diff_drive_controller/src/diff_drive_controller.cpp
Juliaj: @christophfroehlich https://github.com/ros-controls/ros2_controllers/issues/2035.

PR #1855, opened 2025-08-07
Implement realtime-safe stall detection for parallel gripper controller

### LINKED ISSUE #1853: Potential realtime thread safety violations in parallel gripper action controller
**Describe the bug**
The `last_movement_time_` variable is accessed from both realtime (`check_for_success()` called from `update()`) and non-realtime contexts (`accepted_callback()`) without proper synchronization.

**Expected behavior**
My guess is that we should use (I could be super wrong here.)

```
realtime_tools::RealtimeThreadSafeBox<rclcpp::Time> last_movement_time_;
```

 - OS: Ubuntu
 - Version: Jazzy
-  branch: main

**Additional context**
1. A reason I can think of why current implementation works is that when `accepted_callback()` is called the active goal is preempted and the `check_for_success()` is never called.

ashwinsushil:
Convert last_movement_time_ from direct rclcpp::Time to RealtimeThreadSafeBox to ensure thread-safe communication between realtime and non-realtime contexts.

Changes:
- Replace rclcpp::Time last_movement_time_ with RealtimeThreadSafeBox<rclcpp::Time>
- Initialize RealtimeThreadSafeBox in on_activate() method

Fixes potential realtime violations in gripper stall timeout mechanism.

closes #1853

### DIFF
--- a/parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
+++ b/parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
@@ -385,7 +385,7 @@ controller_interface::CallbackReturn GripperActionController::on_activate(
   pre_alloc_result_->reached_goal = false;
   pre_alloc_result_->stalled = false;
 
-  last_movement_time_.try_set(rclcpp::Time(0, 0, RCL_ROS_TIME));
+  last_movement_time_.try_set(rclcpp::Time(0, 0, RCL_CLOCK_UNINITIALIZED));
 
   // Action interface
   action_server_ = rclcpp_action::create_server<control_msgs::action::ParallelGripperCommand>(


parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
```diff
@@ -378,6 +385,9 @@ controller_interface::CallbackReturn GripperActionController::on_activate(
   pre_alloc_result_->reached_goal = false;
   pre_alloc_result_->stalled = false;
 
+  // Initialize last movement time
+  last_movement_time_.set(rclcpp::Time(0, 0, RCL_ROS_TIME));
```
christophfroehlich: on_activate is in the RT loop, we should use try_set here, too.
maybe we have to think what happens if try_set fails.

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
```diff
@@ -237,27 +237,34 @@ void GripperActionController::check_for_success(
...
+          RCLCPP_DEBUG(
+            get_node()->get_logger(), "Stall detected moving to goal. Returning success.");
+          active_goal->setSucceeded(pre_alloc_result_);
+        }
+        else
+        {
+          RCLCPP_DEBUG(get_node()->get_logger(), "Stall detected moving to goal. Aborting action!");
+          active_goal->setAborted(pre_alloc_result_);
+        }
+        rt_active_goal_.set([](RealtimeGoalHandlePtr & stored_value)
```
christophfroehlich: not related to the original PR, but rt_active_goal_ also uses non-RT get and set?

christophfroehlich (changes requested):
Thanks for working on this, some comments below

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
```diff
@@ -237,27 +237,34 @@ void GripperActionController::check_for_success(
   {
     if (fabs(current_velocity) > params_.stall_velocity_threshold)
     {
-      last_movement_time_ = time;
+      last_movement_time_.try_set(time);
```
saikishor, line 240: Should we use `set` here? @christophfroehlich

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
```diff
@@ -237,27 +237,34 @@ void GripperActionController::check_for_success(
...
-      pre_alloc_result_->reached_goal = false;
-      pre_alloc_result_->stalled = true;
-
-      if (params_.allow_stalling)
-      {
-        RCLCPP_DEBUG(get_node()->get_logger(), "Stall detected moving to goal. Returning success.");
-        active_goal->setSucceeded(pre_alloc_result_);
-      }
-      else
+      auto last_time_opt = last_movement_time_.try_get();
```
saikishor, line 244: `get` here too?

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
ashwinsushil: Oh I see, I was under the impression that `on_activate` runs in non-RT and only `update()` is RT. I will change this.

If `last_movement_time_.try_set()`  fails during initialization in `on_activate()`, it's a serious issue that should be handled properly. I propose returning `controller_interface::CallbackReturn::ERROR`. What are your thoughts @christophfroehlich ?

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
ashwinsushil: I'm happy you brought this up. I think, `rt_active_goal_` should use `try_get()` and `try_set() `as well, I was not sure why it uses `set()` and `get()`.

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
ashwinsushil: `check_for_success` is called inside the RT, so I was thinking it should be `try_set()`. Correct me if I'm wrong please.

christophfroehlich: > `check_for_success` is called inside the RT, so I was thinking it should be `try_set()`. 

I think so, too.

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
christophfroehlich: the try_set will only fail if the non-RT locked it. This means that it will be updated from the callback, right?

ashwinsushil: Of course, makes sense!  Fixed in 8292c8084f29b4c68c07bc407c8e487a62e6d9eb

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
ashwinsushil: Fixed in c373c15816e39c77ad5ebb1d299b4d1ecc374d0b.

Still not sure, what behavior we should have if try_set() or try_get() fails here. :/

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
```diff
@@ -216,7 +216,7 @@ void GripperActionController::check_for_success(
   double current_velocity)
 {
   RealtimeGoalHandlePtr active_goal;
-  rt_active_goal_.get([&](const RealtimeGoalHandlePtr & goal) { active_goal = goal; });
+  rt_active_goal_.try_get([&](const RealtimeGoalHandlePtr & goal) { active_goal = goal; });
   if (!active_goal)
```
christophfroehlich, line 220:
```suggestion
  const bool got_goal =
    rt_active_goal_.try_get([&](const RealtimeGoalHandlePtr & goal) { active_goal = goal; });
  if (!got_goal || !active_goal)
```
something like this. but if the try_set fails, we can't do much about it. it will just be re-evaluated at the next update call.

ashwinsushil: Would you mind if we talked this through a bit please?

When `rt_active_goal_.try_get([&](const RealtimeGoalHandlePtr & goal) { active_goal = goal; }); `returns false, it means the operation failed to acquire the lock on the realtime-safe container without blocking.

In this specific context (the `check_for_success` function), if `try_get` returns false, the `active_goal` variable remains uninitialized (or rather, it remains as a default-constructed RealtimeGoalHandlePtr, which would be a null pointer).

The code correctly handles this scenario with the following check:
```
if (!active_goal)
{
  return;
}
```

Now...

I'm more worried if the `try_set` fails, the goal remains active in `rt_active_goal_`, and on the next control cycle, `check_for_success()` will be called again with the same goal that was already marked as succeeded/aborted. This could lead to duplicate completion notifications - calling setSucceeded/setAborted multiple times on the same goal.

Then I was thinking maybe only call `setSucceeded/setAborted` only when the try_set succeeds:
```
  if (fabs(error_position) < params_.goal_tolerance)
  {
    pre_alloc_result_->state.effort[0] = computed_command_;
    pre_alloc_result_->state.position[0] = current_position;
    pre_alloc_result_->reached_goal = true;
    pre_alloc_result_->stalled = false;
    
    // Only complete the goal if we can also clear it atomically
    if (rt_active_goal_.try_set([](RealtimeGoalHandlePtr & stored_value)
                                { stored_value = RealtimeGoalHandlePtr(); }))
    {
      RCLCPP_DEBUG(get_node()->get_logger(), "Successfully moved to goal.");
      active_goal->setSucceeded(pre_alloc_result_);
    }
    else
    {
      RCLCPP_DEBUG(get_node()->get_logger(), "Goal reached but cleanup failed, will retry next cycle");
    }
  }
```

But the problem of course here is that we are delaying goal completion notification until the `try_set()` succeeds.

christophfroehlich: > Would you mind if we talked this through a bit please?
> 
> When `rt_active_goal_.try_get([&](const RealtimeGoalHandlePtr & goal) { active_goal = goal; }); `returns false, it means the operation failed to acquire the lock on the realtime-safe container without blocking.
> 
> In this specific context (the `check_for_success` function), if `try_get` returns false, the `active_goal` variable remains uninitialized (or rather, it remains as a default-constructed RealtimeGoalHandlePtr, which would be a null pointer).
> 
> The code correctly handles this scenario with the following check:
> 
> ```
> if (!active_goal)
> {
>   return;
> }
> ```

You are right, I haven't thought about the goal being a shared ptr the last time.

> Now...
> 
> I'm more worried if the `try_set` fails, the goal remains active in `rt_active_goal_`, and on the next control cycle, `check_for_success()` will be called again with the same goal that was already marked as succeeded/aborted. This could lead to duplicate completion notifications - calling setSucceeded/setAborted multiple times on the same goal.
> 
> Then I was thinking maybe only call `setSucceeded/setAborted` only when the try_set succeeds:
> 
> ```
>   if (fabs(error_position) < params_.goal_tolerance)
>   {
>     pre_alloc_result_->state.effort[0] = computed_command_;
>     pre_alloc_result_->state.position[0] = current_position;
>     pre_alloc_result_->reached_goal = true;
>     pre_alloc_result_->stalled = false;
>     
>     // Only complete the goal if we can also clear it atomically
>     if (rt_active_goal_.try_set([](RealtimeGoalHandlePtr & stored_value)
>                                 { stored_value = RealtimeGoalHandlePtr(); }))
>     {
>       RCLCPP_DEBUG(get_node()->get_logger(), "Successfully moved to goal.");
>       active_goal->setSucceeded(pre_alloc_result_);
>     }
>     else
>     {
>       RCLCPP_DEBUG(get_node()->get_logger(), "Goal reached but cleanup failed, will retry next cycle");
>     }
>   }
> ```
> 
> But the problem of course here is that we are delaying goal completion notification until the `try_set()` succeeds.

In which cases try_set will fail? When any other blocking set/get command is imminent. This means that if we cancelled/aborted the current goal in check_for_success, then it is very likely that the content of rt_active_goal_ will be different in the next update call? This would mean that a failing try_set is fine here?

ashwinsushil: > In which cases try_set will fail? When any other blocking set/get command is imminent. This means that if we cancelled/aborted the current goal in check_for_success, then it is very likely that the content of rt_active_goal_ will be different in the next update call? This would mean that a failing try_set is fine here?

Oh yes! You have a point, I din't think it through. Yes, failing `try_set` is fine then. I don't think we need to do anything more. What do you think?

If all is well, all the comments can be resolved?

christophfroehlich (dismissed):
LGTM

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
```diff
@@ -378,6 +385,8 @@ controller_interface::CallbackReturn GripperActionController::on_activate(
   pre_alloc_result_->reached_goal = false;
   pre_alloc_result_->stalled = false;
 
+  last_movement_time_.try_set(rclcpp::Time(0, 0, RCL_ROS_TIME));
```
saikishor:
```suggestion
  last_movement_time_.try_set(rclcpp::Time(0, 0, RCL_CLOCK_UNINITIALIZED));
```

Sorry, it's just this one. Initializing with ROS time is still a valid time, better set an invalid type

saikishor (approved):
Rest looks good to me

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
ashwinsushil: @saikishor Of course! No worries. I'll commit this suggestion real quick.

May I ask what’s the main reason for this change? You don’t want the time to be valid in `on_activate()`? It doesn’t really matter here since in `accepted_callback()` we set it with `last_movement_time_.set(get_node()->now());`.

saikishor: We never know what can happen, it's better to initialize with invalid clock 😅

saikishor (approved):

christophfroehlich (approved):

PR #1857, opened 2025-08-13
Refactoring to motion_primitives_base_controller to support PR #1858

mathias31415:
This pull request supports PR #1858 . Shared functionality of the motion_primitives forward and from_trajectory controllers has been moved to a common base controller. To make reviewing easier, this first PR only extracts the functionality of the forward controller. In a subsequent PR, the from_trajectory controller will then be added.

This PR is marked as a draft because it depends on PR #1636 , which has not yet been merged.


motion_primitives_controllers/CMakeLists.txt
```diff
@@ -46,38 +47,43 @@ target_link_libraries(motion_primitives_forward_controller PUBLIC
   ${std_srvs_TARGETS}
 )
 
-target_compile_definitions(motion_primitives_forward_controller PRIVATE "MOTION_PRIMITIVES_FORWARD_CONTROLLER_BUILDING_DLL")
+target_compile_definitions(${PROJECT_NAME} PRIVATE "MOTION_PRIMITIVES_CONTROLLERS_BUILDING_DLL")
```
christophfroehlich:
```suggestion
```
we can remove that completely, as we don't have the visibility_control file any more.

motion_primitives_controllers/CMakeLists.txt
```diff
@@ -46,38 +47,43 @@ target_link_libraries(motion_primitives_forward_controller PUBLIC
...
-  ${THIS_PACKAGE_INCLUDE_DEPENDS}
+
+install(FILES controller_plugins.xml
+  DESTINATION share/${PROJECT_NAME}
 )
+
+ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS})
+ament_export_include_directories(include)
+ament_export_libraries(${PROJECT_NAME})
+
```
christophfroehlich:
```suggestion
```
a duplicate

motion_primitives_controllers/CMakeLists.txt
```diff
@@ -46,38 +47,43 @@ target_link_libraries(motion_primitives_forward_controller PUBLIC
...
-  include
+install(DIRECTORY include/
+  DESTINATION include
 )
-ament_export_dependencies(
-  ${THIS_PACKAGE_INCLUDE_DEPENDS}
+
+install(FILES controller_plugins.xml
+  DESTINATION share/${PROJECT_NAME}
 )
```
christophfroehlich:
```suggestion
```
Do we really need this? Afaik we don't export this manually in other packages, and `pluginlib_export_plugin_description_file` should do this for us?

motion_primitives_controllers/src/motion_primitives_forward_controller.cpp
```diff
@@ -159,7 +98,7 @@ controller_interface::return_type MotionPrimitivesForwardController::update(
   {
     case ExecutionState::IDLE:
       print_error_once_ = true;
-      was_executing_ = false;
+      // was_executing_ = false;
```
christophfroehlich: why this change?

mathias31415: The action was sometimes not properly finished with SUCCESSFUL, because was_executing_ was reset too early.
When transitioning from EXECUTING → SUCCESS, the action should end with SUCCESSFUL, which is why the flag was_executing_ is checked here.
The flag should only be reset if the action is canceled (STOPPED) or in case of an error (ERROR).

However, I should remove this completely and not just comment it out.

christophfroehlich (approved):

christophfroehlich:
I merge this without another review as it does not change the behavior but only refactors some stuff.

christophfroehlich:
@mergifyio backport mergify/bp/jazzy/pr-1636

PR #1866, opened 2025-08-18
chore: tf2_ros to hpp headers

Timple:


Timple:
I was unaware of that styleguide! I thought "" was prefered for headers included in the package itself and <> for 'external' includes.

This was just a (too lazy) regex.

christophfroehlich:
> I was unaware of that styleguide! I thought "" was prefered for headers included in the package itself and <> for 'external' includes.

no worries :) 

> Headers should only be included using an angle-bracketed path if the library requires you to do so.

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

christophfroehlich (approved):

PR #1888, opened 2025-09-05
Add Battery State Broadcaster controller

YaraShahin:
This Pull Request introduces the **Battery State Broadcaster**, a controller for publishing battery status information in ROS2. The broadcaster reads battery-related state interfaces from hardware and exposes them in standardized ROS messages for easy integration with monitoring tools, logging systems, and higher-level decision-making nodes.

Dependency: This PR depends on [control_msgs#250](https://github.com/ros-controls/control_msgs/pull/250) which introduces the BatteryStates message.

## Features

- Aggregated and Raw Outputs:

    - Publishes a combined `sensor_msgs::msg::BatteryState` message representing the overall system status.

    - Publishes per-joint `control_msgs::msg::BatteryStates` messages containing raw values.

- Flexible Interface Support: Reads from interfaces such as `battery_voltage`, `battery_current`, `battery_temperature`, `battery_charge`, `battery_percentage`, and others.

- Parameterization: Configurable through YAML using [generate_parameter_library](https://github.com/PickNikRobotics/generate_parameter_library?utm_source=chatgpt.com).

## Interfaces

Published Topics

- `~/battery_state` (`sensor_msgs::msg::BatteryState`) — aggregated battery status across all configured joints.

- `~/raw_battery_states` (`control_msgs::msg::BatteryStates`) — raw per-joint battery state values.

Contributions via pull requests are much appreciated. Before sending us a pull request, please ensure that:

1. Limited scope. Your PR should do one thing or one set of things. Avoid adding “random fixes” to PRs. Put those on separate PRs.
2. Give your PR a descriptive title. Add a short summary, if required.
3. Make sure the pipeline is green.
4. Don’t be afraid to request reviews from maintainers.
5. New code = new tests. If you are adding new functionality, always make sure to add some tests exercising the code and serving as live documentation of your original intention.

- [x] Fork the repository.


battery_state_broadcaster/doc/userdoc.rst
```diff
@@ -0,0 +1,88 @@
+:github_url: https://github.com/ros-controls/ros2_controllers/blob/{REPOS_FILE_BRANCH}/battery_state_broadcaster/doc/userdoc.rst
+
+.. _battery_state_broadcaster_userdoc:
+
+Battery State Broadcaster
+--------------------------------
+The *Battery State Broadcaster* is a ROS 2 controller that publishes battery status information as ``sensor_msgs/msg/BatteryState`` messages.
```
bmagyar:
```suggestion
The *Battery State Broadcaster* publishes battery status information as ``sensor_msgs/msg/BatteryState`` messages.
```

christophfroehlich (changes requested):
Thank you for your contribution.
Unfortunately, there is already an existing package with this name.
https://index.ros.org/p/battery_state_broadcaster/#rolling

Can you please highlight the difference between yours and the existing one? If there is a benefit, we could either rename yours or ask the maintainers/authors of the original one to move it to our repository and merge your additions on top. As an alternative, you could also open a PR there.

battery_state_broadcaster/package.xml
```diff
@@ -0,0 +1,41 @@
...
+  <description>ros2 control battery state broadcaster controller</description>
+  <author email="yara.shahin@b-robotized.com">Yara Shahin</author>
+
+  <maintainer email="denis.stogl@b-robotized.com">Dr. Denis Štogl</maintainer>
+
+  <license>Apache License 2.0</license>
+
+  <url type="website">https://control.ros.org</url>
+  <url type="bugtracker">https://github.com/ros-controls/ros2_controllers/issues</url>
+  <url type="repository">https://github.com/ros-controls/ros2_controllers/</url>
```
bmagyar: please use the usual set of maintainers here from other packages

battery_state_broadcaster/package.xml
```diff
@@ -0,0 +1,41 @@
...
+  <depend>rclcpp</depend>
+  <depend>controller_interface</depend>
+
+  <exec_depend>rosidl_default_runtime</exec_depend>
+  <exec_depend>sensor_msgs</exec_depend>
+  <exec_depend>control_msgs</exec_depend>
+
+  <test_depend>ament_cmake_gmock</test_depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
```
bmagyar:
```suggestion
```
strict no to undeclared linters and linters in the cmake in general ;)

battery_state_broadcaster/package.xml
```diff
@@ -0,0 +1,41 @@
...
+  <url type="website">https://control.ros.org</url>
+  <url type="bugtracker">https://github.com/ros-controls/ros2_controllers/issues</url>
+  <url type="repository">https://github.com/ros-controls/ros2_controllers/</url>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <build_depend>ros2_control_cmake</build_depend>
+  <build_depend>rosidl_default_generators</build_depend>
+  <build_depend>sensor_msgs</build_depend>
+  <build_depend>control_msgs</build_depend>
```
bmagyar: anything you find in "build_depend" and also "exec_depend", it could be shortened to just a single "depend" entry

battery_state_broadcaster/src/battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,442 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "controller_interface/helpers.hpp"
+
+namespace battery_state_broadcaster
+{
+const auto kUninitializedValue = std::numeric_limits<double>::quiet_NaN();
+const size_t MAX_LENGTH = 64;
```
bmagyar:
```suggestion
const size_t MAX_LENGTH = 64;  // maximum length of strings to reserve
```
it's generally recommended to document magic numbers if the name isn't fully explanatory

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
```diff
@@ -0,0 +1,137 @@
...
+#include "sensor_msgs/msg/battery_state.hpp"
+
+namespace battery_state_broadcaster
+{
+// // name constants for state interfaces
+// static constexpr size_t STATE_MY_ITFS = 0;
+
+// // name constants for command interfaces
+// static constexpr size_t CMD_MY_ITFS = 0;
+
```
bmagyar:
```suggestion
```

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
```diff
@@ -0,0 +1,137 @@
...
+    const rclcpp_lifecycle::State & previous_state) override;
+
+  controller_interface::CallbackReturn on_deactivate(
+    const rclcpp_lifecycle::State & previous_state) override;
+
+  controller_interface::return_type update(
+    const rclcpp::Time & time, const rclcpp::Duration & period) override;
+
+  float get_or_nan(int interface_cnt);
+  char get_or_unknown(int interface_cnt);
```
bmagyar: please make sure that the return type from these align with where and how you use them. Internally we use doubles for a lot of stuff, using `float` may cause a narrowing conversion.

for char, I see you used `uint8_t` below. If that is what you are looking for, I suggest using that here as return type too. 

Also please double check if the windows compilation (probably on the Ci is enough) is fine with the `uint8_t` type or maybe you need to include stdint.h

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
```diff
@@ -0,0 +1,137 @@
...
+    float current_cnt = 0.0f;
+    float percentage_cnt = 0.0f;
+  };
+
+  BatteryInterfaceSums sums_;
+  BatteryInterfaceCounts counts_;
+
+  std::vector<bool> battery_presence_;
+
+private:
```
bmagyar, line 115: I bet some of the above could be moved to the private section ;)

battery_state_broadcaster/src/battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,442 @@
...
+            raw_battery_states_msg.battery_states[i].power_supply_health;
+        }
+        interface_cnt++;
+      }
+      if (interfaces.battery_present)
+      {
+        auto opt = state_interfaces_[interface_cnt].get_optional<bool>();
+        if (opt.has_value())
+        {
+          raw_battery_states_msg.battery_states[i].present = static_cast<bool>(*opt);
```
bmagyar: you shouldn't need to cast if you used the `value()` accessor, it returns the right type

```suggestion
          raw_battery_states_msg.battery_states[i].present = opt->value();
```

battery_state_broadcaster/src/battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+  return controller_interface::return_type::OK;
+}
+
+float BatteryStateBroadcaster::get_or_nan(int interface_cnt)
+{
+  auto opt = state_interfaces_[interface_cnt].get_optional<double>();
+  if (opt.has_value())
+  {
+    return static_cast<float>(*opt);
```
bmagyar:
```suggestion
    return opt->value();
```

I won't highlight all of them, please fix the rest of the static_casts

battery_state_broadcaster/src/battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,442 @@
...
+}
+
+char BatteryStateBroadcaster::get_or_unknown(int interface_cnt)
+{
+  auto opt = state_interfaces_[interface_cnt].get_optional<double>();
+  if (opt.has_value())
+  {
+    return static_cast<char>(*opt);
+  }
+  return 0;
```
bmagyar: I'd think a little more carefully about this. What does `return 0;` mean in the scope of a `char`? I understand how it works in the number-crunching way of computers and bytes but what does it tell us from a semantic perspective? How do you want to handle an error from here vs an actual 0 value?

bmagyar (changes requested):
A few small cleanup steps, otherwise looks good!

YaraShahin:
> Thank you for your contribution. Unfortunately, there is already an existing package with this name. https://index.ros.org/p/battery_state_broadcaster/#rolling
> 
> Can you please highlight the difference between yours and the existing one? If there is a benefit, we could either rename yours or ask the maintainers/authors of the original one to move it to our repository and merge your additions on top. As an alternative, you could also open a PR there.

Thank you @christophfroehlich for your feedback. I’ve reached out to the maintainer of the original `battery_state_broadcaster` package by email to discuss whether they intend to release it into `ros2_controllers` and how we could best align efforts.

In the meantime, I’ve prepared a summary of the differences and benefits of this implementation, as follows:

- **Full `sensor_msgs::msg::BatteryState` coverage**:
    The existing broadcaster is currently focused on publishing the voltage interface with parameters for `power_supply_technology` and `design_capacity`. This version extends by also broadcasting additional `BatteryState` interfaces (current, charge, temperature, percentage, `power_supply_status`, `power_supply_health`, and presence), while filling the remaining fields via parameters.

- **Flexible parameterization**: Provides per-joint parameters to enable or disable specific battery interfaces, making it adaptable to different hardware setups. It also supports parameters for fixed battery properties.

- **Multi-battery support**: While the existing package is designed around a single battery, this version broadens usage to multi-battery systems, publishing both per-battery states and an aggregated battery state topic (`control_msgs::msg::BatteryStates`).

- **Battery analytics**: 
    Adds analytics features on top of the basic fields, including:
    - Percentage calculation (if no direct percentage interface is available) using voltage and min/max parameters.
    - Presence detection (via `battery_present` interface or inferred from voltage).
    - Aggregation across multiple batteries (averages, sums, and combined status/health).

- **Consistency with `ros2_controllers` style**: Follows established controller/broadcaster structure, and documented best practices.

- **Test coverage**: Includes unit tests aligned with the `ros2_controllers` testing style, ensuring reliability.

ottojo:
Hi! Creator of https://github.com/ipa320/ros_battery_monitoring here! As already realized my package is pretty bare bones, and admittedly not in active use by us at the moment (in a previous robot we only had battery state information from the motor controller - now we have a dedicated package receiving CAN messages from the battery and publishing that as BatteryState).
There is a PR https://github.com/ipa320/ros_battery_monitoring/pull/3 still open which adds more functionality to my package, but i have not had the time/priority to look at it. So i think leaving the entire topic to the ros2_control project seems reasonable.

I have only briefly skimmed over this PR, a technical question that i still have is: Is the battery state now coupled to joints? IIRC, i had one hardware interface plugin which implemented both the joint interfaces and the battery data interfaces as "sensor" tags beside the "joint"s in the XML. But maybe my setup is the one out of the ordinary here...

Anyways, i myself don't have an issue giving the battery_state_broadcaster package name to the ros2_control project, if thats the question, although i do wonder if anybody is using my package and will have their battery monitoring unexpectedly broken...

christophfroehlich:
> Anyways, i myself don't have an issue giving the battery_state_broadcaster package name to the ros2_control project, if thats the question, although i do wonder if anybody is using my package and will have their battery monitoring unexpectedly broken...

Thanks for that offer. According to index.ros.org there is no released package dependent on it at least. I haven't checked if the new proposed broadcaster can be rewritten to support the existing broadcaster config? Maybe @YaraShahin can evaluate that. If this does not make sense, we can break it on rolling but leave the other distros as they are now.

YaraShahin:
> I have only briefly skimmed over this PR, a technical question that i still have is: Is the battery state now coupled to joints? IIRC, i had one hardware interface plugin which implemented both the joint interfaces and the battery data interfaces as "sensor" tags beside the "joint"s in the XML. But maybe my setup is the one out of the ordinary here...

Thanks a lot @ottojo for your feedback 🙏

To your question: in our implementation the battery state is indeed exposed via state interfaces that are grouped under state_joints. This keeps it consistent with other broadcasters in ros2_control (IMU, force-torque, etc.), but it does mean batteries are defined alongside joints in the URDF. If I understand correctly, your case should still be supported, since you could list your sensor_name under state_joints parameter. The state interface exposed is the same in both cases, for example: `<sensor-or-joint-name>/voltage`.

I’ll also take a look at the differences between our current implementation and the PR you mentioned in your repo, to see if there’s anything we should carry over here.

Thanks again for the clarification and for being open to handing over the package name — we’ll make sure to handle the transition carefully so existing users aren’t left behind.

YaraShahin:
> Thanks for that offer. According to index.ros.org there is no released package dependent on it at least. I haven't checked if the new proposed broadcaster can be rewritten to support the existing broadcaster config? Maybe @YaraShahin can evaluate that. If this does not make sense, we can break it on rolling but leave the other distros as they are now.

Thanks @christophfroehlich for checking the dependency situation. I’ll review whether the configuration style from `ipa320/ros_battery_monitoring` can be supported directly. If keeping full compatibility doesn’t make sense, then I’ll add a migration note in the docs for Rolling so existing users can adapt their configs.

christophfroehlich:
@YaraShahin any updates?

YaraShahin:
> @YaraShahin any updates?

Thanks for the reminder, I’ll push the fixes this week.

battery_state_broadcaster/package.xml
YaraShahin: Done

battery_state_broadcaster/package.xml
YaraShahin: Done

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
YaraShahin: Thank you for pointing this out. I mainly have to do these conversions to match the types of the [BatteryState msg](https://github.com/ros2/common_interfaces/blob/rolling/sensor_msgs/msg/BatteryState.msg) which uses `float` and `int8`.

Regarding your suggestions:

- For `get_or_nan`, I now use `value_or` with a cast to `float`, so the internal double stays until assignment to the message field:
```
static_cast<float>(state_interfaces_[interface_cnt].get_optional<double>().value_or(kUninitializedValue));
```

- For `get_or_unknown`, I cast to `uint8_t` instead of `char`:
```
static_cast<uint8_t>(state_interfaces_[interface_cnt].get_optional<double>().value_or(
            sensor_msgs::msg::BatteryState::POWER_SUPPLY_STATUS_UNKNOWN));
```

- Replaced all `char` by `uint8_t`

- Since other broadcasters are also using uint8_t, I believe it should be fine. I added the `<cstdint>` anyways for safety.

This should keep the double type used by ros2 control till the end where it has to be assigned to the float/int8 message field.

battery_state_broadcaster/src/battery_state_broadcaster.cpp
YaraShahin: For all the interfaces in which I was using this function, 0 was the enum value for UNKNOWN. So, that's its semantic value as well in case of an error.

After your recent comment, I changed this to 

```
static_cast<uint8_t>(state_interfaces_[interface_cnt].get_optional<double>().value_or(
            sensor_msgs::msg::BatteryState::POWER_SUPPLY_STATUS_UNKNOWN));
```
which illustrates this concept better.

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
YaraShahin: I just moved what I could to private; unfortunately I am using some of them in test so they have to stay protected.

battery_state_broadcaster/src/battery_state_broadcaster.cpp
YaraShahin: Fixed as mentioned in [the comment above](https://github.com/ros-controls/ros2_controllers/pull/1888#discussion_r2393242278). Thanks for the tip!

YaraShahin:
> A few small cleanup steps, otherwise looks good!

Thanks for the feedback! I’ve fixed the points you mentioned.

YaraShahin:
Hi @christophfroehlich @ottojo,

As per our conversation, I have updated and tested the code with the following changes:

- The broadcaster now fully supports the legacy parameters from your package. You can specify `sensor_name` instead of `state_joints`. In this mode, users get the full behavior of your original implementation (same topic name, same required hardware interface `sensor_name/voltage`), but naturally only a subset of the new implementation’s extended features.

- A migration guide was added to the docs, including examples for users who wish to migrate to the new configuration using `state_joints`.

- I also reviewed [the PR you mentioned](https://github.com/ipa320/ros_battery_monitoring/pull/3) and didn’t find any gaps compared to this implementation.

Please let me know if you have any feedback on these changes. Thanks a lot for your patience!

YaraShahin:
Moved the `BatteryStates` message to `sensor_msgs` as discussed with @saikishor in [control_msgs#250](https://github.com/ros-controls/control_msgs/pull/250).

This PR now depends on ros2/common_interfaces#302

christophfroehlich:
> Hi @christophfroehlich @ottojo,
> 
> As per our conversation, I have updated and tested the code with the following changes:
> 
>     * The broadcaster now fully supports the legacy parameters from your package. You can specify `sensor_name` instead of `state_joints`. In this mode, users get the full behavior of your original implementation (same topic name, same required hardware interface `sensor_name/voltage`), but naturally only a subset of the new implementation’s extended features.
> 
>     * A migration guide was added to the docs, including examples for users who wish to migrate to the new configuration using `state_joints`.
> 
>     * I also reviewed [the PR you mentioned](https://github.com/ipa320/ros_battery_monitoring/pull/3) and didn’t find any gaps compared to this implementation.
> 
> 
> Please let me know if you have any feedback on these changes. Thanks a lot for your patience!

It remains unclear how we proceed with the new message types (and how long that will take). What do you think, should we first move the original implementation under the ros2_controllers repo and bring that through rosdistro? then we can more easily review the support of the legacy parameters.

YaraShahin:
> It remains unclear how we proceed with the new message types (and how long that will take)

Since it's unclear when (or if) the BatteryStateArray will be merged into sensor_msgs, I could revert back to the original approach of adding it to control_msgs as proposed in ros-controls/control_msgs#250. As @saikishor mentioned in that PR, this is the appropriate fallback if the sensor_msgs timeline is uncertain.

> What do you think, should we first move the original implementation under the ros2_controllers repo and bring that through rosdistro? then we can more easily review the support of the legacy parameters.

I understand the concern about timing. My preference would be to move forward with the complete implementation if that works for you, since it already has support for legacy parameters tested and documented, and I believe others might benefit from the extended functionality as well. However, I'm happy to follow whichever approach you think is best for the project.

If we proceed with the complete implementation, would it be possible to reopen ros-controls/control_msgs#250? That would allow us to move forward without blocking on the sensor_msgs uncertainty.

christophfroehlich:
> > What do you think, should we first move the original implementation under the ros2_controllers repo and bring that through rosdistro? then we can more easily review the support of the legacy parameters.
> 
> I understand the concern about timing. My preference would be to move forward with the complete implementation if that works for you, since it already has support for legacy parameters tested and documented, and I believe others might benefit from the extended functionality as well. However, I'm happy to follow whichever approach you think is best for the project.

But it would be easier for us to review if the legacy behavior remains unchanged if the current state would exist already in this repo.

YaraShahin:
> But it would be easier for us to review if the legacy behavior remains unchanged if the current state would exist already in this repo.

Sure, I understand. My concern is that since I didn't work on the original ipa320 implementation, I'm not sure I'm the right person to move it over.

Would it make more sense for @ottojo to submit the original if he is available, or do you want me to give it a shot? Either way, I can then follow up with the extended features on top.
What do you think?

christophfroehlich:
We need his approval if we remove his repository from rosdistro, that process will take a while anyways. Who is opening the PR in ros2_controllers does not matter that much and should not be a lot of work.

destogl:
> We need his approval if we remove his repository from rosdistro, that process will take a while anyways. Who is opening the PR in ros2_controllers does not matter that much and should not be a lot of work.

@christophfroehlich only concern here is bringing the previous package to the structure and patterns of ros2_control. This might take time, and the question is, do we want to invest in it? Another approach might be to release this package to the versions that the old package is not released.

christophfroehlich:
> > We need his approval if we remove his repository from rosdistro, that process will take a while anyways. Who is opening the PR in ros2_controllers does not matter that much and should not be a lot of work.
> 
> @christophfroehlich only concern here is bringing the previous package to the structure and patterns of ros2_control. This might take time, and the question is, do we want to invest in it? Another approach might be to release this package to the versions that the old package is not released.

It is already [released](https://index.ros.org/?pkgs=battery_state_broadcaster&search_packages=true#rolling) to jazzy,kilted,rolling. I don't think we have to invest much time for refactoring the current implementation as it is short-living and will be replaced by this PR here. But this was just a suggestion, as you prefer..

christophfroehlich:
As we have decided in todays PMC meeting:

- let's include the current state of the released package first, but don't put too much effort in adding tests (maybe only the simple load_controller?)
- put this PR on top, using the message from https://github.com/ros-controls/control_msgs/pull/250

ottojo:
Hi, i've not been following closely here, but i get that you would like to add a simple battery state broadcaster to ros2_control, and then iterate from there? Is there anything for me to do? I will gladly hand over the package name `battery_state_broadcaster` to ros-controls, but i don't know what the process for that is. Should i prepare a release in our repository, removing the battery_state_broadcaster (there is still battery_state_rviz_overlay in the repo, which can stay there for all i care about), and then at some point you make the rosdistro PRs which at the same time remove the package from my repo and add it to yours?

christophfroehlich:
Thanks for reaching out. Yes this could work out. We have to remove the package from rosdistro before we release it from this repo as bloom-release checks if the package already exists.
I suggest that you release your current state, and we handle the rest. I'll tag you once I've opened the rosdistro PR so that the ROS boss know that we have your approval.

christophfroehlich:
@YaraShahin can you please rebase your changes on top of https://github.com/ros-controls/ros2_controllers/pull/2086 and update the target branch of this PR? Then we can merge all at once. (maybe its easer to just delete the folder with one commit, and then rebase your branch on top).

YaraShahin:
> @YaraShahin can you please rebase your changes on top of #2086 and update the target branch of this PR? Then we can merge all at once. (maybe its easer to just delete the folder with one commit, and then rebase your branch on top).

Thank you @christophfroehlich for starting the new PR. I added the load_controller test as you suggested before deleting the folder in a separate commit.

battery_state_broadcaster/CHANGELOG.rst
christophfroehlich, line 1: maybe we should leave the legacy version of the CHANGELOG

battery_state_broadcaster/battery_state_broadcaster.xml
christophfroehlich, line 1: we've had this discussion somewhere already: We don't have copyright claims in xml and yaml files (as they are not required by the linters)

battery_state_broadcaster/include/battery_state_broadcaster/BatteryStateBroadcaster.hpp
christophfroehlich, line 1: We may be break API anyways, but as this header is installed: Should we leave it as a shim to the new headerfile, with a deprecation message?

battery_state_broadcaster/package.xml
```diff
@@ -2,18 +2,42 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>battery_state_broadcaster</name>
-  <version>1.1.0</version>
-  <description>ROS2 Control boradcaster for battery state sensors.</description>
-  <maintainer email="jonas.otto@ipa.fraunhofer.de">Jonas Otto</maintainer>
-  <license>MIT</license>
+  <version>0.1.0</version>
+  <description>ros2 control battery state broadcaster controller</description>
```
christophfroehlich:
```suggestion
  <description>ros2_control battery state broadcaster controller</description>
```

battery_state_broadcaster/src/battery_state_broadcaster.yaml
```diff
@@ -0,0 +1,114 @@
...
+  design_capacity: {
+    type: double,
+    default_value: .nan,
+    description: "Design capacity of the battery [Ah] for the sensor_name mode (If unmeasured NaN).",
+    read_only: true,
+  }
+  power_supply_technology: {
+    type: int,
+    default_value: 0,
+    description: "Battery chemistry type as an enum for the sensor_name mode. If unmeasured, the technology is set to unknown.",
```
christophfroehlich: maybe a link to the message definition helps here.

battery_state_broadcaster/src/battery_state_broadcaster_parameters.yaml
```diff
@@ -0,0 +1,114 @@
...
+  }
+  power_supply_technology: {
+    type: int,
+    default_value: 0,
+    description: "Battery chemistry type as an enum for the sensor_name mode. If unmeasured, the technology is set to unknown.",
+    read_only: true,
+    validation: {
+      bounds<>: [0, 8]
+    }
+  }
```
christophfroehlich, line 114: maybe we should mark them already as deprecated?

battery_state_broadcaster/doc/userdoc.rst
```diff
@@ -3,53 +3,123 @@
 .. _battery_state_broadcaster_userdoc:
 
 Battery State Broadcaster
+--------------------------------
+The *Battery State Broadcaster* publishes battery status information as ``sensor_msgs/msg/BatteryState`` messages.
+
+It reads battery-related state interfaces from one or more joints and exposes them in a standard ROS 2 message format. This allows easy integration with monitoring tools, logging systems, and higher-level decision-making nodes.
```
christophfroehlich: why "joints"? it would also work with sensor or gpio type?

christophfroehlich (changes requested):
A first quick review. 

Please also update docs/release_notes.rst

battery_state_broadcaster/battery_state_broadcaster.xml
YaraShahin: Yes, thank you; removed from battery_state_broadcaster.xml

battery_state_broadcaster/CHANGELOG.rst
YaraShahin: Done

battery_state_broadcaster/include/battery_state_broadcaster/BatteryStateBroadcaster.hpp
YaraShahin: Yes I agree, I added the header with warning for BatteryStateBroadcaster.hpp

battery_state_broadcaster/src/battery_state_broadcaster.yaml
YaraShahin: Good idea! I added the link to the all the enum fields

battery_state_broadcaster/src/battery_state_broadcaster_parameters.yaml
YaraShahin: Done

battery_state_broadcaster/doc/userdoc.rst
YaraShahin: Good point, technically it could be any hardware component type (joint, sensor,  GPIO). We initially named it state_joints because the primary intended usecase was batteries associated with joints (e.g. wheel motors). I'm open to rename it, maybe to something like **batteries** to make it more semantic and user-friendly, hiding the ros2_control hardware abstraction details.    

What do you think?

YaraShahin:
Hi @christophfroehlich , I've addressed all the comments — sorry for the delay on this one! Please take another look when you get a chance.

battery_state_broadcaster/test/battery_state_broadcaster_params.yaml
```diff
@@ -0,0 +1,38 @@
+test_battery_state_broadcaster:
+  ros__parameters:
+    state_joints:
+      - "left_wheel"
+      - "right_wheel"
+    interfaces:
+      left_wheel:
```
nnarain-cpr: why left_wheel / right_wheel? For a clearer example, maybe just battery0 and battery1.

battery_state_broadcaster/test/battery_state_broadcaster_params.yaml
```diff
@@ -0,0 +1,38 @@
...
+        battery_power_supply_health: true
+        battery_present: false
+    left_wheel:
+      minimum_voltage: 0.0
+      maximum_voltage: 10.0
+      capacity: 12000.0
+      design_capacity: 13000.0
+      power_supply_technology: 3
+      location: "left_slot"
+      serial_number: "left_serial_device"
```
nnarain: This is more of a general question. Have you considered how you might retrieve this from the battery through the hardware interface? From my current understanding, that is not possible. 

Having this information is pretty relevant to our use case because we use the serial number / firmware version of the batteries in our diagnostics and metrics.

nnarain-cpr:
Hello friends. I have a change here https://github.com/b-robotized-forks/ros2_controllers/compare/add-battery-state-broadcaster...clearpathrobotics:ros2_controllers:add-battery-state-broadcaster-cell-voltage?expand=1

that adds voltage and temperature per cell on top of this PR.

not that it has to go into this PR, but I'd like to add this as a follow up

destogl:
> Hello friends. I have a change here https://github.com/b-robotized-forks/ros2_controllers/compare/add-battery-state-broadcaster...clearpathrobotics:ros2_controllers:add-battery-state-broadcaster-cell-voltage?expand=1
> 
> that adds voltage and temperature per cell on top of this PR.
> 
> not that it has to go into this PR, but I'd like to add this as a follow up

we should to this as follow up.

battery_state_broadcaster/doc/userdoc.rst
destogl: Yes, you can do this.

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
destogl: @YaraShahin you can use friending for this to enable testing. We should be using this pattern anyways.

battery_state_broadcaster/doc/userdoc.rst
christophfroehlich: should we consider this before the first release?
This broadcaster also can be used for the usecase of a single battery, right? then the "joints" namespace even makes less sense.

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
christophfroehlich: +1

battery_state_broadcaster/test/battery_state_broadcaster_params.yaml
christophfroehlich: +1, see my comment in the userdoc

battery_state_broadcaster/test/battery_state_broadcaster_params.yaml
christophfroehlich: I understand your usecase. This currently isn't directly possible as we only support handles of numerical type. As this is only interesting during the first startup (or do you consider hot-swapping), we could go with a workaround and exchange data via ROS parameters (hardware components do have a node, and we could query that from the controller) until we support datatypes like this.

battery_state_broadcaster/test/test_battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,309 @@
...
+  params.node_namespace = "";
+  params.node_options = battery_state_broadcaster_->define_custom_node_options();
+  ASSERT_EQ(battery_state_broadcaster_->init(params), controller_interface::return_type::OK);
+
+  ASSERT_EQ(battery_state_broadcaster_->on_configure(rclcpp_lifecycle::State()), NODE_SUCCESS);
+  ASSERT_EQ(battery_state_broadcaster_->on_activate(rclcpp_lifecycle::State()), NODE_FAILURE);
+}
+
+// check all msgs initial values
+// check logic for combined strings for local & serial number and same or none for power supply
```
christophfroehlich:
```suggestion
```

christophfroehlich: or there is a missing test

battery_state_broadcaster/test/test_battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,309 @@
...
+  RawBatteryStatesMsg raw_battery_states_msg;
+  BatteryStateMsg battery_state_msg;
+  subscribe_and_get_messages(raw_battery_states_msg, battery_state_msg);
+
+  ASSERT_EQ(raw_battery_states_msg.battery_states.size(), 2u);
+
+  // Left wheel
+  const auto & left = raw_battery_states_msg.battery_states[0];
+  EXPECT_DOUBLE_EQ(left.voltage, 10.0);
+  // percentage calculated (no interface) = (10.0 - 0.0) * 100 / (10.0 - 0.0) = 50
```
christophfroehlich: ?

battery_state_broadcaster/CHANGELOG.rst
```diff
@@ -2,15 +2,23 @@
 Changelog for package battery_state_broadcaster
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+6.3.0 (2026-xx-xx)
+-------------------
+* Add battery_state_broadcaster to ros2_controllers (`#1888 <https://github.com/ros-controls/ros2_controllers/issues/1888>`_)
+* Contributors: Yara Shahin
```
christophfroehlich:
```suggestion
```

this will be auto-filled by catkin_generate_changelog

battery_state_broadcaster/package.xml
```diff
@@ -2,16 +2,33 @@
...
+  <maintainer email="christoph.froehlich@ait.ac.at">Christoph Froehlich</maintainer>
+  <maintainer email="sai.kishor@pal-robotics.com">Sai Kishor Kothakota</maintainer>
+
+  <license>Apache License 2.0</license>
+
+  <url type="website">https://control.ros.org</url>
+  <url type="bugtracker">https://github.com/ros-controls/ros2_controllers/issues</url>
+  <url type="repository">https://github.com/ros-controls/ros2_controllers/</url>
+
+  <author email="yara.shahin@b-robotized.com">Yara Shahin</author>
```
christophfroehlich, line 19:
```suggestion
  <author email="yara.shahin@b-robotized.com">Yara Shahin</author>
  <author email="jonas.otto@ipa.fraunhofer.de">Jonas Otto</author>
```

christophfroehlich (changes requested):
some comments and still some open conversations

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
YaraShahin: Oh, thanks for the tip. I made `FriendBatteryStateBroadcaster` to be a frined of `BatteryStateBroadcaster`. Then I exposed in `FriendBatteryStateBroadcaster` the private variables that each of the TEST_F classes need, since they are also friends of `FriendBatteryStateBroadcaster` but not direct friends of `BatteryStateBroadcaster`.

battery_state_broadcaster/doc/userdoc.rst
YaraShahin: Renamed state_joints to batteries across all the occurances (parameters, descriptions, and source code).

battery_state_broadcaster/test/battery_state_broadcaster_params.yaml
YaraShahin: Thanks for the suggestion. I renamed accordingly.

battery_state_broadcaster/test/test_battery_state_broadcaster.cpp
YaraShahin: Yeah sorry was wrong formula result there in the comment. I fixed it to 100 instead of 50.

battery_state_broadcaster/test/test_battery_state_broadcaster.cpp
YaraShahin: Removed these comments. The behavior they referred to is already covered by the `publish_status_success` test: it asserts the initial/persistent message values (capacity, design_capacity, location, serial_number, power_supply_technology, frame_id) for each battery, as well as the combined `battery_state` message, including the concatenated location ("slot0, slot1, ") and `serial_number` strings and the combined `power_supply_technology`.

battery_state_broadcaster/include/battery_state_broadcaster/battery_state_broadcaster.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+  struct BatteryInterfaceCounts
+  {
+    float temperature_cnt = 0.0f;
+    float current_cnt = 0.0f;
+    float percentage_cnt = 0.0f;
+  };
+
+private:
+  friend class ::FriendBatteryStateBroadcaster;
```
christophfroehlich: in other controller tests we use a different pattern, expose all necessary members as protected and access it from the test fixture. But if this is what @bmagyar was thinking of, I'm fine with that

christophfroehlich (changes requested):
Thanks for the iteration, parameters etc are perfect IMHO.

Clang job is failing with 
  `
  /home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/src/ros2_controllers/battery_state_broadcaster/src/battery_state_broadcaster.cpp:325:27: error: implicit conversion changes signedness: 'int' to 'size_type' (aka 'unsigned long') [-Werror,-Wsign-conversion]`

amongst others.

YaraShahin:
Thanks @christophfroehlich for following up! I fixed the clang issues & made sure it passed with rolling clang check locally. @destogl or @bmagyar Could you please take a look on the new friending/protected variables pattern and advise me if I need to revert it back? Thanks in advance!

christophfroehlich (approved):
Thanks for your patience.

I applied some changes:

- harmonize the lifecycle transition with other tests
- use a wait set for the subscription tests, the old pattern failed in [CI](https://github.com/ros-controls/ros2_controllers/actions/runs/30029933517/job/89283882389?pr=1888)
- removed the friendclassing for the test, so we have the same pattern as in other packages
- fixed the [failing build](https://github.com/ros-controls/ros2_controllers/actions/runs/30036736961/job/89306855592?pr=1888) of docs
- Update handle API

Regarding Bence's comments on float/int/uint8: IMHO you have addressed them properly, I don't see any remaining issue here now.

PR #1941, opened 2025-10-01
mecanum_drive_controller: Declare missing backward_ros dependency

wentasah:
Since CMakeLists.txt contains "find_package(backward_ros REQUIRED)", without this change, compilation can fail with the following error:

    CMake Error at CMakeLists.txt:25 (find_package):
      By not providing "Findbackward_ros.cmake" in CMAKE_MODULE_PATH this project
      has asked CMake to find a package configuration file provided by
      "backward_ros", but CMake did not find one.

This change fixes that.


christophfroehlich (approved):
Thanks!

PR #1959, opened 2025-10-09
Fix upstream Pid class deprecation warnings

christophfroehlich:
We have a bit of chaos with i clamping on jazzy

in JTC we have the i_clamp_max/min in the parameter yaml, but don't use it in the code.

I fixed the tests to be working with https://github.com/ros-controls/control_toolbox/pull/507 and silence hundred lines of deprecation warnings.

### DIFF
--- a/pid_controller/src/pid_controller.yaml
+++ b/pid_controller/src/pid_controller.yaml
@@ -85,12 +85,12 @@ pid_controller:
       }
       i_clamp_max: {
         type: double,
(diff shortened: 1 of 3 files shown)


christophfroehlich:
@ViktorCVS can you have a look please if this makes sense?

ViktorCVS:
> @ViktorCVS can you have a look please if this makes sense?

Yes, this makes sense, I support this change.

PR #1971, opened 2025-10-21
Use UnorderedElementsAreArray for JSB tests

### LINKED ISSUE #1123: [CI] JointStateBroadcasterTest fails on RHEL/Humble
what is different from humble to later distros? (RHEL 8 vs RHEL 9!)
Humble fails: https://github.com/ros-controls/ros2_controllers/actions/runs/9001159902/job/24726794017?pr=1121
Iron not: https://github.com/ros-controls/ros2_controllers/actions/runs/8994884125/job/24709018370

christophfroehlich:
Sames fix as in https://github.com/ros-controls/ros2_control/pull/730

Closes #1123

I'm not sure if we should apply this patch to newer distros, because we never had this problem an newer RHEL versions.


bmagyar (approved):
thank you!

PR #1982, opened 2025-11-02
updated to urdf parser from xmldom (rqt_joint_trajectory_controller)

### LINKED ISSUE #891: rqt_joint_trajectory_controller: Use urdf_parser
Maybe we can use the upstream [urdf_parser_py](https://github.com/ros/urdf_parser_py/tree/ros2/)? The mentioned issue got fixed years ago..

https://github.com/ros-controls/ros2_controllers/blob/925a690fcb469cb705ec67185cd67c6f1481dd61/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py#L17-L20

Ashray4:
Update joint-limits parsing to use urdf_parser_py

### DIFF
--- a/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
+++ b/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
@@ -23,7 +23,8 @@
 # upper is an optional attribute, so I don't understand what's going on
 # See comments in https://github.com/ros/urdfdom/issues/36
 
-import xml.dom.minidom
+import xml.etree.ElementTree as ET
+from urdf_parser_py import urdf
 from math import pi
 
 import rclpy
@@ -45,7 +46,7 @@ def subscribe_to_robot_description(node, key="robot_description"):
     node.create_subscription(String, key, callback, qos_profile)
 
 
-def get_joint_limits(node, joints_names, use_smallest_joint_limits=True):
+def get_joint_limits(node, use_smallest_joint_limits=True):
     use_small = use_smallest_joint_limits
     use_mimic = True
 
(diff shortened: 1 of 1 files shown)


rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -59,67 +60,63 @@ def get_joint_limits(node, joints_names, use_smallest_joint_limits=True):
     dependent_joints = {}
 
     if description != "":
-        robot = xml.dom.minidom.parseString(description).getElementsByTagName("robot")[0]
+
+        root = ET.fromstring(description)
+        for rc in root.findall("ros2_control"):
+            root.remove(rc)
+
+        cleaned_description = ET.tostring(root, encoding="unicode")
```
christophfroehlich, line 68: can you comment on why we need to remove this?

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -23,7 +23,8 @@
 # upper is an optional attribute, so I don't understand what's going on
 # See comments in https://github.com/ros/urdfdom/issues/36
```
christophfroehlich, line 24: if fixed, please remove L21-24

christophfroehlich (changes requested):
I have some comments.
Apart from that, I think we easily could write a unit test for this class. At least some sort of smoke test with a valid robot_description from the assets we have.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
Ashray4: When i tested with ros2_control_demos , i got these unknown tags 
```
ros2 run rqt_gui rqt_gui --force-discover -s rqt_joint_trajectory_controller
Waiting for the robot_description!
Waiting for the robot_description!
Waiting for the robot_description!
Unknown tag "ros2_control" in /robot[@name='2dof_robot']
```
also while testing with some robot descriptions with inertial tags as well
```
Unknown attribute "iyx" in /robot[@name='ur5e']/link[@name='robotiq_85_base_link']/inertial/inertia
Unknown attribute "izx" in /robot[@name='ur5e']/link[@name='robotiq_85_base_link']/inertial/inertia
Unknown attribute "izy" in /robot[@name='ur5e']/link[@name='robotiq_85_base_link']/inertial/inertia
```

christophfroehlich: ok it seems that the parser is very strict with that. iyx/izx/izy does not exist in the [spec](https://wiki.ros.org/urdf/XML/link), this really is an issue in the URDF. 

Unless there isn't a possibility to switch to a robust mode in the urdf module, your fix seems to be fine!

Ashray4: i will check if there is a possibility of the robust mode, and also write the test case as you mentioned

christophfroehlich:
supersed by #2254

PR #1984, opened 2025-11-02
Fix integer literal in logging macro

christophfroehlich:
Similar to
https://github.com/ros-controls/ros2_controllers/blob/a496412c5a51eda9bf9552f7a8e3a8fd9815f9e1/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp#L319

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
@@ -319,7 +319,7 @@ inline bool check_state_tolerance_per_joint(
   if (show_errors)
   {
     const auto logger = rclcpp::get_logger("tolerances");
-    RCLCPP_ERROR(logger, "State tolerances failed for joint %lu:", joint_idx);
+    RCLCPP_ERROR(logger, "State tolerances failed for joint %zu:", joint_idx);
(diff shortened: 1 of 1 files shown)


joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -319,7 +319,7 @@ inline bool check_state_tolerance_per_joint(
   if (show_errors)
   {
     const auto logger = rclcpp::get_logger("tolerances");
-    RCLCPP_ERROR(logger, "State tolerances failed for joint %d:", joint_idx);
+    RCLCPP_ERROR(logger, "State tolerances failed for joint %lu:", joint_idx);
```
saikishor:
```suggestion
    RCLCPP_ERROR(logger, "State tolerances failed for joint %zu:", joint_idx);
```

christophfroehlich: but size_t is unsigned long?

saikishor: Ok I updated it, it should be `zu` isn't it?

PR #1820, opened 2025-07-17
[JTC] added time_from_start to action feedback (#1755)

bmagyar:
Supersedes #1756

PR #1985, opened 2025-11-02
Fix JTC state_msg

christophfroehlich:
on humble, I found some issues while trying to fix #1820:

From #1172 

- the msg field of the wrong state_publisher was updated -> change had actually no effect
- velocity+acceleration is not updated, as commented https://github.com/ros-controls/ros2_controllers/pull/1172#discussion_r1643401579

and #1755 

- only `feedback`+`reference` fields got updated, but not the deprecated `actual`/`desired`

Additional change:
`desired_state.time_from_start` and `current_state_.time_from_start` (subsequently, also entries of feedback msg) is now updated with current time as on rolling, see #1191 for the change


MarqRazz (approved):
Changes look good but I'm not using Humble any more so I don't have anything I can test it on.

PR #1992, opened 2025-11-07
Add missing dependency rclcpp_action

Noel215:
Hi,

I got the following error running some tests that use the JTC
```
[ros2_control_node-1] /opt/ros/rolling/lib/controller_manager/ros2_control_node: symbol lookup error: /opt/ros/rolling/lib/libjoint_trajectory_controller.so: undefined symbol: _ZN13rclcpp_action20ServerGoalHandleBase12try_abortingEv
[ERROR] [ros2_control_node-1]: process has died [pid 5247, exit code 127, cmd '/opt/ros/rolling/lib/controller_manager/ros2_control_node --ros-args --params-file /tmp/launch_params_uh9rx9w0 --params-file /builds/app-tools/play_motion2/play_motion2/test/rrbot/controllers.yaml'].
```
It seems that is not linking properly against `rclcpp_action`

I'm adding a missing dependency that is not on the following (active) distributions:

- [rolling](https://github.com/ros-controls/ros2_controllers/blob/master/joint_state_broadcaster/package.xml)
- [kilted](https://github.com/ros-controls/ros2_controllers/blob/kilted/joint_state_broadcaster/package.xml)
- [jazzy](https://github.com/ros-controls/ros2_controllers/blob/jazzy/joint_state_broadcaster/package.xml)
- [humble](https://github.com/ros-controls/ros2_controllers/blob/humble/joint_state_broadcaster/package.xml)

This change should be backported

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -14,6 +14,7 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   hardware_interface
   pluginlib
   rclcpp
+  rclcpp_action
   rclcpp_lifecycle
   realtime_tools
   rsl
@@ -52,6 +53,7 @@ target_link_libraries(joint_trajectory_controller PUBLIC
                       hardware_interface::hardware_interface
                       pluginlib::pluginlib
                       rclcpp::rclcpp
+                      rclcpp_action::rclcpp_action
                       rclcpp_lifecycle::rclcpp_lifecycle
                       realtime_tools::realtime_tools
                       rsl::rsl


saikishor (approved):

christophfroehlich (approved):
what have you done differently, so that we never saw this issue earlier?

Noel215:
Actually nothing. It is a test that it was passing for humble, jazzy and rolling last week.

Some days ago it started failing for rolling because of the API change for hardware interface (see [here](https://github.com/ros-controls/ros2_control/commit/8c6e2e9b0ea5c9dc1c8a0133953f72582eb5d20f)), that got [released last week](https://github.com/ros/rosdistro/pull/48354).

When I fixed the issues for that API change, I got this issue, that it is only failing for rolling.

bmagyar:
Could you please also explicitly include it in the cmake? It seems we still get it cmake-included transitively through something but better be explicit

saikishor (approved):
Thank you Noel

PR #1996, opened 2025-11-08
Rename Odometry Class to SteeringKinematics

kumar-sanjeeev:
This PR includes the following changes:

- Renamed the `Odometry` class to `SteeringKinematics` to better represent its support for both FK and IK.
- Added deprecation warnings in the old `steering_odometry.hpp` header file to maintain backward compatibility.
- Kept the `steering_odometry.cpp` for backward compatibility and redirected actual implementation to `steering_odometry.cpp`
- Updated comments in the header file to reflect the new class name.

### DIFF
--- a/steering_controllers_library/include/steering_controllers_library/steering_odometry.hpp
+++ b/steering_controllers_library/include/steering_controllers_library/steering_odometry.hpp
@@ -60,101 +60,61 @@ inline bool is_close_to_zero(double val) { return steering_kinematics::is_close_
 class [[deprecated("Use steering_kinematics::SteeringKinematics instead")]] SteeringOdometry
 {
 public:
-  explicit SteeringOdometry(size_t velocity_rolling_window_size = 10)
-  : sk_impl_(velocity_rolling_window_size)
-  {
-  }
+  explicit SteeringOdometry(size_t velocity_rolling_window_size = 10);
 
-  void init(const rclcpp::Time & time) { sk_impl_.init(time); }
+  void init(const rclcpp::Time & time);
 
   bool update_from_position(
-    const double traction_wheel_pos, const double steer_pos, const double dt)
-  {
-    sk_impl_.update_from_position(traction_wheel_pos, steer_pos, dt);
-  }
+    const double traction_wheel_pos, const double steer_pos, const double dt);
 
   bool update_from_position(
     const double right_traction_wheel_pos, const double left_traction_wheel_pos,
-    const double steer_pos, const double dt)
-  {
-    sk_impl_.update_from_position(right_traction_wheel_pos, left_traction_wheel_pos, steer_pos, dt);
-  }
+    const double steer_pos, const double dt);
 
   bool update_from_position(
     const double right_traction_wheel_pos, const double left_traction_wheel_pos,
-    const double right_steer_pos, const double left_steer_pos, const double dt)
-  {
-    sk_impl_.update_from_position(
-      right_traction_wheel_pos, left_traction_wheel_pos, right_steer_pos, left_steer_pos, dt);
-  }
+    const double right_steer_pos, const double left_steer_pos, const double dt);
 
   bool update_from_velocity(
-    const double traction_wheel_vel, const double steer_pos, const double dt)
-  {
-    sk_impl_.update_from_velocity(traction_wheel_vel, steer_pos, dt);
-  }
+    const double traction_wheel_vel, const double steer_pos, const double dt);
 
   bool update_from_velocity(
     const double right_traction_wheel_vel, const double left_traction_wheel_vel,
-    const double steer_pos, const double dt)
-  {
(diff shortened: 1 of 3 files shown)


steering_controllers_library/include/steering_controllers_library/steering_odometry.hpp
```diff
@@ -18,6 +18,9 @@
 #ifndef STEERING_CONTROLLERS_LIBRARY__STEERING_ODOMETRY_HPP_
 #define STEERING_CONTROLLERS_LIBRARY__STEERING_ODOMETRY_HPP_
 
+#warning "This header is deprecated. Please include steering_kinematics.hpp."
```
christophfroehlich: We need a portable version here
` D:\a\ros2_controllers\ros2_controllers\src\ros-controls\ros2_controllers\steering_controllers_library\include\steering_controllers_library\steering_odometry.hpp(21,1): error C1021: invalid preprocessor command 'warning' [D:\a\ros2_controllers\ros2_controllers\build\steering_controllers_library\steering_controllers_library.vcxproj]`

```suggestion
#ifdef _WIN32
#pragma message( \
  "This header is deprecated. Please update your code to use 'steering_kinematics.hpp' header.")  //NOLINT
#else
#warning \
  "This header is deprecated. Please update your code to use 'steering_kinematics.hpp' header." //NOLINT
#endif
```

kumar-sanjeeev: Hi @christophfroehlich,

Thanks for reviewing the PR. I have updated the warning to make it portable as suggested. Due to this refactoring, a couple of warnings are now generated while building the package. Should I also update the previous header (steering_odometry.hpp) to the new one wherever it is used?

christophfroehlich: yes, please do so

kumar-sanjeeev:
Hi @christophfroehlich,

I am unable to figure out why the Rolling-ABI CI/CD check is failing, however it builds fine locally. Also, I have updated the header files in all locations where steering_odometry.hpp was used to suppress the deprecation warnings.

christophfroehlich:
> Hi @christophfroehlich,
> 
> I am unable to figure out why the Rolling-ABI CI/CD check is failing, however it builds fine locally. Also, I have updated the header files in all locations where steering_odometry.hpp was used to suppress the deprecation warnings.

You are changing the ABI with your PR, so this test is supposed to fail. See the [test report](https://github.com/ros-controls/ros2_controllers/actions/runs/19206533266/artifacts/4511330071). Instead of using type aliases, you could leave the old methods but just call the new one from inside. This should leave ABI untouched. 
The current state is fine for rolling, but if we backport this ABI break here is subject for discussion.

kumar-sanjeeev:
> > Hi @christophfroehlich,
> > I am unable to figure out why the Rolling-ABI CI/CD check is failing, however it builds fine locally. Also, I have updated the header files in all locations where steering_odometry.hpp was used to suppress the deprecation warnings.
> 
> You are changing the ABI with your PR, so this test is supposed to fail. See the [test report](https://github.com/ros-controls/ros2_controllers/actions/runs/19206533266/artifacts/4511330071). Instead of using type aliases, you could leave the old methods but just call the new one from inside. This should leave ABI untouched. The current state is fine for rolling, but if we backport this ABI break here is subject for discussion.

I understand the concern, and I also liked your idea to prevent the ABI check from failing. Regarding your suggestion` Instead of using type aliases, you could leave the old methods but just call the new one from inside`,  should I start working on this, or wait for further discussion?

christophfroehlich:
Please do so. I now that it is a bit of a tedious work, but then we can easily backport it to kilted.

kumar-sanjeeev:
> Please do so. I now that it is a bit of a tedious work, but then we can easily backport it to kilted.

understood, i will do this.

kumar-sanjeeev:
> Please do so. I now that it is a bit of a tedious work, but then we can easily backport it to kilted.

Hi @christophfroehlich ,

Is there a recommended way to run the GitHub Actions workflow locally for testing?

christophfroehlich:
you can use `act` to run them locally, a brief description is [here](https://github.com/ros-controls/control.ros.org/blob/rolling/README.md#testing-github-workflows).

kumar-sanjeeev:
> you can use `act` to run them locally, a brief description is [here](https://github.com/ros-controls/control.ros.org/blob/rolling/README.md#testing-github-workflows).

thanks for letting me know. The ABI check is still failing with the new push. I’m digging into logs to find a reason and will update with  fix. I’ll let you know once everything is ready from myside.

christophfroehlich (approved):
LGTM, and ABI checker is also fine with it!

saikishor (approved):
The code looks fine. Didn't pay too much attention to the logic. I'll trust @christophfroehlich 😉

PR #1997, opened 2025-11-09
TF prefix helper for diff drive controller

kuralme:
Used new tf frame prefix helper in differential drive controller to test out & showcase the helper. This is the first part of the changes, the other controllers will come next:
- mecanum_drive_controller
- omni_wheel_drive_controller

Related [issue](https://github.com/ros-controls/ros2_controllers/issues/1745)

Needs https://github.com/ros-controls/ros2_control/pull/2803

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
@@ -1,22 +1,24 @@
 // Copyright 2013 PAL Robotics S.L.
 // All rights reserved.
 //
-// Software License Agreement (BSD License 2.0)
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
-//   * Redistributions of source code must retain the above copyright notice,
-//     this list of conditions and the following disclaimer.
-//   * Redistributions in binary form must reproduce the above copyright
-//     notice, this list of conditions and the following disclaimer in the
-//     documentation and/or other materials provided with the distribution.
-//   * Neither the name of PAL Robotics S.L. nor the names of its
-//     contributors may be used to endorse or promote products derived from
-//     this software without specific prior written permission.
+//
+//    * Redistributions of source code must retain the above copyright
+//      notice, this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of the PAL Robotics S.L. nor the names of its
+//      contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
 //
 // THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 // AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 // IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
-// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
 // LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
 // CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
 // SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -722,7 +722,7 @@ void JointTrajectoryController::query_state_service(
 {
   const auto logger = get_node()->get_logger();
   // Preconditions
-  if (get_lifecycle_state().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+  if (get_lifecycle_id() != lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
   {
     RCLCPP_ERROR(logger, "Can't sample trajectory. Controller is not active.");
     response->success = false;
@@ -1360,7 +1360,7 @@ rclcpp_action::GoalResponse JointTrajectoryController::goal_received_callback(
   RCLCPP_INFO(get_node()->get_logger(), "Received new action goal");
 
   // Precondition: Running controller
-  if (get_lifecycle_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+  if (get_lifecycle_id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
   {
     RCLCPP_ERROR(
       get_node()->get_logger(), "Can't accept new action goals. Controller is not running.");
--- a/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
@@ -427,9 +427,7 @@ public:
   {
     if (traj_controller_)
     {
-      if (
-        traj_controller_->get_lifecycle_state().id() ==
-        lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+      if (traj_controller_->get_lifecycle_id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
       {
         EXPECT_EQ(
           traj_controller_->get_node()->deactivate().id(),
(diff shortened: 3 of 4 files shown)


christophfroehlich:
> I'm not sure if it makes sense to have it in the control_toolbox package. What about controllers_interfaces/helpers.hpp existing file?

OK, maybe you are right and we should put it [here](https://github.com/ros-controls/ros2_control/blob/master/controller_interface/include/controller_interface/helpers.hpp).

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -411,32 +412,17 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
...
-  }
-
-  const auto odom_frame_id = tf_prefix + params_.odom_frame_id;
-  const auto base_frame_id = tf_prefix + params_.base_frame_id;
+  // apply the TF prefix if not empty, otherwise use the node namespace
+  const auto odom_frame_id =
+    params_.tf_frame_prefix_enable
+      ? control_toolbox::apply_tf_prefix(
+          params_.tf_frame_prefix, get_node()->get_namespace(), params_.odom_frame_id)
+      : params_.odom_frame_id;
```
christophfroehlich: should we change the API to only return the prefix, and just append the  odom_rame_id in any case?
something like this?
```suggestion
    (params_.tf_frame_prefix_enable
      ? control_toolbox::get_tf_prefix(
          params_.tf_frame_prefix, get_node()->get_namespace()) : "")
      + params_.odom_frame_id;
```

christophfroehlich: or we readd the boolean `tf_frame_prefix_enable` as argument (sorry for my earlier feedback about that).

But as it is now, it feels a bit overcomplicating things.

kuralme: Sure no problem

kuralme:
> > I'm not sure if it makes sense to have it in the control_toolbox package. What about controllers_interfaces/helpers.hpp existing file?
> 
> OK, maybe you are right and we should put it [here](https://github.com/ros-controls/ros2_control/blob/master/controller_interface/include/controller_interface/helpers.hpp).

I will start working on it

christophfroehlich (changes requested):
There is a behavior-breaking change now, right?
Can we do that gracefully? e.g., 

- Deprecate the  `tf_frame_prefix_enable` parameter in the yaml, and log a warning message if it is set to a value different than default. And we should maintain the behavior and still use the value.
- Please update the description of the  `tf_frame_prefix`  parameter
https://github.com/ros-controls/ros2_controllers/blob/master/diff_drive_controller/src/diff_drive_controller_parameter.yaml#L49-L53
- Please add [release](https://github.com/ros-controls/ros2_controllers/blob/master/doc/release_notes.rst) and [migration](https://github.com/ros-controls/ros2_controllers/blob/master/doc/migration.rst) notes.

kuralme:
A note about my last [commit](https://github.com/ros-controls/ros2_controllers/pull/1997/commits/55ac4411b08d075643dd899f61842c231807b540): This is for the deprecations. I am aware of Bence's [request](https://github.com/ros-controls/ros2_control/pull/2803#discussion_r2529670979), I will apply if a change needed to the helper source.

doc/release_notes.rst
```diff
@@ -4,3 +4,8 @@ Release Notes: Kilted Kaiju to Lyrical Luth
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 This list summarizes important changes between Kilted Kaiju (previous) and Lyrical Luth (current) releases.
+
+diff_drive_controller
+*****************************
+* Remove unused parameter ``tf_frame_prefix_enable``, use ``tf_frame_prefix`` instead. (`#1997 <https://github.com/ros-controls/ros2_controllers/pull/1997>`_).
```
christophfroehlich:
```suggestion
* Parameter ``tf_frame_prefix_enable`` got deprecated and will be removed in a future release (`#1997 <https://github.com/ros-controls/ros2_controllers/pull/1997>`_).
```

doc/migration.rst
```diff
@@ -4,3 +4,8 @@ Migration Guides: Kilted Kaiju to Lyrical Luth
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 This list summarizes important changes between Kilted Kaiju (previous) and Lyrical Luth (current) releases, where changes to user code might be necessary.
+
+diff_drive_controller
+*****************************
+* Remove unused parameter ``tf_frame_prefix_enable``, use ``tf_frame_prefix`` instead. (`#1997 <https://github.com/ros-controls/ros2_controllers/pull/1997>`_).
```
christophfroehlich:
```suggestion
* Instead of using ``tf_frame_prefix_enable:=false``, set an empty ``tf_frame_prefix:=""`` parameter instead. (`#1997 <https://github.com/ros-controls/ros2_controllers/pull/1997>`_).
```

doc/migration.rst
```diff
@@ -4,3 +4,8 @@ Migration Guides: Kilted Kaiju to Lyrical Luth
...
+* Now any tilde ("~") character in ``tf_frame_prefix`` is substituted with node namespace. (`#1997 <https://github.com/ros-controls/ros2_controllers/pull/1997>`_).
```
christophfroehlich:
```suggestion
* For using node namespace as tf prefix: Set ``tf_frame_prefix:="~"``, where the ("~") character is substituted with node namespace. (`#1997 <https://github.com/ros-controls/ros2_controllers/pull/1997>`_).
```

christophfroehlich (changes requested):
`tf_frame_prefix_enable` was not unused before? Instead, during the deprecation phase, we still _should_ use this parameter before we remove it.

I'm not sure how we can gracefully transition the usage of node namespace? Currently, if tf_frame_prefix is not set, then automatically the node namespace is used.

doc/migration.rst
kuralme: Sorry I am unfamiliar with these. Applied your suggestions.

christophfroehlich: No worries.
Just for future PRs, you can directly commit suggestions from the github UI. this makes it easer for re-review :)

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -410,37 +411,29 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
...
-      tf_prefix = tf_prefix + "/";
-    }
-    if (tf_prefix.front() == '/')
-    {
-      tf_prefix.erase(0, 1);
-    }
-  }
+  // resolve prefix: substitute tilde (~) with the namespace if contains and normalize slashes (/)
+  const auto tf_prefix =
+    controller_interface::resolve_tf_prefix(params_.tf_frame_prefix, get_node()->get_namespace());
```
christophfroehlich:
```suggestion
  std::string tf_prefix = "";
  if (params_.tf_frame_prefix_enable)
  {
    if (params_.tf_frame_prefix != "")
    {
      tf_prefix =
    controller_interface::resolve_tf_prefix(params_.tf_frame_prefix, get_node()->get_namespace());
    }
    else
    {
      tf_prefix = std::string(get_node()->get_namespace());
    }    
  }
```
In this way, we get backward compatibility? currently we have the deprecation notice, but it has no effect any more.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -410,37 +411,40 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
...
-  // Append the tf prefix if there is one
+  // resolve prefix: substitute tilde (~) with the namespace if contains and normalize slashes (/)
   std::string tf_prefix = "";
   if (params_.tf_frame_prefix_enable)
   {
     if (params_.tf_frame_prefix != "")
     {
-      tf_prefix = params_.tf_frame_prefix;
+      tf_prefix =
+    controller_interface::resolve_tf_prefix(params_.tf_frame_prefix, get_node()->get_namespace());
```
christophfroehlich:
```suggestion
      tf_prefix = controller_interface::resolve_tf_prefix(
        params_.tf_frame_prefix, get_node()->get_namespace());
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -410,37 +411,40 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
...
   realtime_odometry_publisher_ =
     std::make_shared<realtime_tools::RealtimePublisher<nav_msgs::msg::Odometry>>(
       odometry_publisher_);
 
-  // Append the tf prefix if there is one
+  // resolve prefix: substitute tilde (~) with the namespace if contains and normalize slashes (/)
   std::string tf_prefix = "";
   if (params_.tf_frame_prefix_enable)
   {
     if (params_.tf_frame_prefix != "")
```
kuralme, line 436: This inner if block is wrong imo. It skips slash normalization so the test fail, and the new tilde-replace-namespace not applied as well. That should also address [#1999](https://github.com/ros-controls/ros2_controllers/issues/1999) with the tilde usage.

christophfroehlich: But this still breaks current behavior, right? That's what I'm concerned about. Any idea how to deprecate that gracefully? We could add a new parameter to opt-in the new behavior, together with a warning when it is not set. This can then be fully removed one release after next.

kuralme: How about this? I added the if block back with slash normalization addition. It doesnt look pretty but retains the current behavior. I may add extra test to ensure that if you prefer

christophfroehlich:
@kuralme Thanks for iterating this. Would be great to rollout this change to meanum, omni, and all the steering controllers now..

PR #2017, opened 2025-11-20
Fix links to diff_drive_controller doc file.

l00p3:
Reading the repo I just found these links not working in the doc for the diff_drive_controller. With relative paths they also should remain specific for the branch.


l00p3:
Uh ok, I thought it has to be read in github itself. My bad. I'll close the request, I'm sorry.

PR #2020, opened 2025-11-21
Export pid gains as reference

kumar-sanjeeev:
This PR includes the following changes:

- exported p, i, d gains as reference.
- Updated test cases.


pid_controller/src/pid_controller.cpp
```diff
@@ -495,6 +517,49 @@ controller_interface::return_type PidController::update_and_write_commands(
...
+      {
+        const size_t buffer_index = gains_start_index + i + j * dof_;
+        const double new_gain_value = reference_interfaces_[buffer_index];
+        if (std::isfinite(new_gain_value))
+        {
+          const size_t gain_type = GAIN_TYPES_INDEX[j];
+          switch (gain_type)
+          {
+            case 0:  // P gain
+              pids_[i]->set_gains(
```
christophfroehlich: Am I missing something? I think there is no functionality to skip setting the gain by passing a NaN value?

christophfroehlich (commented):
In principle, this looks fine for me.

pid_controller/include/pid_controller/pid_controller.hpp
```diff
@@ -76,6 +76,9 @@ class PidController : public controller_interface::ChainableControllerInterface
   size_t dof_;
   std::vector<double> measured_state_values_;
 
+  inline static const std::vector<std::string> GAIN_INTERFACES = {"p", "i", "d"};
+  inline static const std::vector<size_t> GAIN_TYPES_INDEX = {0, 1, 2};
```
thedevmystic: A quick question about convention of ROS Controls, do we follow,

`[Storage Specificier] [Modifier] [Type] [Variable Name]`
Like: static constexpr bool is_ros = true;

Or
`[Modifier] [Storage Specificier] [Type] [Variable Name]`

Like: constexpr static bool is_ros = true;

If we follow the above convention, then I think you should change the order of specifier. 

Am I missing something, or is the code convention different. 

I assumed the first convention because it is used pre-dominantely in other codebase.

Thanks!

pid_controller/src/pid_controller.cpp
kumar-sanjeeev: @christophfroehlich , you are right. I am thinking of following fix. What do you think ?
```suggestion
      auto current_pid_gains = pids_[i]->get_gains();
      for (size_t j = 0; j < GAIN_INTERFACES.size(); ++j)
      {
        const size_t buffer_index = gains_start_index + i + j * dof_;
        const double new_gain_value = reference_interfaces_[buffer_index];
        if (std::isfinite(new_gain_value))
        {
          const size_t gain_type = GAIN_TYPES_INDEX[j];
          switch (gain_type)
          {
            case 0:  // P gain
              current_pid_gains.p_gain_ = new_gain_value;
              pids_[i]->set_gains(current_pid_gains);```

File: pid_controller/include/pid_controller/pid_controller.hpp (hunk 2 above)
Comment (kumar-sanjeeev) [reply pid_controller/include/pid_controller/pid_controller.hpp]: hi @TheDevMystic,

I checked your recommendation, and you are right. I am going to follow the first convention.
```suggestion
  static inline const std::vector<std::string> GAIN_INTERFACES = {"p", "i", "d"};
  static inline const std::vector<size_t> GAIN_TYPES_INDEX = {0, 1, 2};```

**Comment by kumar-sanjeeev** [comment]
> In principle, this looks fine for me.

understood. I am trying to understand one thing, in my current implementation, when I set `export_gain_references=true`, the `reference_interfaces_ `vector allocates additional memory for the PID gains for each DOF, as expected:

```cpp
  if (params_.export_gain_references)
  {
    total_reference_size += dof_ * GAIN_INTERFACES.size();
  }
  reference_interfaces_.resize(total_reference_size, std::numeric_limits<double>::quiet_NaN());
```

Because of this, some of the existing tests are failing. If this implementation is correct, should I update the tests so that they handle both cases `(export_gain_references=true and export_gain_references=false)`? Or is there something else I might be missing? Or is there any other recommended way to cater this ?

Test config that I used to reproduce the failure case, along with the corresponding log:

```yaml
test_pid_controller:
  ros__parameters:
    dof_names:
      - joint1

    command_interface: position

    reference_and_state_interfaces: ["position"]

    export_gain_references: true

    gains:
      joint1: {p: 1.0, i: 2.0, d: 3.0, u_clamp_max: 5.0, u_clamp_min: -5.0}

```


```bash
/workspaces/ros2_control_rolling_dev_docker_ws/src/ros-controls/ros2_controllers/pid_controller/test/test_pid_controller.cpp:88: Failure
Expected equality of these values:
  ref_if_conf.size()
    Which is: 4
  dof_state_values_.size()
    Which is: 1

[  FAILED  ] PidControllerTest.check_exported_interfaces (6 ms)
```

thedevmystic:
I also think implementation is correct @kumar-sanjeeev, 
But if implementation is correct then, why are we failing tests?

Overall, I think some tests need to be updated, but this is just my assumption.

kumar-sanjeeev:
hi @christophfroehlich, I have tried few different ways to fix the ABI check, but still haven’t succeeded. Do you have any recommendations? Or is it okay in this feature request, if it breaks ?

pid_controller/src/pid_controller.cpp
```diff
@@ -495,6 +517,43 @@ controller_interface::return_type PidController::update_and_write_commands(
...
+
+  if (params_.export_params.gain_references)
+  {
+    size_t gains_start_index = dof_reference_size;
+    for (size_t i = 0; i < dof_; ++i)
+    {
+      auto current_pid_gains = pids_[i]->get_gains();
+      for (size_t j = 0; j < GAIN_INTERFACES.size(); ++j)
+      {
+        const size_t buffer_index = gains_start_index + i + j * dof_;
```
thedevmystic:
```suggestion
        const size_t buffer_index = gains_start_index + (j * dof_) + i;
```

As the standard formula for index calculation is,
`index = base_offset + (row_index * row_size) + column_index`.

While the previous one is perfectly fine, this improves general readability and precedence readability.

pid_controller/src/pid_controller.yaml
```diff
@@ -126,3 +126,10 @@ pid_controller:
         default_value: false,
         description: "Individual state publisher activation for each DOF. If true, the controller will publish the state of each DOF to the topic `/<controller_name>/<dof_name>/pid_state`."
       }
+  export_params:
+      gain_references: {
+        type: bool,
+        default_value: false,
+        description: "If true, exports P, I, D gains as reference interfaces to be claimed by preceding controllers",
```
thedevmystic: If you don't mind me nitpicking on your descriptions, I'll suggest this.

```suggestion
        description: "If true, exports P, I, and D gains as reference interfaces to be claimed by preceding controllers",
```

Just for clarity and grammatical consistency.

kumar-sanjeeev:
Hi @christophfroehlich,

The implementation is complete from my side. Whenever you have time to review it, please let me know if there is anything I may have missed or should improve.

pid_controller/src/pid_controller.cpp
```diff
@@ -495,6 +517,43 @@ controller_interface::return_type PidController::update_and_write_commands(
...
+              break;
+            case 1:  // I gain
+              current_pid_gains.i_gain_ = new_gain_value;
+              pids_[i]->set_gains(current_pid_gains);
+              break;
+            case 2:  // D gain
+              current_pid_gains.d_gain_ = new_gain_value;
+              pids_[i]->set_gains(current_pid_gains);
+              break;
+          }
```
christophfroehlich: This is part is not covered by tests. Can you please think of a testcase in doing so? Friending the class and checking pids_ variable might be possible.

pid_controller/src/pid_controller.yaml
```diff
@@ -126,3 +126,10 @@ pid_controller:
         default_value: false,
         description: "Individual state publisher activation for each DOF. If true, the controller will publish the state of each DOF to the topic `/<controller_name>/<dof_name>/pid_state`."
       }
+  export_params:
+      gain_references: {
```
christophfroehlich: do we really need this parameter? can't they always be exported, and if they are not claimed they just remain nan?

christophfroehlich (changes requested):
Another quick look and question

pid_controller/src/pid_controller.cpp
kumar-sanjeeev: Hi @christophfroehlich,

I was on vacation, will look into comments and get back to you

christophfroehlich:
@kumar-sanjeeev any updates on this?

kumar-sanjeeev:
> @kumar-sanjeeev any updates on this?

@christophfroehlich apologies, this PR slipped through my mind. I'll work on it in coming days and get it closed.

kumar-sanjeeev:
@christophfroehlich,

Sorry for the delay. Based on your feedback, I have removed the added parameter and modified the logic. Also added a corresponding test case for the mentioned code.

kumar-sanjeeev:
hi @christophfroehlich,

when you will get time, can you please provide feedback on last committed changes.

pid_controller/src/pid_controller.cpp
```diff
@@ -533,6 +548,40 @@ controller_interface::return_type PidController::update_and_write_commands(
...
+  {
+    auto current_pid_gains = pids_[i]->get_gains();
+    for (size_t j = 0; j < GAIN_INTERFACES.size(); ++j)
+    {
+      const size_t buffer_index = gains_start_index + (j * dof_) + i;
+      const double new_gain_value = reference_interfaces_[buffer_index];
+      if (std::isfinite(new_gain_value))
+      {
+        const size_t gain_type = GAIN_TYPES_INDEX[j];
+        switch (gain_type)
```
christophfroehlich: this could be done by changing GAIN_INTERFACES to a `std::vector<char>`, which then could be used in the switch-case directly without the need of GAIN_TYPES_INDEX?

christophfroehlich (commented):
Please never do force pushes if PRs are already under review. It breaks the github diff view and just takes more time for review.

Overall LGTM, one suggestion for further cleanup. what do you think?

christophfroehlich (approved):

Issue #2027, opened 2025-11-26
[force_torque_sensor_broadcaster] wrench_transformer_node handles target frames as positional arguments


Juliaj:
* It could be interesting to have a node like [tf2_echo](https://github.com/ros2/geometry2/blob/054b8091c5f6f0b33abbac032f54e01f73bda0d0/tf2_ros/src/tf2_echo.cpp#L70-L77), so it would look like something like following
```
ros2 run force_torque_sensor_broadcaster wrench_transformer_node frame1 frame2 --ros-args -r __ns:=fts -r input:=wrench
```

_Originally posted by @saikishor in https://github.com/ros-controls/ros2_controllers/issues/2021#issuecomment-3577856144_

thedevmystic:
Indeed @Juliaj, I also think it could be quite interesting to have a node like *tf2_echo*.

Whadya say, about its current implementation?

vedh1234:
Hi! I’d like to work on this.  
Could you please assign it to me?

vedh1234:
Hi @Juliaj @christophfroehlich!
I've opened a PR for this issue, please do check. 
Happy to receive any feedback.

PR #2040, opened 2025-12-01
Add support for positional target frame arguments for transform wrench node

vedh1234:
This PR adds support for passing target frames as positional CLI arguments to `wrench_transformer_node`, similar to `tf2_echo`. ( Issue #2027 ) Positional arguments override the existing `target_frames` parameter when both are provided.

Includes:
- CLI parsing for non-ROS arguments
- Parameter override logic
- Updated documentation
- Test for positional arguments


force_torque_sensor_broadcaster/doc/userdoc.rst
```diff
@@ -44,11 +44,20 @@ The node subscribes to wrench messages from the broadcaster (either raw or filte
...
+   ros2 run force_torque_sensor_broadcaster wrench_transformer_node frame1 frame2
+
+Target frames may also be set via the ``target_frames`` parameter:
+
+.. code-block:: bash
+
+   ros2 run force_torque_sensor_broadcaster wrench_transformer_node \
+     --ros-args -p target_frames:="['frame1','frame2']"
+
+Positional arguments override the parameter value when both are provided.
```
thedevmystic, line 61: I would suggest why don't we keep the standalone executable comment too, and introduce tf2_echo frames option as an "option". So that previous usage of the controller can be still used. And the previous documentation is still valid.

force_torque_sensor_broadcaster/src/wrench_transformer.cpp
```diff
@@ -186,8 +187,17 @@ void WrenchTransformer::setup_publishers()
...
+  std::vector<std::string> target_frames_cli;
+  for (size_t i = 1; i < non_ros_args.size(); ++i)
+  {
+    target_frames_cli.push_back(non_ros_args[i]);
+  }
   rclcpp::NodeOptions options;
+  if (!target_frames_cli.empty())
+  {
+    options.append_parameter_override("target_frames", rclcpp::ParameterValue(target_frames_cli));
+  }
```
thedevmystic:
```suggestion
  std::vector<std::string> target_frames_args;
  for (size_t i = 1; i < non_ros_args.size(); ++i)
  {
    target_frames_args.push_back(non_ros_args[i]);
  }
   rclcpp::NodeOptions options;
  if (!target_frames_args.empty())
  {
    options.append_parameter_override("target_frames", rclcpp::ParameterValue(target_frames_args));
  }
```

Here I think, `target_frames_args` is more suitable choice, as it clearly say that it is not a cli but a argument that comes from the cli.

thedevmystic (commented):
Phenomenal job on adding this features. Appreciate it, @vedh1234

force_torque_sensor_broadcaster/src/wrench_transformer.cpp
vedh1234: Hi, thanks for the suggestion. You are right and I have updated it please check.

thedevmystic: Great work!

force_torque_sensor_broadcaster/doc/userdoc.rst
vedh1234: The node cannot run without specifying target_frames (it throws an initialization error). I saw the tf2_echo doc which also mentions frame usage, that's why I thought of adding it and removing the older comment. So do you think it's still valid to keep the standalone executable comment or we are supposed to add what successfully runs ?

thedevmystic: If the older implementation of force_torque_sensor_broadcaster can run without specifying the tf_echo argument.

Then, I think we shouldn't throw an error if it is missing.

We should introduce `tf_echo` as an "option", not a requirement. To keep the backward compatibility, when tf_echo wasn't introduced.

Am I missing something, if so, feel free to inform me.

thedevmystic: If the older implementation of force_torque_sensor_broadcaster can run without specifying the tf_echo argument.

Then, I think we shouldn't throw an error if it is missing.

We should introduce tf_echo as an "option", not a requirement. To keep the backward compatibility, when tf_echo wasn't introduced.

We should make it optional, by providing a default value when none is provided, but currently it is a requirement. This will break the usage of old existing users.

Am I missing something, if so, feel free to inform me.

vedh1234: Thanks for the clarification!

Just to confirm — the previous implementation of wrench_transformer_node also required target_frames to be non-empty. The parameter had a not_empty<> validator, so launching the node without specifying target_frames already resulted in a parameter validation error.

So the node was never runnable without frames — only the documentation implied it.

This PR doesn’t change that requirement, it just adds an additional way (positional args) to provide the same frames. The parameter-based approach is still fully supported and backward compatible.

thedevmystic: Thanks for clarification.

But this means the `not_empty<>` is incorrect, as the original documentation explicitly said, it can be used as an standalone executable, without specifying the target frames. As also target frames is initialized as empty list (default value). I think it is a long run inconsistency.

Because, something that is forbidden to be empty, can't have a default value.

---

### Orginal documentation:

The package provides a standalone ROS 2 node wrench_transformer_node that transforms wrench messages published by the ForceTorqueSensorBroadcaster controller to different target frames using TF2. This is useful when applications need force/torque data in coordinate frames other than the sensor frame.

The node subscribes to wrench messages from the broadcaster (either raw or filtered) and publishes transformed versions to separate topics for each target frame.

Usage
The wrench transformer node can be launched as a standalone executable:

```bash
ros2 run force_torque_sensor_broadcaster wrench_transformer_node
```

---

As we can see, we don't need to specify target frames, do we?

So, I think `wrench_transformer_node` was intended to be used without explicit `target_frames`. And `tf_echo` is a option, not an requirement.

So, i think we should remove the validation, as default value is provided, if it is empty.

```yaml
force_torque_wrench_transformer:
  target_frames: {
    type: string_array,
    default_value: [],
    description: "...",
    # validation: {
    #   not_empty<>: null
    # }
  }
```

What do you say, @vedh1234?

christophfroehlich: what would be the usage of the wrench_transformer_node without target frames? that was not the original intention. The current state of the documentation fixes this, thanks!

thedevmystic: Okay, so it was the other way around. The documentation was incorrect, and the implementation was right. I was thinking the documentation is correct, and implementation has inconsistency. Thanks for clarification, @christophfroehlich.

christophfroehlich:
why not using both?

saikishor:
> why not using both? 

Sure.

saikishor (approved):

PR #2019, opened 2025-11-21
Refactor interpolation_method class

thedevmystic:
Hello, there ros2_controllers maintainers!
Again this is Surya! Here this is first batch of joint_trajectory_controller library refactor.

As in the issue #2018, I have discussed with @christophfroehlich, to do refactor in batches.

This batch focus on `interpolation_method.hpp`

I have refactored the following,
1. Add extensive doxygen-style comments, so future maintainance can go smoothly.
2. ~~Refactored Maps and added reverse map.~~
3. Refactored `from_string` function.
4. ~~Changed reference of `InterpolationMethodMap` in joint_trajectory_controller.cpp to `ReverseInterpolationMethodMap` as it makes much more sense.~~

If it is good, I can move forward to refactor `trajectory.hpp` and `trajectory.cpp` to handle edge cases, enhance readability, add comments. As when I was using it, I got segmentation fault (Trying to access memory [nil]).

PR #2041, opened 2025-12-02
Cleanup deprecated InterpolationMethodMap

thedevmystic:
Hello, respected maintainers, and reviewers!!
Again, this is Surya :)

This follow-up PR on interpolation_methods.hpp, cleans up the deprecated `InterpolationMethodMap`, for the next release cycle.

This PR addresses,
1. Removed deprecated `InterpolationMethodMap`.


saikishor:
> The change itself seems fine, but what is the policy on deprecating this? Having only one sync in between seems less than what we used to do previously.

Yes, we will have to hold it for sometime

thedevmystic:
Yes, I also think so. We should wait few release syncs before cleaning it up. 👍

christophfroehlich:
Yes, that's why I labeled it as hold but will likely merge it once the deprecation got synced.
I'm just annoyed by the deprecation warning we have now, see https://github.com/ros-controls/ros2_controllers/pull/2019#issuecomment-3601665735

thedevmystic:
This comment is to remove the `stale` label.

christophfroehlich (approved):
We can merge this now after today's rolling sync.

thedevmystic:
Yeah!!!

PR #2043, opened 2025-12-02
Fill point_before_trajectory with same information as trajectory

### LINKED ISSUE #2015: [JTC] Trajectory with accelerations has jerk if HW doesn't have an acceleration interface
**Describe the bug**
When passing a trajectory to the JTC, it will add a segment from the current state (or the last commanded state based on the config) to the first point of the trajectory. If the hardware doesn't provide an acceleration in its state interface, the `has_accel` check will be false for the first segment, leading to a quadratic spline parametrization for that segment which I feel to be unexpected, as I did specify accelerations for the first trajectory point.

**Screenshots**
The following picture explains that: There is a trajectory with three setpoints. At each setpoints a velocity and acceleration vector of 0 is specified. The second and third segment show a smooth rampup of velocity while the first segment has a sudden change in velocity and therefore a jump in acceleration.

Image

**Expected behavior**
Is that desired behavior? I would expect that the first segment has a smooth rampup, as well. I made a quick check and when adding a 0 acceleration vector into the `state_before_traj_msg_` in case where the first trajectory point has accelerations indeed results in a smooth velocity profile.

**Environment**
 - Version: Rolling dev

fmauch:
When executing a trajectory with a start time in the future, the trajectory interpolation will create a setpoint at the robot's current configuration based on the state interface information.

However, when the state interface for example doesn't have acceleration information, but the trajectory does contain position, velocity and acceleration, the segment from the current state to the first trajectory point will not use quintic spline interpolation, though the rest of the trajectory will.

This commit fills the current state with velocity and acceleration information if it is not given in the state interfaces but it is defined in the trajectory's first point.

This fixes #2015 

Velocity profile with first point in the future / at other position **without** this PR applied:
velocity_default

Velocity profile with first point in the future / at other position **with** this PR applied:
velocity_with_acceleration

I know that this is strictly speaking behavior changing, but I would nevertheless consider this a bugfix. This could, of course get an opt-in parameter on released versions to avoid the behavior change if desired.

### DIFF
--- a/force_torque_sensor_broadcaster/doc/userdoc.rst
+++ b/force_torque_sensor_broadcaster/doc/userdoc.rst
@@ -44,11 +44,20 @@ The node subscribes to wrench messages from the broadcaster (either raw or filte
 
 Usage
 ^^^^^
-The wrench transformer node can be launched as a standalone executable:
+The wrench transformer can be launched with target frames passed directly as positional arguments:
 
 .. code-block:: bash
 
-   ros2 run force_torque_sensor_broadcaster wrench_transformer_node
+   ros2 run force_torque_sensor_broadcaster wrench_transformer_node frame1 frame2
+
+Target frames may also be set via the ``target_frames`` parameter:
+
+.. code-block:: bash
+
+   ros2 run force_torque_sensor_broadcaster wrench_transformer_node \
+     --ros-args -p target_frames:="['frame1','frame2']"
+
+Positional arguments override the parameter value when both are provided.
 
 Wrench Transformer Parameters
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


fmauch:
I don't understand what the abi checker is complaining about, either.

I think backporting it all the way back to humble might make sense. As I wrote earlier, an opt-in flag could be added do not surprisingly change behavior for users. Not that I would expecmany users would notice. Though trajectory starts will be less jerky which might be noticable in some cases.

christophfroehlich:
Ok, I'm fine with that. Would you mind adding something to the release notes then?

fmauch:
Sure thing. I'll try to do it tonight or tomorrow. But just to be sure: What would be the plan to backport this? Should we add the opt-in flag or just backport as is? In the first case I would note that as a change to Lyrical that this is now the default behavior.

saikishor:
> I think backporting it all the way back to humble might make sense. As I wrote earlier, an opt-in flag could be added do not surprisingly change behavior for users. Not that I would expecmany users would notice. Though trajectory starts will be less jerky which might be noticable in some cases.

I personally think this is a very valid behaviour and can be backported

fmauch:
Added a release notes entry.

christophfroehlich (approved):

Issue #2044, opened 2025-12-03
Deprecate dynamic_joint_states


christophfroehlich:
> We decided to not do this, but
> - add a parameter back to humble to disable dynamic_joint_states
> - deprecate dynamic_joint_states on rolling, with migration notes to gpio_command_controller, pal_statistics, and [interface_state_broadcaster](https://github.com/ros-controls/ros2_controllers/pull/2006)
> - remove it in an upcoming release
> 

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_controllers/issues/1783#issuecomment-3608329199_

PR #2064, opened 2025-12-09
Add parameter for deactivating dynamic_joint_states

christophfroehlich:
- Remove unused method
- Make parameters read-only.
- Add parameter to deactivate dynamic_joint_states

First step of #2044

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -16,8 +16,8 @@ diff_drive_controller
 
 joint_state_broadcaster
 ************************
-* Make all parameters read-only (the never got re-evaluated after initialization anyway). (`#2064 <https://github.com/ros-controls/ros2_controllers/pull/2064>`_).
-* Added parameter ``publish_dynamic_joint_states`` to enable/disable publishing of dynamic joint states. (`#2064 <https://github.com/ros-controls/ros2_controllers/pull/2064>`_).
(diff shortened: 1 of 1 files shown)


joint_state_broadcaster/src/joint_state_broadcaster_parameters.yaml
```diff
@@ -30,24 +34,35 @@ joint_state_broadcaster:
...
   }
   frame_id: {
     type: string,
     default_value: "base_link",
+    read_only: true,
     description: "The frame_id to be used in the published joint states. This parameter allows rviz2 to visualize the effort of the joints."
   }
+  publish_dynamic_joint_states: {
+    type: bool,
+    default_value: true,
```
saikishor, line 65:
```suggestion
    default_value: false,
```

We can change this in the backports

saikishor (approved):
The changes look good to me

joint_state_broadcaster/src/joint_state_broadcaster_parameters.yaml
christophfroehlich: I'd like to deprecate that first before deactivating it. Can I do that in an upcoming PR? so we can backport this one 1:1

saikishor: Whatever you say boss 🫡

PR #2074, opened 2025-12-14
RateLimiter: Don't update parameters before input checks #437

hjh059:
Fixes https://github.com/ros-controls/control_toolbox/issues/437

I submitted two pull requests: one in the ros2_controllers repository and another in the control_toolbox repository.
The current PR addressed the following:
1. Evaluated which parameters should be modifiable while the controller is active
2. Handled exceptions that may be thrown by RateLimiter::set_params


christophfroehlich (commented):
If we set all parameters as read-only, there would not be the need for the changes in https://github.com/ros-controls/control_toolbox/pull/554

I think that we could allow to change the limits during runtime as mentioned [here](https://github.com/ros-controls/ros2_controllers/issues/1696) and [here](https://github.com/ros-controls/ros2_controllers/pull/1781#discussion_r2243642450). what do you think?

hjh059:
@christophfroehlich Hello!
My apologies for the late reply - I should have responded sooner.

Thanks for your feedback! You're absolutely right. I've reconsidered the approach and have now removed the read-only restriction on parameters as you suggested.

The changes are ready for your review. Thanks for your patience and the excellent suggestion.

PR #2080, opened 2025-12-19
tf prefix helper used in steering controllers library

kuralme:
Used tf frame prefix [helper](https://github.com/ros-controls/ros2_control/pull/2803) similar to the other 3 controller, but without the enable flag.

### DIFF
--- a/steering_controllers_library/src/steering_controllers_library.cpp
+++ b/steering_controllers_library/src/steering_controllers_library.cpp
@@ -277,6 +277,7 @@ controller_interface::CallbackReturn SteeringControllersLibrary::on_configure(
   }
 
   // resolve prefix: substitute tilde (~) with the namespace if contains and normalize slashes (/)
+  // Note: resolve_tf_prefix handles empty tf_frame_prefix by returning an empty string
   const std::string tf_prefix =
     controller_interface::resolve_tf_prefix(params_.tf_frame_prefix, get_node()->get_namespace());
 


steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -275,9 +276,37 @@ controller_interface::CallbackReturn SteeringControllersLibrary::on_configure(
...
+    tf_prefix = std::string(get_node()->get_namespace());
+    if (tf_prefix.back() != '/')
+    {
+      tf_prefix = tf_prefix + "/";
+    }
+    if (tf_prefix.front() == '/')
+    {
+      tf_prefix.erase(0, 1);
+    }
+  }
```
christophfroehlich: can't we remove this branch completely?

kuralme: Its redundant but I kept it to make the usage similar to the other controllers. Is it not better define like this and remove altogether when its time?

doc/migration.rst
```diff
@@ -30,3 +30,7 @@ omni_wheel_drive_controller
 *****************************
 * Instead of using ``tf_frame_prefix_enable:=false``, set an empty ``tf_frame_prefix:=""`` parameter instead. (`#2073 <https://github.com/ros-controls/ros2_controllers/pull/2073>`_).
 * For using node namespace as tf prefix: Set ``tf_frame_prefix:="~"``, where the ("~") character is substituted with node namespace. (`#2073 <https://github.com/ros-controls/ros2_controllers/pull/2073>`_).
+
+steering_controllers_library
+*****************************
+* Parameter ``tf_frame_prefix`` added with the similar functionality to other controllers. (`#2080 <https://github.com/ros-controls/ros2_controllers/pull/2080>`_).
```
christophfroehlich:
```suggestion
```

As this functionality was not present before, there is nothing to "migrate".

steering_controllers_library/src/steering_controllers_library.cpp
christophfroehlich: But this warning will be thrown for every existing config, because the default value of `tf_frame_prefix` is empty?

kuralme: Yeah it doesn't make much sense without the outer if block. I will just remove it.

doc/migration.rst
```diff
@@ -29,4 +29,4 @@ mecanum_drive_controller
 omni_wheel_drive_controller
 *****************************
 * Instead of using ``tf_frame_prefix_enable:=false``, set an empty ``tf_frame_prefix:=""`` parameter instead. (`#2073 <https://github.com/ros-controls/ros2_controllers/pull/2073>`_).
-* For using node namespace as tf prefix: Set ``tf_frame_prefix:="~"``, where the ("~") character is substituted with node namespace. (`#2073 <https://github.com/ros-controls/ros2_controllers/pull/2073>`_).
+* For using node namespace as tf prefix: Set ``tf_frame_prefix:="~"``, where the ("~") character is substituted with node namespace. (`#2073 <https://github.com/ros-controls/ros2_controllers/pull/2073>`_).
```
christophfroehlich:
```suggestion
```
nothing to migrate here :)

christophfroehlich (dismissed):
Thanks again, now LGTM!

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -275,9 +276,17 @@ controller_interface::CallbackReturn SteeringControllersLibrary::on_configure(
     return controller_interface::CallbackReturn::ERROR;
   }
 
+  // resolve prefix: substitute tilde (~) with the namespace if contains and normalize slashes (/)
```
Juliaj, line 279: nit, add a comment here to make the edge case (empty param) handling apparent to reader.

```suggestion
  // resolve prefix: substitute tilde (~) with the namespace if contains and normalize slashes (/)
  // Note: resolve_tf_prefix handles empty tf_frame_prefix by returning an empty string
```

Issue #942, opened 2023-12-22
resetOdometry as a ros2 srv


Fixit-Davide:
**Is your feature request related to a problem? Please describe.**
During debugging or testing session it could be useful to reset to odometry of a moving robot by simply calling a ROS2 service.

**Describe the solution you'd like**
A simple ROS2 service (std_srvs/srv/Empty or std_srvs/srv/Trigger) that, once triggered, calls for the method resetOdometry() of the implemented controller.

christophfroehlich:
In ros2_control, we already have a different thread which will process the service call. 

We already have a controller having implemented this as a service
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/tricycle_controller/src/tricycle_controller.cpp#L348-L352
But I see a problem that we don't have a thread-save implementation there to exchange data with the realtime loop. See the pid_controller, where we use a RealtimeBuffer for that (but a RealtimeBox would be better suited for that).
https://github.com/ros-controls/ros2_controllers/blob/d1844e6da63f25d384a0d81bf8a48fdc9a7861d2/pid_controller/src/pid_controller.cpp#L241-L255

Do you want to take this and implement this properly in all mobile robot controllers?

kuralme:
I can work on this

christophfroehlich:
@kuralme is this resolved, or is there something missing? steering_controllers_library?

kuralme:
Forgot that one, I'll work on it

christophfroehlich:
Can you please open a draft PR in removing all the deprecated reset methods? We can merge that after the next release of ros2_controllers

kuralme:
Sure

PR #2081, opened 2025-12-24
reset odometry service update - tricycle controller

kuralme:
First part of changes for addressing the [issue](https://github.com/ros-controls/ros2_controllers/issues/942):
- atomic reset flag
- reset is now checked in the update loop
- testing

Note: I redefined the  [odometry_](https://github.com/kuralme/ros2_controllers/blob/9dcdce6c6ba8aa607b782d0e469cad4419370211/tricycle_controller/test/test_tricycle_controller.cpp#L53) object in the test class. If this doesnt seem ok i can change it to something else, like adding an odom getter method

### DIFF
--- a/.github/workflows/rolling-semi-binary-build-win.yml
+++ b/.github/workflows/rolling-semi-binary-build-win.yml
@@ -29,4 +29,3 @@ jobs:
       ros_distro: rolling
       pixi_dependencies: typeguard jinja2 boost compilers octomap
       ninja_packages: rsl
-      target_cmake_args: -DBUILD_TESTING=OFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,7 +64,7 @@ repos:
 
   # CPP hooks
   - repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v21.1.8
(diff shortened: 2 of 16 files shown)


kuralme:
Happy new year.

Do I not need to mention this in release notes? If not, I will continue impementing this into the rest of the controllers

christophfroehlich:
> Happy new year.

Also a happy new year!

> Do I not need to mention this in release notes? If not, I will continue impementing this into the rest of the controllers

No because there is no new functionality here, you just have improved it by making it thread safe?

kuralme:
I guess it'll be new for other controllers, not this one. I'll continue with controllers that has odometry implementations

kuralme:
I forgot about this one. Unlike other three drive controllers this one still has reset odometry service. Should we merge it or close? @christophfroehlich 
Saying based on your previous statement:

> We plan to integrate tricycle_controller into the steering controller lib and deprecate it (at least before ROS-L is the goal) -> Let's don't spend too much time on it, but when we add this service to the the other controllers we could use the new service. The service definition can be added to control_msgs.

christophfroehlich:
As the work is already done, I'd say we can merge this as it is.

PR #2083, opened 2025-12-27
fix(steering_controllers): handle NaN/Inf values in odometry update

Ishan1923:
## Description
~~
This PR addresses a safety issue where invalid sensor data (NaN or Infinity) could corrupt the odometry state in steering controllers.

**Problem:**
Previously, if a sensor returned `NaN` or `Inf` (e.g., due to a hardware fault or connection glitch), the `SteeringOdometry` and `SteeringKinematics` update loops would perform calculations with these invalid values. This caused the robot's odometry to become invalid (`NaN`), potentially destabilizing the control loop or causing unpredictable behavior.

**Solution:**
1.  **Library Update:** Added `try_update_from_position` and `try_update_from_velocity` methods to `SteeringKinematics` (and the legacy `SteeringOdometry` wrapper). These methods perform a validity check on the inputs before updating the internal state.
2.  **Controller Update:** Updated `AckermannSteeringController` to use these new safe update methods. If the update fails (returns `false`), the controller logs a warning and skips the odometry update for that cycle, preserving the last known valid state.
~~
Edit: This PR addresses an issue that odometry is updated even if a reference interface (due to a timeout for example) is NaN.

## Testing
- [x] Compiled successfully with `colcon build`.
- [x] Verified unit tests pass with `colcon test`.
- [x] Ran static analysis and formatting checks via `pre-commit run`.

## Checklist
- [x] Fork the repository.
- [x] Modify the source; focus on the specific change.
- [x] Ensure local tests pass (`colcon test` and `pre-commit run`).
- [x] Send a pull request, answering any default questions.
- [x] Pay attention to automated CI failures.


saikishor (changes requested):
Why not implement the protection logic in the original methods instead of new ones?

Ishan1923:
Thank you for the review! I originally created new methods to be conservative and ensure I didn't break existing behavior/ABI. However, I agree that hardening the original methods is a much cleaner approach. I will update the PR with those changes shortly.

Ishan1923:
I have refactored the code based on the review.

**Hardened Existing Methods:** I moved the NaN/Inf checks inside the original update_from_position and update_from_velocity methods as requested. They now return false if invalid data is detected, keeping the API clean.

**Open Loop Handling:** The original update_open_loop returns void, so changing it to bool to propagate errors would break ABI compatibility. To solve this:

- I added a silent safety check inside the original update_open_loop (it just returns early on failure).

- I added a helper try_update_open_loop that returns bool, which allows the controller to detect failures explicitly without modifying the existing function signature.

ackermann_steering_controller/src/ackermann_steering_controller.cpp
```diff
@@ -59,7 +59,9 @@ bool AckermannSteeringController::update_odometry(const rclcpp::Duration & perio
 
   if (params_.open_loop)
   {
-    odometry_.update_open_loop(last_linear_velocity_, last_angular_velocity_, period.seconds());
+    if(!odometry_.try_update_open_loop(
+          last_linear_velocity_, last_angular_velocity_, period.seconds()
+        ))return false;
```
saikishor: Apply pre-commit changes

steering_controllers_library/include/steering_controllers_library/steering_odometry.hpp
```diff
@@ -116,6 +116,8 @@ class [[deprecated("Use steering_kinematics::SteeringKinematics instead")]] Stee
 
   void reset_odometry();
 
+  bool try_update_open_loop(double linear, double angular, double delTime);
```
saikishor: I'm not sure about the naming as we will always have a successful update call, but it may or maynot fail, so not sure about the naming here

@christophfroehlich what do you think about adding this new method? or we simply change the void return type to bool and have it all¿+?

saikishor (changes requested):

christophfroehlich (changes requested):
I don't see the problem as described in the initial post. All implementations have the pattern 
```
if (isfinite(state))
  odometry.update()
```
Where exactly is the odometry updated with infinite values?

Ishan1923:
> I don't see the problem as described in the initial post. All implementations have the pattern
> 
> ```
> if (isfinite(state))
>   odometry.update()
> ```
> 
> Where exactly is the odometry updated with infinite values?

@christophfroehlich I can point to the exact location. You are correct that the sensor update methods guard against this, but the Open Loop command path does not.

Here is the execution chain in the current master branch that leads to the issue:

1. In steering_kinematics.cpp, the integrate_fk method performs math directly on its inputs without checking them:

```
void SteeringKinematics::integrate_fk(const double v_bx, const double omega_bz, const double dt)
{
  const double delta_x_b = v_bx * dt; // <--- If v_bx is NaN, delta_x_b becomes NaN
  // ...
  // x_ += ... (Odometry state becomes permanently NaN here)
}
```

2. The function that calls this math, update_open_loop, acts as a pass-through. It does not check isfinite() before calling the integration:
```
void SteeringKinematics::update_open_loop(const double v_bx, const double omega_bz, const double dt)
{
    / <--- MISSING: if (!isfinite(v_bx)) return;
  
     linear_ = v_bx;
     angular_ = omega_bz;
     integrate_fk(v_bx, omega_bz, dt); // <--- Passes NaN inputs directly to the math above
}
```

3. In the controller's update loop, if open_loop is enabled, it passes the last command velocity directly to the library:
```
// (In AckermannSteeringController::update_odometry)
if (params_.open_loop) {
     // Uses command reference (which can be NaN from an upstream topic)
      odometry_.update_open_loop(last_linear_velocity_, last_angular_velocity_, period.seconds());
}
```

Because of this chain, a single NaN in the command topic (when in open-loop mode) bypasses all checks and permanently destroys the odometry state. 
My PR guards this specific entry point.

steering_controllers_library/include/steering_controllers_library/steering_odometry.hpp
Ishan1923: @saikishor Regarding the naming and the new method(which is redundant, yes):
 I introduced **try_update_open_loop** _(which returns bool)_ alongside the existing **update_open_loop** _(returning void)_ specifically to preserve _ABI compatibility._ Changing the return type of the existing function from void to bool would break binary compatibility for any downstream users linking against steering_controllers_library.

However, if you and @christophfroehlich prefer a cleaner API and are okay with breaking ABI for this release, I am happy to remove the try_ wrapper and modify the original update_open_loop to return bool directly. Just let me know your preference.

christophfroehlich:
In open_loop mode, there is no sensor involved.

Even more, see #2087, open_loop was broken and never updated without my fix. But I see the issue now if the reference interfaces are NaN. 

Please rebase your changes on top of mine.
This could be solved way simpler by just not updating the last*velocity variables if they are not finite, what do you think?

Please also add appropriate tests to steering_controllers_library, where it should be checked if the odometry is updated in open_loop mode, including the case of a timeout.

Ishan1923:
> In open_loop mode, there is no sensor involved.
> 
> Even more, see #2087, open_loop was broken and never updated without my fix. But I see the issue now if the reference interfaces are NaN.
> 
> Please rebase your changes on top of mine. This could be solved way simpler by just not updating the last*velocity variables if they are not finite, what do you think?
> 
> Please also add appropriate tests to steering_controllers_library, where it should be checked if the odometry is updated in open_loop mode, including the case of a timeout.

_@christophfroehlich thanks for the review and for the fix #2087_ ; yes, I think this way, it could be way simpler.

Working on this now:
1. Rebase on top of your latest changes.
2. Revert the library changes and implement the isfinite check directly in the controller when updating last_*_velocity_.
3. Add a test case to verify the behavior with NaN inputs as requested.

Ishan1923:
@christophfroehlich I have updated the PR based on your feedback.

Changes in this revision:
1. Reverted all modifications to `AckermannSteeringController`, `SteeringKinematics`, and `SteeringOdometry` to rely on the upstream fixes in #2087.
2. Implemented the `isfinite` check directly in `steering_controllers_library.cpp` (in `update_and_write_commands`) to prevent invalid values from overwriting the last known velocity.
3. Updated the regression test to verify that the controller safely stops (issuing `0.0` commands) when `NaN` input is received, rather than persisting the last value.

Ready for re-review!

Ishan1923:
Hi @christophfroehlich ,

I have updated the PR with the requested changes.
- Test Coverage: I added a new test case, test_open_loop_update_timeout, to steering_controllers_library.
- This test successfully reproduced the issue: before the fix, the robot continued moving at the last valid velocity even after the timeout occurred.
- The Fix: I updated update_and_write_commands to explicitly check for NaN values in the reference interfaces (which occur during a timeout). Now, if NaN is detected while in open-loop mode, the internal velocity states (last_linear_velocity_, etc.) are safely reset to 0.0.
   

```cpp
  if (std::isfinite(reference_interfaces_[0]))
  {
    last_linear_velocity_ = reference_interfaces_[0];
  }
  else{
    last_linear_velocity_ = 0.0;
  }
  if (std::isfinite(reference_interfaces_[1]))
  {
    last_angular_velocity_ = reference_interfaces_[1];
  }
  else{
    last_angular_velocity_ = 0.0;
  }
  update_odometry(period);
  
```

christophfroehlich (changes requested):
Please install pre-commit in your workspace, and activate it for this repository (pre-commit install). To fix the failing tests now, simply run `pre-commit run --all`

Ishan1923:
Pushed the style fixes via pre-commit. The linting checks should pass now. Ready for review!

christophfroehlich:
Please don't do force pushes if a PR is already under review, as this just complicates new reviews.

Ishan1923:
Apologies for the force push. I wasn't aware that it disrupts the review history. I wanted to keep the commit log clean after the pre-commit fixes, but I will stick to regular commits/pushes for future updates to preserve the diff context. I again genuinely apologize.

Ishan1923:
I have added the missing <cmath> header to fix the strict build compilation error on Linux. I verified locally that the strict build passes now.

Regarding the other failure: I noticed it failed with an Authentication required error _(Invoke-WebRequest : Authentication required)_  during a download, which I don't see is related to my changes.

Could you please re-trigger the workflows?

christophfroehlich:
> Regarding the other failure: I noticed it failed with an Authentication required error _(Invoke-WebRequest : Authentication required)_ during a download, which I don't see is related to my changes.

Failing windows build is known and not related.

christophfroehlich:
> Apologies for the force push. I wasn't aware that it disrupts the review history. I wanted to keep the commit log clean after the pre-commit fixes, but I will stick to regular commits/pushes for future updates to preserve the diff context. I again genuinely apologize.

There is no need for a linear history, as we use the "squash" strategy when PRs get merged.

Ishan1923:
Thanks for the clarification on the force-push policy and the Windows build flake. I noticed the pre-commit check failed on the previous push (likely due to header sorting), so I've pushed a quick style fix. Everything should be green now!

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -386,6 +386,90 @@ TEST_F(SteeringControllersLibraryTest, test_velocity_feedback_ref_timeout)
...
+
+  pub_controller->input_ref_.set(nan_msg);
+
+  controller_->update_reference_from_subscribers(
+    controller_->get_node()->now(), rclcpp::Duration::from_seconds(0.01));
+  controller_->update_and_write_commands(
+    controller_->get_node()->now(), rclcpp::Duration::from_seconds(0.01));
+
+  // The wheel speed should stay exactly the same
+  EXPECT_DOUBLE_EQ(pub_controller->command_interfaces_[0].get_optional().value(), 0.0);
```
christophfroehlich:
```suggestion
  // The wheel speed should have been reset to 0.0
  EXPECT_DOUBLE_EQ(pub_controller->command_interfaces_[0].get_optional().value(), 0.0);
```
I think the comment is wrong?

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -520,8 +521,22 @@ controller_interface::return_type SteeringControllersLibrary::update_and_write_c
   auto logger = get_node()->get_logger();
 
   // store current ref (for open loop odometry) and update odometry
-  last_linear_velocity_ = reference_interfaces_[0];
-  last_angular_velocity_ = reference_interfaces_[1];
+  if (std::isfinite(reference_interfaces_[0]))
```
christophfroehlich: Do we need to cover infinite values here?
If we just care about non-nan, we could move these assignments in the part below, and move the update_odometry() to the end of the update_and_write_commands() method.

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -386,6 +386,90 @@ TEST_F(SteeringControllersLibraryTest, test_velocity_feedback_ref_timeout)
...
+  pub_controller->input_ref_.set(command_msg);
+
+  controller_->update_reference_from_subscribers(
+    controller_->get_node()->now(), rclcpp::Duration::from_seconds(0.01));
+  controller_->update_and_write_commands(
+    controller_->get_node()->now(), rclcpp::Duration::from_seconds(0.01));
+
+  double wheel_speed_1 = pub_controller->command_interfaces_[0].get_optional().value();
+
+  ASSERT_GT(wheel_speed_1, 0.1);
```
christophfroehlich:
```suggestion
  ASSERT_GT(pub_controller->command_interfaces_[0].get_optional().value(), 0.1);
```

steering_controllers_library/test/test_steering_controllers_library.cpp
```diff
@@ -386,6 +386,90 @@ TEST_F(SteeringControllersLibraryTest, test_velocity_feedback_ref_timeout)
...
+
+  controller_->update(controller_->get_node()->now(), rclcpp::Duration::from_seconds(0.1));
+
+  EXPECT_DOUBLE_EQ(controller_->last_linear_velocity_, 5.0);
+
+  rclcpp::Time future_time = controller_->get_node()->now() + rclcpp::Duration::from_seconds(2.0);
+
+  controller_->update(future_time, rclcpp::Duration::from_seconds(0.1));
+
+  EXPECT_DOUBLE_EQ(controller_->last_linear_velocity_, 0.0);
```
christophfroehlich, line 482: Maybe we can also test the content of 
controller_->odom_state_msg_? It should remain the same even after another two seconds.

christophfroehlich (changes requested):
The logic now LGTM, thanks for also adding tests.

I have only some minor comments for improvement.

Ishan1923:
@christophfroehlich I have updated the PR with the requested changes:

- Refactored the logic to ensure update_odometry happens at the end.
- Added the safety check for isfinite.
- Updated the tests to verify the public message state (odom_state_msg_) and fixed the comments.

Ready for re-review!

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -661,19 +657,36 @@ controller_interface::return_type SteeringControllersLibrary::update_and_write_c
         RCLCPP_DEBUG(
           logger, "Unable to retrieve %s for steering wheel %zu",
           !state_interface_value_op.has_value() ? "state interface value"
-                                                : "command interface value",
-          i);
+                                                : "command interface value", );
```
christophfroehlich:
```suggestion
                                                : "command interface value", i);
```

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -661,19 +657,36 @@ controller_interface::return_type SteeringControllersLibrary::update_and_write_c
         RCLCPP_DEBUG(
           logger, "Unable to retrieve %s for steering wheel %zu",
           !state_interface_value_op.has_value() ? "state interface value"
-                                                : "command interface value",
-          i);
+                                                : "command interface value", i);
```
christophfroehlich:
```suggestion
                                                : "command interface value",
          i);
```

saikishor (dismissed):
Looks good to me

steering_controllers_library/test/test_steering_controllers_library.cpp
christophfroehlich: We should also check above if `controller_->odom_state_msg_.twist.twist.linear.x` was nonzero before.

christophfroehlich (changes requested):
I fixed the compilation- and pre-commit errors. One minor request for the tests, then this is good to merge

Ishan1923:
I added some checks in :
- **test_position_feedback_ref_timeout**, which now checks for the velocity if greater than 0.0, before the condition _age_of_last_command > ref_timeout_  gets checked.
- similarly in **test_velocity_feedback_ref_timeout**; velocity gets check before the condition _age_of_last_command > ref_timeout_  get checked.
- In **test_open_loop_update_timeout**, i removed the last line _EXPECT_DOUBLE_EQ(controller_->odom_state_msg_.twist.twist.linear.x, 0.0);_ which checking for the wrong velocity value [1 cycle behind / lag] [The internal variable _last_linear_velocity__ updates immediately].

_Also, in steering_controllers_library.cpp, at line 656 in the RCLCPP_DEBUG, the variable 'i' was missing, so i added it._ 

Pls check. Ready of the review :)

steering_controllers_library/test/test_steering_controllers_library.cpp
christophfroehlich: @Ishan1923 This is still missing, right?

christophfroehlich:
@Ishan1923 could you please merge the current master into your PR and fix the conflicts?

Ishan1923:
@christophfroehlich The master branch is merged and conflicts are resolved.
- Added move_robot(0.0, 0.0, 0.0); to the odometry_set_service test to clear simulated wheel feedback.
- This prevents the open-loop math from integrating leftover feedback and causing Y-axis drift.
- All local builds and tests are passing successfully.

steering_controllers_library/test/test_steering_controllers_library.cpp
Ishan1923: Addressed in commit 91c84ff (Feb 8); added a second update cycle before the timeout jump and verified `odom_state_msg_.twist.twist.linear.x > 0.0` in the `test_open_loop_update_timeout test`.
Please let me know if anything is still missing.

Ishan1923:
@christophfroehlich the move_robot argument bug in the test has been fixed and pushed. The remaining CI failures appear to be infrastructure-related (rosdep cannot resolve packages for OS version 'resolute'). Could you please re-review when you get a chance?

christophfroehlich (changes requested):
There was a change in test patterns with https://github.com/ros-controls/ros2_controllers/pull/2399
Please update the tests added by this PR.

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -727,10 +728,28 @@ controller_interface::return_type SteeringControllersLibrary::update_and_write_c
...
   }
 
+  // store current ref (for open loop odometry) and update odometry
+  if (std::isfinite(reference_interfaces_[0]))
+  {
+    last_linear_velocity_ = reference_interfaces_[0];
+  }
+  else
+  {
+    last_linear_velocity_ = 0.0;
```
MarqRazz: Shouldn't we warn the user we got an invalid state from the hardware and that we are assuming it is not moving?

PR #2086, opened 2025-12-31
Add battery_state_broadcaster

christophfroehlich:
This moves https://github.com/ipa320/ros_battery_monitoring/tree/main/battery_state_broadcaster to our repository.

For discsussion about this, see #1888 and especially https://github.com/ros-controls/ros2_controllers/pull/1888#issuecomment-3667106815


MichaelT64:
@christophfroehlich are there any further plans for this?

christophfroehlich:
Please have a look in #1888 and leave your review there.

battery_state_broadcaster/test/test_battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,306 @@
...
+TEST_F(BatteryStateBroadcasterTest, deactivate_success)
+{
+  SetUpBatteryStateBroadcaster();
+
+  ASSERT_TRUE(configure_succeeds(battery_state_broadcaster_));
+  ASSERT_TRUE(activate_succeeds(battery_state_broadcaster_));
+  ASSERT_TRUE(deactivate_succeeds(battery_state_broadcaster_));
+}
+
+TEST_F(BatteryStateBroadcasterTest, check_exported_intefaces)
```
christophfroehlich:
```suggestion
TEST_F(BatteryStateBroadcasterTest, check_exported_interfaces)
```

battery_state_broadcaster/test/test_battery_state_broadcaster.hpp
```diff
@@ -0,0 +1,246 @@
...
+  using BatteryStateBroadcaster::counts_;
+  using BatteryStateBroadcaster::params_;
+  using BatteryStateBroadcaster::sums_;
+
+  FRIEND_TEST(BatteryStateBroadcasterTest, init_success);
+  FRIEND_TEST(BatteryStateBroadcasterTest, all_parameters_set_configure_success);
+  FRIEND_TEST(BatteryStateBroadcasterTest, no_interfaces_set_activate_fail);
+  FRIEND_TEST(BatteryStateBroadcasterTest, activate_success);
+  FRIEND_TEST(BatteryStateBroadcasterTest, deactivate_success);
+  FRIEND_TEST(BatteryStateBroadcasterTest, check_exported_intefaces);
```
christophfroehlich:
```suggestion
  FRIEND_TEST(BatteryStateBroadcasterTest, check_exported_interfaces);
```

battery_state_broadcaster/package.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>battery_state_broadcaster</name>
+  <version>0.1.0</version>
```
christophfroehlich:
```suggestion
  <version>6.8.0</version>
```

.github/workflows/kilted-binary-build.yml
```diff
@@ -7,8 +7,6 @@ on:
   pull_request: &event
     branches:
       - kilted
-      - '*feature*'
-      - '*feature/**'
```
saikishor: Should we seperate CI changes into a separate PR?

battery_state_broadcaster/src/battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,488 @@
...
+    raw_battery_states_msg.battery_states.emplace_back(std::move(battery));
+  }
+
+  // Get count of enabled batteries for each interface
+  for (size_t i = 0; i < params_.batteries.size(); ++i)
+  {
+    const auto & battery_name = params_.batteries.at(i);
+    const auto interfaces_it = params_.interfaces.batteries_map.find(battery_name);
+    const auto props_it = params_.batteries_map.find(battery_name);
+    if (interfaces_it == params_.interfaces.batteries_map.end() || props_it == params_.batteries_map.end())
```
christophfroehlich:
```suggestion
    if (
      interfaces_it == params_.interfaces.batteries_map.end() ||
      props_it == params_.batteries_map.end())
```

.github/workflows/kilted-binary-build.yml
christophfroehlich: I just wanted to deactivate the builds in this PR. but I can cherry-pick it into a separate one

battery_state_broadcaster/src/battery_state_broadcaster_parameters.yaml
```diff
@@ -0,0 +1,121 @@
+battery_state_broadcaster:
+  batteries: {
+    type: string_array,
+    description: "List of batteries from which battery state interfaces will be read.",
+    read_only: true,
+    default_value: [],
```
saikishor:
```suggestion
```

battery_state_broadcaster/src/battery_state_broadcaster_parameters.yaml
```diff
@@ -0,0 +1,121 @@
...
+    validation: {
+      unique<>: null
```
saikishor, line 7: Should we also add the not_empty validator?

battery_state_broadcaster/src/battery_state_broadcaster.cpp
```diff
@@ -0,0 +1,492 @@
...
+
+      raw_battery_states_msg_.battery_states[i].voltage = static_cast<float>(
+        state_interfaces_[interface_cnt].get_optional<double>().value_or(kUninitializedValue));
+      sums_.voltage_sum += raw_battery_states_msg_.battery_states[i].voltage;
+      interface_cnt++;
+
+      if (interfaces.battery_temperature)
+      {
+        raw_battery_states_msg_.battery_states[i].temperature = static_cast<float>(
+          state_interfaces_[interface_cnt].get_optional<double>().value_or(kUninitializedValue));
```
saikishor, line 348: In the math here, I've a question, what if one of the battery doesn't have the interface, then we are summing NaN and resulting in the bad mean right?

saikishor (changes requested):
Some nitpicks

battery_state_broadcaster/src/battery_state_broadcaster_parameters.yaml
christophfroehlich: This is not possible for the compatibility with the configuration of the old plugin.

battery_state_broadcaster/src/battery_state_broadcaster.cpp
christophfroehlich: But `interfaces.battery_temperature` only is true if the interface exists?

battery_state_broadcaster/src/battery_state_broadcaster_parameters.yaml
saikishor: Ok, but then removing the default value will cause same issue? Forcing users to define it?. IIRC, empty vector is something not compatible to define in yaml, it simply fails with not set and no clear instructions

battery_state_broadcaster/src/battery_state_broadcaster.cpp
saikishor: What I mean is, what if it might be true for some of the batteries but not each of them?

saikishor (commented):
Left some comments

battery_state_broadcaster/src/battery_state_broadcaster.cpp
christophfroehlich: `interfaces` is a map with an entry for every battery?

battery_state_broadcaster/src/battery_state_broadcaster_parameters.yaml
christophfroehlich: I was curious now, but I don't see that this shouldn't work.
The difference in the generated code is

`        std::vector<std::string> batteries = {};`
vs
`        std::vector<std::string> batteries;`.
and an assignment during parameter declaration.

I added a test for the legacy config where `batteries` parameter is empty, and it succeeds.

saikishor: Sureee. If you say so, then ok. I had issues in the past with this approach. If I add a emptyone in the yaml, it used to crash

battery_state_broadcaster/src/battery_state_broadcaster.cpp
saikishor: Image

I believe so

christophfroehlich: that's what I mean: If it is not available for certain batteries and configured properly, than it won't accumulate NaN here?

battery_state_broadcaster/src/battery_state_broadcaster_parameters.yaml
christophfroehlich: please report upstream if you encounter any issue in the future

saikishor: Ok

saikishor (approved):

PR #2096, opened 2026-01-02
set odometry service addition - diff drive controller

kuralme:
adds service to reset odometry, works similar to the [updated](https://github.com/ros-controls/ros2_controllers/pull/2081) one in tricycle controller. Addressing the [issue](https://github.com/ros-controls/ros2_controllers/issues/942). Key additions:
- atomic reset flag
- reset is executed in the update loop after service call
- testing

### DIFF
--- a/diff_drive_controller/CMakeLists.txt
+++ b/diff_drive_controller/CMakeLists.txt
@@ -18,7 +18,6 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   rclcpp_lifecycle
   rcpputils
   realtime_tools
-  std_srvs
   tf2
   tf2_msgs
 )
@@ -63,7 +62,6 @@ target_link_libraries(diff_drive_controller
     ${tf2_msgs_TARGETS}
     ${geometry_msgs_TARGETS}
     ${nav_msgs_TARGETS}
-    ${std_srvs_TARGETS}
     ${control_msgs_TARGETS})
 pluginlib_export_plugin_description_file(controller_interface diff_drive_plugin.xml)
 
--- a/diff_drive_controller/package.xml
+++ b/diff_drive_controller/package.xml
@@ -37,7 +37,6 @@
   <depend>rclcpp_lifecycle</depend>
   <depend>rcpputils</depend>
   <depend>realtime_tools</depend>
-  <depend>std_srvs</depend>
   <depend>tf2</depend>
   <depend>tf2_msgs</depend>
 
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -172,7 +172,7 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
     if (param_op.has_value())
     {
       auto params = param_op.value();
-      odometry_.setOdometry(params->x, params->y, params->yaw);
+      odometry_.setOdometry(params.x, params.y, params.yaw);
       odometry_updated = true;
       set_odom_requested_.store(false);
     }


kuralme:
That sounds like a nice feature. Shall i start working on it starting with tricycle?
How about adding it as a new service called set_odometry and keep the reset?

christophfroehlich:
We plan to integrate tricycle_controller into the steering controller lib and deprecate it (at least before ROS-L is the goal) -> Let's don't spend too much time on it, but when we add this service to the the other controllers we could use the new service. The service definition can be added to control_msgs.

kuralme:
Can you review this? @christophfroehlich @saikishor 

I will update [#2110](https://github.com/ros-controls/ros2_controllers/pull/2110) and create PR for omni drive next

diff_drive_controller/CMakeLists.txt
```diff
@@ -60,7 +62,8 @@ target_link_libraries(diff_drive_controller
     tf2::tf2
     ${tf2_msgs_TARGETS}
     ${geometry_msgs_TARGETS}
-    ${nav_msgs_TARGETS})
+    ${nav_msgs_TARGETS}
+    ${std_srvs_TARGETS})
```
saikishor:
```suggestion
    ${std_srvs_TARGETS}
    ${control_msgs_TARGETS})
```

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
```diff
@@ -154,8 +159,13 @@ class DiffDriveController : public controller_interface::ChainableControllerInte
...
   rclcpp::Time previous_publish_timestamp_{0, 0, RCL_CLOCK_UNINITIALIZED};
 
+  rclcpp::Service<control_msgs::srv::SetOdometry>::SharedPtr set_odom_service_;
   rclcpp::Service<std_srvs::srv::Empty>::SharedPtr reset_odom_service_;
-  std::atomic<bool> reset_odom_{false};
+  std::atomic<bool> set_odom_request_{false}, reset_odom_request_{false};
+  struct
+  {
+    double x, y, yaw;
+  } requested_odom_params_;
```
christophfroehlich: Should we use a RealtimeBox for this struct?

diff_drive_controller/include/diff_drive_controller/odometry.hpp
```diff
@@ -55,6 +55,7 @@ class Odometry
   bool update_from_pos(double left_pos, double right_pos, double dt);
   bool update_from_vel(double left_vel, double right_vel, double dt);
   bool try_update_open_loop(double linear_vel, double angular_vel, double dt);
+  void setOdometry(double x, double y, double heading);
   void resetOdometry();
```
christophfroehlich:
```suggestion
  [[deprecated("Use setOdometry(0.0, 0.0, 0.0) instead")]] void resetOdometry();
```
We could deprecate this, just to simplify things?

kuralme:
IMHO reset odom is useful, simpler to call and likely to be used more often then set. But can remove it anyway if you think its redundant

christophfroehlich:
> IMHO reset odom is useful, simpler to call and likely to be used more often then set. But can remove it anyway if you think its redundant

@saikishor do you have an opinion about that?

diff_drive_controller/include/diff_drive_controller/diff_drive_controller.hpp
```diff
@@ -147,6 +159,17 @@ class DiffDriveController : public controller_interface::ChainableControllerInte
...
+  rclcpp::Service<control_msgs::srv::SetOdometry>::SharedPtr set_odom_service_;
+  rclcpp::Service<std_srvs::srv::Empty>::SharedPtr reset_odom_service_;
+  std::atomic<bool> set_odom_request_{false}, reset_odom_request_{false};
+  struct OdomParams
+  {
+    double x = 0.0;
+    double y = 0.0;
+    double yaw = 0.0;
+  };
+  realtime_tools::RealtimeThreadSafeBox<OdomParams> requested_odom_params_;
```
christophfroehlich:
```suggestion
  realtime_tools::RealtimeThreadSafeBox<control_msgs::srv::SetOdometry::Request> requested_odom_params_;
```
we maybe could just use the message struct?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -556,6 +588,30 @@ controller_interface::CallbackReturn DiffDriveController::on_error(const rclcpp_
...
+void DiffDriveController::reset_odometry(
+  const std::shared_ptr<rmw_request_id_t> /*request_header*/,
+  const std::shared_ptr<std_srvs::srv::Empty::Request> /*req*/,
+  std::shared_ptr<std_srvs::srv::Empty::Response> /*res*/)
+{
+  // flip the flag for thread-safe odom reset in the control loop
+  reset_odom_request_.store(true);
+  RCLCPP_INFO(get_node()->get_logger(), "Odometry reset requested");
+}
+
```
saikishor, line 604: I'm not sure if I like the logic here. Because, you set the variable to True and you simple return the service as it is finished. You are not waiting or not even checking if the controller is active.

saikishor:
> > IMHO reset odom is useful, simpler to call and likely to be used more often then set. But can remove it anyway if you think its redundant
> 
> @saikishor do you have an opinion about that?

I would leave it for legacy reasons at least. If I remember well, there are some rqt interfaces or Rviz2 plugins that might use this service.

christophfroehlich:
> > > IMHO reset odom is useful, simpler to call and likely to be used more often then set. But can remove it anyway if you think its redundant
> > 
> > 
> > @saikishor do you have an opinion about that?
> 
> I would leave it for legacy reasons at least. If I remember well, there are some rqt interfaces or Rviz2 plugins that might use this service.

This service did not exit before ;) But I'm fine with that.

saikishor:
@christophfroehlich LoL i thought it did, in this case i agree with @christophfroehlich let's remove it. And anyway calling set with default arguments is basically resetting it

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -161,56 +164,75 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
-  if (params_.open_loop)
+
+  // check if odometry set or reset was requested by non-RT thread
+  if (set_odom_request_.exchange(false))
+  {
+    control_msgs::srv::SetOdometry::Request params;
+    requested_odom_params_.get(params);
+
+    odometry_.setOdometry(params.x, params.y, params.yaw);
+    odometry_updated = true;
```
christophfroehlich: let's use the non-blocking call here. we could leave the atomic variable then an try in the next loop
```suggestion
    if (requested_odom_params_.try_get(params))
    {
      odometry_.setOdometry(params.x, params.y, params.yaw);
      odometry_updated = true;
    }
```

diff_drive_controller/src/diff_drive_controller.cpp
kuralme: I added controller state check to the service, but opt out of adding a timeout loop (to wait odom setting done). That would create problems for a single threaded executor (or is there a method I am not aware of?). 

I may switch to action instead of service but I doubt you would want that.

diff_drive_controller/include/diff_drive_controller/odometry.hpp
christophfroehlich: I think this still applies.

diff_drive_controller/CMakeLists.txt
```diff
@@ -60,7 +62,9 @@ target_link_libraries(diff_drive_controller
     tf2::tf2
     ${tf2_msgs_TARGETS}
     ${geometry_msgs_TARGETS}
-    ${nav_msgs_TARGETS})
+    ${nav_msgs_TARGETS}
+    ${std_srvs_TARGETS}
```
christophfroehlich:
```suggestion
```

diff_drive_controller/CMakeLists.txt
```diff
@@ -17,6 +18,7 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   rclcpp_lifecycle
   rcpputils
   realtime_tools
+  std_srvs
```
christophfroehlich:
```suggestion
```

diff_drive_controller/package.xml
```diff
@@ -36,6 +37,7 @@
   <depend>rclcpp_lifecycle</depend>
   <depend>rcpputils</depend>
   <depend>realtime_tools</depend>
+  <depend>std_srvs</depend>
```
christophfroehlich:
```suggestion
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -161,56 +163,70 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
+  // check if odometry set or reset was requested by non-RT thread
+  if (set_odom_requested_.exchange(false))
   {
-    odometry_updated =
-      odometry_.try_update_open_loop(linear_command, angular_command, period.seconds());
+    if (auto params = requested_odom_params_.try_get())
+    {
+      odometry_.setOdometry(params->x, params->y, params->yaw);
+      odometry_updated = true;
+    }
```
christophfroehlich: Oh sorry, I had a look at the wrong template specialization. It returns an optional, this would be safer

```suggestion
    auto param_op = requested_odom_params_.try_get();
    if (param_op.has_value())
    {
      auto params = param_op.value();
      odometry_.setOdometry(params->x, params->y, params->yaw);
      odometry_updated = true;
      set_odom_requested_.store(false);
    }
```

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -161,56 +163,70 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
   const double left_wheel_radius = params_.left_wheel_radius_multiplier * params_.wheel_radius;
   const double right_wheel_radius = params_.right_wheel_radius_multiplier * params_.wheel_radius;
 
-  // Update odometry
   bool odometry_updated = false;
-  if (params_.open_loop)
+
+  // check if odometry set or reset was requested by non-RT thread
+  if (set_odom_requested_.exchange(false))
```
christophfroehlich: Let's clear the flag only if the box access succeeded
```suggestion
  if (set_odom_requested_.load())
```

diff_drive_controller/src/diff_drive_controller.cpp
christophfroehlich: I would not overcomplicate things here. the response from the service is "reset requested", which is exactly what it does.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -161,56 +163,73 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
...
+  // check if odometry set or reset was requested by non-RT thread
+  if (set_odom_requested_.load())
   {
-    odometry_updated =
-      odometry_.try_update_open_loop(linear_command, angular_command, period.seconds());
+    auto param_op = requested_odom_params_.try_get();
+    if (param_op.has_value())
+    {
+      auto params = param_op.value();
+      odometry_.setOdometry(params->x, params->y, params->yaw);
```
christophfroehlich: sorry :D 
```suggestion
      odometry_.setOdometry(params.x, params.y, params.yaw);
```

diff_drive_controller/src/diff_drive_controller.cpp
saikishor: IMO the service shouldn't return success unless it has completed doing so. If you are fine with it, fine for me.

P.S. there won't be a problem with single threaded executor here as the update runs in a realtime thread, so you won't have any issues

saikishor (approved):

PR #2101, opened 2026-01-08
fix(joint-trajectory-controller): use active tolerances in update step

### LINKED ISSUE #2100: Joint Trajectory Controller tolerances
**Describe the bug**
PR https://github.com/ros-controls/ros2_controllers/pull/716 introduces the concept of *active tolerances* to the Joint Trajectory Controller, in order to respect the tolerances sent with the action goal. From that point on, the tolerances that were used on the `update` step of the controller were taken from a `RealtimeBuffer`. There seems to be just one place where this change wasn't propagated and where the *live* tolerances are being used.

**To Reproduce**
1. Go to https://github.com/ros-controls/ros2_controllers/blob/c703235f0c904d3f4aeebcbf26caf582b9303267/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L258
  This is where the active tolerances for the current update step are read.
2. Scroll down to https://github.com/ros-controls/ros2_controllers/blob/c703235f0c904d3f4aeebcbf26caf582b9303267/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L303
  This is where the live `default_tolerances_` are still being used, even though all other checks around this use `active_tol`

**Expected behavior**
I would expect this to be
```
              check_state_tolerance_per_joint(
                state_error_, index, active_tol->goal_state_tolerance[index],
                true /* show_errors */);
```
Otherwise, it would technically be possible to change the goal state tolerances while the trajectory is being executed, something that the controller is trying to avoid in the current implementation.

domire8:
This PR fixes issue #2100


domire8:
I believe this would still be valid @christophfroehlich ?

MarqRazz (commented):
What happens when a goal includes tolerances and then a follow up one does not? Will it fall back to the `default_tolerances_` when they are not specified?

domire8:
> What happens when a goal includes tolerances and then a follow up one does not? Will it fall back to the `default_tolerances_` when they are not specified?

The way I understand [this part](https://github.com/ros-controls/ros2_controllers/blob/9d2e196734e7b44812e0529db21f0aed92f71a32/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1425), yes.

domire8:
This PR is not stale

christophfroehlich (approved):
LGTM, thank you!

PR #2103, opened 2026-01-12
[gpio_controllers] Fix crash on NaN state values and sanitize inputs

### LINKED ISSUE #1970: Boolean data type signals are not supported by GPIO Controller
**Describe the bug**

After upgrading my ros2_control version from v 4.37.0 to this commit [hash](https://github.com/ros-controls/ros2_control/commit/2c186d2a88e1a95c2f215b39382008db77c094fa) where boolean data types are handle, I noticed that gpio controller is triggering the following error:

```bash
[ros2_control_node-2] Exception thrown during reading the state of: GeneralSafetyIn/RunOk
[ros2_control_node-2] Exception thrown during reading the state of: PhysDigitalInputs/TruckOnState
[ros2_control_node-2] Exception thrown during reading the state of: PhysDigitalInputs/LoadFlap
```

That error comes from [here](https://github.com/ros-controls/ros2_controllers/blob/jazzy/gpio_controllers/src/gpio_command_controller.cpp#L412-L439) where it didn't consider between different data types:

How do I fixed?

apply_command function
```cpp
    auto data_type = command_interface.get_data_type();
    bool success;
    if (data_type == hardware_interface::HandleDataType::DOUBLE) {
      success = command_interface.set_value<double>(
        gpio_commands.interface_values[gpio_index].values[command_interface_index]);
    } else if (data_type == hardware_interface::HandleDataType::BOOL) {
      // To be discussed with Jonas
      success = command_interface.set_value<bool>(
        gpio_commands.interface_values[gpio_index].values[command_interface_index]);
    } else {
      RCLCPP_ERROR(
        get_node()->get_logger(), "Unsupported data type for interface %s",
        full_command_interface_name.c_str());
      return;
    }
    (void)success;
  } catch (const std::exception & e) {
    fprintf(
      stderr, "Exception thrown during applying command stage of %s with message: %s \n",
      full_command_interface_name.c_str(), e.what());
  }
```
apply_command

```cpp
    auto & command_interface = command_interfaces_map_.at(full_command_interface_name).get();
    auto data_type = command_interface.get_data_type();
    bool success;
    if (data_type == hardware_interface::HandleDataType::DOUBLE) {
      success = command_interface.set_value<double>(
        gpio_commands.interface_values[gpio_index].values[command_interface_index]);
    } else if (data_type == hardware_interface::HandleDataType::BOOL) {
      // To be discussed with Jonas
      success = command_interface.set_value<bool>(
        gpio_commands.interface_values[gpio_index].values[command_interface_index] < 0.5 ? false : true);
    } else {
      RCLCPP_ERROR(
        get_node()->get_logger(), "Unsupported data type for interface %s",
        full_command_interface_name.c_str());
      return;
    }
```

ros2_control.config.yaml
```yaml
gpio_controller:
  ros__parameters:
    type: gpio_controllers/GpioCommandController
    gpios:
      - PhysDigitalOutputs
      - GeneralSafetyIn
      - PhysDigitalInputs
      - PushThrResp
    command_interfaces:
      PhysDigitalOutputs:
        - interfaces:
          - TruckOn
    state_interfaces:
      GeneralSafetyIn:
        - interfaces:
          - RunOK
      PhysDigitalInputs:
        - interfaces:
          - TruckOnState
          - LoadFlap
      PushThrResp:
        - interfaces:
          - PushThroughDist
```

urdf example
```
<gpio name="GeneralSafetyIn">
      <state_interface name="RunOK" data_type="bool"></state_interface>
</gpio>
```

**Expected behavior**
No error thrown when a bool signal arrives

**Screenshots**

 - OS: 24.04
 - Version Jazzy
- Anything that may be unusual about your environment

**Additional context**
Add any other context about the problem here, especially include any modifications to ros2_control that relate to this issue.

Ishan1923:
This PR attempts to address Issue #1970 ("Boolean data type signals are not supported by GPIO Controller"). Users were encountering crashes and exceptions when reading states, likely due to uninitialized values or type mismatches.

I have introduced a `sanitize_double` helper function to handle state readings.

**Approach & Reasoning:**
Instead of a hard cast to bool, this function detects `NaN` (uninitialized memory) or invalid values and defaults them safely to `0.0`. This stops the crashes while preserving the original `double` value if valid.

This supports both use cases:
* **Boolean logic:** Works as expected (Values like 0.0 and 1.0 pass through).
* **Analog logic:** Remains supported (Values like 3.1 or 5.0 are preserved).

I have also updated the `ReproduceBadCastCrash` test case to `EXPECT_NO_THROW` to verify that these edge cases are now handled gracefully.

Resolves #1970


JavierIntermodalicsKion:
Any news on this? Thanks in advance for all the effort @Ishan1923

Ishan1923:
> Any news on this? Thanks in advance for all the effort @Ishan1923

Hi @JavierIntermodalicsKion, thanks for checking in! The PR is fully ready and tested on my end. Currently, I am just waiting for a maintainer to approve the CI workflows (since I am a new contributor, they don't run automatically) and for a code review. Fingers crossed we can get this merged soon!

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -264,7 +275,8 @@ void GpioCommandController::initialize_gpio_state_msg()
       get_gpios_state_interfaces_names(gpio_name);
     gpio_state_msg_.interface_values[gpio_index].values = std::vector<double>(
       gpio_state_msg_.interface_values[gpio_index].interface_names.size(),
-      std::numeric_limits<double>::quiet_NaN());
+      sanitize_double(std::numeric_limits<double>::quiet_NaN()));  ////////////////////////////////////////////////
+                                                                   ///Change1
```
Juliaj: Please remove the comment here.

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -429,7 +441,8 @@ void GpioCommandController::apply_state_value(
     else
     {
       state_msg.interface_values[gpio_index].values[interface_index] =
-        state_msg_interface_value_op.value();
+
+        sanitize_double(state_msg_interface_value_op.value());  /////////////////////change2
```
Juliaj: Same here, remove the comment.

gpio_controllers/test/test_load_gpio_command_controller.cpp
```diff
@@ -35,3 +65,53 @@ TEST(TestLoadGpioCommandController, load_controller)
...
   rclcpp::shutdown();
 }
+
+class TestableGpioController : public gpio_controllers::GpioCommandController
+{
+public:
+  using gpio_controllers::GpioCommandController::assign_interfaces;
+};
+
+TEST(TestLoadGpioCommandController, ReproduceBadCastCrash)
```
Juliaj: Consider renaming the test to describe the behavior it verifies. For example: UpdateBoolGpioInterfaces.

gpio_controllers/test/test_load_gpio_command_controller.cpp
```diff
@@ -35,3 +65,53 @@ TEST(TestLoadGpioCommandController, load_controller)
...
+
+  ASSERT_EQ(controller->on_activate(rclcpp_lifecycle::State()), controller_interface::CallbackReturn::SUCCESS);
+  
+  // This verifies that the controller no longer crashes on update
+  EXPECT_NO_THROW(
+    controller->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01))
+  );
+
+  rclcpp::shutdown();
+}
```
Juliaj: Make sure you run "pre-commit run -a" with your PR changes.

Ishan1923:
@Juliaj Thanks for the review! I have addressed your comments:
- Removed the temporary comments.
- Renamed the test case to UpdateBoolGpioInterfaces.
- Verified pre-commit passes.

_(Note: I encountered a local dependency race condition with the parameter generation during build, but the code changes themselves are verified. I was just trying to re-verify my changes by rebuilding.)_

Juliaj (approved):
LGTM

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -264,7 +274,7 @@ void GpioCommandController::initialize_gpio_state_msg()
       get_gpios_state_interfaces_names(gpio_name);
     gpio_state_msg_.interface_values[gpio_index].values = std::vector<double>(
       gpio_state_msg_.interface_values[gpio_index].interface_names.size(),
-      std::numeric_limits<double>::quiet_NaN());
+      sanitize_double(std::numeric_limits<double>::quiet_NaN()));
```
christophfroehlich: what is the purpose of this change? unset interfaces are per convention NaN, why not keeping this also in the messages? And even if we want to set it zero, why not setting it directly to zero instead of sanitize(nan) logic?

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -429,7 +439,8 @@ void GpioCommandController::apply_state_value(
     else
     {
       state_msg.interface_values[gpio_index].values[interface_index] =
-        state_msg_interface_value_op.value();
+
+        sanitize_double(state_msg_interface_value_op.value());
```
christophfroehlich: I don't think that zero is a good default choice here. If it is NaN, then either don't publish or publish NaN. why should we publish zeros?

gpio_controllers/test/test_load_gpio_command_controller.cpp
christophfroehlich, line 1: please move new tests to test_gpio_command_controller.cpp. This has nothing to do with the test_load_* idea.

christophfroehlich (changes requested):
I don't think that this fixes the reported problem. The exceptions happens in 
`state_interfaces_map_.at(interface_name).get().get_optional()` with default template parameter (which is double), but only on jazzy:

On rolling, we do auto-casting to double with a single warning per interface
https://github.com/ros-controls/ros2_control/blob/b59bd9c48067ad989caf4a719afc529b8f61045b/hardware_interface/include/hardware_interface/handle.hpp#L339-L354

on Jazzy, we throw an exception instead
https://github.com/ros-controls/ros2_control/blob/43d058f1fae58a2cecd5a708e7cce61f554a9bb0/hardware_interface/include/hardware_interface/handle.hpp#L361-L374

gpio_controllers/test/test_load_gpio_command_controller.cpp
christophfroehlich: The build even fails because the test is not linked to the parameters library.

Juliaj:
> I don't think that this fixes the reported problem. The exceptions happens in
> `state_interfaces_map_.at(interface_name).get().get_optional()` with default template parameter (which is double), but only on jazzy:

Thanks @christophfroehlich, I wasn't aware of this difference between Jazzy and rolling. At a high level, how should this be handled ?

christophfroehlich:
> > I don't think that this fixes the reported problem. The exceptions happens in
> > `state_interfaces_map_.at(interface_name).get().get_optional()` with default template parameter (which is double), but only on jazzy:
> 
> Thanks @christophfroehlich, I wasn't aware of this difference between Jazzy and rolling. At a high level, how should this be handled ?

Currently:  Either skip non-double interfaces, or do explicit casting (needs correct template of the get_optional() method)

I'd vote for auto-casting of interfaces to make these custom controllers more reusable. This is an always recurring discussion with @saikishor ;)

saikishor:
I wouldn't vote for this approach. IMO the controller should handle them. If you know you cannot deal with certain types, fail the activation or do something. If we have arrays in future, then it gets complicated easily. 

I would say instead, add a check what are and can be supported by the controller. If this casting is only for publishing the data that's where the issue is. Add a standalone function that does the casting and let's see how many ros2_controllers will need it. 

I understand that GPIO will need it as it is a more generic controller for the interfaces.

Ishan1923:
Thanks for the clarification on the Rolling vs. Jazzy versions. I apologize for the oversight.
I have updated apply_state_value in gpio_command_controller.cpp with the following changes:
- **sequential Logic**: The code now checks for a double interface first. If that fails, it checks for a bool interface (casting it to 0.0/1.0).
- **error Handling**: If the interface is neither double nor bool, it logs a single debug message stating data could not be retrieved.
- **tests**: Moved new tests to test_gpio_command_controller.cpp.
- **build**: linked the parameters library

Regarding @saikishor's point on future complexity (e.g., handling arrays): The current sequential approach solves the immediate crash efficiently without unnecessary lookups. I was unsure about the arrays, and other datatypes, as the current problem addresses bool/double issue.

Ready for review.

christophfroehlich (changes requested):
Please make the logic explicitly dependent on the data type using
`state_interfaces_map_.at(interface_name).get().get_data_type()`

Furthermore, please install pre-commit, activate it for this repo (`pre-commit install`) and reformat your changes.

Ishan1923:
- **Explicit Type Checking**: I now check interface.get_data_type() against hardware_interface::HandleDataType::DOUBLE and BOOL before attempting to read.
- **Safety**: It no longer guesses/casts sequentially. If a type is unsupported (not double/bool), it logs a specific debug message and skips it.
 - **Formatting**: pre-commit has been run on the codebase.

_Local tests passed successfully. Ready for re-review!_

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -417,20 +419,35 @@ void GpioCommandController::apply_state_value(
...
+      if (val_opt.has_value())
+      {
+        state_value = val_opt.value() ? 1.0 : 0.0;
+      }
     }
     else
     {
-      state_msg.interface_values[gpio_index].values[interface_index] =
-        state_msg_interface_value_op.value();
+      RCLCPP_DEBUG_ONCE(
```
christophfroehlich:
```suggestion
      RCLCPP_INFO_ONCE(
```
this can be info level. But there is a problem with ONCE macro, it does not work in a for loop for different interfaces. You'd need to add a bool vector/map warned_once per interface.

christophfroehlich (changes requested):
This is going into the right direction :)

Two things:

- ONCE macro does not work like this
- compiling on humble fails with
```
    [ RUN      ] GpioCommandControllerTestSuite.UpdateBoolGpioInterfaces
    [INFO] [1770019124.709374624] [test_gpio_command_controller]: configure successful
    [INFO] [1770019124.709482175] [test_gpio_command_controller]: activate successful
    terminate called after throwing an instance of 'std::logic_error'
      what():  basic_string::_M_construct null not valid
  >>>
```

gpio_controllers/src/gpio_command_controller.cpp
Ishan1923: Thank you @christophfroehlich 
I will look into it

Ishan1923:
I have added a unordered_set for storing nan interfaces and used _RCLCPP_INFO_, and for checking only once, there is a if condition statement that checks for any existing interface, matching with the new one in O(1) TC.

I removed the _type.to_string()_ call in the log message, which was causing the _basic_string::_M_construct_ null crash.

Local tests (_test_gpio_command_controller_) are now passing successfully.

christophfroehlich (approved):
LGTM, thank you!

christophfroehlich (changes requested):
idk why, but the ros2_control stack including this patch fails in the compatibility-build on humble:
```
[6193](https://github.com/ros-controls/ros2_controllers/actions/runs/21919550699/job/63317000871?pr=2103#step:6:16206)
    [ RUN      ] GpioCommandControllerTestSuite.UpdateBoolGpioInterfaces
    [INFO] [1770849228.540379897] [test_gpio_command_controller]: configure successful
    [INFO] [1770849228.540464415] [test_gpio_command_controller]: activate successful
    terminate called after throwing an instance of 'std::logic_error'
      what():  basic_string::_M_construct null not valid
```

This job uses [this](https://github.com/ros-controls/ros2_controllers/blob/master/ros2_controllers.rolling.repos) repos file, but builds it on humble distro.

Can you have a look please?

Ishan1923:
> idk why, but the ros2_control stack including this patch fails in the compatibility-build on humble:
> 
> ```
> [6193](https://github.com/ros-controls/ros2_controllers/actions/runs/21919550699/job/63317000871?pr=2103#step:6:16206)
>     [ RUN      ] GpioCommandControllerTestSuite.UpdateBoolGpioInterfaces
>     [INFO] [1770849228.540379897] [test_gpio_command_controller]: configure successful
>     [INFO] [1770849228.540464415] [test_gpio_command_controller]: activate successful
>     terminate called after throwing an instance of 'std::logic_error'
>       what():  basic_string::_M_construct null not valid
> ```
> 
> This job uses [this](https://github.com/ros-controls/ros2_controllers/blob/master/ros2_controllers.rolling.repos) repos file, but builds it on humble distro.
> 
> Can you have a look please?

I looked into it; I think that this is most likely related to an ABI break, I added _undordered_set nan_interfaces__ to keep the records of the joints, the CI library must be linking my changes with the old compiled library, which must be causing a **memory layout mismatch** or **failure of ODR (One-Defination-Rule)** implementation.

I propose that the nan_interfaces be removed from header file and introduced as a local variable in the cpp file or we can use DEBUG or RCLCPP_WARN_THROTTLE in the cpp file. _Pushing the fix shortly_.

christophfroehlich:
It seems that this issue in the compatibility build was introduced somewhere else, as it is failing everywhere in the CI jobs now. Not sure why this happens suddenly and only in the compatibility builds. I'll come back once we have a fixed them

Ishan1923:
Thanks for the update! I was worried I broke the whole stack! 
I'll hold off on any further changes until the CI issues are resolved.  Thanks!

Ishan1923:
> It seems that this issue in the compatibility build was introduced somewhere else, as it is failing everywhere in the CI jobs now. Not sure why this happens suddenly and only in the compatibility builds. I'll come back once we have a fixed them

any updates?

johanubbink:
Thank you to everyone for the effort on this, it is much appreciated!

I have tried this fix and noticed that this PR modifies the `apply_state_value` function to support both `bool` and `double`. However, the `apply_command` function is not updated. With the current implementation, the controller seems to fail if the command interface is a `bool`, because it always try to write a `double` (at least on my ros2_control version 4.43.0). I think the original issue #1970 also mentioned this?

Maybe the `apply_command` function can be updated along the line of:
```cpp
    bool success {false};
    if (type == hardware_interface::HandleDataType::DOUBLE) {
      success = interface.set_value<double>(command_value);
    } else if (type == hardware_interface::HandleDataType::BOOL) {
      // Assuming some convert_to_bool logic is available
      success = interface.set_value<bool>(convert_to_bool(command_value));
    } else {
      RCLCPP_WARN_THROTTLE(
        get_node()->get_logger(), *get_node()->get_clock(), 10000,
        "Interface '%s' has unsupported type. Only 'double' and 'bool' are supported.",
        interface_name.c_str());
    }
```
_disclaimer_: I'm still finding my way around ros2 control, so I might be missing something in the bigger picture... But please let me know if I can be of any help.

Ishan1923:
@christophfroehlich @johanubbink You're absolutely right ; I missed 
the `apply_command` side of the fix, this was a big blunder. I'll update the PR to handle 
`bool` interfaces in `apply_command` as well, along with a test case 
to cover it. Sorry for the oversight, will push the fix shortly.

Ishan1923:
@christophfroehlich @johanubbink I have updated the PR to address the missing logic in `apply_command` and also updated tests (`UpdateBoolGpioInterfaces`, `UpdateDoubleGpioInterfaces`). Please review.

gpio_controllers/src/gpio_command_controller.cpp
```diff
@@ -417,20 +442,35 @@ void GpioCommandController::apply_state_value(
...
   try
   {
-    auto state_msg_interface_value_op =
-      state_interfaces_map_.at(interface_name).get().get_optional();
+    double state_value = std::numeric_limits<double>::quiet_NaN();
+    const auto & interface = state_interfaces_map_.at(interface_name).get();
+    const auto & type = interface.get_data_type();
 
-    if (!state_msg_interface_value_op.has_value())
+    if (type == hardware_interface::HandleDataType::DOUBLE)
```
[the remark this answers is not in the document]
christophfroehlich: good catch, what exactly do we want here? @Ishan1923 please update either the PR description/title or the code. I personally would not hallucinate values here and report what the state interface is carrying.
For the case that the optional returns empty, I would prefer to use a local copy and republish the old value (what is the current behavior with gpio_state_msg_, right?)

christophfroehlich (changes requested):
See comments in copilot reviews, and please fix


```
  /home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/src/ros2_controllers/gpio_controllers/src/gpio_command_controller.cpp: In lambda function:
  /home/runner/work/ros2_controllers/ros2_controllers/.work/target_ws/src/ros2_controllers/gpio_controllers/src/gpio_command_controller.cpp:467:7: error: passing ‘const rclcpp::Clock’ as ‘this’ argument discards qualifiers [-fpermissive]
    467 |       RCLCPP_WARN_THROTTLE(
        |       ^~~~~~~~~~~~~~~~~~~~

```

PR #2109, opened 2026-01-17
Populate pose covariance correctly in steering controllers

### LINKED ISSUE #2098: Steering controller library does not populate pose covariance
Issue:
Steering controller library does computes odometry and reads in a parameter for the covariance for the odometry (both, pose and twist). On the published message, however, only the twist contains the covariance and not the pose.

1. Run an example such as the bicycle steering controller example from ros2-control-demos using `ros2 launch ros2_control_demo_example_11 carlikebot.launch.py remap_odometry_tf:=true`. Use default parameters.
2. echo `ros2 topic echo /bicycle_steering_controller/odometry --field pose.covariance`
3. echo `ros2 topic echo /bicycle_steering_controller/odometry --field twist.covariance`
4. Repeat if required with different covariance parameters
5. Check for covariance

Expected Behavior:
The published odometry message covariance should reflect the parameter assigned in the controller's config file

 - OS: Ubuntu
 - Version: Humble

vedh1234:
Fixes #2098 
Fixes a bug where pose_covariance_diagonal values were incorrectly written to twist.covariance due to a reference alias error.

### DIFF
--- a/steering_controllers_library/src/steering_controllers_library.cpp
+++ b/steering_controllers_library/src/steering_controllers_library.cpp
@@ -293,11 +293,11 @@ controller_interface::CallbackReturn SteeringControllersLibrary::on_configure(
   const size_t NUM_DIMENSIONS = 6;
   auto & pose_cov = odom_state_msg_.pose.covariance;
   auto & twist_cov = odom_state_msg_.twist.covariance;
-  for (size_t i = 0; i < NUM_DIMENSIONS; ++i) 
+  for (size_t i = 0; i < NUM_DIMENSIONS; ++i)
   {
     // indices of the diagonal: 0, 7, 14, 21, 28, 35
     const size_t index = (NUM_DIMENSIONS + 1) * i;
-    pose_cov[index]  = params_.pose_covariance_diagonal[i];
+    pose_cov[index] = params_.pose_covariance_diagonal[i];
     twist_cov[index] = params_.twist_covariance_diagonal[i];
   }
 


steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -290,14 +290,11 @@ controller_interface::CallbackReturn SteeringControllersLibrary::on_configure(
...
-    // 0, 7, 14, 21, 28, 35
-    const size_t diagonal_index = NUM_DIMENSIONS * index + index;
-    covariance[diagonal_index] = params_.pose_covariance_diagonal[index];
-    covariance[diagonal_index] = params_.twist_covariance_diagonal[index];
+  for (size_t i = 0; i < 6; ++i)
+  {
+    const size_t index = i * 6 + i;
+    odom_state_msg_.pose.covariance[index] = params_.pose_covariance_diagonal[i];
+    odom_state_msg_.twist.covariance[index] = params_.twist_covariance_diagonal[i];
   }
```
thedevmystic: I think we should use NUM_DIMENSIONS because currently we're using a magic number. It will be hard to maintain in future without knowing what 6 really is.

In short, we shouldn't sacrifice readability for brevity here.

I would suggest:
```suggestion
  const size_t NUM_DIMENSIONS = 6;
  auto & pose_cov = odom_state_msg_.pose.covariance;
  auto & twist_cov = odom_state_msg_.twist.covariance;
  for (size_t i = 0; i < NUM_DIMENSIONS; ++i) 
  {
    // 0, 7, 14, 21, 28, 35
    const size_t index = (NUM_DIMENSIONS + 1) * i;
    pose_cov[index]  = params_.pose_covariance_diagonal[i];
    twist_cov[index] = params_.twist_covariance_diagonal[i];
  }
```

bmagyar: Well a 6dof pose covariance isnt magic, it is a pretty fixed size, exactly 6 :D but good point re. readability

thedevmystic: Pretty fair point. 6 really isn't a magic in a "6"dof pose cov.

Well I know that, and I didn't mean in that way. It seems like I exaggerated it little bit 😅.

But don't you think using NUM_DIMENSIONS is better than just using 6?

Like that way code is self documenting?

vedh1234: Thanks,
readability is a fair point, I'll commit your suggested changes

steering_controllers_library/src/steering_controllers_library.cpp
```diff
@@ -290,14 +290,15 @@ controller_interface::CallbackReturn SteeringControllersLibrary::on_configure(
...
 
-  auto & covariance = odom_state_msg_.twist.covariance;
-  constexpr size_t NUM_DIMENSIONS = 6;
-  for (size_t index = 0; index < 6; ++index)
+  const size_t NUM_DIMENSIONS = 6;
+  auto & pose_cov = odom_state_msg_.pose.covariance;
+  auto & twist_cov = odom_state_msg_.twist.covariance;
+  for (size_t i = 0; i < NUM_DIMENSIONS; ++i) 
   {
     // 0, 7, 14, 21, 28, 35
```
bmagyar:
```suggestion
    // indices of the diagonal: 0, 7, 14, 21, 28, 35
```

bmagyar (approved):

PR #2110, opened 2026-01-19
set odometry service addition - mecanum drive controller

kuralme:
Adds odometry set/reset services, similar to the [previous](https://github.com/ros-controls/ros2_controllers/pull/2096) one. Features:
- atomic set/reset flags
- executed in the update loop
- testing

### DIFF
--- a/mecanum_drive_controller/test/test_mecanum_drive_controller.cpp
+++ b/mecanum_drive_controller/test/test_mecanum_drive_controller.cpp
@@ -805,7 +805,7 @@ TEST_F(MecanumDriveControllerTest, SideToSideAndRotationOdometryTest)
   EXPECT_LT(std::abs(controller_->odometry_.getRz()), M_PI);
 }
 
-TEST_F(MecanumDriveControllerTest, odometry_set_reset_services)
+TEST_F(MecanumDriveControllerTest, odometry_set_service)
 {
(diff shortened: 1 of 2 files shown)


mecanum_drive_controller/include/mecanum_drive_controller/odometry.hpp
```diff
@@ -81,6 +81,14 @@ class Odometry
   void setWheelsParams(
     const double sum_of_robot_center_projection_on_X_Y_axis, const double wheels_radius);
 
+  /// \brief Sets the odometry to given 2D pose values
+  /// \param x position (x component) [m]
+  /// \param y position (y component) [m]
+  /// \param heading orientation (z component) [rad]
+  void setOdometry(double x, double y, double heading);
+  /// \brief Sets the odometry parameters to zero
+  void resetOdometry();
```
christophfroehlich:
```suggestion
```

This method is not used anywhere, right?

kuralme: Yes its not used anymore, I'll remove it

kuralme:
The [issue](https://github.com/ros-controls/ros2_controllers/issues/942) to mention. I'll create the pr for omni wheel next

PR #2115, opened 2026-01-21
Add test to check JSB is not throwing when there is a boolean interface

Noel215:
Hi,

I was having some issue on the update method that was throwing an exception when using a boolean interface. Now I see that got fixed in https://github.com/ros-controls/ros2_control/pull/2360 and I was using an older version of `ros2_control`.

Anyway, I've done this test for testing exactly that issue. I leave it here just in case you want to add it.

### DIFF
--- a/joint_state_broadcaster/test/test_joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/test/test_joint_state_broadcaster.cpp
@@ -1342,7 +1342,8 @@ TEST_F(JointStateBroadcasterTest, NoThrowWithBooleanInterfaceTest)
   ASSERT_EQ(state_broadcaster_->on_activate(rclcpp_lifecycle::State()), NODE_SUCCESS);
 
   // update should not throw
-  ASSERT_NO_THROW(state_broadcaster_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
+  ASSERT_NO_THROW(
+    state_broadcaster_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
 
   const auto & dynamic_joint_state_msg = state_broadcaster_->dynamic_joint_state_msg_;
   ASSERT_EQ(dynamic_joint_state_msg.header.frame_id, frame_id_);
@@ -1369,7 +1370,8 @@ TEST_F(JointStateBroadcasterTest, NoThrowWithBooleanAndDoubleInterfaceTest)
   ASSERT_EQ(state_broadcaster_->on_activate(rclcpp_lifecycle::State()), NODE_SUCCESS);
 
   // update should not throw
-  ASSERT_NO_THROW(state_broadcaster_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
+  ASSERT_NO_THROW(
+    state_broadcaster_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
 
   const auto & dynamic_joint_state_msg = state_broadcaster_->dynamic_joint_state_msg_;
   ASSERT_EQ(dynamic_joint_state_msg.header.frame_id, frame_id_);


Noel215:
Digging a bit more, I've found that https://github.com/ros-controls/ros2_control/pull/2360 was never backported to `jazzy`, so the issue is happening on that distro, maybe also others. 

Opened the issue in https://github.com/ros-controls/ros2_controllers/issues/2116

christophfroehlich:
@saikishor any objections in backporting the mentioned PR?

saikishor:
> @saikishor any objections in backporting the mentioned PR?

@christophfroehlich as we added also more data types now, I would better go for a more aligned method that can get the double data directly instead. I'm writing something like the following directly in the handle.hpp. The controllers who needs double can use it directly. What do you think?

```
template <typename InterfaceType>
  double cast_interface_value_to_double(const InterfaceType & value)
```

saikishor:
Sorry my bad. We discussed this approach in the past and discarded it completely as we said every controller should handle them by itself. If not everyone simply uses these functions to bypass variants

joint_state_broadcaster/test/test_joint_state_broadcaster.cpp
```diff
@@ -1324,3 +1324,66 @@ TEST_F(JointStateBroadcasterTest, ExtraJointStatePublishTest)
...
+  std::vector<LoanedStateInterface> state_ifs;
+  state_ifs.emplace_back(joint_1_moving_state_);
+  state_broadcaster_->assign_interfaces({}, std::move(state_ifs));
+
+  // configure and activate ok
+  ASSERT_EQ(state_broadcaster_->on_configure(rclcpp_lifecycle::State()), NODE_SUCCESS);
+  ASSERT_EQ(state_broadcaster_->on_activate(rclcpp_lifecycle::State()), NODE_SUCCESS);
+
+  // update should not throw
+  state_broadcaster_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01));
```
Noel215:
```suggestion
  ASSERT_NO_THROW(state_broadcaster_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
```
I missed this, leaving it as a suggestion.

joint_state_broadcaster/test/test_joint_state_broadcaster.cpp
```diff
@@ -1324,3 +1324,66 @@ TEST_F(JointStateBroadcasterTest, ExtraJointStatePublishTest)
...
+  state_ifs.emplace_back(joint_1_vel_state_);
+  state_ifs.emplace_back(joint_1_eff_state_);
+  state_broadcaster_->assign_interfaces({}, std::move(state_ifs));
+
+  // configure and activate ok
+  ASSERT_EQ(state_broadcaster_->on_configure(rclcpp_lifecycle::State()), NODE_SUCCESS);
+  ASSERT_EQ(state_broadcaster_->on_activate(rclcpp_lifecycle::State()), NODE_SUCCESS);
+
+  // update should not throw
+  state_broadcaster_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01));
```
Noel215:
```suggestion
  ASSERT_NO_THROW(state_broadcaster_->update(rclcpp::Time(0), rclcpp::Duration::from_seconds(0.01)));
```

Noel215:
Rest LGTM. It makes sense to ignore interfaces that are not using double type.

saikishor (dismissed):

saikishor (approved):

bmagyar (approved):
thank you

christophfroehlich (commented):
@saikishor For my understanding: Earlier it used the booleans with a WARN_ONCE for accessing them from double getter. Now we change the behavior of just skipping it?

`[ros2_control_node-18] [WARN] [1769696718.989911944] [manual_inputs/left.button2]: Casting bool to double for interface 'manual_inputs/left.button2'. Better use get_optional<bool>().`

christophfroehlich:
Thanks for the discussion Sai: On rolling on kilted we auto-cast the boolean, but not on jazzy where we throw an exception
https://github.com/ros-controls/ros2_control/blob/0e2134ad7324c5bfbf5ec9096d1fb5db1c61b8ae/hardware_interface/include/hardware_interface/handle.hpp#L361-L374

this means that this PR is fine for jazzy, but not for kilted (as it is a behavior change). Rolling is up for discussion, as we will remove the dynamic_joint_states anyways.

PR #2127, opened 2026-01-25
dynamically link JTC parameter validation instead of header only

thedevmystic:
Hello respected maintainers and reviewers! Again this is Surya!

This PR addresses,
  - Seperate validate_jtc_parameters into hpp and cpp.
  - Added doxygen-style comments.
 
When I was refactoring tolerances.hpp I got linker error of multiple definitions of this header file. Because it defines functions directly in header without any inline, so I seperated them in .hpp and .cpp

I also added some doxygen-style comments for documentation of these functions.

Regards,
Surya

### DIFF
--- a/joint_trajectory_controller/CMakeLists.txt
+++ b/joint_trajectory_controller/CMakeLists.txt
@@ -37,6 +37,7 @@ generate_parameter_library(joint_trajectory_controller_parameters
 add_library(joint_trajectory_controller SHARED
   src/joint_trajectory_controller.cpp
   src/trajectory.cpp
+  src/validate_jtc_parameters.cpp
 )
 target_compile_features(joint_trajectory_controller PUBLIC cxx_std_17)
 if(WIN32)
--- a/joint_trajectory_controller/include/joint_trajectory_controller/validate_jtc_parameters.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/validate_jtc_parameters.hpp
@@ -29,16 +29,18 @@ namespace joint_trajectory_controller
  * \brief Validate command interface type combinations for joint trajectory controller.
  *
  * \param[in] parameter The rclcpp parameter containing the command interface types.
- * \return tl::expected<void, std::string> An empty expected on success, or an error message on failure.
+ * \return tl::expected<void, std::string> An empty expected on success, or an error message on
+ * failure.
  */
 tl::expected<void, std::string> command_interface_type_combinations(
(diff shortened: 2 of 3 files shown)


saikishor (approved):

bmagyar (approved):
thank you!

bmagyar:
@christophfroehlich I've removed the backport labels as this may cause issues to people inheriting from the JTC. Happy to discuss though

christophfroehlich (commented):
I am not sure about this here, as the joint_trajectory_controller_parameters library will be broken now.
People might have been able to use the parameter library without linking against the joint_trajectory_controller, which is now not possible.
Sorry, haven't had a proper look before.

christophfroehlich:
> When I was refactoring tolerances.hpp I got linker error of multiple definitions of this header file. Because it defines functions directly in header without any inline, so I seperated them in .hpp and .cpp

I don't understand why. This means that the generate_parameter_library does not have proper include guards, because we don't include the validation file anywhere else?

thedevmystic:
> > When I was refactoring tolerances.hpp I got linker error of multiple definitions of this header file. Because it defines functions directly in header without any inline, so I seperated them in .hpp and .cpp
> 
> I don't understand why. This means that the generate_parameter_library does not have proper include guards, because we don't include the validation file anywhere else?

I think you didn't get me, what I said. I meant I'm reorganizing, rewriting and adding something to make the tolerances.hpp faster to compile and better. That's why I got linker error. As validate_jtc_parameters.hpp defines 2 functions without inline this violated ODR (I'm adding tolerances.cpp for better compilation speed and decoupling from .hpp).

thedevmystic:
So, tolerances.hpp also includes JTC Params header file and joint_trajectory_controller.hpp also includes it. So, viola! We have multiple definitions in 2 .cpp (if we add my Refactored tolerances.cpp).

So, you are 100% correct that:
> we don't include the validation file anywhere else?

And I'll say **Yet**.

christophfroehlich:
- Idk why we would use the validate_jtc_parameter.hpp file elsewhere?
- We have included joint_trajectory_controller_parameters.hpp twice already but in the same compilation unit, what will you change there?

thedevmystic:
I'm about to initiate that PR in like a hour or so. So, I think instead of explaining here, just see it!

Edit: PR opened as #2133.

thedevmystic:
i see that it was reverted but i did it because without this we are getting errors in #2133

PR #2128, opened 2026-01-25
fix(gpio_controllers): resolve build race condition in tests

### LINKED ISSUE #2117: [gpio_controllers] Build race condition: test_load_gpio_command_controller missing dependency on generated parameters
When building gpio_controllers from a clean workspace (or after deleting build/ and install/), the build frequently fails with a missing header error _(fatal error: gpio_controllers/gpio_command_controller_parameters.hpp: No such file or directory)_ in the test file _test_load_gpio_command_controller.cpp_.

This appears to be a build race condition in CMakeLists.txt. The test_load_gpio_command_controller target does not explicitly link against the main library (gpio_controllers) or the generated parameter library. Consequently, the build system attempts to compile the test before the parameter header generation is complete.

To Reproduce Steps to reproduce the behavior:

1. Clean the workspace: rm -rf build/ install/ log/

2. Build the package from scratch (forcing parallel execution makes it more likely to fail): 

> `colcon build --packages-select gpio_controllers --parallel-workers 8`
 
3. Observe the build failure:    

> ```
> /src/ros2_controllers/gpio_controllers/include/gpio_controllers/gpio_command_controller.hpp:30:10: 
>     fatal error: gpio_controllers/gpio_command_controller_parameters.hpp: No such file or directory
> 
> ```

**Expected behavior**: The package and its tests should build successfully from a clean state without dependency race conditions.

**Environment**:
- OS: Linux (Pop!_OS / Ubuntu 24.04 based)
- Version: Rolling
- Build Tool: colcon
_Additional context The issue seems to be missing a dependency link in gpio_controllers/CMakeLists.txt._

Currently, the test target is defined as:

```cmake
  ament_add_gmock(
    test_load_gpio_command_controller
    test/test_load_gpio_command_controller.cpp
  )

  target_link_libraries(test_load_gpio_command_controller
    controller_manager::controller_manager
    hardware_interface::hardware_interface
    ros2_control_test_assets::ros2_control_test_assets
    # MISSING LINK to ${PROJECT_NAME} or gpio_controllers
  )
```

_Adding gpio_controllers to target_link_libraries for this test target resolves the race condition by enforcing the correct build order._

Ishan1923:
This PR fixes the build failure reported in #2117.

Fixes #2117

### DIFF
--- a/diff_drive_controller/src/odometry.cpp
+++ b/diff_drive_controller/src/odometry.cpp
@@ -68,7 +68,11 @@ bool Odometry::update(double left_pos, double right_pos, const rclcpp::Time & ti
   left_wheel_old_pos_ = left_wheel_cur_pos;
   right_wheel_old_pos_ = right_wheel_cur_pos;
 
+// Disable deprecated warnings
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
   updateFromVelocity(left_wheel_est_vel, right_wheel_est_vel, time);
+#pragma GCC diagnostic pop
 
   return true;
 }
@@ -105,7 +109,11 @@ bool Odometry::updateFromVelocity(double left_vel, double right_vel, const rclcp
   const double angular = (right_vel - left_vel) / wheel_separation_;
 
   // Integrate odometry:
+// Disable deprecated warnings
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
   integrateExact(linear, angular);
+#pragma GCC diagnostic pop
 
   timestamp_ = time;
 
@@ -152,7 +160,11 @@ void Odometry::updateOpenLoop(double linear, double angular, const rclcpp::Time 
   /// Integrate odometry:
   const double dt = time.seconds() - timestamp_.seconds();
   timestamp_ = time;
+// Disable deprecated warnings
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
   integrateExact(linear * dt, angular * dt);
+#pragma GCC diagnostic pop
 }
 
 bool Odometry::try_update_open_loop(double linear_vel, double angular_vel, double dt)
@@ -211,7 +223,11 @@ void Odometry::integrateExact(double linear, double angular)
 {
   if (fabs(angular) < 1e-6)
   {
+// Disable deprecated warnings
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
     integrateRungeKutta2(linear, angular);
+#pragma GCC diagnostic pop
   }
   else
   {


gpio_controllers/CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ if(BUILD_TESTING)
     controller_manager::controller_manager
     hardware_interface::hardware_interface
     ros2_control_test_assets::ros2_control_test_assets
+    ${PROJECT_NAME}
```
saikishor:
```suggestion
    gpio_controllers
```

Let's change it to the name of the library for now, to have consistency with other controllers.

I agree it is better with ${PROJECT_NAME} though

Ishan1923: Ok i will just do this instead

**change**: ${PROJECT_NAME} -> gpio_controllers

christophfroehlich:
Maybe `target_include_directories(test_load_gpio_command_controller PRIVATE include)` is the culprit, do we need the include folder here? We have this in some of the test_load_* tests, but not in every package here.

saikishor:
> We never link against the library itself (aka the controller plugins) in `test_load_*` tests of other controllers. Why should this be necessary here?

I agree. I also thought the same at first and then I verified other controllers and there we are linking too. Not sure why

christophfroehlich:
> > We never link against the library itself (aka the controller plugins) in `test_load_*` tests of other controllers. Why should this be necessary here?
> 
> I agree. I also thought the same at first and then I verified other controllers and there we are linking too. Not sure why

In which ones are we linking against? I only had a look in a few..

saikishor:
> In which ones are we linking against? I only had a look in a few..

https://github.com/ros-controls/ros2_controllers/blob/981747527c7b6b655c6f3dddcac614fb6eafb9f3/joint_state_broadcaster/CMakeLists.txt#L58-L66

https://github.com/ros-controls/ros2_controllers/blob/981747527c7b6b655c6f3dddcac614fb6eafb9f3/joint_trajectory_controller/CMakeLists.txt#L96-L103

https://github.com/ros-controls/ros2_controllers/blob/981747527c7b6b655c6f3dddcac614fb6eafb9f3/position_controllers/CMakeLists.txt#L42-L49

There might be few more with the linking, I didn't go through all of them. I saw few without too, but I agree that it shouldn't be linked.

Ishan1923:
@christophfroehlich Thanks for the clarification!

The reason I added the link is to resolve a build race condition: this test includes headers generated by `generate_parameter_library` within `gpio_controllers`. Without an explicit dependency, the test tries to compile before those headers are generated.

Since we prefer not to link the plugin library directly, I can replace `target_link_libraries` with `add_dependencies`:

```cmake
add_dependencies(test_load_gpio_command_controller gpio_controllers)
```

_what should I do?_

christophfroehlich:
> @christophfroehlich Thanks for the clarification!
> 
> The reason I added the link is to resolve a build race condition: this test includes headers generated by `generate_parameter_library` within `gpio_controllers`. Without an explicit dependency, the test tries to compile before those headers are generated.

No this is not correct, it does not use any package-related headers (or library) at compile time. It only loads the plugin at test-time.

> _what should I do?_

Have you tried to remove the `target_include_directories`?

Ishan1923:
This was overlooked by me. The headers included via this:

```cmake
target_include_directories(test_load_gpio_command_controller PRIVATE include)
```

is not required at the compile time, but in the runtime; for this pluginlib will handle it. It was extra-safety line which was not needed, as it was creating unnecessary compile time dependency.

gpio_controllers/CMakeLists.txt
```diff
@@ -62,11 +62,11 @@ if(BUILD_TESTING)
     test/test_load_gpio_command_controller.cpp
   )
 
-  target_include_directories(test_load_gpio_command_controller PRIVATE include)
   target_link_libraries(test_load_gpio_command_controller
     controller_manager::controller_manager
     hardware_interface::hardware_interface
     ros2_control_test_assets::ros2_control_test_assets
+    gpio_controllers
```
christophfroehlich: this is not needed now, right?
```suggestion
```

Ishan1923:
I have removed _gpio_controllers_.
Apologies for the oversight.

PR #2134, opened 2026-01-29
Add a hint to the docs about non-double interfaces in JSB

christophfroehlich:
The behavior change was not documented in #2115


christophfroehlich (commented):
https://github.com/ros-controls/ros2_controllers/actions/runs/21508033206

PR #2148, opened 2026-02-05
set odometry service addition - omni wheel drive controller

kuralme:
Adds odometry set service, similar to the [other](https://github.com/ros-controls/ros2_controllers/pull/2096) 2 controllers. Related [issue](https://github.com/ros-controls/ros2_controllers/issues/942). 
Features:
- atomic set flag, RealtimeThreadSafeBox usage
- executed in the update loop
- testing

### DIFF
--- a/omni_wheel_drive_controller/include/omni_wheel_drive_controller/odometry.hpp
+++ b/omni_wheel_drive_controller/include/omni_wheel_drive_controller/odometry.hpp
@@ -32,7 +32,7 @@ public:
   bool updateOpenLoop(
     const double & linear_x_vel, const double & linear_y_vel, const double & angular_vel,
     const rclcpp::Time & time);
-  void resetOdometry();
+  [[deprecated("Use setOdometry(0.0, 0.0, 0.0) instead")]] void resetOdometry();
   void setOdometry(const double & x, const double & y, const double & heading);
 
   double getX() const { return x_; }


omni_wheel_drive_controller/include/omni_wheel_drive_controller/odometry.hpp
```diff
@@ -33,6 +33,7 @@ class Odometry
     const double & linear_x_vel, const double & linear_y_vel, const double & angular_vel,
     const rclcpp::Time & time);
   void resetOdometry();
```
christophfroehlich: let's deprecate this and use the `set_odometry` method instead?

omni_wheel_drive_controller/src/omni_wheel_drive_controller.cpp
```diff
@@ -436,6 +459,28 @@ void OmniWheelDriveController::compute_and_set_wheel_velocities()
...
+  // flip the flag for thread-safe odom set in the control loop
+  set_odom_requested_.store(true);
+
+  res->success = true;
+  res->message = "Odometry set request accepted";
+}
+
 bool OmniWheelDriveController::reset()
 {
   odometry_.resetOdometry();
```
christophfroehlich:
```suggestion
  odometry_.setOdometry(0.0, 0.0, 0.0);
```

omni_wheel_drive_controller/include/omni_wheel_drive_controller/odometry.hpp
kuralme: Sure

PR #1695, opened 2025-05-21
Integrate tricycle_controller into steering_controller_library

christophfroehlich:
Closes #850 and others.

Tested with https://github.com/ros-controls/gz_ros2_control/pull/865

PR #2149, opened 2026-02-07
feat: add convex tricycle speed scaling

thedevmystic:
Hello, respected maintainers and reviewers!
This is Surya!

This PR addresses:
  - add convex speed scaling in tricycle controller.

now, the scale factor is smooth out instead of being janky!


thedevmystic:
Well, I didn't knew it was going to be deprecated. But I've already finished the work on this. Well we can complete the TODO and deprecate it, can't we? Because `tricycle_controller` is still not deprecated if I'm right.

Anyways, this is the data.

new_vs_old

> old vs New

exponents_comparison

> different exponential powers

Again, @christophfroehlich, I didn't knew about the depreciation, but take a look at it :)

christophfroehlich:
> Again, @christophfroehlich, I didn't knew about the depreciation, but take a look at it :)

I haven't deprecated it until https://github.com/ros-controls/ros2_controllers/pull/1695 is merged. You can help to leave a review there.

As this is kind of a behavior change, I'm unsure if we should merge this into tricycle_controller; or at least don't backport it to the stable branches.

thedevmystic:
Well I can advise one thing, if `tricycle_controller` is going to deprecated very soon, then we shouldn't merge it. But if `tricycle_controller` is going to be around for some more time, then there is no bad thing to merge this, as it completes the TODO.

That's just my thoughts :)

thedevmystic:
I'm closing this PR off, @christophfroehlich. Is it okay?

PR #2150, opened 2026-02-07
fix: fix dynamic allocation in loop

thedevmystic:
Hello, respected maintainers and reviewers!
This is Surya!

This PR addresses:
  - dynamic allocation in real-time loop.

Previously in admittance controller there was a dynamic allocation which can be fixed by reserving memory and allocating only when string changes.

### DIFF
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -81,6 +81,9 @@ controller_interface::return_type AdmittanceRule::configure(
     return controller_interface::return_type::ERROR;
   }
 
+  // configure force torque sensor frame in state message
+  state_message_.ft_sensor_frame.data = parameters_.ft_sensor.frame.id;
+
   return controller_interface::return_type::OK;
 }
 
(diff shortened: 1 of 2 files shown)


saikishor (commented):
Check if it can be assigned in the on_configure method and then you set it only once. IIRC, this cannot be changed online, so it should be fine.

thedevmystic:
I increased the reserved space from 64 to 128, because I thought 64 is a bit restrictive. Check if it's fine.

And yes @saikishor, this parameter can't change once the controller is online. I checked to make sure that in `on_configure` there is no assignment to it.

In `on_configure`, there are only some data extraction from `ft_sensor`.

thedevmystic:
I think it's because `AdmittanceRule::get_controller_state` returns the full state, I guess.

We can't remove it from there, thus can't remove it from loop.

Because we use,
```
 // Initialize state message
/*363*/   state_msg_ = admittance_->get_controller_state();

// This is inside on_configure
```


If I'm going in wrong direction, feel free to correct me.

thedevmystic:
Well, now when I look at it, we can completely remove the `ft_sensor_frame` from the `get_controller_state` and copy the data in `on_configure`. You're right!

So, now I'm thinking I'll remove it from `Admittance::get_controller_state()` and add it to `on_configure()`. If it's fine with you, @saikishor?

thedevmystic:
Done, @saikishor. Does it need any other tweaking?

saikishor (approved):
LGTM

christophfroehlich (approved):
LGTM

PR #2151, opened 2026-02-07
fix(admittance): remove real-time memory allocations in update loop

### LINKED ISSUE #2130: [admittance_controller] Real-time safety violation: Dynamic allocation in update loop
The admittance_controller violates real-time safety guarantees by performing dynamic memory allocation (heap allocation) inside the high-frequency update loop.

Specifically, a std::string assignment/copy is performed every cycle in get_controller_state(), which is called by the main update loop. Using std::string assignment can trigger malloc/new calls, leading to non-deterministic latency (jitter) in the control loop.

**To Reproduce :**
1. Open ros2_controllers/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp.
2. Locate the function get_controller_state() (approx line 437).
3. Observe the following line:

   ` state_message_.ft_sensor_frame.data = admittance_state_.ft_sensor_frame;`

    _Note the explicit TODO comment acknowledging the issue: // TODO(anyone) remove dynamic allocation here._

    Run the tests (colcon test) or a simulation; this code path executes on every update cycle.

**Environment**
    OS: Ubuntu 22.04 / Pop!_OS
    Version: Rolling (Source build)

If this is confirmed as an issue, I would be interested in working on a fix.

Ishan1923:
This PR resolves real-time safety violations in the admittance_controller. 

Previously, the update_and_write_commands() loop was creating new geometry_msgs objects and calling get_node()->get_logger() every cycle. These operations involve dynamic memory allocation and potential mutex locking, which are not real-time safe.

**Changes**:
- **Member Variables**: Promoted _offsetted_ft_values_ to a private class member (_offsetted_ft_values__) to ensure zero-copy/zero-allocation during runtime.
- **Cached Logger**: Added _logger__ as a class member initialized in _on_init()_ to avoid retrieving the logger node-side in the update loop.
- **Helper Optimization**: Refactored the _add_wrenches()_ helper function to use output references (void return) instead of returning objects by value.
- **Cleanup**: Removed a duplicate check/assignment of _wrench_command_msg__ inside the update loop.

**Testing & Verification**

I have verified that these changes maintain strict backward compatibility and pass all existing tests.
- Build: Passed (colcon build --packages-up-to admittance_controller)
- Tests: Passed 100% (colcon test --packages-select admittance_controller) - 38/38 tests passed.

Fixes #2130

### DIFF
--- a/admittance_controller/include/admittance_controller/admittance_controller.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_controller.hpp
@@ -27,8 +27,10 @@
 #include "admittance_controller/admittance_rule.hpp"
 #include "control_msgs/msg/admittance_controller_state.hpp"
 #include "controller_interface/chainable_controller_interface.hpp"
+#include "geometry_msgs/msg/wrench.hpp"
 #include "hardware_interface/types/hardware_interface_type_values.hpp"
 #include "rclcpp/duration.hpp"
+#include "rclcpp/logging.hpp"
 #include "rclcpp/time.hpp"
 #include "rclcpp_lifecycle/state.hpp"
 #include "realtime_tools/realtime_publisher.hpp"
@@ -156,6 +158,10 @@ protected:
    * @brief Write values from state_command to claimed hardware interfaces
    */
   void write_state_to_hardware(const trajectory_msgs::msg::JointTrajectoryPoint & state_command);
+
+private:
+  geometry_msgs::msg::Wrench offsetted_ft_values_;
+  rclcpp::Logger logger_ = rclcpp::get_logger("AdmittanceController");
 };
 
 }  // namespace admittance_controller
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -103,6 +103,8 @@ controller_interface::return_type AdmittanceRule::reset(const size_t num_joints)
   state_message_.admittance_velocity.header.frame_id = parameters_.kinematics.base;
   state_message_.admittance_acceleration.header.frame_id = parameters_.kinematics.base;
 
+  state_message_.ft_sensor_frame.data.resize(256);  // preallocate space
+
   // reset admittance state
   admittance_state_ = AdmittanceState(num_joints);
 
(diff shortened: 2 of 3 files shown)


admittance_controller/include/admittance_controller/admittance_controller.hpp
```diff
@@ -156,6 +158,10 @@ class AdmittanceController : public controller_interface::ChainableControllerInt
    * @brief Write values from state_command to claimed hardware interfaces
    */
   void write_state_to_hardware(const trajectory_msgs::msg::JointTrajectoryPoint & state_command);
+
+private:
+  geometry_msgs::msg::Wrench offsetted_ft_values_;
+  rclcpp::Logger logger_ = rclcpp::get_logger("AdmittanceController");
 };
```
saikishor, line 165: I don't think this is needed. Wrench has no dynamically allocated variables

admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
```diff
@@ -395,8 +397,16 @@ const control_msgs::msg::AdmittanceControllerState & AdmittanceRule::get_control
...
+  auto & dst = state_message_.ft_sensor_frame.data;
+  const auto & src = admittance_state_.ft_sensor_frame;
+  if (src.size() <= dst.capacity())
+  {
+    dst.assign(src.c_str(), src.size());
+  }
+  else
+  {
+    dst.assign(src.c_str(), dst.capacity());  // truncate to reserved size
+  }
```
saikishor, line 409: Why this approach¿?

admittance_controller/src/admittance_controller.cpp
```diff
@@ -640,7 +638,7 @@ void AdmittanceController::write_state_to_hardware(
     }
     if (!success)
     {
-      RCLCPP_WARN(logger, "Error while setting command for joint %zu.", joint_ind);
+      RCLCPP_WARN(logger_, "Error while setting command for joint %zu.", joint_ind);
```
saikishor, line 641:
```suggestion
      RCLCPP_WARN(get_node()->get_logger(), "Error while setting command for joint %zu.", joint_ind);
```

admittance_controller/include/admittance_controller/admittance_controller.hpp
Ishan1923: Currently geometry_msgs::msg::Wrench objects are being created on the stack on every cycle in the update while they don't allocate heap memory, _promoting them to members, avoid unnecessary stack construction/destruction overhead in the RT loop._

admittance_controller/src/admittance_controller.cpp
Ishan1923: I thought caching the logger would be a better choice; as normally, calling get_node()->get_logger() will involve pointer derefrencing with potentially, internel mutexes/allocations in rclcpp

Ishan1923:
> The dynamic allocation part is already being handled in this PR: #2150

@saikishor Thanks for the review.

I agree that #2150 is the cleaner solution for the string allocation issue since it handles it in the configuration phase.

Regarding the other optimizations (cached logger and stack allocation): I see your points that the overhead is likely negligible. Since the critical real-time violation is being fixed in #2150, I will close this PR.

PR #2153, opened 2026-02-09
Fix JTC test speed scaling publisher

MarqRazz:
I'm working on adding a test to the JTC but when running `master` locally it fails on my machine. I use CycloneDDS in my setup so that is likely why CI is passing (because FastDDS isn't enforcing this) but the output from the failure pointed to an easy fix.

When I run locally you can see the test publisher to `/test_joint_trajectory_controller/speed_scaling_input` has the wrong QoS settings...
```
3: [test_joint_trajectory_controller 1770662071.626853898]: Setting initial scaling factor to 0.123000
3: [test_joint_trajectory_controller 1770662071.626858865]: Using scaling state from the hardware from interface speed_scaling/speed_scaling_factor.
3: [test_joint_trajectory_controller 1770662071.627182014]: New publisher discovered on topic '/test_joint_trajectory_controller/speed_scaling_input', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: DURABILITY_QOS_POLICY
3: /root/control_ws/src/ros2_controllers/joint_trajectory_controller/test/test_trajectory_controller.cpp:2835: Failure
3: Expected equality of these values:
3:   state->speed_scaling_factor
3:     Which is: 0.123
3:   0.765
3: 
3: [  FAILED  ] TrajectoryControllerTest.scaling_command_interface_sets_value (832 ms)
-------
3: [  FAILED  ] 2 tests, listed below:
3: [  FAILED  ] TrajectoryControllerTest.setting_scaling_factor_works_correctly
3: [  FAILED  ] TrajectoryControllerTest.scaling_command_interface_sets_value
```
because using `rclcpp::SystemDefaultsQoS()` lets the underlying RMW choose the settings, which almost always defaults to Volatile.

To fix this I set the publisher to `Transient Local` durability to ensure it doesn't miss the value changes published in the test.

### DIFF
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -81,6 +81,9 @@ controller_interface::return_type AdmittanceRule::configure(
     return controller_interface::return_type::ERROR;
   }
 
+  // configure force torque sensor frame in state message
+  state_message_.ft_sensor_frame.data = parameters_.ft_sensor.frame.id;
+
   return controller_interface::return_type::OK;
 }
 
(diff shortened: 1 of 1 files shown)


saikishor:
Thanks for fixing it

PR #2158, opened 2026-02-10
Add test to check JSB is not throwing when there is a boolean interface (backport #2115)

Noel215:
Backporting #2115, as it is broken for `jazzy`

closes #2131


Noel215:
@christophfroehlich @saikishor I've done the backport manually because it needed some changes in the declaration of the `joint_1_moving_state_` variable in the tests.

PR #2159, opened 2026-02-10
Optimize joint state broadcaster execution time

saikishor:
Hello!

When testing the joint state broadcaster on high-frequency systems, it is observed that it is taking around 120-140 microseconds of execution, and it doesn't improve even when we choose to set the parameter `publish_dynamic_joint_states: false` . So, upon closer look, it seems like it always does the whole map lookup of the interfaces and it is consuming quite some time. When running 2 of these broadcaster it consumes half of our cycle time.

This PR proposes minor improvements that reduce the execution time by 70%

**With the current state:**
image

**With the proposed fix:**
image

Tested on TALOS robot by setting parameter `publish_dynamic_joint_states: false`


joint_state_broadcaster/src/joint_state_broadcaster.cpp
```diff
@@ -419,16 +435,32 @@ bool JointStateBroadcaster::use_urdf_joint_interfaces() const
 controller_interface::return_type JointStateBroadcaster::update(
   const rclcpp::Time & time, const rclcpp::Duration & /*period*/)
 {
-  size_t map_index = 0u;
-  for (auto i = 0u; i < state_interfaces_.size(); ++i)
+  if (realtime_dynamic_joint_state_publisher_)
```
christophfroehlich: this got removed. do we need the else-branch now?

christophfroehlich (changes requested):

christophfroehlich (dismissed):
I fixed a bad earlier merge. Now LGTM, haven't benchmarked it though.

christophfroehlich (approved):

saikishor:
Thank you

PR #2160, opened 2026-02-10
Use preallocated feedback from JTC to avoid heap allocation

saikishor:
This PR adds a preallocated feedback variable to avoid head allocation in the RT loop. This PR was tested along with https://github.com/ros-controls/realtime_tools/pull/486 to improve the realtimeness of the JTC

Thank you


joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1938,6 +1935,10 @@ void JointTrajectoryController::init_hold_position_msg()
   {
     hold_position_msg_ptr_->points[0].effort.resize(dof_, 0.0);
   }
+
+  // Pre-allocate feedback message to avoid heap allocation in RT loop
+  preallocated_feedback_ = std::make_shared<FollowJTrajAction::Feedback>();
+  preallocated_feedback_->joint_names = params_.joints;
```
christophfroehlich: This has nothing to do with the hold_position message.
let's move this to the on_configure, or rename the method.

christophfroehlich (changes requested):
The idea LGTM, but I suggest moving the initialization to a different place

joint_trajectory_controller/src/joint_trajectory_controller.cpp
saikishor: You are right. Sorry

bmagyar (dismissed):
thank you!

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -509,11 +509,9 @@ controller_interface::return_type JointTrajectoryController::update(
       // trajectory has not started yet, so the old trajectory's progress must not succeed/abort it.
       if (active_goal && !rt_active_goal_deferred_)
       {
-        // send feedback
-        auto feedback = std::make_shared<FollowJTrajAction::Feedback>();
+        // send feedback, reusing the goal handle's preallocated message to avoid allocating here
```
bmagyar:
```suggestion
        // send feedback
```

saikishor:
@bmagyar Looks spot on. Thanks for the nice insights

bmagyar (approved):
thank you for spotting this!

christophfroehlich (approved):
LGTM, I didn't know that this even existed 😁

Issue #2111, opened 2026-01-21
soft-stop functionality for JTC


pastoriomarco:
Hi!  
I think that the on-path soft-stop is a very important functionality for any manipulator. Since the speed scaling came out in rolling I thought I could leverage that to implement a soft-stop, but I fear that a ramp-down that is sent with a burst of messages could suffer from latency and not be reliable enough in industrial applications. I'd rather have the joint_trajectory_controller handle the ramp-down.

There are several decisions to make, and I tried to turn them into requisites. For each I give my opinion, which corresponds to the way it's done in the sample implementation below.

### REQUISITES:

- Ramp down shape? Linear, sinusoid, quintic? I'd go with sinusoid
- Should the goal be kept "active" on soft stop (pause function)? I think so, we shouldn't terminate the active goal or this functionality would be a controlled cancel/abort. The cancel can be issued by the application upon receiving the stopped status.
- Speed scaling works only on position command interface, then should the soft stop apply to that interface only? I think so, for the same good reasons that are valid for the speed scaling. The soft stop could use other mechanisms to make it work anyway, but only if it actually makes sense and is reliable.
- Should we use the speed scaling logic directly or should we implement a parallel/similar logic? How would they interact? I think the stop logic should have its own logic and should add to the speed scaling by further tuning the speed, as they are separate concerns. 
- Soft stop should be applied/applicable also on on-robot scaling? I think so, but only if there's no velocity command_interface, otherwise we'd have to use a separate mechanism.
- What if a cancel/abort/preempt arrives during soft stop? Since the soft stop keeps the goal active then we just apply the standard behavior and apply the incoming action to an active goal as usual
- What happens if a new speed scaling factor is issued during soft stop? I'd apply it cumulatively
- What's the default soft stop time? Ideally, I'd calculate the minimum possible given the current joint acceleration/jerk limits, if available. But JTC doesn't read robot's limits if I'm not wrong. The user should still be able to set a custom stop time per-call, so he can handle the stop time to respect robot limits. I'll go with a default ramp-down duration and a selectable ramp-down in the command.
- Soft stop can't be partial, it's either 0 or 1. If the target is not 0, then is automatically set to 1. This is not to overlap with speed scaling logic and scope
[...]

PR #2163, opened 2026-02-10
Add decelerate to stop functionality when trajectory is canceled or preempted

MarqRazz:
Title says it all. The old behavior of the JTC is to set the robots command position to its current pose at the time the trajectory is canceled. 
image

For my hardware this frequently causes it to fault because it instantly sets the joints velocity to zero no matter how fast it is moving. Below is a plot of the new behavior...
image

I added a few parameters to the `constraints` struct to opt into the new functionality but once the URDF supports specifying joint acceleration we could use those values if these are not specified.
```yaml
constraints:
  decelerate_on_cancel: true
  stopped_velocity_tolerance: 0.008
  goal_time: 0.0
  joint_1:
    goal: 0.005
    max_deceleration_on_cancel: 10.0
  joint_2:
    goal: 0.005
    max_deceleration_on_cancel: 3.0
  joint_3:
    goal: 0.005
    max_deceleration_on_cancel: 6.0
  joint_4:
    goal: 0.005
    max_deceleration_on_cancel: 10.0
  joint_5:
    goal: 0.005
    max_deceleration_on_cancel: 10.0
  joint_6:
    goal: 0.005
    max_deceleration_on_cancel: 17.0
```

The new tests check for ...
- Invalid max_deceleration:
```
5: [test_joint_trajectory_controller 1770759765.603637493]: Canceling active action goal because cancel callback received.
5: [test_joint_trajectory_controller 1770759765.603646091]: Joint [joint1] invalid max_deceleration_on_cancel [0.0]. Falling back to hold position.
```
- Hardware does not have a velocity state interface:
```
5: [test_joint_trajectory_controller 1770761095.581896746]: Canceling active action goal because cancel callback received.
5: [test_joint_trajectory_controller 1770761095.581907052]: Hardware does not support velocity state interface, falling back to hold position.
```
- Hardware is already stopped (or is reporting zeros like several of the tests do)
```
5: [test_joint_trajectory_controller 1770761097.789958878]: Canceling active action goal because cancel callback received.
5: [test_joint_trajectory_controller 1770761097.789967357]: Joint [joint1] decel [10.000], stop dist [0.0000], initial vel [0.0000], initial pos [2.0826], hold pos [2.0826], time to stop [0.0000]
5: [test_joint_trajectory_controller 1770761097.789995287]: Joint [joint2] decel [10.000], stop dist [0.0000], initial vel [0.0000], initial pos [2.9553], hold pos [2.9553], time to stop [0.0000]
5: [test_joint_trajectory_controller 1770761097.790008351]: Joint [joint3] decel [10.000], stop dist [0.0000], initial vel [0.0000], initial pos [3.8280], hold pos [3.8280], time to stop [0.0000]
5: [test_joint_trajectory_controller 1770761097.790021810]: Created ramped stop trajectory with [1] points and max time to stop [0.000] sec
```
- Hardware has non-zero velocity and needs a ramp generated to stop gracefully
```
5: [test_joint_trajectory_controller 1770761179.755913527]: Canceling active action goal because cancel callback received.
5: [test_joint_trajectory_controller 1770761179.755922035]: Joint [joint1] decel [10.000], stop dist [0.5749], initial vel [3.3908], initial pos [2.0849], hold pos [2.6598], time to stop [0.3391]
5: [test_joint_trajectory_controller 1770761179.755956492]: Joint [joint2] decel [10.000], stop dist [0.4967], initial vel [3.1517], initial pos [2.9575], hold pos [3.4541], time to stop [0.3152]
5: [test_joint_trajectory_controller 1770761179.755969589]: Joint [joint3] decel [10.000], stop dist [0.4241], initial vel [2.9125], initial pos [3.8300], hold pos [4.2541], time to stop [0.2912]
5: [test_joint_trajectory_controller 1770761179.755984729]: Created ramped stop trajectory with [35] points and max time to stop [0.339] sec
```

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -151,7 +151,7 @@ protected:
   std::vector<double> tmp_command_;
 
   // If true, enable calculations to stop all joints using constant deceleration
-  bool decelerate_on_cancel_requested_ = false;
+  bool should_decelerate_on_cancel_ = false;
   // reserved storage for the max deceleration values
   std::vector<double> max_decel_;
   // reserved storage for each joints max stopping time
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -115,9 +115,9 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
     hold_position_.resize(params_.joints.size(), 0.0);
     stop_direction_.resize(params_.joints.size(), 0.0);
 
-    decelerate_on_cancel_requested_ = true;
+    should_decelerate_on_cancel_ = true;
     // check for valid max_deceleration values for each joint
-    for (size_t i = 0; i < params_.joints.size() && decelerate_on_cancel_requested_; ++i)
+    for (size_t i = 0; i < params_.joints.size() && should_decelerate_on_cancel_; ++i)
     {
       max_decel_[i] =
         params_.constraints.joints_map.at(params_.joints[i]).max_deceleration_on_cancel;
@@ -129,7 +129,7 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
           "Joint [%s] invalid max_deceleration_on_cancel [%.1f]. "
           "Falling back to hold position on cancel.",
           params_.joints[i].c_str(), max_decel_[i]);
-        decelerate_on_cancel_requested_ = false;
+        should_decelerate_on_cancel_ = false;
       }
       if (max_joint_vel[i] <= 0.0)
       {
@@ -138,11 +138,11 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
           "Joint [%s] has invalid joint velocity defined in URDF [%.1f]. "
           "Falling back to hold position on cancel.",
           params_.joints[i].c_str(), max_joint_vel[i]);
-        decelerate_on_cancel_requested_ = false;
+        should_decelerate_on_cancel_ = false;
       }
     }
     // if everything is valid reserve space for the stop trajectory on cancel
-    if (decelerate_on_cancel_requested_)
+    if (should_decelerate_on_cancel_)
     {
       double max_t_stop = 0.0;
       // find the joint with the largest max time to stop
@@ -336,7 +336,7 @@ controller_interface::return_type JointTrajectoryController::update(
         RCLCPP_WARN(logger, "Aborted due to command timeout");
 
         new_trajectory_msg_.reset();
-        if (decelerate_on_cancel_requested_)
+        if (should_decelerate_on_cancel_)
         {
           // calculate stopping position based on max deceleration
           new_trajectory_msg_.initRT(decelerate_to_hold_position());
@@ -472,7 +472,7 @@ controller_interface::return_type JointTrajectoryController::update(
           RCLCPP_WARN(logger, "Aborted due to state tolerance violation");
 
           new_trajectory_msg_.reset();
-          if (decelerate_on_cancel_requested_)
+          if (should_decelerate_on_cancel_)
           {
             // calculate stopping position based on max deceleration
             new_trajectory_msg_.initRT(decelerate_to_hold_position());
@@ -519,7 +519,7 @@ controller_interface::return_type JointTrajectoryController::update(
             RCLCPP_WARN(logger, "%s", error_string.c_str());
 
             new_trajectory_msg_.reset();
-            if (decelerate_on_cancel_requested_)
+            if (should_decelerate_on_cancel_)
             {
               // calculate stopping position based on max deceleration
               new_trajectory_msg_.initRT(decelerate_to_hold_position());
@@ -538,7 +538,7 @@ controller_interface::return_type JointTrajectoryController::update(
         RCLCPP_ERROR(logger, "Holding position due to state tolerance violation");
 
         new_trajectory_msg_.reset();
-        if (decelerate_on_cancel_requested_)
+        if (should_decelerate_on_cancel_)
         {
           // calculate stopping position based on max deceleration
           new_trajectory_msg_.initRT(decelerate_to_hold_position());
@@ -554,7 +554,7 @@ controller_interface::return_type JointTrajectoryController::update(
         RCLCPP_ERROR(logger, "Exceeded goal_time_tolerance: holding position...");
 
         new_trajectory_msg_.reset();
-        if (decelerate_on_cancel_requested_)
+        if (should_decelerate_on_cancel_)
         {
           // calculate stopping position based on max deceleration
           new_trajectory_msg_.initRT(decelerate_to_hold_position());
@@ -1030,13 +1030,13 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
   }
 
   // velocity state interface is required to calculate ramped stop trajectories
-  if (decelerate_on_cancel_requested_ && !has_velocity_state_interface_)
+  if (should_decelerate_on_cancel_ && !has_velocity_state_interface_)
   {
     RCLCPP_WARN(
       get_node()->get_logger(),
       "Decelerate on cancel is enabled but hardware does not support velocity state interface. "
       "Falling back to hold position on cancel.");
-    decelerate_on_cancel_requested_ = false;
+    should_decelerate_on_cancel_ = false;
   }
 
   auto get_interface_list = [](const std::vector<std::string> & interface_types)
@@ -1519,7 +1519,7 @@ rclcpp_action::CancelResponse JointTrajectoryController::goal_cancelled_callback
     active_goal->setCanceled(action_res);
     rt_active_goal_.writeFromNonRT(RealtimeGoalHandlePtr());
 
-    if (decelerate_on_cancel_requested_)
+    if (should_decelerate_on_cancel_)
     {
       // calculate stopping position based on max deceleration
       add_new_trajectory_msg(decelerate_to_hold_position());
--- a/joint_trajectory_controller/test/test_trajectory_controller.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_controller.cpp
@@ -2921,7 +2921,7 @@ TEST_F(TrajectoryControllerTest, decelerate_to_hold_position_fallback_no_velocit
 
 /**
  * @brief When max_deceleration_on_cancel is 0.0 (the default) for any joint, the
- * controller disables decelerate_on_cancel_requested_ internally during configure and falls back
+ * controller disables should_decelerate_on_cancel_ internally during configure and falls back
  * to set_hold_position on timeout.
  */
 TEST_F(TrajectoryControllerTest, decelerate_to_hold_position_fallback_zero_max_decel)
@@ -2929,7 +2929,7 @@ TEST_F(TrajectoryControllerTest, decelerate_to_hold_position_fallback_zero_max_d
   rclcpp::executors::MultiThreadedExecutor executor;
   constexpr double cmd_timeout = 0.1;
   // decelerate_on_cancel = true but no max_deceleration_on_cancel set (defaults to 0.0)
-  // -> controller disables decelerate_on_cancel_requested_ and falls back to set_hold_position
+  // -> controller disables should_decelerate_on_cancel_ and falls back to set_hold_position
   std::vector<rclcpp::Parameter> params = {
     rclcpp::Parameter("cmd_timeout", cmd_timeout),
     rclcpp::Parameter("constraints.decelerate_on_cancel", true)};


joint_trajectory_controller/CMakeLists.txt
```diff
@@ -106,7 +106,7 @@ if(BUILD_TESTING)
   ament_add_gmock(test_trajectory_actions
     test/test_trajectory_actions.cpp
   )
-  set_tests_properties(test_trajectory_actions PROPERTIES TIMEOUT 300)
+  set_tests_properties(test_trajectory_actions PROPERTIES TIMEOUT 330)
```
MarqRazz, line 109: Tests on my hardware take a little longer than the old limit so I gave it a little margin.
```
Total Test time (real) = 302.84 sec
Finished <<< joint_trajectory_controller [5min 3s]
```

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -226,3 +233,11 @@ joint_trajectory_controller:
         default_value: 0.0,
         description: "Per-joint trajectory offset tolerance at the goal position.",
       }
+      max_deceleration_on_cancel: {
+        type: double,
+        default_value: 0.0,
+        description: "Per-joint max acceleration used to calculate the stopping position when a request is canceled or preempted.",
+        validation: {
+          gt_eq: [0.0]
+        }
```
saikishor, line 244: Why not set the default to infinity?, that way we can skip the other `deceleration_on_cancel` Boolean parameter right?

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -238,6 +238,11 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
    */
   std::shared_ptr<trajectory_msgs::msg::JointTrajectory> set_hold_position();
 
+  /** @brief decelerate at constant rate to a holding position with
+   * zero velocity and acceleration as new command
+   */
+  std::shared_ptr<trajectory_msgs::msg::JointTrajectory> decelerate_to_hold_position();
```
saikishor, line 257: Is there a reason for returning a shared_ptr ?

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
```diff
@@ -209,6 +209,20 @@ void wraparound_joint(
...
 
+/** @brief Creates a trajectory to decelerate at constant rate to a stop
+ * \param state_current The current state (requires velocity state for each joint) to use for the
+ *      starting point of the stop trajectory
+ * \param joint_accelerations The acceleration values used to calculate how fast each joint can stop
+ * \param joints_angle_wraparound Vector of boolean where true value corresponds to a joint that
+ *      wraps around (ie. is continuous).
+ * \param sample_period The time period to sample the stop trajectory at
+ */
+std::shared_ptr<trajectory_msgs::msg::JointTrajectory> decelerate_to_stop(
```
saikishor: Maybe it is better to pass an object  by reference where the method can fill in information, this way we can avoid memory allocation in RT loop

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
...
+    RCLCPP_ERROR(
+      kLogger,
+      "Decelerate to stop calculations failed because input states did not match in size or the "
+      "sample period was zero.");
+    return nullptr;
+  }
+  auto num_joints = joint_names.size();
+  // Precompute per-joint decel, stop time, and hold position
+  std::vector<double> p0(num_joints), v0(num_joints), t_stop(num_joints), p_hold(num_joints),
+    sgn(num_joints);
```
saikishor: We could avoid these with pre allocated Trajectory variable

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
   return has_trajectory_msg() && trajectory_msg_->points.size() > 1;
 }
 
+std::shared_ptr<trajectory_msgs::msg::JointTrajectory> decelerate_to_stop(
+  const std::vector<std::string> & joint_names,
+  const trajectory_msgs::msg::JointTrajectoryPoint & state_current,
+  const std::vector<double> & joint_accelerations,
```
saikishor:
```suggestion
  const std::vector<double> & joint_decelerations,
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
...
+  const std::vector<bool> & joints_angle_wraparound, const double sample_period)
+{
+  // validate inputs are the same size and the sample_period is not zero
+  const size_t expected_size = joint_names.size();
+  if (
+    state_current.positions.size() != expected_size ||
+    state_current.velocities.size() != expected_size ||
+    joint_accelerations.size() != expected_size ||
```
saikishor:
```suggestion
    joint_decelerations.size() != expected_size ||
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
...
+  double max_t_stop = 0.0;
+  for (size_t i = 0; i < num_joints; ++i)
+  {
+    p0[i] = state_current.positions[i];
+    v0[i] = state_current.velocities[i];
+
+    sgn[i] = (v0[i] >= 0.0) ? 1.0 : -1.0;
+
+    // Time to stop (constant decel)
+    t_stop[i] = std::abs(v0[i]) / joint_accelerations[i];
```
saikishor:
```suggestion
    t_stop[i] = std::abs(v0[i]) / joint_decelerations[i];
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
...
+    max_t_stop = std::max(max_t_stop, t_stop[i]);
+
+    // Analytical stop distance and hold position
+    const double stop_distance = (v0[i] * v0[i]) / (2.0 * joint_accelerations[i]);
```
saikishor:
```suggestion
    const double stop_distance = (v0[i] * v0[i]) / (2.0 * joint_decelerations[i]);
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
...
+    p_hold[i] = p0[i] + sgn[i] * stop_distance;
+
+    RCLCPP_DEBUG(
+      kLogger,
+      "Joint [%s] decel [%.3f], stop dist [%.4f], initial vel [%.4f], initial pos [%.4f], hold pos "
+      "[%.4f], time to stop [%.4f]",
+      joint_names[i].c_str(), joint_accelerations[i], stop_distance, v0[i], p0[i], p_hold[i],
```
saikishor:
```suggestion
      joint_names[i].c_str(), joint_decelerations[i], stop_distance, v0[i], p0[i], p_hold[i],
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
...
+  traj->points.reserve(num_points);
+
+  // Build traj points that ramp to zero velocity
+  for (size_t k = 0; k < num_points; ++k)
+  {
+    const double t = static_cast<double>(k) * sample_period;
+    trajectory_msgs::msg::JointTrajectoryPoint pt;
+    pt.positions.resize(num_joints);
+    pt.velocities.resize(num_joints);
+    pt.accelerations.resize(num_joints);
```
saikishor: Let's move all these to preallocated variable

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
...
+        pt.accelerations[i] = 0.0;
+      }
+    }
+    // wrap the each stopping point around if the joint is continuous
+    wraparound_joint(pt.positions, state_current.positions, joints_angle_wraparound);
+    pt.time_from_start = rclcpp::Duration::from_seconds(t);
+    traj->points.emplace_back(std::move(pt));
+  }
+
+  RCLCPP_INFO(
```
saikishor:
```suggestion
  RCLCPP_DEBUG(
```

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -440,4 +445,111 @@ bool Trajectory::has_nontrivial_msg() const
...
+    kLogger, "Created ramped stop trajectory with [%zu] points and max time to stop [%.3f] sec",
+    traj->points.size(), max_t_stop);
+
+  RCLCPP_INFO(
```
saikishor:
```suggestion
  RCLCPP_DEBUG(
```

saikishor (changes requested):
Very nice approach.
Thanks for such a nice implementation.

I've some nitpicks and comments here and there.


I'm curious to know what happens, if the declaration_on_cancel is only set for some joints of the JTC, but not all.

Thank you

pastoriomarco:
Hi!
I tested this feature with moveit (quite painful to set it up with ros2_controllers rolling... if you have some instructions to load it all more simply it'd be great). 

For what I see the decelerated stop works well, and a decelerated stop surely is a way better approach than a no-ramp hard stop.

I think this feature is a **must-have**. It's very common in any industrial robot to have a stop functionality that ramps down with a given max acceleration for each axis. I don't know if I can do anything more than give my feedback here to help approve this PR, the code and comments look good to me.

This said, if I may, the biggest problem I see here is that it's not on-path. After testing with different accelerations (even very low ones to amplify the edge cases), it's clear that this could lead to undesired effects, such as path overshoots and off-path positions.
On modern robots the standard for a normal stop function (and even a safety-stop function in some brands) is an on-path stop.

I tried to implement an on-path soft-stop function in my branch, see #2111 for details and link to code. It works on humble, jazzy, kilted and rolling, but to be honest it's not PR-ready and would need a good polish (lots of copy paste of same code structure in various parts that I'd like to refine). It would be very useful to at least understand if there's interest in merging this kind of functionality and I'll work on it (after discussing the implementation details required, obviously)

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
bmagyar: While using infinity sounds compact & cool, I think the boolean is a more clear-cut approach. With the boolean we can also be stricter on decelaration constraints

MarqRazz: The Boolean parameter is to make it super easy to opt into the functionality. If we always want to use the decelerate ramp then we don't need it and only requires the user set non-zero values for `max_deceleration_on_cancel`.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
MarqRazz: To make it compatible with `std::shared_ptr<trajectory_msgs::msg::JointTrajectory> set_hold_position();`

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
MarqRazz: Something is going to need to preallocate the max size for a worst case stop trajectory which could be quite large depending on the users settings. For example if they are running at 1kHz vs 100 Hz they will need a lot more space for it. Also depends on how fast each joint can go so we will need to plan for the fastest joint and user defined max deceleration.
```cpp
const size_t num_points = static_cast<size_t>(std::ceil(max_time_to_stop / sample_period)) + 1;
traj->points.reserve(num_points);
```

Right now `decelerate_to_stop` is a free function in the trajectory.cpp file. Who should be the owner of this memory? The JTC or the Trajectory class (assuming we move it into the class).

joint_trajectory_controller/src/trajectory.cpp
MarqRazz: Should we move this free function into a class? If so which one?

MarqRazz:
> Hi! I tested this feature with moveit (quite painful to set it up with ros2_controllers rolling... if you have some instructions to load it all more simply it'd be great).

Sorry that is not something we really have for ros2_control. We don't maintain configurations between ros2_control and other packages that use it and my companies robot configuration is private. In my first post to the PR I shared the configuration values you need to set in your control.yaml for the JTC so hopefully that made it a little easier.

> This said, if I may, the biggest problem I see here is that it's not on-path. After testing with different accelerations (even very low ones to amplify the edge cases), it's clear that this could lead to undesired effects, such as path overshoots and off-path positions. On modern robots the standard for a normal stop function (and even a safety-stop function in some brands) is an on-path stop.

This functionality is similar to a Protective or Emergency stop event except the brakes don't engage or power off the servos. This is the safest thing to do, you said stop!

I can't protect users from specifying poor values for deceleration. If the trajectory requested to execute has larger accelerations than they set in the JTC which one should it choose? The user set specific settings for how to handle a cancel event shouldn't they always be applied? Technically the JTC does not enforce or validate a given trajectory against the URDF (which doesn't contain all of the joints dynamic limits) it is running on before it executes it so users are allowed to ask for things that the hardware can struggle or fail to execute.

MarqRazz:
> I'm curious to know what happens, if the declaration_on_cancel is only set for some joints of the JTC, but not all.

If [any of the joints](https://github.com/MarqRazz/ros2_controllers/blob/48fbcc9b5842e023a942963b534a02cde78b381a/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1859-L1873) have zero values for `max_deceleration_on_cancel` it falls back to set_hold_position.

pastoriomarco:
> Sorry that is not something we really have for ros2_control. We don't maintain configurations between ros2_control and other packages that use it and my companies robot configuration is private. In my first post to the PR I shared the configuration values you need to set in your control.yaml for the JTC so hopefully that made it a little easier.

Yep, it made it a lot easier indeed! The hard part was to have all compile and work together with latest(ish) rolling ros2_controller, especially MoveIt and related packages. 

> This functionality is similar to a Protective or Emergency stop event except the brakes don't engage or power off the servos. This is the safest thing to do, you said stop!
> 
> I can't protect users from specifying poor values for deceleration. If the trajectory requested to execute has larger accelerations than they set in the JTC which one should it choose? The user set specific settings for how to handle a cancel event shouldn't they always be applied? Technically the JTC does not enforce or validate a given trajectory against the URDF (which doesn't contain all of the joints dynamic limits) it is running on before it executes it so users are allowed to ask for things that the hardware can struggle or fail to execute.

As I said, this functionality is simply a must-have, no doubt about it! It's one of the available modes of the protective/emergency stop in modern robots indeed. It may be ok for an emergency stop, but surely it's not something I want on a standard stop: it's not on-path, potentially risky for attached workpieces and arguably too hard on the hardware for a standard stop (depending on payload and on the tradeoff between as close to on-path as possible and softer deceleration). I tried too soft accelerations just to confirm the behavior, but even setting everything perfectly it just has several downsides for non-emergency stops.

This kind of stop is actually something that I'm not using in any of my applications, not even for emergency stops, simply because I design the layout so it's never necessary to decelerate with max acceleration: simplifying, what's needed for certification is the robot to be stopped within a certain time and for it to never exit a determined area during operation.  

That's why I feel the need for a soft-stop function too (not _instead_), it's really a primary functionality of any robotic application.

joint_trajectory_controller/include/joint_trajectory_controller/trajectory.hpp
MarqRazz: I pre-allocated the variables based on the URDF defined joint limits and added the ability to dynamically grow if users request trajectories that are faster than defined there (most of the tests do this).

MarqRazz:
Is there anything else you guys would like me to address here?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -79,6 +80,7 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
       for (size_t i = 0; i < params_.joints.size(); ++i)
       {
         auto urdf_joint = model.getJoint(params_.joints[i]);
+        max_joint_vel[i] = urdf_joint->limits->velocity;
```
saikishor, line 83:
```suggestion
        max_joint_vel[i] = std::abss(urdf_joint->limits->velocity);
```

Just curious do we need this?, because right now the URDF doesn't check for positive value

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -97,6 +99,74 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
     RCLCPP_DEBUG(get_node()->get_logger(), "No URDF file given");
   }
 
+  // read the update_period_ for this controller
+  if (get_update_rate() == 0)
+  {
+    throw std::runtime_error("Controller's update rate is set to 0. This should not happen!");
+  }
+  update_period_ =
+    rclcpp::Duration(0.0, static_cast<uint32_t>(1.0e9 / static_cast<double>(get_update_rate())));
```
saikishor, line 108: Just curious, what do you think about using the period from the update callback. 
I'm fine with this appraoch, but just asking your thoughts about it

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -150,6 +150,19 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   // reserved storage for result of the command when closed loop pid adapter is used
   std::vector<double> tmp_command_;
 
+  // If true, enable calculations to stop all joints using constant deceleration
+  bool decelerate_on_cancel_ = false;
```
saikishor: Hey @MarqRazz can't we use the parameter params_.constraints.decelerate_on_cancel directly?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -97,6 +99,74 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
...
+      }
+      if (max_joint_vel[i] <= 0.0)
+      {
+        RCLCPP_ERROR(
+          get_node()->get_logger(),
+          "Joint [%s] has invalid joint velocity defined in URDF [%.1f]. "
+          "Falling back to hold position on cancel.",
+          params_.joints[i].c_str(), max_joint_vel[i]);
+        decelerate_on_cancel_ = false;
+      }
```
saikishor, line 142: ok I see that you are checking for negative velocities here

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -97,6 +99,74 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
...
+    }
+    // if everything is valid reserve space for the stop trajectory on cancel
+    if (decelerate_on_cancel_)
+    {
+      double max_t_stop = 0.0;
+      // find the joint with the largest max time to stop
+      for (size_t i = 0; i < params_.joints.size(); ++i)
+      {
+        stop_time_[i] = std::abs(max_joint_vel[i]) / max_decel_[i];
```
```suggestion
        stop_time_[i] = max_joint_vel[i] / max_decel_[i];
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -266,7 +336,16 @@ controller_interface::return_type JointTrajectoryController::update(
         RCLCPP_WARN(logger, "Aborted due to command timeout");
 
         new_trajectory_msg_.reset();
-        new_trajectory_msg_.initRT(set_hold_position());
+        if (decelerate_on_cancel_)
```
```suggestion
        if (params_.constraints.decelerate_on_cancel)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -393,7 +472,16 @@ controller_interface::return_type JointTrajectoryController::update(
           RCLCPP_WARN(logger, "Aborted due to state tolerance violation");
 
           new_trajectory_msg_.reset();
-          new_trajectory_msg_.initRT(set_hold_position());
+          if (decelerate_on_cancel_)
```
```suggestion
          if (params_.constraints.decelerate_on_cancel)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -431,7 +519,16 @@ controller_interface::return_type JointTrajectoryController::update(
             RCLCPP_WARN(logger, "%s", error_string.c_str());
 
             new_trajectory_msg_.reset();
-            new_trajectory_msg_.initRT(set_hold_position());
+            if (decelerate_on_cancel_)
```
```suggestion
            if (params_.constraints.decelerate_on_cancel)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -441,14 +538,32 @@ controller_interface::return_type JointTrajectoryController::update(
         RCLCPP_ERROR(logger, "Holding position due to state tolerance violation");
 
         new_trajectory_msg_.reset();
-        new_trajectory_msg_.initRT(set_hold_position());
+        if (decelerate_on_cancel_)
```
```suggestion
        if (params_.constraints.decelerate_on_cancel)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -441,14 +538,32 @@ controller_interface::return_type JointTrajectoryController::update(
...
+          new_trajectory_msg_.initRT(set_hold_position());
+        }
       }
       else if (!before_last_point && !within_goal_time && !rt_has_pending_goal_)
       {
         RCLCPP_ERROR(logger, "Exceeded goal_time_tolerance: holding position...");
 
         new_trajectory_msg_.reset();
-        new_trajectory_msg_.initRT(set_hold_position());
+        if (decelerate_on_cancel_)
```
```suggestion
        if (params_.constraints.decelerate_on_cancel)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1394,8 +1509,16 @@ rclcpp_action::CancelResponse JointTrajectoryController::goal_cancelled_callback
     active_goal->setCanceled(action_res);
     rt_active_goal_.writeFromNonRT(RealtimeGoalHandlePtr());
 
-    // Enter hold current position mode
-    add_new_trajectory_msg(set_hold_position());
+    if (decelerate_on_cancel_)
```
```suggestion
    if (params_.constraints.decelerate_on_cancel)
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1791,6 +1914,103 @@ JointTrajectoryController::set_hold_position()
...
+{
+  // if we don't know the velocity of the joints we can't calculate a ramped stop trajectory
+  if (!has_velocity_state_interface_)
+  {
+    RCLCPP_WARN(
+      get_node()->get_logger(),
+      "Hardware does not support velocity state interface. Falling back to hold position on "
+      "cancel.");
+    return set_hold_position();
+  }
```
saikishor: Maybe this better checked in configure?

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -215,6 +215,13 @@ joint_trajectory_controller:
         gt_eq: [0.0],
       }
     }
+    decelerate_on_cancel: {
+      type: bool,
+      default_value: false,
+      description: "If true and each joints max_deceleration_on_cancel is greater than 0,
```
saikishor, line 221: Should this be read_only?, if not, this is good

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -226,3 +233,11 @@ joint_trajectory_controller:
         default_value: 0.0,
         description: "Per-joint trajectory offset tolerance at the goal position.",
       }
+      max_deceleration_on_cancel: {
+        type: double,
+        default_value: 0.0,
+        description: "Per-joint max acceleration used to calculate the stopping position when a request is canceled or preempted.",
```
saikishor, line 240:
```suggestion
        default_value: 0.0,
        read_only: true,
        description: "Per-joint max acceleration used to calculate the stopping position when a request is canceled or preempted.",
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
MarqRazz: I needed a way to check if it was enabled but the configuration was not valid so the functionality is disabled. I'm also worried about the JTC updating the parameters (even if it is read only) and toggling the parameter when the configuration isn't valid.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
MarqRazz: Yep it is checked below.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
MarqRazz: `decelerate_to_hold_position()` is not run in the update_loop so the period is not available. The calculations use this to sample the stop trajectory at the update_period so the interpolation is smooth.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
MarqRazz: Good call, thanks for the suggestion.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
MarqRazz: Yes because this is only validated in `on_init` it should be `read_only`.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
saikishor: If it is read-only, you won't be able to toggle it right?

MarqRazz: If the user has selected it to be enabled but we discover that the configuration is invalid and change the parameter to false then the next update of the parameters could switch it back to true.

christophfroehlich: should we rename this variable to something different than the parameter? See the open conversations from Sai, I think this is just misleading.

I don't have a good suggestion, but a mix of pre/postfixes like `possible`, `should`, `requested`

bmagyar:
I think pushing this back to Jazzy would be very good!

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
bmagyar: `decelerate_on_cancel_requested_` it is.

bmagyar:
@christophfroehlich I've addressed your comments

bmagyar:
I've also stolen the content from Marq's OP and put it into a new docs page :D

PR #2183, opened 2026-02-24
Fix the teardown of the diff_drive_controller test

saikishor:
Should probably fix https://github.com/ros-controls/ros2_control_ci/issues/665

### DIFF
--- a/omni_wheel_drive_controller/test/test_omni_wheel_drive_controller.hpp
+++ b/omni_wheel_drive_controller/test/test_omni_wheel_drive_controller.hpp
@@ -107,7 +107,7 @@ public:
     controller_.reset();
   }
 
-  static void TearDownTestCase() {}
+  static void TearDownTestCase() { rclcpp::shutdown(); }
 
 protected:
   void publish_twist(double linear_x = 1.0, double linear_y = 1.0, double angular = 1.0)


christophfroehlich (dismissed):

omni_wheel_drive_controller/test/test_omni_wheel_drive_controller.hpp
```diff
@@ -98,7 +98,16 @@ class OmniWheelDriveControllerFixture : public ::testing::Test
         "/test_omni_wheel_drive_controller/cmd_vel", rclcpp::SystemDefaultsQoS());
   }
 
-  static void TearDownTestCase() { rclcpp::shutdown(); }
```
Amronos, line 101: Is ``rclcpp::shutdown();`` needed? It's present in diff_drive in this PR, but removed here.

christophfroehlich: I saw that we have a real mess in the different controller tests ;)
Mostly, we don't shutdown the context after ever test case but only after all tests within a file
https://github.com/ros-controls/ros2_controllers/blob/master/diff_drive_controller/test/test_diff_drive_controller.cpp#L1247

omni_wheel_drive_controller/test/test_omni_wheel_drive_controller.hpp
```diff
@@ -98,7 +98,16 @@ class OmniWheelDriveControllerFixture : public ::testing::Test
...
+  void TearDown() override
+  {
+    // Reset the controller before the fixture is destroyed to ensure the controller's
+    // shutdown transition (which clears loaned interfaces) runs while the underlying
+    // StateInterface/CommandInterface objects are still alive. LoanedStateInterface stores
+    // a const reference (not a shared_ptr), so destruction order matters.
+    controller_.reset();
+  }
+
+  static void TearDownTestCase() {}
```
christophfroehlich:
```suggestion
  static void TearDownTestCase() { rclcpp::shutdown(); }
```

omni_wheel_drive_controller/test/test_omni_wheel_drive_controller.hpp
christophfroehlich: but you are right, this should maybe go into a different PR in refactoring all the tests.

Amronos (approved):

PR #2187, opened 2026-02-25
[JSB] Remove dynamic_joint_states and use INDIVIDUAL_BEST_EFFORT for interfaces

saikishor:
This cleans up the deprecated dynamic joint states and then uses the INDIVIDUAL_BEST_EFFORT part to handle the joint_states topic

NOT TO BE BACKPORTED


bmagyar (dismissed):

christophfroehlich (changes requested):
Changes LGTM. This is possibly a behavior change, if the urdf does not contain a joint otherwise published with `interface_configuration_type::ALL`?
At least a hint in the release notes would be great.

saikishor:
> Changes LGTM. This is possibly a behavior change, if the urdf does not contain a joint otherwise published with `interface_configuration_type::ALL`?
> At least a hint in the release notes would be great.

Right now, it does use filter URDF joints parameter in doing so. Sure, I can add to the release notes

christophfroehlich (changes requested):
```
In file included from /workspaces/ros2_ipc_ws/src/ros2_controllers/joint_state_broadcaster/test/test_joint_state_broadcaster.cpp:32:
/workspaces/ros2_ipc_ws/src/ros2_controllers/joint_state_broadcaster/test/test_joint_state_broadcaster.hpp:57:36: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
   57 | constexpr char * kThreeJointURDF = R"(
      |                                    ^~~
...
---
Finished <<< joint_state_broadcaster [27.1s]
```

christophfroehlich (approved):

PR #2204, opened 2026-03-08
Add queue_size to the forward command controller

saikishor:
This adds the queue_size to the forward command controller, so that setting a proper value you can avoid the commands sent from nonRT loop from being missed

### DIFF
--- a/forward_command_controller/include/forward_command_controller/forward_controllers_base.hpp
+++ b/forward_command_controller/include/forward_command_controller/forward_controllers_base.hpp
@@ -86,6 +86,8 @@ protected:
 
   std::vector<std::string> command_interface_types_;
 
+  size_t history_depth_ = 1u;
+
   // the realtime container to exchange the reference from subscriber
   realtime_tools::RealtimeThreadSafeBox<CmdType> rt_command_;
   // save the last reference in case of unable to get value from box
--- a/forward_command_controller/src/forward_command_controller.cpp
+++ b/forward_command_controller/src/forward_command_controller.cpp
@@ -51,6 +51,8 @@ controller_interface::CallbackReturn ForwardCommandController::read_parameters()
     command_interface_types_.push_back(joint + "/" + params_.interface_name);
   }
 
+  history_depth_ = static_cast<size_t>(params_.queue_size);
+
   return controller_interface::CallbackReturn::SUCCESS;
 }
 
--- a/forward_command_controller/src/forward_command_controller_parameters.yaml
+++ b/forward_command_controller/src/forward_command_controller_parameters.yaml
@@ -15,3 +15,12 @@ forward_command_controller:
     },
     read_only: true
   }
+  queue_size: {
+    type: int,
+    description: "Size of the command queue",
+    default_value: 1,
+    read_only: true,
+    validation: {
+      gt_eq<>: 1
+    },
+  }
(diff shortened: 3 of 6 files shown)


Juliaj:
> This adds the queue_size to the forward command controller, so that setting a proper value you can avoid the commands sent from nonRT loop from being missed

Thanks for this addition, the `queue_size` addition makes sense to avoid message loss at the DDS layer.

Is one of the intended use cases supporting action chunking (e.g. batched action sequences from a learned policy)? If so, one friction that I am seeing: FCC currently doesn't seem to have a way to consume a chunk as an ordered sequence at the intended execution cadence. To compensate for that, a policy node could:

- Publish the chunk as a single burst of N messages — all callbacks fire nearly simultaneously, each overwriting `rt_command_`, so only the last action in the chunk survives.
- Publish one action at a time, externally rate-limited to match execution cadence — technically works, but the policy node is now owning the timing, and the chunk semantics are effectively lost.
- ...

For action chunking to work correctly, more work may be needed. If the use case is in scope, happy to discuss more.

saikishor:
> > This adds the queue_size to the forward command controller, so that setting a proper value you can avoid the commands sent from nonRT loop from being missed
> 
> Thanks for this addition, the `queue_size` addition makes sense to avoid message loss at the DDS layer.
> 
> Is one of the intended use cases supporting action chunking (e.g. batched action sequences from a learned policy)? If so, one friction that I am seeing: FCC currently doesn't seem to have a way to consume a chunk as an ordered sequence at the intended execution cadence. To compensate for that, a policy node could:
> 
> * Publish the chunk as a single burst of N messages — all callbacks fire nearly simultaneously, each overwriting `rt_command_`, so only the last action in the chunk survives.
> * Publish one action at a time, externally rate-limited to match execution cadence — technically works, but the policy node is now owning the timing, and the chunk semantics are effectively lost.
> * ...
> 
> For action chunking to work correctly, more work may be needed. If the use case is in scope, happy to discuss more.

@Juliaj the changes proposed in this PR has nothing to do with action chunking. That's completely different topic. It depends on the usecase, I have a usecase where every point of the Trajectory is important and I cannot miss it, and I can use this queue_size to obtain that

Juliaj:
> I have a usecase where every point of the Trajectory is important and I cannot miss it, and I can use this queue_size to obtain that

@saikishor thanks for clarifying.  Just to make sure I understand, is the missing command issue caused by command bursts or a slower FCC update cycle? The fix reduces message loss at the DDS layer, but since `rt_command_` is a single slot, only the last written value is guaranteed to execute.

I'm guessing you considered a `LockFreeQueue` on the RT side but decided not going there. If so, what were the thoughts?

saikishor:
This is because with forwarded command controller maybe you are interested in all commands like me instead of only the latest value. The problem with using latest value is that it causes jerks.

christophfroehlich (changes requested):
release notes are missing. The SystemDefaultsQoS depends on the middleware, but what is default for fastrtps for example? keep_last(10)?

saikishor:
> but what is default for fastrtps for example? keep_last(10)?

No clue

saikishor:
Self note: Change to add the RTBuffer to have more info stored and to not have losses between RT and nonRT loop

christophfroehlich:
With the default value 1 of the newly introduced parameter, you just disable the queue actually?

PR #2205, opened 2026-03-08
[PID Controllers] Set first set point to current measurement

saikishor:
This modifies the PID Controller to use the current measurement as the starting set point to the PID controller. This way the robot might stay in place.


christophfroehlich (commented):
Makes sense, might there be a usecase for the old behavior? should we make that parameterizable?

saikishor:
> Makes sense, might there be a usecase for the old behavior? should we make that parameterizable?

I agree with you. I'll make the change. Maybe for rolling we can set it to this behaviour and for backports we can set it to false?

PR #1146, opened 2024-05-23
[RQT-JTC] limits from jtc controlled joints

delihus:
The further explanation of this issue is here https://github.com/ros-controls/ros2_controllers/pull/1131

PR #2231, opened 2026-03-24
rqt_jtc: Check for interface type when adding joint names

urfeex:
The JTC RQT gui extracts joint names from the required state interfaces of a controller matching the JTC name pattern. If that controller doesn't contain limits for all joints, that controller will be ignored. Therefore, of the controller has setup a speed scaling state interface, that will show up in the list of joints, while there are no limits configured.

Thus, the GUI cannot be used if the controller is setup with a speed scaling state interface.

This commit only adds state interfaces which are of type "position" or "velocity" to the list of joints.

### DIFF
--- a/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
+++ b/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
@@ -467,7 +467,7 @@ def _jtc_joint_names(jtc_info):
         name = "/".join(interface.split("/")[:-1])
         interface_type = interface.split("/")[-1]
         if name not in joint_names:
-            if interface_type == "position" or interface_type == "velocity":
+            if interface_type == "position":
                 joint_names.append(name)
 
     return joint_names


christophfroehlich:
Have you seen https://github.com/ros-controls/ros2_controllers/pull/1131, is this related?

urfeex:
> Have you seen #1131, is this related?

Yes, and yes, but no. The problem described there, is similar, as in the past, basically all joints in the URDF were checked. That was fixed in #1146 as a breakout of #1131 as far as I can see. Since then, only joints in the controller's `required_state_interfaces` (which is effectively what `_jtc_joint_names(jtc_info)` returns) are checked.

The issue described here is that the speed scaling interface is a required state interface of the controller if reading speed scaling is enabled.

Another approach would be to use the required command interfaces, instead, but that would still leave the problem if one configured setting speed scaling from the controller.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_trajectory_controller.py
```diff
@@ -465,8 +465,10 @@ def _jtc_joint_names(jtc_info):
     joint_names = []
     for interface in jtc_info.required_state_interfaces:
         name = "/".join(interface.split("/")[:-1])
+        interface_type = interface.split("/")[-1]
         if name not in joint_names:
-            joint_names.append(name)
+            if interface_type == "position" or interface_type == "velocity":
```
christophfroehlich:
```suggestion
            if interface_type == "position":
```

PR #2237, opened 2026-03-25
[JTC] Return INVALID_GOAL error code for rejected trajectories (#700)

### LINKED ISSUE #700: [JTC] Error codes for trajectory message rejections
I copied the comment from @gavanderhoorn into a new issue, open for discussions.

              > The message will be rejected, which means that nothing moves and the user will have a look in the command prompt.

I guess I would phrase this as:

> The message will be rejected, which means that nothing moves and the user will _**have to**_ look in the command prompt.

that's what I meant when I wrote:

> A single error line in the controller's terminal is probably not sufficient to prevent surprises / confused users.

I've been such a confused user in the past, hence my comment.

> Using action interface, `rclcpp_action::GoalResponse::REJECT` is the result for the action client.

Last time I checked, there was no support for setting the [error_code](https://github.com/ros-controls/control_msgs/blob/c9ce53dfb5e87b3dfd60bdd85b4826b67989ced4/control_msgs/action/FollowJointTrajectory.action#L33) or the [error_string](https://github.com/ros-controls/control_msgs/blob/c9ce53dfb5e87b3dfd60bdd85b4826b67989ced4/control_msgs/action/FollowJointTrajectory.action#L49) as part of the rejection, as you could in ROS 1 (https://github.com/ros2/rclc/issues/271).

That complicates things, but a work-around I've seen is to accept the goal, then 'abort' by never processing it and immediately sending the final result, in which you set the `error_code` and `error_string` to a meaningful value.

Clearly a work-around, and I still consider the limitation discussed in the linked issue a serious design flaw, but it seems without some work, the best you can do.

The advantage of doing something like this is that clients would have machine readable information about what happened to their goal.

"the human needs to look at the command line" is not a viable error handling / recovery strategy in many cases, so should be avoided as much as possible. The `error_code` field could be used by clients to implement automated recovery, or at the very least promote the error returned by the JTC in a way easily visible to users (using some form of UI, or perhaps still an `ERROR` in a log somewhere, but at least not hidden in the rest of the output).

_Originally posted by @gavanderhoorn in https://github.com/ros-controls/ros2_controllers/issues/699#issuecomment-1635574490_

AdvityaCode:
**Problem**
When the Joint Trajectory Controller rejects an invalid trajectory (empty points, wrong joint names, non-monotonic timestamps, etc.), the action client receives an opaque REJECT response with no structured feedback. The only signal available to the client is a log line in the controller's terminal — which is not a viable error-handling strategy for automated systems.

This is a limitation of action protocol (tracked upstream at ros2/rclc#271): GoalResponse::REJECT provides no mechanism to set error_code or error_string on the result.

Closes #700.

**Proposed Solution**
This is my suggested approach — I'm open to feedback and happy to revise the implementation. It is just what I thought would be the best approach.

The workaround discussed in #699 is to always accept goals at the action protocol level, then immediately abort with a populated error_code and error_string if validation fails. I followed that direction here, but I wasn't sure on some of the design choices so please let me know if any of these should be done differently:

- goal_received_callback now unconditionally returns ACCEPT_AND_EXECUTE. Validation is deferred to goal_accepted_callback, where a structured result can be returned to the client.
- goal_accepted_callback performs two checks before processing a goal:
          -  Whether the controller is active (PRIMARY_STATE_INACTIVE → abort with INVALID_GOAL)
          -  Whether the trajectory message is valid (all existing checks → abort with INVALID_GOAL + descriptive error_string)
- validate_trajectory_msg and validate_trajectory_point_field now return std::string instead of bool. An empty string means valid; a non-empty string is the human-readable rejection reason. Each path still logs via RCLCPP_ERROR for operator visibility.
- The topic-based trajectory interface (~/joint_trajectory) is unaffected in behavior.

I wasn't certain whether changing the return type of validate_trajectory_msg to std::string was the right call, or whether an out-parameter or a separate struct would be preferred. Happy to change this if there's a style the project prefers.

**Tests**
- All existing validate_trajectory_msg unit tests updated for the new return type.
- Two new parameterized action tests added:
     - test_goal_aborted_invalid_trajectory_empty_points
     - test_goal_aborted_invalid_trajectory_wrong_joint_names


joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -233,10 +233,15 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   // sorts the joints of the incoming message to our local order
   void sort_to_local_joint_order(
     std::shared_ptr<trajectory_msgs::msg::JointTrajectory> trajectory_msg) const;
-  bool validate_trajectory_msg(const trajectory_msgs::msg::JointTrajectory & trajectory) const;
+  /// Validate a trajectory message. Returns an empty string if valid, or a
+  /// human-readable error description if the trajectory should be rejected.
+  std::string validate_trajectory_msg(
```
christophfroehlich: what do you think of using tl-expected here? Similar to the validators in https://github.com/PickNikRobotics/generate_parameter_library?tab=readme-ov-file#custom-validator-functions

christophfroehlich (changes requested):
Thank you for picking this. 

We will have to discuss this in the next PMC meeting if the accept-and-immediately-reject pattern is something everyone is happy with.

I propose that we use tl::expected (until std::expected is available) instead of returning an empty string.

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
AdvityaCode: I agree with using tl-expected! Just updated the pr: switched validate_trajectory_msg and validate_trajectory_point_field to return tl::expected<void, std::string>. The tl_expected header was already available via validate_jtc_parameters.hpp so no new dependencies were needed. Let me know your thoughts and if there's anything else you'd like changed.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1449,24 +1449,11 @@ void JointTrajectoryController::topic_callback(
...
-
-  if (!validate_trajectory_msg(goal->trajectory))
-  {
-    return rclcpp_action::GoalResponse::REJECT;
-  }
-
-  RCLCPP_INFO(get_node()->get_logger(), "Accepted new action goal");
+  // Always accept so that the client receives a structured result with error_code and
+  // error_string on failure, instead of an opaque REJECT with no feedback (ros2/rclc#271).
   return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE;
```
christophfroehlich, line 1457: maybe we should still check and return `ACCEPT_AND_DEFER` if conditions are not met
```suggestion
  // Precondition: controller must be running
  if (get_lifecycle_id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
  {
    return rclcpp_action::GoalResponse::ACCEPT_AND_DEFER;
  }

  // Validate the trajectory message itself
  const auto trajectory_result = validate_trajectory_msg(goal_handle->get_goal()->trajectory);
  if (!trajectory_result)
  {
    return rclcpp_action::GoalResponse::ACCEPT_AND_DEFER;
  }
  
  return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE;
```

christophfroehlich (commented):
I fixed the API, the tl::expected library changed a bit.
LGTM, I have a suggestion left and want to have a second opinion here.

PR #2244, opened 2026-03-26
set odometry service addition - steering controllers library

kuralme:
Adds odometry set service. Related to the [issue](https://github.com/ros-controls/ros2_controllers/issues/942).
Main features:
- atomic flag to enable set
- executed in the update loop
- testing

Extra: Added `reset()` function with `on_error` and `on_cleanup` lifecycle support; integrated reset call into `on_configure`. Follows similar pattern of [`DiffDriveController::reset()`](https://github.com/ros-controls/ros2_controllers/blob/28cdcb75da77489d5943a55cad7e69367314537f/diff_drive_controller/src/diff_drive_controller.cpp#L605)

### DIFF
--- a/steering_controllers_library/include/steering_controllers_library/steering_controllers_library.hpp
+++ b/steering_controllers_library/include/steering_controllers_library/steering_controllers_library.hpp
@@ -60,6 +60,8 @@ public:
     const std::shared_ptr<control_msgs::srv::SetOdometry::Request> req,
     std::shared_ptr<control_msgs::srv::SetOdometry::Response> res);
 
+  bool reset();
+
   controller_interface::CallbackReturn on_configure(
     const rclcpp_lifecycle::State & previous_state) override;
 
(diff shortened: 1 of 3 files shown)


steering_controllers_library/include/steering_controllers_library/steering_kinematics.hpp
```diff
@@ -218,6 +218,14 @@ class SteeringKinematics
    */
   void reset_odometry();
```
christophfroehlich: same note as with the old PRs, shall we deprecate it and use set_odometry(0.0, 0.0, 0.0) instead?

kuralme: I noticed reset exist on `steering_odometry.hpp` but gave up. later forgot to add to `steering_kinematics.hpp`.

christophfroehlich (commented):
I see that reset was never used actually. In diff_drive_controller it is called in on_cleanup, on_error, on_configure, should we add this here?

kuralme:
Yes, I can implement something like [`DiffDriveController::reset()`](https://github.com/ros-controls/ros2_controllers/blob/28cdcb75da77489d5943a55cad7e69367314537f/diff_drive_controller/src/diff_drive_controller.cpp#L605)to this one as well ... if you are fine its going out of scope of this PR

Issue #2035, opened 2025-12-01
[diff-drive-controller]  Move previous_publish_timestamp_ initialization out of exception block


Juliaj:
## Background

Based on the review from https://github.com/ros-controls/ros2_controllers/pull/1854#discussion_r2573155408, the `previous_publish_timestamp_` was initialized with `RCL_CLOCK_UNINITIALIZED` and handled via exception catching when performing time arithmetic operations. This approach uses exceptions for normal control flow (first-time initialization). It could be improved with an approach similar to how `previous_update_timestamp_` is initialized in `on_configure()`.

As @Amronos pointed out, other controllers may have similar patterns. Consider reviewing and updating them accordingly.

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_control/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/rolling/Tutorials/Workspace/Creating-A-Workspace.html), for Step 3 use "Download Source Code" section with [these instructions](https://control.ros.org/master/doc/getting_started/getting_started.html#building-from-source).

- [ ] 🍴 **Fork the repository** using the handy button at the top of the repository page and **clone** it into `~/ws_ros2_control/src/ros-controls/ros2_control`, [here is a guide that you can follow](https://guides.github.com/activities/forking/) (You will have to remove or empty the existing `ros2_control` folder before cloning your own fork)

- [ ] **Checkout a new branch** using `git checkout -b <branch_name>`

- [ ] 🤖 **Apply `pre-commit`** auto formatting, by running `pip3 install pre-commit` and running `pre-commit install` in the ros2_control repo.

[...]

PR #2245, opened 2026-03-27
Deprecate publish_rate and resolve initialization anti-pattern in diff_drive_controller

Bhavin-umatiya:
### Description
This PR addresses issue #2035 by fixing the initialization of `previous_publish_timestamp_` in `on_configure` to avoid exception-based control flow for first-time loop initialization.

Additionally, following maintainer feedback for architectural consistency and backward compatibility:
1. The `publish_rate` parameter is marked as **deprecated** in the YAML configuration, header, and source files.
2. A generic deprecation warning is added to inform users of its future removal.
3. The exception-based clock check in the control loop has been replaced with an explicit clock-type verification to ensure real-time safety and stability.

This approach preserves backward compatibility while signaling the intent to remove the parameter in a future release.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -240,14 +240,26 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
 
     // TODO(bhavin-umatiya): Remove publish rate functionality
     bool should_publish = false;
-    if (previous_publish_timestamp_ + publish_period_ <= time)
+    if (previous_publish_timestamp_.get_clock_type() != time.get_clock_type())
     {
-      previous_publish_timestamp_ += publish_period_;
+      should_publish = true;
+    }
+    else if (previous_publish_timestamp_ + publish_period_ <= time)
+    {
       should_publish = true;
     }
 
     if (should_publish)
     {
+      if (previous_publish_timestamp_.get_clock_type() != time.get_clock_type())
+      {
+        previous_publish_timestamp_ = time;
+      }
+      else
+      {
+        previous_publish_timestamp_ += publish_period_;
+      }
+
       if (realtime_odometry_publisher_)
       {
         odometry_message_.header.stamp = time;
--- a/diff_drive_controller/src/diff_drive_controller_parameter.yaml
+++ b/diff_drive_controller/src/diff_drive_controller_parameter.yaml
@@ -106,12 +106,12 @@ diff_drive_controller:
     default_value: 10,
     description: "Size of the rolling window for calculation of mean velocity use in odometry.",
   }
-    # TODO(bhavin-umatiya): Remove this parameter as it is deprecated
-    publish_rate: {
-      type: double,
-      default_value: 50.0, # Hz
-      description: "Publishing rate (Hz) of the odometry and TF messages. This parameter is deprecated and will be removed in a future release.",
-    }
+  # TODO(bhavin-umatiya): Remove this parameter as it is deprecated
+  publish_rate: {
+    type: double,
+    default_value: 50.0, # Hz
+    description: "Publishing rate (Hz) of the odometry and TF messages. This parameter is deprecated and will be removed in a future release.",
+  }
   linear:
     x:
       max_velocity: {


bmagyar:
Could you please elaborate on the reason?

Bhavin-umatiya:
Hi @bmagyar,

I made this change to resolve Issue #2035. 

As described in the issue, `previous_publish_timestamp_` was being initialized with `RCL_CLOCK_UNINITIALIZED`. Because of this, the very first time `previous_publish_timestamp_ + publish_period_` was evaluated in the control loop, it threw a `std::runtime_error` which was then caught to do the first-time initialization.

Using exceptions for normal control flow (first-time loop initialization) is generally an anti-pattern. To improve this, I initialized `previous_publish_timestamp_` inside `on_configure()` (just like `previous_update_timestamp_`). Since it is now properly initialized before the control loop starts, the `try-catch` block around the arithmetic is no longer needed.

thedevmystic:
This looks good. `update_and_write_commands` is intended to be real-time safe, using exceptions was indeed an anti-pattern.

Amronos:
Could you change the PR title and description for future reviewers?

Bhavin-umatiya:
@Amronos   Done! I have updated the title and description.

diff_drive_controller/src/diff_drive_controller_parameter.yaml
```diff
@@ -106,11 +106,6 @@ diff_drive_controller:
     default_value: 10,
     description: "Size of the rolling window for calculation of mean velocity use in odometry.",
   }
-  publish_rate: {
```
thedevmystic, line 109: Why remove publish_rate?

It'll break backward compatibility too. Even when removing we first deprecate then remove.

Bhavin-umatiya: @thedevmystic Thank you for pointing that out! I completely agree that removing publish_rate was a mistake since it breaks backward compatibility and the ABI checks.

christophfroehlich: In general, better not to self-resolve threads. This should be done by the reviewers instead.

Bhavin-umatiya:
@christophfroehlich  Thank you for guidance.

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -486,6 +478,15 @@ controller_interface::CallbackReturn DiffDriveController::on_configure(
   publish_rate_ = params_.publish_rate;
   publish_period_ = rclcpp::Duration::from_seconds(1.0 / publish_rate_);
 
+  // TODO(bhavin-umatiya): Remove this warning
+  if (publish_rate_ > 0.0 && !std::isnan(publish_rate_))
+  {
+    RCLCPP_WARN(
+      get_node()->get_logger(),
+      "[deprecated] publish_rate parameter is deprecated, please set the value to 0.0. "
```
thedevmystic:
```suggestion
      "[deprecated] publish_rate parameter is deprecated. "
```

diff_drive_controller/src/diff_drive_controller_parameter.yaml
```diff
@@ -106,11 +106,12 @@ diff_drive_controller:
...
-    type: double,
-    default_value: 50.0, # Hz
-    description: "Publishing rate (Hz) of the odometry and TF messages.",
-  }
+    # TODO(bhavin-umatiya): Remove this parameter as it is deprecated
+    publish_rate: {
+      type: double,
+      default_value: 50.0, # Hz
+      description: "Publishing rate (Hz) of the odometry and TF messages. This parameter is deprecated and limiting should not be performed.",
```
thedevmystic:
```suggestion
      description: "[Deprecated] Publishing rate (Hz) of the odometry and TF messages. This parameter is deprecated and limiting should not be performed.",
```

diff_drive_controller/src/diff_drive_controller.cpp
thedevmystic: As it's deprecated better just tell it'll be removed than saying to use a specific value.

Though, it doesn't cause any issue.

Bhavin-umatiya: Thank you for the suggestion! I've updated the messages to explicitly state that it will be removed.

thedevmystic (approved):
LGTM! Thanks!

diff_drive_controller/src/diff_drive_controller_parameter.yaml
christophfroehlich, line 1: The indentation is wrong here, that's why all tests fail.
Please install and use pre-commit for every repository in ros-controls. (see the contribution guidelines).

Bhavin-umatiya: Thank you for pointing that out! I've corrected the indentation in the YAML file and ensured it passes yamllint. I've also run pre-commit locally to verify the formatting across all modified files. I've also updated the PR title and description as requested.

thedevmystic (changes requested):

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -238,24 +238,28 @@ controller_interface::return_type DiffDriveController::update_and_write_commands
     tf2::Quaternion orientation;
     orientation.setRPY(0.0, 0.0, odometry_.getHeading());
 
+    // TODO(bhavin-umatiya): Remove publish rate functionality
     bool should_publish = false;
-    try
+    if (previous_publish_timestamp_.get_clock_type() != time.get_clock_type())
```
christophfroehlich, line 243: no this is not true. the comparison throws an error if it does not have the same type. previous_publish_timestamp_ will be set to the same clock type later, and next run succeeds.

thedevmystic:
Oh, that's so. Well, then it's good!

Bhavin-umatiya:
Thank you for the merge, @christophfroehlich! I'm glad we could get the initialization fix and deprecation into the project.

PR #2252, opened 2026-03-28
Remove deprecated odometry reset methods

kuralme:
Removes deprecated `reset_odometry` methods that became obsolate with `set_odometry` additions. Related [issue](https://github.com/ros-controls/ros2_controllers/issues/942)

Note: `SteeringOdometry::reset_odometry()` removed as well since it has never referenced, and `SteeringKinematics::set_odometry()` is available now


christophfroehlich (approved):
Thank you!

Issue #2253, opened 2026-03-28
rqt_joint_trajectory_controller: Add unit tests


christophfroehlich:
## Background

rqt_joint_trajectory_controller is a rqt plugin which lacks any tests.

Before adding any integration tests as mentioned https://github.com/ros-controls/ros2_controllers/pull/1982#pullrequestreview-3410210683 I suggest to write basic pytests for the underlying code.

Start by collecting usecases by having a look at the documentation and [recent patches ](https://github.com/ros-controls/ros2_controllers/commits/master/rqt_joint_trajectory_controller/rqt_joint_trajectory_controller) and post a list here. Then check which features are needed and can be unit tested.

If the code is not testable, feel free to refactor it to split rqt/ROS related code from the actual logic of the parser.

## Instructions
Hi, this is a `good-first-issue` issue. This means we've worked to make it more legible to people who either **haven't contributed to our codebase before, or even folks who haven't contributed to open source before**.

We're interested in helping you take the first step, and can answer questions and help you out along the way. Note that we're especially interested in contributions from underrepresented groups!

We know that creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

If you have contributed before, **consider leaving this PR for someone new**, and looking through our general [bug](https://github.com/ros-controls/ros2_control/labels/bug) issues. Thanks!

### 🤔 What you will need to know.

Nothing. This issue is meant to welcome you to Open Source :) We are happy to walk you through the process.

### 📋 Step by Step

- [ ] 🙋 **Claim this issue**: Comment below. If someone else has claimed it, ask if they've opened a pull request already and if they're stuck -- maybe you can help them solve a problem or move it along!

- [ ] 🗄️ **Create a local workspace** for making your changes and testing [following these instructions](https://docs.ros.org/en/rolling/Tutorials/Workspace/Creating-A-Workspace.html), for Step 3 use "Download Source Code" section with [these instructions](https://control.ros.org/master/doc/getting_started/getting_started.html#building-from-source).

- [ ] 🍴 **Fork the repository** using the handy button at the top of the repository page and **clone** it into `~/ws_ros2_control/src/ros-controls/ros2_control`, [here is a guide that you can follow](https://guides.github.com/activities/forking/) (You will have to remove or empty the existing `ros2_control` folder before cloning your own fork)

- [ ] **Checkout a new branch** using `git checkout -b <branch_name>`

[...]

PR #2254, opened 2026-03-29
[rqt_jtc] Use urdf_parser_py

### LINKED ISSUE #891: rqt_joint_trajectory_controller: Use urdf_parser

lakhmanisahil:
Closes #891

# Description

This PR replaces the manual XML DOM parsing in `joint_limits_urdf.py` with `urdf_parser_py.urdf.Robot`, resolving the long-standing TODO in the code. 

This is a fresh implementation inspired by the stale PR #1982. It incorporates the review feedback and also fixes an additional bug that was not handled in that earlier attempt.

## Why this change?

The previous implementation used `xml.dom.minidom` to manually parse URDF XML, extract attributes, and handle missing values using exceptions. This approach was fragile and unnecessarily complex.

The code already had a TODO suggesting migration to `urdf_parser_py`, but an earlier attempt failed due to: `Required attribute not set in XML: upper.` This issue has now been resolved in newer versions of `urdf_parser_py`, making the migration possible.

### `joint_limits_urdf.py`

- Replaced manual XML parsing (`xml.dom.minidom`) with `urdf_parser_py`  
  (using `Robot.from_xml_string()` and `robot.joints`)

- Added preprocessing to remove `<ros2_control>` tags before parsing  
  (needed because `urdf_parser_py` throws an error on unknown tags)

- Fixed how missing joint limits are handled  
  - earlier: missing values caused errors  
  - now: parser sets missing limits to `0.0`  
  - this can create invalid ranges (min = max) and crash the slider  
  - fixed by checking `minval >= maxval`

- Removed the old TODO comment.

### `package.xml`

- Added the runtime dependency:
  ```xml
  <exec_depend>python3-urdf-parser-py</exec_depend>
  ```

### Setup with RRBot example 1

```bash
# Terminal 1: Launch the demo robot
ros2 launch ros2_control_demo_example_1 rrbot.launch.py

# Terminal 2: Deactivate the forward controller if it is holding the same interfaces
ros2 control set_controller_state forward_position_controller inactive

# Terminal 2: Spawn the joint trajectory controller
ros2 run controller_manager spawner joint_trajectory_position_controller   --param-file $(ros2 pkg prefix ros2_control_demo_example_1)/share/ros2_control_demo_example_1/config/rrbot_jtc.yaml

# Terminal 2: Verify the active controllers
ros2 control list_controllers
```

Expected output should show:

- `joint_trajectory_position_controller` as `active`
- `joint_state_broadcaster` as `active`
- `forward_position_controller` as `inactive`

### Launch the GUI

```bash
# Terminal 3
ros2 run rqt_joint_trajectory_controller rqt_joint_trajectory_controller
```

### After this PR

- Joint sliders appear correctly.
- Limits are derived from the URDF using `urdf_parser_py`.
- The GUI behaves as expected for supported robots and controller setups.
- The robot can be controlled through the sliders once the controller is active.

Screenshot from 2026-03-29 04-37-08
Screenshot from 2026-03-29 04-36-48
Screenshot from 2026-03-29 04-36-40
Screenshot from 2026-03-29 04-36-36


rqt_joint_trajectory_controller/package.xml
```diff
@@ -25,6 +25,7 @@
   <exec_depend>controller_manager_msgs</exec_depend>
   <exec_depend>python_qt_binding</exec_depend>
   <exec_depend>python3-rospkg</exec_depend>
+  <exec_depend>python3-urdf-parser-py</exec_depend>
```
christophfroehlich: https://index.ros.org/p/urdfdom_py/#rolling

should be this one, right?
```suggestion
  <exec_depend>urdfdom_py</exec_depend>
```

lakhmanisahil: Thanks for pointing that out - I missed it earlier.
I've updated the dependency to urdfdom_py and verified the changes locally.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -59,67 +55,77 @@ def get_joint_limits(node, joints_names, use_smallest_joint_limits=True):
     dependent_joints = {}
 
     if description != "":
-        robot = xml.dom.minidom.parseString(description).getElementsByTagName("robot")[0]
+        try:
+            # urdf_parser_py does not recognize non-URDF tags such as
+            # <ros2_control> and raises an AssertionError.
```
christophfroehlich: are `<gazebo>` tags fine here?

christophfroehlich (commented):
Thank you! I think it is time for adding unittest with #2253 first, then it's easier to check if this PR is not causing any troubles.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
lakhmanisahil: I checked it, in the the current minidom implementation `<gazebo>` tags are fine because the parser only looks at localName == "joint" nodes — everything else is silently skipped.
But you're right that in this PR , only `<ros2_control>` is being stripped. I hit that error when testing with `ros2_control_demos` so I stripped it specifically.

lakhmanisahil: I am thinking to update it to use a whitelist of known URDF tags (link, joint, transmission, etc) and strip everything else — that way `<gazebo>` and any other non-standard tags are handled automatically without needing to list them one by one.
Is this a good approach to go?

lakhmanisahil:
Hi @christophfroehlich, since issue #2253 is now closed, should I continue working on this PR or close it and create a new one based on the recent changes made to the `joint_limits_urdf.py` file?

christophfroehlich:
Why not continuing here?

lakhmanisahil:
> Why not continuing here?

Got it, thanks. 
I just wanted to confirm if continuing with this PR is the preferred approach.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
christophfroehlich: whitelisting sounds good here

lakhmanisahil:
Hi @christophfroehlich, I have rebased onto the current master with the whitelist preprocessing. Standard URDF tags (link, joint, transmission, material) are kept, and everything else is stripped before parsing. 

21/22 unit tests pass. The one failure is test_missing_velocity_raises, which asserted on the message "Missing velocity limits" produced by our old application code. `urdf_parser_py` now catches the same condition during parsing and raises a ParseError with its own message "Required attribute not set in XML: velocity" before our code gets to inspect the joint.

Screenshot from 2026-04-13 04-57-28

Screenshot from 2026-04-13 05-20-09

lakhmanisahil:
> ParseError with its own message "Required attribute not set in XML: velocity" before our code gets to inspect the joint.


There are two ways I am thinking of handling this:

1. Update the test to expect the new error message from the library.
2. Catch the ParseError and re-raise it with the old message.

I’m leaning toward the first option since it reflects the new behavior. I just wanted to confirm before changing a test that recently merged.

Please let me know what you prefer.

christophfroehlich:
Sure, just change the test if it is no behavior change but only a different message.

lakhmanisahil:
Hi @christophfroehlich, I’ve updated the changes and successfully tested them locally. PTAL.
After the final changes, I’ll update the PR description accordingly.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -78,87 +100,75 @@ def parse_joint_limits(urdf_string, joints_names, use_smallest_joint_limits=True
...
-            joint["has_position_limits"] = jtype != "continuous"
-            joint["max_velocity"] = maxvel
-            free_joints[name] = joint
+
+        # urdf_parser_py defaults lower/upper to 0.0 when the attributes are
+        # absent in the URDF. We treat min >= max as "limits missing" and
+        # either fall back to -pi..pi for continuous joints or raise for
+        # revolute joints where valid bounds are required.
+        minval = joint.limit.lower if joint.limit.lower is not None else 0.0
+        maxval = joint.limit.upper if joint.limit.upper is not None else 0.0
```
christophfroehlich, line 130: if it is 0.0 when absent, when can it be None?

rqt_joint_trajectory_controller/test/test_joint_limits_urdf.py
christophfroehlich, line 272: please update all comments like this one, they do not apply anymore.

christophfroehlich (changes requested):
Thanks for the follow-up. I have some comments, see below

rqt_joint_trajectory_controller/test/test_joint_limits_urdf.py
lakhmanisahil: Hi, do i have to update all the comments in this PR only or a separate PR is needed for the test files?

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
lakhmanisahil: Yes, you are right, i was doing initially to play safe,
I checked this now and urdf_parser_py returns 0 for missing lower/upper, not None. The conditional was dead code. 
Will push the changes soon.

rqt_joint_trajectory_controller/test/test_joint_limits_urdf.py
christophfroehlich: do everything in the same PR please

lakhmanisahil:
Hello @christophfroehlich, updated and addressed the reviews, PTAL.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -45,6 +49,24 @@ def subscribe_to_robot_description(node, key="robot_description"):
...
+    defined by the URDF specification and drop the rest.
+    """
+    root = ET.fromstring(urdf_string)
+    # Iterate over a list copy because we are mutating root during the loop
+    for child in list(root):
+        if child.tag not in _URDF_STANDARD_TAGS:
+            root.remove(child)
+    return ET.tostring(root, encoding="unicode")
+
+
```
saikishor, line 69: Does it really throw exception with urdfdom_py library?

Sorry, If I am coming in middle and asking new questions. I still have to go through the rest of the comments 😅

lakhmanisahil: Hi, I tested this. urdf_parser_py (v1.2.1) does not throw an exception on unknown tags. It prints a warning to stderr like Unknown tag "ros2_control" in /robot[@name='rrbot'] and continues parsing successfully. <gazebo> tags don't even produce a warning.
I initially added the stripping function because during an earlier review @christophfroehlich asked about handling <gazebo> and other vendor extensions, and at the time I believed the library rejected them. That turned out to be wrong , it now only warns.

lakhmanisahil: I tested with and without the function: all 22 unit tests pass in both cases, and the parsed joint data is identical. The only difference is that without stripping, stderr gets warning messages for tags like <ros2_control>.
So the function is not required for correctness, here it just keeps logs clean.

lakhmanisahil: Happy to remove it if you prefer, or keep it with a corrected comment. 
Please, let me know.

saikishor: @christophfroehlich what should we do? Do we need our own parser here?

christophfroehlich: We should not print stderr messages for a correct setup. Either we can catch the stderr messages otherwise, or we strip it like suggested.

saikishor: Got it. @lakhmanisahil look if you can capture the stderr's. If not, this PR approach is fine with me

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -78,87 +100,75 @@ def parse_joint_limits(urdf_string, joints_names, use_smallest_joint_limits=True
...
+                entry["factor"] = joint.mimic.multiplier
+            if joint.mimic.offset is not None:
+                entry["offset"] = joint.mimic.offset
+            dependent_joints[name] = entry
+            continue
+
+        free_joints[name] = {
+            "min_position": minval,
+            "max_position": maxval,
+            "has_position_limits": joint.type != "continuous",
```
saikishor, line 169:
```suggestion
            "has_position_limits": joint.type != "continuous" and joint.type != "fixed",
```

saikishor (commented):
Just one minor nitpick, rest LGTM

saikishor (dismissed):

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
christophfroehlich: we should correct the docstring at least

christophfroehlich (commented):
please see the open threads (and please don't resolve them by yourself, this should be done by reviewers only)

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
saikishor: We can go with trimming the URDF. it's fine for me

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
saikishor: Resolving it as the fixed links are already filtered above

lakhmanisahil:
> please see the open threads (and please don't resolve them by yourself, this should be done by reviewers only)

Hi, apologies for that. I’ve reopened all the threads I had marked as resolved and reviewed the open threads. 
i have made the requested updates in the thread. PTAL.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
christophfroehlich: fine, but this is wrong, right?
> urdf_parser_py validates strictly and raises AssertionError on any unknown tag

christophfroehlich (approved):
Thank you, LGTM! The failing tests are not related, so let's move on

PR #2264, opened 2026-03-30
fix: correct ASSERT_EQ to ASSERT_NE in admittance controller load test

### LINKED ISSUE #583: Fix admittance_controller load test

### LINKED ISSUE #1258: test_load_controller tests always return true

```
```

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_controllers/issues/1256#issuecomment-2299717499_

SouriRishik:
Fixes #1258

## Problem
The `test_load_controller` test in `admittance_controller` was using `ASSERT_EQ` to check against `nullptr`, which means the test was asserting that the controller *fails* to load. Since the controller loads successfully with the params file provided, the test was always passing, even though it was checking the wrong condition.

## Fix
Changed `ASSERT_EQ` to `ASSERT_NE` so the test correctly asserts that the controller loaded successfully (i.e. the result is not nullptr).

Note: `range_sensor_broadcaster` already has the correct `ASSERT_NE` in its load test and does not need changes.

no

## Did you use Generative AI?
no


thedevmystic:
Looks good, but if previous `ASSERT_EQ()` was passing so now this `ASSERT_NE()` would certainly fail, won't it?

SouriRishik:
> Looks good, but if previous `ASSERT_EQ()` was passing so now this `ASSERT_NE()` would certainly fail, won't it?

Good question. The previous `ASSERT_EQ(..., nullptr)` was the bug because it was asserting that the controller fails to load, which is the wrong condition. Since the controller does load successfully with the params file provided, the result is a valid pointer and not nullptr. `ASSERT_NE(..., nullptr)` correctly verifies that the controller loaded successfully. The CI will confirm this.

saikishor (commented):
That's weird. How come the tests were passing till now? 🤔

thedevmystic:
> That's weird. How come the tests were passing till now?

I think the controller must be failing to load (returning nullptr). It's the only way to pass this. That means the there is also a bug.

SouriRishik:
If the previous `ASSERT_EQ(..., nullptr)` was passing, that suggests `load_controller()` might actually be returning nullptr, meaning the controller isn't loading successfully.

I'll verify the actual return value and investigate whether this is due to a configuration/plugin issue or something in the test setup itself, rather than just the assertion.

I'll update the PR once I confirm the behavior.

SouriRishik:
Verified that `load_controller()` returns `nullptr`, so the controller is not loading in this setup.

Reverted the assertion and added an invalid controller test to ensure failure cases are properly covered.

thedevmystic:
The `ASSERT_NE()` in first test is correct as per of logic. So, reverting is not right as it is a bug.

SouriRishik:
okay I'll change that again but what exactly do i follow up, I added additional test function for an invalid controller configuration to explicitly verify failure cases, so both success and failure paths are covered. What do you think of that?

thedevmystic:
From what I see, we have a bug somewhere. Since `ASSERT_EQ(cm.load_controller("load_admittance_controller"), nullptr);` was passing, `ASSERT_NE()` will fail. But the catch is `ASSERT_NE()` is logically correct.

Let's see what maintainers has to say :)

admittance_controller/test/test_load_admittance_controller.cpp
```diff
@@ -36,7 +36,20 @@ TEST(TestLoadAdmittanceController, load_controller)
   cm.set_parameter(
     {"load_admittance_controller.type", "admittance_controller/AdmittanceController"});
 
-  ASSERT_EQ(cm.load_controller("load_admittance_controller"), nullptr);
+  ASSERT_NE(cm.load_controller("load_admittance_controller"), nullptr);
+}
+
+TEST(TestLoadAdmittanceController, load_invalid_controller_fails)
```
thedevmystic:
```suggestion
TEST(TestLoadAdmittanceController, load_invalid_controller)
```

I think we don't use and need `*_fails` prefix (as far I see).

Furthermore, a contrasting check is okay here (though redundant).

admittance_controller/test/test_load_admittance_controller.cpp
```diff
@@ -36,7 +36,20 @@ TEST(TestLoadAdmittanceController, load_controller)
   cm.set_parameter(
     {"load_admittance_controller.type", "admittance_controller/AdmittanceController"});
 
-  ASSERT_EQ(cm.load_controller("load_admittance_controller"), nullptr);
+  ASSERT_NE(cm.load_controller("load_admittance_controller"), nullptr);
```
christophfroehlich: the log output is rather clear:
```
1: [ERROR] [1776276101.257635204] [load_admittance_controller]: Exception thrown during init stage with message: parameter 'kinematics.plugin_name' is not initialized 
1: 
1: [ERROR] [1776276101.257691393] [load_admittance_controller]: Calling shutdown transition of controller node 'load_admittance_controller' due to init failure.
1: [ERROR] [1776276101.257757266] [test_controller_manager]: Could not initialize the controller named 'load_admittance_controller'
```

that's why it was failing. Please try to configure it properly and fix the assertion then.
With https://github.com/ros-controls/ros2_control/pull/3032 in place we can use the 6D robot from the test assets.

admittance_controller/test/test_load_admittance_controller.cpp
```diff
@@ -36,7 +36,20 @@ TEST(TestLoadAdmittanceController, load_controller)
...
+{
+  std::shared_ptr<rclcpp::Executor> executor =
+    std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+
+  controller_manager::ControllerManager cm(
+    executor, ros2_control_test_assets::minimal_robot_urdf, true, "test_controller_manager");
+
+  cm.set_parameter({"bad_controller.type", "invalid/Controller"});
+
+  ASSERT_EQ(cm.load_controller("bad_controller"), nullptr);
```
christophfroehlich:
```suggestion
```
This test does not make sense here, it just tests the controller manager but not the admittance controller.

christophfroehlich (changes requested):

admittance_controller/test/test_params.yaml
```diff
@@ -16,6 +16,12 @@ load_admittance_controller:
       - position
       - velocity
 
+    kinematics:
+      plugin_name: kinematics_interface_kdl/KinematicsInterfaceKDL
+      plugin_package: kinematics_interface
+      base: base_link
+      tip: link_2
+      group_name: arm
```
kamal2730, line 90: Add all this  missing parameters to load controllers properly:

```yaml
ft_sensor:
  name: ft_sensor_name
  frame:
    id: link_6 # tool0 Wrench measurements are in this frame
    external: false # force torque frame exists within URDF kinematic chain
  filter_coefficient: 0.005

control:
  frame:
    id: tool0 # Admittance calcs (displacement etc) are done in this frame. Usually the tool or end-effector
    external: false # control frame exists within URDF kinematic chain

fixed_world_frame: # Gravity points down (neg. Z) in this frame (Usually: world or base_link)
  frame:
    id: base_link # Admittance calcs (displacement etc) are done in this frame. Usually the tool or end-effector
    external: false # control frame exists within URDF kinematic chain

gravity_compensation:
  frame:
    id: tool0
    external: false

  CoG: # specifies the center of gravity of the end effector
    pos:
      - 0.1 # x
      - 0.0 # y
      - 0.0 # z
    force: 23.0 # mass * 9.81

admittance:
  selected_axes:
    - true # x
    - true # y
    - true # z
    - true # rx
    - true # ry
    - true # rz

  # Having ".0" at the end is MUST, otherwise there is a loading error
  # F = M*a + D*v + S*(x - x_d)
  mass:
    - 5.5
    - 6.6
    - 7.7
    - 8.8
    - 9.9
    - 10.10

  damping_ratio: # damping can be used instead: zeta = D / (2 * sqrt( M * S ))
    - 2.828427 # x
    - 2.828427 # y
    - 2.828427 # z
    - 2.828427 # rx
    - 2.828427 # ry
    - 2.828427 # rz

  stiffness:
    - 214.1
    - 214.2
    - 214.3
    - 214.4
    - 214.5
    - 214.6
```

kamal2730 (changes requested):
Add missing parameters and test locally by running:
```bash
colcon test --packages-select admittance_controller --event-handlers console_direct+
```

admittance_controller/test/test_params.yaml
christophfroehlich: thank you for the review!

SouriRishik: yeah i noticed that i didn't push my remaining changes, thanks for pointing out

SouriRishik:
Hello @christophfroehlich, I have made the required changes. Could you please review?

kamal2730 (approved):
LGTM!

PR #2276, opened 2026-04-01
fix(joint_state_broadcaster): suppress confusing warning for standard interfaces

### LINKED ISSUE #2261: confusing joint_state_broadcaster warning
**Describe the bug**
I have a simple `JointStateBroadcaster` configuration:
```yaml
controller_manager:
  ros__parameters:
    rotor_state_broadcaster:
      type: joint_state_broadcaster/JointStateBroadcaster
      
rotor_state_broadcaster:
  ros__parameters:
    joints: [rotor/1, rotor/2, rotor/3, rotor/4]
    interfaces: [velocity]
```
to broadcast the interfaces `rotor/1/velocity`, etc., on topic `/joint_states`. This works, and I am getting the values in the `velocity` field of `sensor_msgs/msg/JointState` on `/joint_states`.

However, the `rotor_state_broadcaster` node is throwing this confusing warning:
```
Mapping from 'velocity' to interface 'velocity' will not be done, because 'velocity' is defined in 'interface' parameter.
```

What does it mean that this mapping will not be done? Obviously, the `velocity` interfaces are correctly mapped to the `velocity` field of `sensor_msgs/msg/JointState`.

What settings do I have to change to inhibit this warning?

**To Reproduce**
1. define and configure a `joint_state_broadcaster/JointStateBroadcaster` controller as shown above
2. launch and observe error message

**Expected behavior**
As this behaves as expected, no warning should be shown. If this is actually not behaving as expected, the warning message should state the error properly.

**Screenshots**

 - OS: Ubuntu
 - Version jazzy

**Additional context**
Add any other context about the problem here, especially include any modifications to ros2_control that relate to this issue.

greencookie-afk:
## Description

This PR fixes the confusing warning message from `JointStateBroadcaster` when using standard interfaces like `velocity` with the default mapping.

## Problem

When configuring the broadcaster with standard interfaces:
```yaml
rotor_state_broadcaster:
  ros__parameters:
    joints: [rotor/1, rotor/2, rotor/3, rotor/4]
    interfaces: [velocity]
```

The following confusing warning was displayed:
```
Mapping from 'velocity' to interface 'velocity' will not be done, because 'velocity' is defined in 'interface' parameter.
```

This warning is confusing because:
1. The user didn't configure any custom mapping
2. The data is being published correctly to the `velocity` field
3. The warning suggests something is wrong when everything works as expected

## Solution

The warning is now only displayed when there's an actual custom mapping being ignored (i.e., when the interface name differs from the JointState field name).

For example, if someone had:
```yaml
interfaces: [velocity]
map_interface_to_joint_state:
  velocity: custom_velocity_interface
```

Then the warning would still be shown because the custom mapping `custom_velocity_interface` -> `velocity` would be ignored.

But for the standard case where `interfaces: [velocity]` and the default mapping `velocity` -> `velocity`, no warning is shown because no mapping is needed.

## Changes

- Modified `joint_state_broadcaster.cpp` to only print the warning when `interface != interface_to_map`

## Related Issue

Fixes #2261

### DIFF
--- a/joint_state_broadcaster/src/joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/src/joint_state_broadcaster.cpp
@@ -124,8 +124,7 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_configure(
       params_.interfaces.end())
     {
       map_interface_to_joint_state_[interface] = interface;
-      // Only warn if there's a custom mapping being ignored (interface != interface_to_map)
-      // When they're equal (e.g., both "velocity"), it's the standard case and no warning is needed
+      // Warn if custom mapping is being ignored
       if (interface != interface_to_map)
       {
         RCLCPP_WARN(
--- a/joint_state_broadcaster/src/joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/src/joint_state_broadcaster.cpp
@@ -125,14 +125,11 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_configure(
     {
       map_interface_to_joint_state_[interface] = interface;
       // Warn if custom mapping is being ignored
-      if (interface != interface_to_map)
-      {
-        RCLCPP_WARN(
-          get_node()->get_logger(),
-          "Mapping from '%s' to interface '%s' will not be done, because '%s' is defined "
-          "in 'interface' parameter.",
-          interface_to_map.c_str(), interface.c_str(), interface.c_str());
-      }
+      RCLCPP_WARN_EXPRESSION(
+        get_node()->get_logger(), interface != interface_to_map,
+        "Mapping from '%s' to interface '%s' will not be done, because '%s' is defined "
+        "in 'interface' parameter.",
+        interface_to_map.c_str(), interface.c_str(), interface.c_str());
     }
     else
     {
--- a/joint_state_broadcaster/src/joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/src/joint_state_broadcaster.cpp
@@ -124,11 +124,16 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_configure(
       params_.interfaces.end())
     {
       map_interface_to_joint_state_[interface] = interface;
-      RCLCPP_WARN(
-        get_node()->get_logger(),
-        "Mapping from '%s' to interface '%s' will not be done, because '%s' is defined "
-        "in 'interface' parameter.",
-        interface_to_map.c_str(), interface.c_str(), interface.c_str());
+      // Only warn if there's a custom mapping being ignored (interface != interface_to_map)
+      // When they're equal (e.g., both "velocity"), it's the standard case and no warning is needed
+      if (interface != interface_to_map)
+      {
+        RCLCPP_WARN(
+          get_node()->get_logger(),
+          "Mapping from '%s' to interface '%s' will not be done, because '%s' is defined "
+          "in 'interface' parameter.",
+          interface_to_map.c_str(), interface.c_str(), interface.c_str());
+      }
     }
     else
     {
--- a/joint_state_broadcaster/doc/joint_state_broadcaster_parameter_context.yml
+++ b/joint_state_broadcaster/doc/joint_state_broadcaster_parameter_context.yml
@@ -18,6 +18,11 @@ map_interface_to_joint_state: |
         \t\tvelocity: <custom_interface>
         \t\teffort: <custom_interface>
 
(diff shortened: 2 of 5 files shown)


cookified:
@christophfroehlich, I opened a PR for this issue. Would you be able to review when you have a chance? I am a beginner so any feedback will be appreciated.

joint_state_broadcaster/src/joint_state_broadcaster.cpp
```diff
@@ -124,11 +124,16 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_configure(
       params_.interfaces.end())
     {
       map_interface_to_joint_state_[interface] = interface;
-      RCLCPP_WARN(
-        get_node()->get_logger(),
-        "Mapping from '%s' to interface '%s' will not be done, because '%s' is defined "
-        "in 'interface' parameter.",
-        interface_to_map.c_str(), interface.c_str(), interface.c_str());
+      // Only warn if there's a custom mapping being ignored (interface != interface_to_map)
+      // When they're equal (e.g., both "velocity"), it's the standard case and no warning is needed
```
thedevmystic:
```suggestion
      // Warn if custom mapping is being ignored
```

Current comment is bit too wordy than needed. We can directly cut the chase to the point.

thedevmystic (changes requested):
Looks good!

thedevmystic:
> @christophfroehlich, I opened a PR for this issue. Would you be able to review when you have a chance? I am a beginner so any feedback will be appreciated.

We're here for the community! Either beginner or expert. Welcome aboard to the project, @greencookie-afk!

greencookie-afk:
@thedevmystic thanks buddy i updated it.!!

christophfroehlich (changes requested):
Thank you. The changes LGTM, but I ask you to 

- clarify the wrong configuration (i.e., when the warning is printed) in the parameter description of `interfaces` and joint_state_broadcaster_parameter_context.yml
- Add a test for this case to have coverage. You could copy TestCustomInterfaceMapping but run it with a velocity interface parameter.

bmagyar:
@greencookie-afk :point_up: we still miss the follow up action on this

cookified:
> @greencookie-afk ☝️ we still miss the follow up action on this

i am working on it.

christophfroehlich:
@greencookie-afk
> 6. Once a PR got reviewed, please don't do force pushes as this will break github UI and subsequent reviews will take more time.

christophfroehlich (changes requested):
The newly added tests do not even build, please have a look

cookified:
> The newly added tests do not even build, please have a look

I'm sorry I'll have a look at it again in a cleaner and better way. Kind of new here so it takes time to adjust.

christophfroehlich:
> > The newly added tests do not even build, please have a look
> 
> I'm sorry I'll have a look at it again in a cleaner and better way. Kind of new here so it takes time to adjust.

We are happy to help, but please check as a minimum if your changes build and tests succeed. See https://control.ros.org/rolling/doc/contributing/contributing.html

joint_state_broadcaster/src/joint_state_broadcaster.cpp
```diff
@@ -124,11 +124,15 @@ controller_interface::CallbackReturn JointStateBroadcaster::on_configure(
...
-        interface_to_map.c_str(), interface.c_str(), interface.c_str());
+      // Warn if custom mapping is being ignored
+      if (interface != interface_to_map)
+      {
+        RCLCPP_WARN(
+          get_node()->get_logger(),
+          "Mapping from '%s' to interface '%s' will not be done, because '%s' is defined "
+          "in 'interface' parameter.",
+          interface_to_map.c_str(), interface.c_str(), interface.c_str());
+      }
```
saikishor: Can you use RCLCPP_WARN_EXPRESSION here instead of the if condition and then printing?. It does the same work

saikishor (changes requested):

joint_state_broadcaster/src/joint_state_broadcaster.cpp
cookified: okay wait lemme try

cookified: Thanks @saikishor, I updated this to use `RCLCPP_WARN_EXPRESSION` here and reran the relevant `joint_state_broadcaster` checks locally in the rolling container (`pre-commit`, `colcon build`, and `colcon test`).

saikishor (approved):
LGTM

christophfroehlich (approved):
Thank you!

cookified:
> Thank you!

Thank you so much everyone

PR #2099, opened 2026-01-07
Unit tests for the new odometry implementation in diff_drive_controller

vedh1234:
Fixes #2033 

**Description**
This PR adds unit tests for the `Odometry` class in `diff_drive_controller` to verify the new mathematical logic introduced in #1854.

**Changes**
- Added `test/test_odometry.cpp` covering:
  - Exact Arc integration (Curved motion)
  - Linear and Pure Rotation motion
  - Open Loop updates
  - Handling of small `dt`
- Updated `CMakeLists.txt` to build the new test.

**Testing**
- Verified locally with `colcon test`: 8 tests passed, 0 failures.

Issue #2039, opened 2025-12-01
Roll-out changes to odometry calculation to all controllers


christophfroehlich:
There were some updates with
https://github.com/ros-controls/ros2_controllers/pull/1854

- Check for which mobile base controllers the changes can be applied
- Comment here if none
- or open a PR for the others

Related to #2033, which should be included as well.

Ishan1923:
Hi @christophfroehlich, I have investigated ackermann_steering_controller. It currently uses the old odometry_.update_... methods without return value checks (unsafe).

I would like to claim the task of porting this controller to the new try_update pattern (similar to PR #1854). I will also verify if the underlying SteeringOdometry library supports the new API or if it needs to be added there as well.

Ishan1923:
I have implemented the odometry safety fixes for the Ackermann controller in PR #2083. It is currently under review.

Devdoot57:
Hi @christophfroehlich , I noticed omni_wheel_drive_controller still uses rclcpp::Time for odometry.
I actually have a local branch ready that updates it to use dt (matching #1854) and adds the missing odometry tests (matching #2033).
Should I go ahead and open a PR for this, or hold off because of the upcoming Odometry API unification (#2251)?

christophfroehlich:
You can open a PR now. If there is any unification taking place, then it's also easier if all controllers already have the same structure.

PR #2286, opened 2026-04-11
Omni wheel drive odometry update

Devdoot57:
## Description
This PR updates the `omni_wheel_drive_controller` odometry implementation to use `double dt` for integration instead of relying on `rclcpp::Time` (matching #1854). It also introduces a comprehensive `gtest` suite to verify the holonomic kinematics and integration math (matching #2099).

This rolls out the architectural improvements requested in #2039 for the omni-wheel base.

**Changes:**
* **Odometry API:** Deprecated `updateFromPos`, `updateFromVel`, and `updateOpenLoop`. Replaced with `update_from_pos`, `update_from_vel`, and `update_open_loop`.
* **Controller Loop:** Updated the main controller update to account for the above mentioned changes.
* **Testing:** Added `test_odometry.cpp` covering:
  * Independent holonomic Y and X movements.
  * Exact arc integration for simultaneous X, Y, and rotational velocities.
  * Boundary checks for small `dt` rejection.


omni_wheel_drive_controller/include/omni_wheel_drive_controller/odometry.hpp
```diff
@@ -27,11 +27,26 @@ class Odometry
...
+  [[deprecated(
+    "Replaced by bool try_update_open_loop(const double & linear_x_vel, const double "
+    "& linear_y_vel, const double & angular_vel, double dt).")]]
   bool updateOpenLoop(
     const double & linear_x_vel, const double & linear_y_vel, const double & angular_vel,
     const rclcpp::Time & time);
+
+  bool update_from_pos(const std::vector<double> & wheels_pos, double dt);
+  bool update_from_vel(const std::vector<double> & wheels_vel, double dt);
+  bool try_update_open_loop(
```
christophfroehlich, line 47:
```suggestion
  bool update_open_loop(
```
why the try_ prefix here?

christophfroehlich: I see that this is similar to #1854, what was the intention back then @Amronos? because it won't get updated when dt is close to zero?

Amronos: I don't really remember the reason now, update_open_loop would be better though.

christophfroehlich (commented):
Changes LGTM, but copilot has some valid suggestions. Please address them, and we can merge this soon!

omni_wheel_drive_controller/src/odometry.cpp
```diff
@@ -51,13 +51,46 @@ bool Odometry::updateFromPos(const std::vector<double> & wheels_pos, const rclcp
...
+    return false;
+  }
+
+  // Estimate angular velocity of wheels using old and current position [rads/s]:
+  if (wheels_pos.size() != wheels_old_pos_.size())
+  {
+    return false;
+  }
+  std::vector<double> wheels_vel(wheels_pos.size());
+  for (size_t i = 0; i < static_cast<size_t>(wheels_pos.size()); ++i)
```
christophfroehlich:
```suggestion
  for (size_t i = 0; i < wheels_pos.size(); ++i)
```

PR #2290, opened 2026-04-12
initial joint values for JointStateBroadcaster

christianrauch:
## Description

Add new parameters to set the initial interface values for JointStateBroadcaster.

Fixes # (issue)

Yes. It adds new parameters. The defaults stay the same, i.e. the default behaviour does not change.

No.

### Additional Information

The tests:
```
colcon test --event-handlers=console_direct+ --packages-select joint_state_broadcaster
```
fail. But they currently also fail on `master` (at least for me locally).

## TODOs

- [x] Fork the repository.
- [ ] Ensure local tests pass. (`colcon test` and `pre-commit run` (requires you to install pre-commit by `pip3 install pre-commit`)

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -31,6 +31,7 @@ joint_state_broadcaster
 * Added parameter ``publish_dynamic_joint_states`` to enable/disable publishing of dynamic joint states. (`#2064 <https://github.com/ros-controls/ros2_controllers/pull/2064>`_)
 * Removed interfaces with other data types than double for publishing to ``dynamic_joint_states``. (`#2115 <https://github.com/ros-controls/ros2_controllers/pull/2115>`_)
 * Parameter ``publish_dynamic_joint_states`` is now deprecated (default changed to ``false``).
+* New parameters ``initial_value.position``, ``initial_value.velocity``, and ``initial_value.effort`` to set the initial values for interfaces
 
 omni_wheel_drive_controller
 *****************************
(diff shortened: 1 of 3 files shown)


saikishor (commented):
Could you explain the motivation behind this change?

christianrauch:
> Could you explain the motivation behind this change?

When your system only exposes `velocity` state interfaces, but you need the `position` for the `rotor_state_broadcaster`, it makes snese to initialise these values with 0, for example. This is akin to the `joint_state_publisher`, which allows you to send 0 for joint values.

saikishor (changes requested):
> When your system only exposes `velocity` state interfaces, but you need the `position` for the `rotor_state_broadcaster`, it makes sense to initialise these values with 0, for example. This is akin to the `joint_state_publisher`, which allows you to send 0 for joint values.

I understand the need. If this is what you need, then the intial value should be per joint name instead of everything together, so it is explicit

christianrauch:
> > When your system only exposes `velocity` state interfaces, but you need the `position` for the `rotor_state_broadcaster`, it makes sense to initialise these values with 0, for example. This is akin to the `joint_state_publisher`, which allows you to send 0 for joint values.
> 
> I understand the need. If this is what you need, then the intial value should be per joint name instead of everything together, so it is explicit

This seems a bit excessive to me. For a 7 DoF arm, this amounts to setting 21 values. You do not have to do this with the `joint_state_publisher`. Isn't there an easier way to do this?

saikishor:
You only set those joints, who lack the feedback not for everything right?

christianrauch:
> You only set those joints, who lack the feedback not for everything right?

The initial values of the joints with feedback will be overridden anyway, so you can as well just set all joints at once. But I see your problem. I guess when a robot's default configuration is not all-0, then initialising all joints to the same value does not make sense.

saikishor:
Moreover, 0 is a valid value and I don't think the JSB should initialise them to a valid value by default. There could be a case where 0 is not within your joint limits too.

christianrauch:
I am closing this as my proposal is not a good solution.

PR #2294, opened 2026-04-13
More general initialization of state from command

peter-mitrano-ar:
## Description

When operating a robot which does not always track desired positions accurately (e.g. torque controlled) you may sometimes want the JTC to initialize based on the current/latest command from the hardware interface, rather than the current measured state. This will ensure the "command" is unchanged when activating the controller. Otherwise, if you use measured state, the arm can move unexpectedly (usually, it will droop).

To handle this, JTC has `set_last_command_interface_value_as_state_on_activation` parameter. However, it's implementation is weird to me. Specifically, `bool read_state_from_command_interfaces` will return `False` if you have a `velocity` state interface but no `velocity` command interface. For me, this restriction seems unnecessary. You simply should need a state interface for each `command` interface, but not the other way around. Extra state interfaces are simply ignored.

## Explanation of changes

Now, we always first call `read_state_from_state_interfaces` in `on_activate`.
Then, if the parameter is set, we update this state from the command interfaces.
Thus, we renamed `read_state_from_command_interfaces` to `update_state_from_command_interface` and change its behavior to only update fields that have command interfaces.
To minimize code duplication, I refactored out `interface_has_values` and `assign_point_from_command_interface`. This is the reason we have more deletions than additions :)

Also I updated the tests to reflect the desired change in behavior.

Fixes # (issue)

https://discourse.openrobotics.org/t/handling-missing-command-interfaces-in-read-state-from-command-interface/53551

Yes!

A bit of copilot autocompletion.


christophfroehlich (changes requested):
Your description seems fair for me. Can you please also improve the parameter description in the yaml file, and add a line to the release_notes?

saikishor (commented):
Is there a way to add a test to this usecase?

peter-mitrano-ar:
Thanks for the quick replies everyone!
@christophfroehlich updated the docs/comments
@saikishor I added a new test for this parameter. I also confirmed (locally) that if I set the parameter to false, the test fails which is good :smile: 

In thinking about these features more, I wonder whether it really makes sense to initialize `state_current_` to command interfaces. I definitely want `last_command_state_` set correctly, and ultimate what I want is that when you first update the commands, you set them to the command not the measured state. The tests seem to be only testing internal state of the controller (`last_commanded_state_`), not what actually gets bound to the hardware interface (I think `joint_command_interface_`). Does anyone have thoughts on tihs?

peter-mitrano-ar:
Not sure about the CI failures, looks related to [this](https://github.com/ros-controls/ros2_control/pull/3048) but not sure if I did something wrong or need to somehow update something to get CI to pass.

christophfroehlich:
> Not sure about the CI failures, looks related to [this](https://github.com/ros-controls/ros2_control/pull/3048) but not sure if I did something wrong or need to somehow update something to get CI to pass.

tl;dr not related ;)

You are right. I already released ros2_control, but the build farm had an outtake this week and the binaries are still not available on ros-testing. Windows build fails because of the removal done by #2016 until we have released this.

doc/release_notes.rst
```diff
@@ -40,6 +40,7 @@ omni_wheel_drive_controller
 
 joint_trajectory_controller
 ***************************
+* When using ``set_last_command_interface_value_as_state_on_activation``, it is no longer required to have state and command for the same interface type (e.g. velocity). With this param set, the JTC state and command will be initialized using a command interface value, if available, and will otherwise fallback to the value read from the state interface. This allows you to have position command and position+velocity state, for example, which previously would have been disallowed (with this param set).
```
christophfroehlich:
```suggestion
* When using ``set_last_command_interface_value_as_state_on_activation``, it is no longer required to have state and command for the same interface type (e.g. velocity). With this param set, the JTC state and command will be initialized using a command interface value, if available, and will otherwise fallback to the value read from the state interface. This allows you to have position command and position+velocity state, for example, which previously would have been disallowed (with this param set).  (`#2294
  <https://github.com/ros-controls/ros2_controllers/pull/2294>`_)
```

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -2385,62 +2385,96 @@ TEST_P(TrajectoryControllerTestParameterized, test_hw_states_has_offset_later_co
...
+
+  executor.cancel();
+}
 
+// Testing the behavior of the set_last_command_interface_value_as_state_on_activation parameter.
+// On activation, both state and command should be equal to the current command values, not current
+// state values.
+TEST_P(TrajectoryControllerTestParameterized, test_set_last_command_interface_on_activation)
+{
+  rclcpp::Parameter const set_last_command_on_activation(
```
christophfroehlich, line 2428: This is the default value, right? 
What is the difference to the test above?

Shouldn't there also be a test with setting set_last_command_interface_value_as_state_on_activation=false?

christophfroehlich (changes requested):

joint_trajectory_controller/test/test_trajectory_controller.cpp
peter-mitrano-ar: ah my mistake, I forgot it was true by default. I have switched it to test the false case instead

peter-mitrano-ar:
@destogl since it seems you wrote the initial implementation, maybe you want to check that this logic change makes sense

christophfroehlich (dismissed):
LGTM

peter-mitrano-ar:
should I rebase on master? or will that trigger needless CI and/or dismiss your approval?

christophfroehlich:
Please never do a rebase if there was already any review in the PR, see the hint in the PR template you certainly have read ;)
Updating via merge via UI without having merge conflicts normally doesn't dismiss approvals. But let's wait for another review, then re-trigger CI before merging.

peter-mitrano-ar:
friendly ping in hopes of a review @saikishor

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -43,6 +43,18 @@
 
 namespace joint_trajectory_controller
 {
+
+auto interface_has_values = [](const auto & joint_interface)
```
```suggestion
auto interface_has_values = [](const auto & joint_interface) -> bool
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -43,6 +43,18 @@
...
+{
+  return std::find_if(
+           joint_interface.begin(), joint_interface.end(),
+           [](const auto & interface)
+           {
+             auto interface_op = interface.get().get_optional();
+             return !interface_op.has_value() || std::isnan(interface_op.value());
```
```suggestion
             return !interface_op.has_value() || !std::isfinite(interface_op.value());
```

saikishor (commented):
Apologies for not getting to this PR before

The changes in general look good to me. Just 2 minor nitpicks. 

Let me know what you think.

peter-mitrano-ar:
@saikishor thanks! Changes included.
@christophfroehlich  since review was dismissed

peter-mitrano-ar:
I clicked "rerequest review", since it looked like the new commit means old approvals are dismissed. Sorry if that generated extra emails/notifications!

saikishor (approved):

PeterMitrano:
@saikishor is there a bot that can backport to Jazzy/Kilted...? I am planning to do a backport to Humble but I am pretty sure that has to be manual since the code has changed a lot.

saikishor:
> @saikishor is there a bot that can backport to Jazzy/Kilted...? I am planning to do a backport to Humble but I am pretty sure that has to be manual since the code has changed a lot.

I think there shouldn't be problem backporting it, as it only affects the case when the interface is only available as a command but not as a state. @christophfroehlich what do you think?

peter-mitrano-ar:
Hmm slight correction -- I believe it effects also the case where you have the interface only as state and not command. That's how I discovered this in the first place. At least in humble, if you had `position` command but `position,velocity` state, the controller would initialize from the state not command.

christophfroehlich:
do you need it on humble? if it is too much effort, I would not backport it
users can always compile the rolling version on humble if needed

peter-mitrano-ar:
I would like it in humble. It shouldn't take more than an hour for me to do the backport. My current workaround, unbelievably, is to remove my velocity state interface :laughing: 

Compiling rolling on humble is possible, yes, but I think I wouldn't do it. It involves also bringing many other repos (ros2 control, ros2 control cmake, pal statistics, ...) from rolling as well.

PR #2293, opened 2026-04-13
diff_drive_controller: set parameters as read_only

kamal2730:
Add read_only: true to all 36 parameters in diff_drive_controller for consistency with other drive controllers.
Fixes #1696

## Description

Right now I added read_only: true for all params because currently, the controller does not support runtime parameter updates for SpeedLimiter-related parameters. This would require adding try_update_params() to the RT loop and implementing a set_params() method in the SpeedLimiter class. If needed, I will address this in a follow-up PR.

Fixes #1696 

### Is this user-facing behavior change?
NO

### Did you use Generative AI?
No

## TODOs

To send us a pull request, please:

- [x] Fork the repository.
- [x] Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
- [x] Ensure local tests pass. (`colcon test` and `pre-commit run` (requires you to install pre-commit by `pip3 install pre-commit`)
- [x] Commit to your fork using clear commit messages.
- [x] Send a pull request, answering any default questions in the pull request interface.
- [x] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.

PR #2313, opened 2026-04-16
Added velocity limiting to the mecanum controller.

tonybaltovski:
## Description

Added velocity limiting to linear x,y direction and angular yaw.  Also, added tests to verify them.

Fixes # (issue)

Users can set velocity limits but if they are not set it will not change behaviors.

No.

## TODOs

- [X] Fork the repository.
- [X] Send a pull request, answering any default questions in the pull request interface.
- [X] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.


mecanum_drive_controller/src/mecanum_drive_controller.cpp
```diff
@@ -267,6 +267,55 @@ controller_interface::CallbackReturn MecanumDriveController::on_configure(
 
   RCLCPP_INFO(get_node()->get_logger(), "MecanumDriveController configured successfully");
```
bmagyar: here we say we configured successfully but can fail afterwards. worth moving down

mecanum_drive_controller/src/mecanum_drive_controller.cpp
```diff
@@ -267,6 +267,55 @@ controller_interface::CallbackReturn MecanumDriveController::on_configure(
...
+  {
+    RCLCPP_ERROR(
+      get_node()->get_logger(), "Failed to configure angular z speed limiter: %s", e.what());
+    return controller_interface::CallbackReturn::ERROR;
+  }
+
+  // Initialize previous commands queue for jerk limiting
+  previous_two_commands_ = std::queue<std::array<double, 3>>();
+  previous_two_commands_.push({{0.0, 0.0, 0.0}});
+  previous_two_commands_.push({{0.0, 0.0, 0.0}});
```
bmagyar:
```suggestion
    previous_two_commands_ = std::queue<std::array<double, 3>>({{{0.0, 0.0, 0.0}}, {{0.0, 0.0, 0.0}}});                                                                                                                                                                                     
```
likely too long but I think putting this on a single expression would be nicer

mecanum_drive_controller/include/mecanum_drive_controller/mecanum_drive_controller.hpp
```diff
@@ -158,6 +159,13 @@ class MecanumDriveController : public controller_interface::ChainableControllerI
   double velocity_in_center_frame_linear_x_;   // [m/s]
   double velocity_in_center_frame_linear_y_;   // [m/s]
   double velocity_in_center_frame_angular_z_;  // [rad/s]
+
+  // Speed limiters
+  std::unique_ptr<control_toolbox::RateLimiter<double>> limiter_linear_x_;
+  std::unique_ptr<control_toolbox::RateLimiter<double>> limiter_linear_y_;
+  std::unique_ptr<control_toolbox::RateLimiter<double>> limiter_angular_z_;
+  // Previous two commands for jerk limiting: queue of [linear_x, linear_y, angular_z]
+  std::queue<std::array<double, 3>> previous_two_commands_;
```
bmagyar, line 172: I can't find where we reset these values when the lifecycle state changes. It'd be best to cover that with a test

bmagyar (changes requested):
Thank you!
Left a few notes but I'm mostly concerned about the reset.

mecanum_drive_controller/src/mecanum_drive_controller.cpp
tonybaltovski: Good call!

mecanum_drive_controller/include/mecanum_drive_controller/mecanum_drive_controller.hpp
tonybaltovski: Added!

bmagyar (dismissed):
Thank you for the quick fixes!

tonybaltovski:
> Thank you for the quick fixes!

No worries but let me get these changes test on a robot before merging though.

MarqRazz (dismissed):

christophfroehlich (changes requested):
As the limiter parameters are not read_only, we should also update them in active state
This is also the way we want o go with [diff_drive_controller](https://github.com/ros-controls/ros2_controllers/pull/2293/changes#diff-f8d3a930a6782d85b295ef6bce89dc2a757ae540cc80128e8044aee6ddee5df9)

tonybaltovski:
@christophfroehlich added the requested changes.

christophfroehlich (approved):
Thank you!

bmagyar (approved):

Issue #1229, opened 2024-07-25
parallel gripper: feedback / improvements 


brennand:
@pac48 Hi Paul, thank you for a great controller, the parallel gripper.  I started to use it and how some feedback, improvements that we might want to consider.

### 1) Command interface
I found the current config confusing for the `max_effort_interface`, `max_velocity_interface`.  I had to look into the code to figure out what these should be. for consistancy and also more flexibility in the future what about changing this to:

`command_interfaces: [position, velocity, effort]`

Then in the hpp we can use the same logic as the state_interface:

```
  std::vector<std::string> interface_names;
  for (const auto & interface : params_.state_interfaces)
  {
    interface_names.push_back(params_.joint + "/" + interface);
  }
```

I would be happy to make a PR for this.

### 2) Effort control

Does this currently work?  I had to modify the code to get effort to be forward to my motor controller, or did I have something wrongly configured?  I had to add this code, some minor changes to get it working:

```
  auto effort_command_interface_it = std::find_if(command_interfaces_.begin(), command_interfaces_.end(),
    [](const hardware_interface::LoanedCommandInterface & command_interface)
    { return command_interface.get_interface_name() == hardware_interface::HW_IF_EFFORT; });

  if (effort_command_interface_it == command_interfaces_.end())
  {
    RCLCPP_ERROR_STREAM(
      get_node()->get_logger(),
      "Expected 1 " << hardware_interface::HW_IF_EFFORT << " effort interface");
    return controller_interface::CallbackReturn::ERROR;
  }

  if (effort_command_interface_it->get_prefix_name() != params_.joint)
  {
    RCLCPP_ERROR_STREAM(
      get_node()->get_logger(), "Command interface is different than joint name `"
                                  << effort_command_interface_it->get_prefix_name() << "` != `"
                                  << params_.joint << "`");
    return controller_interface::CallbackReturn::ERROR;
  }
```

### 3) Hold value

Currently when we set the hold position we use this code:

```
void GripperActionController::set_hold_position()
{
  command_struct_.position_cmd_ = joint_position_state_interface_->get().get_value();
  command_struct_.max_effort_ = params_.max_effort;
  command_struct_.max_velocity_ = params_.max_velocity;
```

Wouldn't it be better to use the effort and velocity that the user has set via an action as we would be setting the force to the max available and could be dangers as this isn't we a lot higher then the user wants? 

### 4) Minor var renaming
Maybe rename this to be more consistent and easier to understand:
[...]

PR #2318, opened 2026-04-19
fix(parallel_gripper): fix effort and velocity interface lookup

AkshayArjun:
## Description

Fix effort and velocity command interface lookup in on_activate().

Previously, the controller searched for interfaces using hardcoded strings
"set_gripper_max_effort" and "set_gripper_max_velocity". This meant effort and velocity
interfaces were silently never populated for any other hardware using
standard interface names.("Controller will claim the specified effort interface, e.g. robotiq_85_left_knuckle_joint/max_effort_interface.", "Controller will claim the speed specified interface, e.g. robotiq_85_left_knuckle_joint/max_effort_interface.", -> the previous code would work for these, but not for soemthing like joint/effort_max or something generic) 

The fix uses interface.get_name() compared against
params_.max_effort_interface and params_.max_velocity_interface (
exactly what was declared in command_interface_configuration() and claimed
from hardware, so this should now work for any generically mentioned names in the yaml file.

Also adds missing std::nullopt resets for effort and speed interfaces in
on_deactivate().

Relates to #1229

Yes. Effort and velocity commands will now actually be written to hardware
when max_effort_interface and max_velocity_interface are configured.
Previously they were silently ignored.

Yes, Claude (Anthropic) was used to help understand the codebase and
discuss the changes. The actual code changes were made manually.

### Additional Information

This is the second of multiple PRs addressing issue #1229.
Tests added to verify effort and velocity interface activation and
deactivation behavior.


christophfroehlich (changes requested):
Thank you. Can you fix also the parameter description here? this does not seem to be correct
Image

AkshayArjun:
> Thank you. Can you fix also the parameter description here? this does not seem to be correct.

Yup, just made the changes to the yaml file, must be updated now.

tpoignonec:
Hi, I also ran into the issue for a project.
Was about to open a PR before I saw that one that is essentially the same.

Is there a remaining issue that prevent merging?

christophfroehlich:
> Is there a remaining issue that prevent merging?

Your review would help!

christophfroehlich (dismissed):

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
tpoignonec, line 368: Maybe we should fail on missing required interfaces

```cpp

  if (!params_.max_effort_interface.empty() && !joint_effort_command_interface_.has_value())
  {
    RCLCPP_ERROR(
      get_node()->get_logger(), "Expected max effort command interface `%s` not found",
      params_.max_effort_interface.c_str());
    return controller_interface::CallbackReturn::FAILURE;
  }
  if (!params_.max_velocity_interface.empty() && !joint_speed_command_interface_.has_value())
  {
    RCLCPP_ERROR(
      get_node()->get_logger(), "Expected max velocity command interface `%s` not found",
      params_.max_velocity_interface.c_str());
    return controller_interface::CallbackReturn::FAILURE;
  }
```

parallel_gripper_controller/test/test_parallel_gripper_controller.cpp
```diff
@@ -141,6 +141,108 @@ TEST_F(GripperControllerTest, ActivateDeactivateActivateSuccess)
...
+
+  auto cmd_if_conf = this->controller_->command_interface_configuration();
+  ASSERT_THAT(cmd_if_conf.names, testing::Contains("joint1/effort"));
+  ASSERT_THAT(cmd_if_conf.names, testing::Contains("joint1/velocity"));
+
+  ASSERT_EQ(
+    this->controller_->on_activate(rclcpp_lifecycle::State()),
+    controller_interface::CallbackReturn::SUCCESS);
+}
+
```
tpoignonec, line 220: Test case for missing interfaces?

```cpp
TEST_F(GripperControllerTest, SetupInvalidLimitInterfacesFails)
{
  this->SetUpController("test_gripper_controller_with_effort_and_velocity");

  this->controller_->get_node()->set_parameter({"max_effort_interface", "joint/invalid_max_effort_interface"});

  ASSERT_TRUE(configure_succeeds(controller_));
  // activate fails
  ASSERT_FALSE(activate_succeeds(controller_));
}
```

tpoignonec (changes requested):

parallel_gripper_controller/include/parallel_gripper_controller/parallel_gripper_action_controller_impl.hpp
AkshayArjun: Yes, that defenietly seems necessary. Ill update the file with the suggested changes. Thanks for the review.

MarqRazz (commented):
@pac48 initially set these up and I don't believe they were made to be a command interface. It was a way for you to set the max gripping force so that it stops closing at that point regardless of the position command you give. Is that still the case with these changes? The PR description sounds like it does this but I just wanted to double check.

AkshayArjun:
@tpoignonec sorry for the delay, but I'hve added the changes you requested

tpoignonec (approved):
LGTM

tpoignonec:
@christophfroehlich I don't think I can resolve the two discussions above, but they can be

christophfroehlich (approved):
After reviewing the code, the old on_activate() lookup using hardcoded strings "set_gripper_max_effort" / "set_gripper_max_velocity" was effectively dead code — since command_interface_configuration() only claims these interfaces when max_effort_interface / max_velocity_interface params are non-empty, the hardcoded get_interface_name() match could never find them anyway. So this PR isn't breaking any existing working setups; it's fixing a feature that was never actually functional.

The fix correctly uses interface.get_name() == params_.max_effort_interface (full name match, consistent with how the interface was claimed), adds proper error reporting when a configured interface isn't found, and cleans up the missing std::nullopt resets in on_deactivate(). Tests cover the new activation/deactivation paths.

Looks good to me.

saikishor (approved):

PR #2359, opened 2026-05-16
[JTC] Fix segfault when last trajectory segment is skipped

### LINKED ISSUE #2282: [JTC] Segfault on skipped last segment
**Describe the bug**
The Joint Trajectory Controller allows velocity-only points in trajectories by filling in the missing positions here:

https://github.com/ros-controls/ros2_controllers/blob/fafccb546c26c997ebb683fb9f64662cd805f0e2/joint_trajectory_controller/src/trajectory.cpp#L192

Sampling within a segment will fill in the positions for this segment's endpoint. So starting from the first, this should fill in the positions for all points sequentially. After the trajectory has played out, the last point will be provided as `output_state`. 

All works well, unless the last segment is skipped for some reason. In my case it happened with a trajectory where the last segment was quite small in duration and sometimes not sampled because of the slow(er) controller rate. In this case, `compute_error_for_joint` will segfault trying to access the desired state, whose positions have not been filled in. 

**To Reproduce**

Provide a velocity-only trajectory with a very small last segment duration and run it with a relatively larger controller rate.

**Expected behavior**

Some sort of error handling or check for empty positions. Hard to say at which point of the call stack an intervention could be made to fix the issue. @domire8 has a suggestion of an extra check [here](https://github.com/ros-controls/ros2_controllers/blob/f730c7afd36bd555c86ffee1f326e0f743ed0545/joint_trajectory_controller/src/trajectory.cpp#L208) to see if the positions are empty and return `false`, with the drawback being that there is no actual feedback on the failure. 

**Environment:**
 - OS: Ubuntu 24
 - Jazzy

dparikh79:
## Summary

Fixes #2282.

When the controller rate is slower than the duration of the last trajectory segment, `Trajectory::sample()` can fall through to the "whole animation has played out" branch without ever entering the last segment in its for-loop. The last point's positions never get deduced from velocities (the deduce step only runs when a segment is actually entered for interpolation at `trajectory.cpp:192`). The empty positions vector then gets copied into `output_state` and downstream `JointTrajectoryController::compute_error_for_joint` accesses `desired.positions[index]` against a zero-length vector, which is the segfault at `0x18` the reporter saw.

This PR recovers in the same branch by deducing the last point's positions from the previous point's positions plus the segment velocities, reusing the existing `deduce_from_derivatives` helper. If even that recovery is not possible (no point along the chain has usable positions, e.g. a velocity-only trajectory whose first sample was past the end), `sample()` now returns `false` so the caller does not dereference an empty positions vector. This matches @domire8's suggestion from the issue, with the additional recovery path on top so the common single-skip case continues working.

## Change

`joint_trajectory_controller/src/trajectory.cpp`: inside the "whole animation has played out" branch, before assigning to `output_state`:

1. If `trajectory_msg_->points[last_idx].positions` is empty and the previous point has positions, call `deduce_from_derivatives(prev_point, last_point, ...)` to fill them in.
2. After the copy to `output_state`, if positions is still empty, set the iterators to `end()` and return `false`.

No change to the inner for-loop path or to non-initialize requests.

## Test plan

Two new regression tests in `joint_trajectory_controller/test/test_trajectory.cpp`:

- [x] `sample_velocity_only_skipped_last_segment_deduces_positions`: trajectory with `p1` (position + velocity), then `p2` and `p3` velocity-only with a tiny last segment duration. After anchoring `trajectory_start_time_` and entering segment 0 (which deduces `p2`), a sample past the end with the fix populates `output_state.positions` via the new recovery path. Without the fix, this asserts on empty positions.
- [x] `sample_velocity_only_all_segments_skipped_returns_false`: trajectory with two velocity-only points, prior state also velocity-only. Sample-time far past the end. With the fix, `sample()` returns `false` rather than copying empty positions into `output_state`.

Local build + test (jazzy branch, since master's `FollowJointTrajectory_Feedback.index` requires a control_msgs that has not yet hit the jazzy binary repos; `trajectory.cpp` and `test_trajectory.cpp` are byte-identical between master and jazzy in the touched area):

```
$ colcon build --packages-select joint_trajectory_controller
Summary: 1 package finished [33.8s]

$ ctest -R 'test_trajectory$'
[==========] 16 tests from 2 test suites ran.
[  PASSED  ] 16 tests.
```

clang-format clean. The 4 pre-existing `ament_cpplint` `whitespace/newline` warnings in `trajectory.cpp` (`} else { ... }` style at lines 155, 189, 334, 365) are not in the touched lines.

## Branch choice

Targeted at `master` per the standard ros-controls workflow. Mergify should backport to `jazzy` (where the reporter is) and `humble` / `kilted` as appropriate.

## AI Assistance Disclosure

Analysis and implementation drafted with Claude assistance. I reviewed every line, traced the bug path against the source, validated the fix and tests locally on jazzy, and am the human contributor accountable for this PR.

### DIFF
--- a/joint_trajectory_controller/src/trajectory.cpp
+++ b/joint_trajectory_controller/src/trajectory.cpp
@@ -209,7 +209,34 @@ bool Trajectory::sample(
   start_segment_itr = --end();
   end_segment_itr = end();
   last_sample_idx_ = last_idx;
+
+  // If the last segment was never entered (e.g. sample_time jumped past a very
+  // short last segment because the controller rate is slower than the segment
+  // duration), the last trajectory point's positions may have never been
+  // deduced from velocities. Recover by deducing them from the previous point
+  // if that point has positions; otherwise return false so the caller does not
+  // dereference an empty positions vector downstream. See #2282.
+  if (trajectory_msg_->points[last_idx].positions.empty() && last_idx > 0)
+  {
+    auto & prev_point = trajectory_msg_->points[last_idx - 1];
+    auto & last_point = trajectory_msg_->points[last_idx];
+    if (!prev_point.positions.empty())
+    {
+      const rclcpp::Time t_prev = trajectory_start_time_ + prev_point.time_from_start;
+      const rclcpp::Time t_last = trajectory_start_time_ + last_point.time_from_start;
+      deduce_from_derivatives(
+        prev_point, last_point, state_before_traj_msg_.positions.size(),
+        (t_last - t_prev).seconds());
+    }
+  }
+
   output_state = (*start_segment_itr);
+  if (output_state.positions.empty())
+  {
+    start_segment_itr = end();
+    end_segment_itr = end();
+    return false;
+  }
   // the trajectories in msg may have empty velocities/accel, so resize them
   if (output_state.velocities.empty())
   {
--- a/joint_trajectory_controller/test/test_trajectory.cpp
+++ b/joint_trajectory_controller/test/test_trajectory.cpp
@@ -1046,3 +1046,98 @@ TEST(TestWrapAroundJoint, wraparound_all_joints_no_offset)
   EXPECT_EQ(current_position[1], next_position[1]);
   EXPECT_EQ(current_position[2], next_position[2]);
 }
+// Regression test for #2282: with a velocity-only trajectory whose last
+// segment is skipped (sample_time advances past the last segment without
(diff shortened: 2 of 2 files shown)


domire8 (approved):
Very neat fix, thank you so much!

saikishor (approved):

saikishor:
Tested locally by running tests alone and I'm able to reproduce the issue and then with the fix, it works

PR #2360, opened 2026-05-16
[JTC] Add goal_timeout parameter to abort action when total trajectory time exceeds limit

### LINKED ISSUE #548: JTC action result never returns

kamal2730:
## Description

Implements a `goal_timeout` parameter for `joint_trajectory_controller` that addresses issue #548: when `goal_time_tolerance = 0.0` and the goal position cannot be reached, the action never returns a result and hangs indefinitely.
## What it does
- Adds `goal_timeout` (double, default 0.0 = disabled)
- When set > 0.0, the action is aborted with `GOAL_TOLERANCE_VIOLATED` if total elapsed time from trajectory start exceeds the timeout
- The check is placed **after** all tolerance checks and action feedback processing, so it cannot be overridden by state tolerance or goal tolerance transitions
- Includes validation: warns when `goal_timeout < constraints.goal_time`
- Both action-level and controller-level tests added

Fixes #548

NO

## TODOs

- [x] Fork the repository.


kamal2730:
Hello @christophfroehlich , this PR is ready for review. Kindly review it when you have time.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -584,6 +571,32 @@ controller_interface::return_type JointTrajectoryController::update(
       // else, run another cycle while waiting for outside_goal_tolerance
       // to be satisfied (will stay in this state until new message arrives)
       // or outside_goal_tolerance violated within the goal_time_tolerance
+      // Check if goal_timeout is exceeded (from trajectory start)
+      if (
+        !before_last_point && !rt_is_holding_ && goal_timeout_ > 0.0 &&
+        (traj_time_ - current_trajectory_->time_from_start()).seconds() > goal_timeout_)
+      {
+        RCLCPP_WARN(logger, "Aborted due to goal_timeout exceeded");
```
MarqRazz: I think this is an error not a warning. You failed.

Should we add messaging (like an info_once) that says the timeout is disabled?

MarqRazz (commented):
What happens when a user specifies it in the [goal request](https://github.com/ros-controls/control_msgs/blob/master/control_msgs/action/FollowJointTrajectory.action#L25-L30) and it started disabled? Will that enable it with those settings from that point on? (Like a parameter change would)
```
# If the joints are not within goal_tolerance after "trajectory finish
# time" + goal_time_tolerance, the goal aborts with error_code set to
# GOAL_TOLERANCE_VIOLATED
JointTolerance[] goal_tolerance
JointComponentTolerance[] component_goal_tolerance
builtin_interfaces/Duration goal_time_tolerance
```

MarqRazz:
Can you detail better how this is different than the [constraints:goal_time](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml#L209-L217) parameter? I've used that and it works as expected. The only thing I don't like about it is that you can't make it a function of the trajectory sent (its just a fixed number). I would like functionality like MoveIt where it does commanded trajectory + percentage. So you can say `goal_time=1.3` and that would add a 30% time buffer.

kamal2730:
Hello @MarqRazz , Thanks for the review 


> What happens when a user specifies it in the [goal request](https://github.com/ros-controls/control_msgs/blob/master/control_msgs/action/FollowJointTrajectory.action#L25-L30) and it started disabled? Will that enable it with those settings from that point on? (Like a parameter change would)
> 
> ```
> # If the joints are not within goal_tolerance after "trajectory finish
> # time" + goal_time_tolerance, the goal aborts with error_code set to
> # GOAL_TOLERANCE_VIOLATED
> JointTolerance[] goal_tolerance
> JointComponentTolerance[] component_goal_tolerance
> builtin_interfaces/Duration goal_time_tolerance
> ```

The goal_timeout parameter is independent from the action goal's goal_time_tolerance field. Setting goal_time_tolerance in the goal request has no effect on goal_timeout .

> Can you detail better how this is different than the [constraints:goal_time](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml#L209-L217) parameter? I've used that and it works as expected. The only thing I don't like about it is that you can't make it a function of the trajectory sent (its just a fixed number). I would like functionality like MoveIt where it does commanded trajectory + percentage. So you can say `goal_time=1.3` and that would add a 30% time buffer.


- constraints.goal_time is a grace period that starts **_after_** the trajectory's last point , it only fires when you've reached the end and the robot hasn't settled into goal tolerances.
- goal_timeout is an absolute wall-clock deadline measured **_from trajectory start_** ,it applies even mid-trajectory, acting as a hard cap.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
kamal2730: I've addressed this in the latest commit.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -133,6 +133,18 @@ joint_trajectory_controller:
      ``cmd_timeout`` must be greater than ``constraints.goal_time``, otherwise ignored.
      If zero, timeout is deactivated",
   }
+  goal_timeout: {
```
MarqRazz: What do you think about renaming this parameter to something like `action_execution_timeout`? The fact that it is so similar to `goal_time` makes it confusing which one you might need.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -584,6 +571,32 @@ controller_interface::return_type JointTrajectoryController::update(
       // else, run another cycle while waiting for outside_goal_tolerance
       // to be satisfied (will stay in this state until new message arrives)
       // or outside_goal_tolerance violated within the goal_time_tolerance
+      // Check if goal_timeout is exceeded (from trajectory start)
+      if (
+        !before_last_point && !rt_is_holding_ && goal_timeout_ > 0.0 &&
+        (traj_time_ - current_trajectory_->time_from_start()).seconds() > goal_timeout_)
+      {
+        RCLCPP_ERROR(logger, "Aborted due to goal_timeout exceeded");
```
MarqRazz: Suggestion
```
RCLCPP_ERROR(logger, "Aborted due to action_execution_timeout [current value for action_execution_timeout]  exceeded  [measured execution time]");
```

It would be helpful if you printed the configured and calculated values so that users can easily tune their application without needing to do their own measurements.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1246,6 +1359,22 @@ controller_interface::CallbackReturn JointTrajectoryController::on_activate(
...
 
+  if (params_.goal_timeout > 0.0)
+  {
+    goal_timeout_ = params_.goal_timeout;
+    if (params_.goal_timeout < default_tolerances_.goal_time_tolerance)
+    {
+      RCLCPP_WARN(
+        logger, "goal_timeout (%f) is less than goal_time tolerance (%f)...", params_.goal_timeout,
+        default_tolerances_.goal_time_tolerance);
+    }
```
MarqRazz: Will the system still run with an invalid config? One suggestion if miss-configured would be to tell the user you are disabling one of them.

MarqRazz (commented):
Thanks for taking the time to describe the difference!

Do you think this new parameter should be part of the `constraints` namespace?

kamal2730:
> Thanks for taking the time to describe the difference!
> 
> Do you think this new parameter should be part of the `constraints` namespace?

I think keeping it top-level makes more sense for consistency , cmd_timeout is also a timeout and lives at the top level, not under constraints. The constraints namespace is primarily for per-joint tolerances (trajectory, goal, max_deceleration) plus goal_time and stopped_velocity_tolerance, which are all about checking deviation between desired and actual state. goal_timeout is a different concept , it's a hard execution deadline, not a tolerance check. So constraints would be a bit misleading.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
kamal2730: I felt the same way too, so I renamed `goal_timeout` to `action_execution_timeout` in the latest commit.

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -209,7 +221,7 @@ joint_trajectory_controller:
     goal_time: {
       type: double,
       default_value: 0.0,
-      description: "Time tolerance for achieving trajectory goal before or after commanded time.
+      description:  "Time tolerance for achieving trajectory goal before or after commanded time.
```
MarqRazz:
```suggestion
      description: "Time tolerance for achieving trajectory goal before or after commanded time.
```

MarqRazz (dismissed):
The code changes look good, thanks for adjusting it  to tell the user what is happening!

christophfroehlich (changes requested):
Main changes LGTM, one thing: please add a note [here](https://github.com/ros-controls/ros2_controllers/blob/master/doc/release_notes.rst).

kamal2730:
@christophfroehlich Added release notes. Could you please review?

saikishor (commented):
@kamal2730  @christophfroehlich if the goal_time_tolerance is zero, and then if the position is not reached within the duration specified in the goal or the published message. Then, it should abort immediately after the time is passed right?. Do we really need a parameter here? Or Am i missing something?

christophfroehlich:
problem is that with goal_time being zero that the action never gets cancelled. at least this is what I understood from the issue.

saikishor:
> problem is that with goal_time being zero that the action never gets cancelled. at least this is what I understood from the issue.

Yes me too, that's what I understood, but then it is a bug. I would solve it properly instead of adding yet another new parameters

kamal2730:
> @kamal2730 @christophfroehlich if the goal_time_tolerance is zero, and then if the position is not reached within the duration specified in the goal or the published message. Then, it should abort immediately after the time is passed right?. Do we really need a parameter here? Or Am i missing something?

Hello @saikishor,

Thanks for the review.

You're right - fixing the `!= 0.0` guard so that `goal_time_tolerance = 0` means "abort immediately at trajectory end" would solve the hanging issue entirely. The combination of `time_from_start` (planned duration) and `goal_time_tolerance` (grace period) already covers both cases: aborting immediately at the end of the trajectory or allowing additional settling time.

However, the current documented and intended behavior of `goal_time = 0` is: *"If set to zero, the controller will wait a potentially infinite amount of time."* Some users may explicitly rely on this behavior because the action is considered complete only after all joints satisfy their configured goal tolerances, regardless of how long that takes. Changing the semantics of `goal_time = 0` would therefore be a breaking change for anyone who expects the controller to keep waiting until the goal tolerances are met.

That's the main reason I chose to introduce a new parameter rather than change the existing behavior.

That said, if the maintainers are comfortable with the breaking change, I'm happy to take that approach instead.

saikishor:
> However, the current documented and intended behavior of goal_time = 0 is: "If set to zero, the controller will wait a potentially infinite amount of time." Some users may explicitly rely on this behavior because the action is considered complete only after all joints satisfy their configured goal tolerances, regardless of how long that takes. Changing the semantics of goal_time = 0 would therefore be a breaking change for anyone who expects the controller to keep waiting until the goal tolerances are met.

I don't think waiting an infinite amount of time is the right approach. I don't think we should allow it. For the action, the user will have to cancel it himself right?

kamal2730:
> I don't think waiting an infinite amount of time is the right approach. I don't think we should allow it. For the action, the user will have to cancel it himself right?

Yes, you are correct. However, `goal_time = 0` does not always cause an infinite wait. It only does so in the specific case where the robot can never satisfy the goal tolerances (for example, due to mechanical constraints).

In most real-world scenarios, the robot simply needs a little more time to settle. Maybe the motor is slightly slower than expected, or there is some friction, and the robot will eventually reach the goal if given additional time. That's the normal case, and I believe this is the reason the current behavior was intentionally designed this way.

If we change `goal_time = 0` to abort immediately at the end of the trajectory, we would break those normal cases. The action would fail even though the robot was about to reach the goal. The current behavior, which is to wait as long as necessary for the goal tolerances to be satisfied, is beneficial for these situations.

The real problem is the edge case where the robot can never reach the goal or  satisfy the goal tolerances. That's where the new `action_execution_timeout` parameter helps. It provides a hard safety limit without changing the existing behavior that many users may rely on.

christophfroehlich:
I'm fine with both approaches. We can't just break current behavior on distros other than rolling. so maybe we merge this only in distros from jazzy backwards

bmagyar:
@christophfroehlich @saikishor , my understanding is that the PR as it is fully feature-flags the new functionality and it should be safe to be backported to Jazzy too. Do we agree on that?

destogl:
This is nice, but would it make sense to put default value different than 0? Something like 5 or 10 seconds?

What does it happen with movement in the case where we don't end? Does robot still moves?

christophfroehlich:
> This is nice, but would it make sense to put default value different than 0? Something like 5 or 10 seconds?

this would possibly break behaviors, so let's merge this as it is to older distros and do the proper fix on rolling/lyrical.

> What does it happen with movement in the case where we don't end? Does robot still moves?

it stops

```
        if (should_decelerate_on_cancel_)
        {
          new_trajectory_msg_.initRT(decelerate_to_hold_position());
        }
        else
        {
          new_trajectory_msg_.initRT(set_hold_position());
        }
```

christophfroehlich:
@mergifyio backport jazzy humble

christophfroehlich (approved):

christophfroehlich:
@Mergifyio backport humble

PR #2392, opened 2026-06-07
fix(tricycle_controller): check non-strict inequality for M_PI_2 steering boundary

### LINKED ISSUE #2391: bug(tricycle_controller): strict inequality for M_PI_2 boundary causes speed scale to drop to ~0 instead of 0.01 limit
**Describe the bug**
In `TricycleController::update()`, there is an attenuation mechanism that scales down the wheel speed (`Ws_write`) based on the steering angle error (`alpha_delta`) between the target wheel angle (`alpha_write`) and the current reading (`alpha_read`).

However, there is an edge case at the boundary of `M_PI_2` (90 degrees) where a strict inequality check (`alpha_delta > M_PI_2`) prevents the controller from applying the intended 1% speed limit (`0.01`). Instead, it falls through to the `else` branch and calculates `cos(alpha_delta)` where `alpha_delta` is exactly `M_PI_2`. Due to floating-point precision, `cos(M_PI_2)` yields a value near zero (`~6.12e-17`), which virtually halts the wheel instead of scaling it to the intended 1% creep speed.

**To Reproduce**
1. Command a spin action (set `linear.x = 0.0` m/s, `angular.z = 0.5` rad/s).
2. The controller sets `alpha_write` to exactly `M_PI_2` (~1.5707963268 rad).
3. At the initial transition state (where current steering angle `alpha_read` is `0.0` rad), `alpha_delta` becomes exactly `M_PI_2`.
4. The condition `alpha_delta > M_PI_2` evaluates to `false` because they are equal.
5. The execution flow falls through to the `else` block, calculating `scale = cos(M_PI_2)`.
6. The output wheel speed is scaled to near-zero (e.g., `2.47e-16` rad/s for an unattenuated target speed of `4.04` rad/s) during the first control cycle, instead of the intended 1% speed limit (`0.04` rad/s).

Here is the empirical CSV printout captured from the `cmd_ackermann` topic at a 50Hz control rate:
```csv
# steering_angle, steering_angle_velocity, speed, acceleration, jerk
0.0,0.0,2.915452003479004,0.0,0.0   # straight driving
0.0,0.0,0.0,0.0,0.0                 # timeout stop
1.5707963705062866,0.0,2.4725010844361235e-16,0.0,0.0  # Transition transient (cos boundary error!)
1.5707963705062866,0.0,4.037900924682617,0.0,0.0       # Steady state (angle tracking complete)
```

### Expected behavior

When the steering error is exactly at or larger than `M_PI_2`, the speed scale should be limited to `0.01` to allow a slow creep motion rather than dropping to `10^-16` (virtually a zero-command dead zone).

### Code Reference

https://github.com/ros-controls/ros2_controllers/blob/jazzy/tricycle_controller/src/tricycle_controller.cpp#L185-L188

### Proposed Solution

Change the strict inequality `>` to `>=` in the boundary check to catch the exact boundary case:

```diff
-  else if (alpha_delta > M_PI_2)
+  else if (alpha_delta >= M_PI_2)
   {
     scale = 0.01;
   }
```

### Environment

- **OS**: Ubuntu 24.04 LTS (Noble Numbat)
- **ROS 2 Distro**: Jazzy Jalisco
- **Version or commit hash**: `ros2_controllers` version `4.40.1`

joseph-ros2-dev:
This PR addresses the edge case in the steering error attenuation mechanism of the tricycle controller.

### Description
When the steering angle error (`alpha_delta`) is exactly `M_PI_2` (e.g. at the start of a zero-radius spin maneuver from a straight-line driving state), the strict inequality check `alpha_delta > M_PI_2` evaluates to `false`. The control logic then falls through to the `else` branch, calculating `scale = cos(M_PI_2)`.

Due to double-precision floating-point limits, `cos(M_PI_2)` yields `~6.12e-17` instead of the expected `0.01` (1% speed limit), virtually halting the wheel command entirely on the first control cycle.

### Solution
Changing the strict inequality `>` to `>=` in the boundary check allows the controller to correctly apply the intended `0.01` speed limit when the error is exactly at the `M_PI_2` boundary.

Fixes #2391


saikishor (changes requested):
Hello,

Can you add a test case for this change?

joseph-ros2-dev:
Hello @saikishor,

I have added the unit test `velocity_attenuation_at_pi_2_boundary` to `test_tricycle_controller.cpp` to verify this boundary behavior. The local test successfully passed (100% passed).

Please let me know if you have any further feedback. Thanks!

saikishor:
> Hello @saikishor,
> 
> I have added the unit test `velocity_attenuation_at_pi_2_boundary` to `test_tricycle_controller.cpp` to verify this boundary behavior. The local test successfully passed (100% passed).
> 
> Please let me know if you have any further feedback. Thanks!

Good testing. Thanks for adding it

saikishor (dismissed):

christophfroehlich (commented):
I'm not entirely sure if this state makes sense. This logic just reduces wheel speed until the steering angle is reached. why should it not stop if the delta_angle is Pi/2?

christophfroehlich:
> I'm not entirely sure if this state makes sense. This logic just reduces wheel speed until the steering angle is reached. why should it not stop if the delta_angle is Pi/2?

@joseph-ros2-dev can you comment on this plz?

dlanov (commented):
Thanks for inviting additional review. I traced the current code and its history.

At `alpha_delta == M_PI_2`, `cos(alpha_delta)` is mathematically zero, so changing the result to `0.01` appears to be a control-policy choice rather than only a floating-point boundary fix. Stopping at that exact state and retaining a small nonzero floor both seem defensible, but I could not find a documented requirement for the old controller.

The current head also normalizes by `cos(M_PI / 6)`, which removes the discontinuity at 30 degrees and changes attenuation throughout the 30–90 degree range. That is broader than the PR title and description.

Could the intended behavior be clarified and the PR description updated? It would also help to add tests just below, at, and just above the 30-degree boundary, and clarify whether this implementation should match the clamp and floor used by `steering_controllers_library`.

I confirmed that the existing M_PI_2 test reaches the intended state and correctly verifies the selected `0.01` behavior. My question is about the behavioral contract rather than the test execution.

PR #2400, opened 2026-06-09
Remove deprecated steering_odometry class

christophfroehlich:
## Description

Cleanup after #1996

apart from API break, no

no


MarqRazz (approved):
Thanks for cleaning this up!

Issue #84, opened 2020-08-18
Port over "update existing trajectory" to joint_trajectory_controller


v-lopez:
As implemented in ROS1 control: https://github.com/ros-controls/ros_controllers/blob/melodic-devel/joint_trajectory_controller/include/joint_trajectory_controller/init_joint_trajectory.h#L149

There's a test implemented in test_trajectory_controller.cpp named `test_execute_partial_traj_in_future` that is disabled right now waiting for this feature.

pastoriomarco:
I'm trying to figure out how to port the ros_controllers logic to ros2_controllers maintaining the original behavior and without changing the current one, but I'm wondering:
- The current behavior for trajectory update does work, just not as in the documentation. See [here](https://github.com/ros-controls/ros2_controllers/issues/1455#issuecomment-2920420636) for a workaround.
- The way the current ros2_controllers trajectory update works requires to know the previous trajectory, but so does sending a new trajectory that can be successfully executed within the robot's limits, as the trajectory itself must be constructed to be coherent with the current robot's state. 
- It's been almost 5 years since this issue was opened, so it doesn't seem something urgently needed by the user base, or it may have already been solved otherwise by those who need it. In the latter case, modifying it could break the current user applications that leverage it (like mine, potentially).

I think that the ros1 implementation is more elegant and complete, but is it necessary? I mean, after so much time wouldn't it be safer to update the documentation with the ros2_controllers behavior?

Sure, having the same behavior as in ROS1 would help the transition to ROS2 now that ROS1 is EOL, but it would also need to be consistent with the current ros2_controller's behavior to be safe with the current ROS2 applications.


**PS: Some further considerations:**
Currently, the trajectory substitution on ros_controllers seems to:
- Use absolute time, instead of relative time in ROS2: without keeping track of when the current traj was sent I don't think there's a way to get the actual time_from_start. At least on Humble, the actual.time_from_start field in the feedback's trajectory_msgs/JointTrajectoryPoint is always 0.
- Find the first valid traj point of the new traj that is in the future, and the last valid point that can be executed in the old traj before the first point of the new traj
- Discard everything before 'now'.
- Checks if the robot has joints with more than a complete revolution, and applies offsets to the new traj if needed
- Build a 'bridge' segment from the last point of the old traj to the fist point of the new traj

[...]

PR #2401, opened 2026-06-09
Trajectory blending with new trajectory deferral

vedh1234:
## Description

Partial port of the ROS 1 "update existing trajectory" behavior to ROS 2 joint_trajectory_controller.

Related to #84 

What this PR does:

- Adds a new parameter allow_trajectory_blending (default: true)
- When enabled, a trajectory with a future header.stamp is deferred: the currently active trajectory continues executing until that start time arrives, then the new trajectory is installed
- This eliminates the legacy behavior where a future-stamped trajectory would immediately kill the active trajectory and create a slow spline ramp to the first point
- Deferred trajectories are correctly dropped on goal cancellation, controller deactivation, and controller reset
- Action Server feedback/tolerance checking is suppressed during the deferral period to prevent spurious aborts
- Immediate trajectories (stamp = 0 or stamp <= now) behave exactly as before (no behavioral change)

What this PR does NOT target: ( Compared to ROS1 blending feature )
- Does not allow omitted joints to continue their old trajectory after the new one starts (ROS 1 full blend). Omitted joints freeze at their current position via fill_partial_goal(), same as legacy ROS 2 behavior
- Does not handle speed scaling during the deferral period. The trigger uses wall-clock time, so if speed_scaling != 1.0, there will be a positional mismatch at handoff
- Does not track multiple action goals simultaneously. The new goal immediately preempts the old one at the Action Server level; only the physical trajectory execution is deferred

Yes. When allow_trajectory_blending is true (default), the active trajectory will continue running until the new trajectory's start time.

Yes, Google Gemini and Claude were used to make improvements in manually written code, architectural analysis, edge-case review. 

### Additional Information

- The existing disabled test test_execute_partial_traj_in_future referenced in #84 has been enabled and passes

## TODOs

- [x] Fork the repository.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -48,6 +48,11 @@ joint_trajectory_controller
   up to the first waypoint will use the same interpolation as the rest of the trajectory. (`#2043
   <https://github.com/ros-controls/ros2_controllers/pull/2043>`_)
 * Added decelerate-to-stop functionality when a trajectory is canceled or preempted. Instead of immediately holding position, the controller can now smoothly decelerate each joint to a stop using the per-joint ``max_deceleration_on_cancel`` parameter. (`#2163 <https://github.com/ros-controls/ros2_controllers/pull/2163>`_)
+* Added trajectory deferral support via the ``allow_trajectory_replacement`` parameter. When enabled,
+  a trajectory with a future ``header.stamp`` is deferred: the currently active trajectory
+  continues executing until the start time arrives, then the new trajectory is installed.
+  This is a partial port of the ROS 1 trajectory replacement behavior.
+  (`#2401 <https://github.com/ros-controls/ros2_controllers/pull/2401>`_)
 
 pid_controller
 **************
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -50,7 +50,7 @@ joint_trajectory_controller
 * Added decelerate-to-stop functionality when a trajectory is canceled or preempted. Instead of immediately holding position, the controller can now smoothly decelerate each joint to a stop using the per-joint ``max_deceleration_on_cancel`` parameter. (`#2163 <https://github.com/ros-controls/ros2_controllers/pull/2163>`_)
 * Added trajectory deferral support via the ``allow_trajectory_replacement`` parameter. When enabled,
   a trajectory with a future ``header.stamp`` is deferred: the currently active trajectory
-  continues executing until the start time arrives, then the new trajectory is installed.
+  continues executing until the start time arrives, then the new trajectory is executed.
   This is a partial port of the ROS 1 trajectory replacement behavior.
   (`#2401 <https://github.com/ros-controls/ros2_controllers/pull/2401>`_)
 
--- a/joint_trajectory_controller/doc/trajectory.rst
+++ b/joint_trajectory_controller/doc/trajectory.rst
@@ -119,16 +119,16 @@ Joint trajectory messages allow to specify the time at which a new trajectory sh
   Partial support for this functionality has been ported to ROS 2 via the ``allow_trajectory_replacement`` parameter (see `#84 <https://github.com/ros-controls/ros2_controllers/issues/84>`__).
   When enabled, a trajectory arriving with a future ``header.stamp`` is deferred. The controller
   continues executing the active trajectory until the requested start time is reached, at which
-  point the new trajectory is installed and execution begins.
+  point the new trajectory is handled, and the execution begins.
 
   The current ROS 2 implementation differs from the legacy behavior described below in the following ways:
 
   + The controller does not splice or stitch the current and new trajectories together. At the handoff time, the new trajectory fully replaces the old one.
   + Omitted joints in a partial goal are filled with their current position at the handoff time (hold-in-place), rather than continuing to follow the old trajectory.
   + Only a single deferred trajectory is stored. A newer deferred trajectory overwrites any previously deferred one.
-  + When a new action goal is accepted and deferred, the currently active goal is immediately aborted instead of continuing to emit feedback until the handoff time.
+  + When a new action goal is accepted, the currently active goal is immediately aborted instead of continuing to emit feedback until the handoff time.
   + While a deferred action goal is waiting for its start time, it does not publish action feedback and is not evaluated for path tolerances.
-  + The handoff trigger uses the controller's wall-clock time. If the controller is running with a ``speed_scaling`` other than 1.0, the handoff may occur out of sync with the active trajectory's execution state.
+  + The handoff trigger uses the controller's clock time. If the controller is running with a ``speed_scaling`` other than 1.0, the handoff may occur out of sync with the active trajectory's execution state.
 
 The arrival of a new trajectory command does not necessarily mean that the controller will completely discard the currently running trajectory and substitute it with the new one.
 Rather, the controller will take the useful parts of both and combine them appropriately, yielding a smarter trajectory replacement strategy.
--- a/joint_trajectory_controller/doc/trajectory.rst
+++ b/joint_trajectory_controller/doc/trajectory.rst
@@ -115,10 +115,20 @@ Trajectory Replacement
 
 Joint trajectory messages allow to specify the time at which a new trajectory should start executing by means of the header timestamp, where zero time (the default) means "start now".
 
-.. warning::
+.. note::
+  Partial support for this functionality has been ported to ROS 2 via the ``allow_trajectory_replacement`` parameter (see `#84 <https://github.com/ros-controls/ros2_controllers/issues/84>`__).
+  When enabled, a trajectory arriving with a future ``header.stamp`` is deferred. The controller
+  continues executing the active trajectory until the requested start time is reached, at which
+  point the new trajectory is installed and execution begins.
 
-  As of now, this functionality is not ported to ROS 2, see `this issue <https://github.com/ros-controls/ros2_controllers/issues/84>`__ for more information.
-  The current implementation just forgets the old trajectory.
+  The current ROS 2 implementation differs from the legacy behavior described below in the following ways:
+
+  + The controller does not splice or stitch the current and new trajectories together. At the handoff time, the new trajectory fully replaces the old one.
+  + Omitted joints in a partial goal are filled with their current position at the handoff time (hold-in-place), rather than continuing to follow the old trajectory.
+  + Only a single deferred trajectory is stored. A newer deferred trajectory overwrites any previously deferred one.
+  + When a new action goal is accepted and deferred, the currently active goal is immediately aborted instead of continuing to emit feedback until the handoff time.
+  + While a deferred action goal is waiting for its start time, it does not publish action feedback and is not evaluated for path tolerances.
+  + The handoff trigger uses the controller's wall-clock time. If the controller is running with a ``speed_scaling`` other than 1.0, the handoff may occur out of sync with the active trajectory's execution state.
 
 The arrival of a new trajectory command does not necessarily mean that the controller will completely discard the currently running trajectory and substitute it with the new one.
 Rather, the controller will take the useful parts of both and combine them appropriately, yielding a smarter trajectory replacement strategy.
--- a/joint_trajectory_controller/test/test_trajectory_actions.cpp
+++ b/joint_trajectory_controller/test/test_trajectory_actions.cpp
@@ -1470,7 +1470,7 @@ INSTANTIATE_TEST_SUITE_P(
  */
 TEST_F(TestTrajectoryActions, blend_cancel_deferred_action_goal)
 {
-  SetUpExecutor({rclcpp::Parameter("allow_trajectory_blending", true)});
+  SetUpExecutor({rclcpp::Parameter("allow_trajectory_replacement", true)});
 
   setup_controller_hw_ = true;
   controller_hw_thread_ = std::thread(
@@ -1536,7 +1536,7 @@ TEST_F(TestTrajectoryActions, blend_cancel_deferred_action_goal)
  */
 TEST_F(TestTrajectoryActions, blend_action_replaces_action_with_future_stamp)
 {
-  SetUpExecutor({rclcpp::Parameter("allow_trajectory_blending", true)});
+  SetUpExecutor({rclcpp::Parameter("allow_trajectory_replacement", true)});
 
   setup_controller_hw_ = true;
   controller_hw_thread_ = std::thread(
@@ -1602,7 +1602,7 @@ TEST_F(TestTrajectoryActions, blend_action_replaces_action_with_future_stamp)
  */
 TEST_F(TestTrajectoryActions, blend_topic_preempts_deferred_action_goal)
 {
-  SetUpExecutor({rclcpp::Parameter("allow_trajectory_blending", true)});
+  SetUpExecutor({rclcpp::Parameter("allow_trajectory_replacement", true)});
 
   setup_controller_hw_ = true;
   controller_hw_thread_ = std::thread(
@@ -1670,7 +1670,7 @@ TEST_F(TestTrajectoryActions, blend_topic_preempts_deferred_action_goal)
  */
 TEST_F(TestTrajectoryActions, blend_deactivate_aborts_deferred_action_goal)
 {
-  SetUpExecutor({rclcpp::Parameter("allow_trajectory_blending", true)});
+  SetUpExecutor({rclcpp::Parameter("allow_trajectory_replacement", true)});
 
   setup_controller_hw_ = true;
   controller_hw_thread_ = std::thread(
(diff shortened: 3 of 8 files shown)


christophfroehlich (dismissed):
> Partial port of the ROS 1 "update existing trajectory" behavior to ROS 2 joint_trajectory_controller.
> 
> Fixes #84

For documentation here, what is different from the ROS 1 behavior/what is maybe missing to port?

Please also update the [release_notes](https://github.com/ros-controls/ros2_controllers/blob/master/doc/release_notes.rst) and update the [documentation](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/doc/trajectory.rst#trajectory-replacement).

doc/release_notes.rst
```diff
@@ -48,6 +48,11 @@ joint_trajectory_controller
   up to the first waypoint will use the same interpolation as the rest of the trajectory. (`#2043
   <https://github.com/ros-controls/ros2_controllers/pull/2043>`_)
 * Added decelerate-to-stop functionality when a trajectory is canceled or preempted. Instead of immediately holding position, the controller can now smoothly decelerate each joint to a stop using the per-joint ``max_deceleration_on_cancel`` parameter. (`#2163 <https://github.com/ros-controls/ros2_controllers/pull/2163>`_)
+* Added trajectory deferral support via the ``allow_trajectory_replacement`` parameter. When enabled,
+  a trajectory with a future ``header.stamp`` is deferred: the currently active trajectory
+  continues executing until the start time arrives, then the new trajectory is installed.
```
saikishor:
```suggestion
  continues executing until the start time arrives, then the new trajectory is executed.
```

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
```diff
@@ -589,7 +589,7 @@ class TrajectoryControllerTest : public ::testing::Test
     rclcpp::Time start_time = rclcpp::Time(0, 0, RCL_STEADY_TIME),
     const rclcpp::Duration update_rate = rclcpp::Duration::from_seconds(0.01))
   {
-    if (start_time == rclcpp::Time(0, 0, RCL_STEADY_TIME))
+    if (start_time.nanoseconds() == 0)
```
saikishor, line 597: Why did you have to change this?

saikishor:
I'll review the logic and tests soon. I need time

joint_trajectory_controller/test/test_trajectory_controller_utils.hpp
bmagyar: IMO this is a better comparison as it doesn't involve the type of the clock in a zero check

vedh1234: For the new test I had to pass a future stamp which was required to use ROS_TIME to pass `validate_trajectory_msg`. This caused a clock type mismatch in `updateControllerAsync` because the check `start_time == rclcpp::Time(0, 0, RCL_STEADY_TIME)` throws when start_time is a different clock type. Since the check is purely testing for a zero value, I switched to `start_time.nanoseconds() == 0` which is clock-type-agnostic.

doc/release_notes.rst
saikishor:
```suggestion
  continues executing until the start time arrives, then the new trajectory is executed.
```

joint_trajectory_controller/doc/trajectory.rst
```diff
@@ -115,10 +115,20 @@ Trajectory Replacement
...
 
-.. warning::
-
-  As of now, this functionality is not ported to ROS 2, see `this issue <https://github.com/ros-controls/ros2_controllers/issues/84>`__ for more information.
-  The current implementation just forgets the old trajectory.
+.. note::
+  Partial support for this functionality has been ported to ROS 2 via the ``allow_trajectory_replacement`` parameter (see `#84 <https://github.com/ros-controls/ros2_controllers/issues/84>`__).
+  When enabled, a trajectory arriving with a future ``header.stamp`` is deferred. The controller
+  continues executing the active trajectory until the requested start time is reached, at which
+  point the new trajectory is installed and execution begins.
```
saikishor:
```suggestion
  point the new trajectory is handled, and the execution begins.
```

joint_trajectory_controller/doc/trajectory.rst
```diff
@@ -115,10 +115,20 @@ Trajectory Replacement
...
+
+  The current ROS 2 implementation differs from the legacy behavior described below in the following ways:
+
+  + The controller does not splice or stitch the current and new trajectories together. At the handoff time, the new trajectory fully replaces the old one.
```
saikishor, line 126: We discussed that if the deffered time is in future, but if the trajectory A is still executing by that time, we infer the position, velocity and acceleration from there and continue right?

joint_trajectory_controller/doc/trajectory.rst
```diff
@@ -115,10 +115,20 @@ Trajectory Replacement
...
+  + Omitted joints in a partial goal are filled with their current position at the handoff time (hold-in-place), rather than continuing to follow the old trajectory.
+  + Only a single deferred trajectory is stored. A newer deferred trajectory overwrites any previously deferred one.
+  + When a new action goal is accepted and deferred, the currently active goal is immediately aborted instead of continuing to emit feedback until the handoff time.
```
saikishor:
```suggestion
  + When a new action goal is accepted, the currently active goal is immediately aborted instead of continuing to emit feedback until the handoff time.
```

When it is accepted it is not deferred anymore, we are fully replacing it for action goals

joint_trajectory_controller/doc/trajectory.rst
```diff
@@ -115,10 +115,20 @@ Trajectory Replacement
...
+  + While a deferred action goal is waiting for its start time, it does not publish action feedback and is not evaluated for path tolerances.
+  + The handoff trigger uses the controller's wall-clock time. If the controller is running with a ``speed_scaling`` other than 1.0, the handoff may occur out of sync with the active trajectory's execution state.
```
saikishor:
```suggestion
  + The handoff trigger uses the controller's clock time. If the controller is running with a ``speed_scaling`` other than 1.0, the handoff may occur out of sync with the active trajectory's execution state.
```

saikishor (changes requested):
The main logic looks good. 

I still need to verify the tests though

joint_trajectory_controller/doc/trajectory.rst
vedh1234: Yes, if interpolate_from_desired_state is true, it does use the state (pos/vel/acc) from Trajectory A to interpolate to the new trajectory. I used 'fully replaces' rather than 'splices' because it destroys the rest of Trajectory A's planned path, meaning omitted joints immediately freeze, is that statement fine or it's sounding more like we don't create the spline?

joint_trajectory_controller/doc/trajectory.rst
vedh1234: Yeah you are right, I meant it's trajectory can be deferred, but the goal handle is fully destroyed hence can't give feedback until handoff time.

joint_trajectory_controller/doc/trajectory.rst
saikishor: @vedh1234 Next time use Add suggestion to batch and then commit them, so these duscussions will be auto resolved

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -1463,3 +1463,260 @@ TEST_P(TestTrajectoryActionsTestScalingFactor, test_scaling_sampling_is_correct)
...
+  }
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(150));  // well before 500 ms fire time
+  auto goal_handle_b = gh_b.get();
+  ASSERT_TRUE(goal_handle_b);
+  action_client_->async_cancel_goal(goal_handle_b);
+
+  controller_hw_thread_.join();
+
+  EXPECT_EQ(rclcpp_action::ResultCode::CANCELED, resultcode_b);
```
saikishor, line 1526: once the goal_b is sent you are not verifying the status of goal_a handle

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -1463,3 +1463,260 @@ TEST_P(TestTrajectoryActionsTestScalingFactor, test_scaling_sampling_is_correct)
...
+    GoalOptions opts_b;
+    opts_b.result_callback = [&resultcode_b](const GoalHandle::WrappedResult & r)
+    { resultcode_b = r.code; };
+    action_client_->async_send_goal(goal_b, opts_b);
+  }
+
+  controller_hw_thread_.join();
+
+  EXPECT_EQ(rclcpp_action::ResultCode::ABORTED, resultcode_a);
+  EXPECT_EQ(rclcpp_action::ResultCode::SUCCEEDED, resultcode_b);
```
saikishor, line 1587: I see that this is being triggered, but in this test shouldn't the A start immediately instead of future timestamp and then when B enter it is aborted. This is a nice test

Maybe it is interesting to get the time_now and verify it continuously during the cycle?

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -1463,3 +1463,260 @@ TEST_P(TestTrajectoryActionsTestScalingFactor, test_scaling_sampling_is_correct)
...
+}
+
+/**
+ * @brief A stamp=0 topic trajectory preempts a deferred action goal at the trajectory level:
+ * the topic trajectory installs immediately and the deferred goal never fires.
+ * NOTE: the action goal handle is not formally canceled on topic preemption — that
```
saikishor, line 1593: Maybe it should cancel?

We shall discuss about that behaviour today

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -1463,3 +1463,260 @@ TEST_P(TestTrajectoryActionsTestScalingFactor, test_scaling_sampling_is_correct)
...
+    std::vector<JointTrajectoryPoint> points_a(1);
+    points_a[0].time_from_start = rclcpp::Duration::from_seconds(0.5);
+    points_a[0].positions = {7.0, 8.0, 9.0};
+    control_msgs::action::FollowJointTrajectory_Goal goal_a;
+    goal_a.goal_time_tolerance = rclcpp::Duration::from_seconds(1.0);
+    goal_a.trajectory.joint_names = joint_names_;
+    goal_a.trajectory.points = points_a;
+    goal_a.trajectory.header.stamp =
+      traj_controller_->get_node()->now() + rclcpp::Duration::from_seconds(1.0);
+    action_client_->async_send_goal(goal_a, goal_options_);
```
saikishor, line 1636: Any reasons for sending 2 goals for this test?

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -1463,3 +1463,260 @@ TEST_P(TestTrajectoryActionsTestScalingFactor, test_scaling_sampling_is_correct)
...
+      while (clock.now() < end_time)
+      {
+        now_time = clock.now();
+        traj_controller_->update(now_time, now_time - last_time);
+        last_time = now_time;
+      }
+      traj_controller_->get_node()->deactivate();
+      std::this_thread::sleep_for(std::chrono::milliseconds(200));
+    });
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
saikishor: This may end up flaky test. Increase the time and reduce the sleep?

joint_trajectory_controller/test/test_trajectory_actions.cpp
```diff
@@ -1463,3 +1463,260 @@ TEST_P(TestTrajectoryActionsTestScalingFactor, test_scaling_sampling_is_correct)
...
+    points_b[0].time_from_start = rclcpp::Duration::from_seconds(0.4);
+    points_b[0].positions = {7.0, 8.0, 9.0};
+    control_msgs::action::FollowJointTrajectory_Goal goal_b;
+    goal_b.goal_time_tolerance = rclcpp::Duration::from_seconds(1.0);
+    goal_b.trajectory.joint_names = joint_names_;
+    goal_b.trajectory.points = points_b;
+    goal_b.trajectory.header.stamp =
+      traj_controller_->get_node()->now() + rclcpp::Duration::from_seconds(0.5);
+    action_client_->async_send_goal(goal_b, goal_options_);
+  }
```
saikishor, line 1704: Again why 2 goals here? I don't understand the reasoning behind it

joint_trajectory_controller/test/test_trajectory_controller.cpp
```diff
@@ -3236,3 +3230,75 @@ TEST_F(TrajectoryControllerTest, decelerate_to_hold_position_velocity_command_ra
...
+  updateControllerAsync(rclcpp::Duration::from_seconds(0.8), t_parked);
+
+  // joint3 must be near FIRE-time value (≈4.55), not receive-time value (≈3.73).
+  if (traj_controller_->has_position_command_interface())
+  {
+    auto state = traj_controller_->get_state_reference();
+    EXPECT_NEAR(2.0, state.positions[0], 0.15);
+    EXPECT_NEAR(2.5, state.positions[1], 0.15);
+    EXPECT_NEAR(4.55, state.positions[2], 0.25);
+    EXPECT_GT(state.positions[2], 4.1);  // clearly above receive-time value ≈3.73
```
saikishor: Isn't this condition redundant?

saikishor (changes requested):
Some comments on tests

joint_trajectory_controller/test/test_trajectory_actions.cpp
vedh1234: That had a separate test testing it, but I'll add it here as well.

joint_trajectory_controller/test/test_trajectory_actions.cpp
vedh1234: Goal A already starts immediately (no header.stamp set). Goal B is the future-stamped one that gets deferred. So the behaviour is exactly as you described: A starts immediately, then when B is accepted, A gets preempted and ABORTED while B's trajectory is deferred until its stamp

vedh1234: The final ABORTED/SUCCEEDED check already proves the essential behaviour: A was preempted when B was accepted, and B successfully executed after its deferred start time. Happy to add intermediate checks in a follow-up if you feel it's important for the coverage.

joint_trajectory_controller/test/test_trajectory_actions.cpp
vedh1234: This is a pre-existing issue in JTC, topic_callback has never explicitly aborted the active action goal on topic preemption, even before this PR. The goal handle survives and eventually gets ABORTED maybe via tolerance violation when the robot follows the topic trajectory instead.

should we abort it when a topic traj comes on top of action goal ?
If Yes we can open a separate PR resolving this.

joint_trajectory_controller/test/test_trajectory_actions.cpp
vedh1234: goal_a creates the active trajectory that causes goal_b to be deferred. Without goal_a running, goal_b (future-stamped) would install immediately since has_active_trajectory() would be false.

joint_trajectory_controller/test/test_trajectory_actions.cpp
vedh1234: Okay will do that.

joint_trajectory_controller/test/test_trajectory_actions.cpp
vedh1234: Same reason: goal_a creates the active trajectory that causes goal_b to be deferred. Without goal_a running, goal_b (future-stamped) would install immediately since has_active_trajectory() would be false.

joint_trajectory_controller/test/test_trajectory_controller.cpp
vedh1234: Okay right, it was just to check if the joint moved beyond that point. But agreed, it's redundant, I'll remove the EXPECT_GT.

vedh1234:
@saikishor I have addressed your comments and also made requested changes. While going through I also realised that there is a helper API for sending goals. I've committed that as well.

saikishor (approved):
LGTM

bmagyar (approved):

abuibaid:
Hi everyone, just out of curiosity, will this PR be merged with other ROS_DISTROs  like humble or just for the master? Regards,

christophfroehlich:
I doubt so. But you can compile the whole ros2_control stack with [development versions](https://control.ros.org/humble/doc/getting_started/getting_started.html#building-from-source) on all active distros.

PR #2281, opened 2026-04-03
[RQT_JTC] add unit tests for parse_joint_limits

lakhmanisahil:
Closes #2253

## What this PR does

`get_joint_limits()` read from a global `description` variable set only by a ROS topic subscription — impossible to unit test without a live node.

This PR extracts all parsing logic into a new pure function `parse_joint_limits(urdf_string, ...)` that takes the URDF directly as a string with no ROS dependency. `get_joint_limits()` becomes a thin wrapper that reads the global and delegates to it. The minidom parsing logic itself is **unchanged**.

## Files changed

| File | Change |
|---|---|
| `joint_limits_urdf.py` | Extract `parse_joint_limits()` from `get_joint_limits()` |
| `test/test_joint_limits_urdf.py` | 23 pytest tests, fully ROS-free |
| `test/__init__.py` | New (marks test dir as Python package) |
| `package.xml` | Add `python3-pytest` test dependency |

## Tests (23 total, all passing)

- Revolute joint — correct min/max position, velocity, limit flag
- Continuous joint — defaults to ±π when no bounds given
- Fixed joints — silently ignored
- Multiple mixed joints — all handled correctly
- `safety_controller` soft limits — narrowed when flag is `True`, ignored when `False`
- Mimic joints — excluded from `free_joints`
- Missing `<limit>` for a required joint → raises with clear message
- Missing `lower`/`upper` on revolute → raises (minidom returns `""`, our code raises)
- Missing `velocity` → raises (same reason — our logic, not minidom's)
Screenshot from 2026-04-04 00-14-26

PR #2404, opened 2026-06-09
[RQT-JTC] Allow incomplete joint limits

christophfroehlich:
## Description

For faster review, I manually cherry-picked the changes regarding joint-limits from #1131 
With #2254 and #2281 some of the changes were applied already, and the rest was not possible to be directly merged due to conflicts.
Also, the urdf_parser_py does not allow a `<limit>` tag without velocity limits, so I removed the respective fallback section.

yes, but instead of crashing it should now gracefully try to do something useful and give warnings.

GPT Codex 5.3


christophfroehlich:
@lakhmanisahil I splitted the original PR, this one here is faster to review. As you have refactored the module recently, maybe you can have a quick look if that makes sense? Thank you!

lakhmanisahil:
> @lakhmanisahil I splitted the original PR, this one here is faster to review. As you have refactored the module recently, maybe you can have a quick look if that makes sense? Thank you!

Hello, I tested it on Lyrical: colcon test passes 28/28, including the 6 new cases, and the parser behaves as expected in both strict and tolerant modes. The new default allow_incomplete_joints=False preserves the existing behavior.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -113,6 +123,19 @@ def parse_joint_limits(urdf_string, joints_names, use_smallest_joint_limits=True
...
+                if logger is not None:
+                    logger.warn(
+                        f"Joint '{name}' has no <limit> tag in the robot_description. "
+                        "Slider will be displayed but disabled."
+                    )
+                free_joints[name] = {
+                    "min_position": -2 * pi,
+                    "max_position": 2 * pi,
+                    "has_position_limits": False,
+                    "max_velocity": 1.0,
```
MarqRazz, line 139: Are these good defaults for every system? 1 rad/sec seems pretty fast for an unknown actuator (the limits also seem large). What if we made all of the values small/slow to encourage users to set them to valid numbers for their system.

If the user can't command the joint maybe they are okay. If we are only displaying state readings in RQT then really restrictive values here don't sound like they would hurt anything and would make it more obvious that they need to update their URDF.

rqt_joint_trajectory_controller/test/test_joint_limits_urdf.py
```diff
@@ -23,6 +23,7 @@
 """
 
 import math
+
```
MarqRazz:
```suggestion
```

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
christophfroehlich: When has_position_limits is False, then max_velocity does not have a real meaning. and afaik the min/max_position is used to scale the slider even if it is disabled, so the range of 4pi might cover most of revolute and nearly every prismatic joint I guess.

MarqRazz: How does this differ from a continuous joint? Is it just the fact that the joint is disabled/grayed out in RQT when `has_position_limits=False`?

For example here is what the Kinova Gen3 looks like (and several of its joints are [continuous](https://github.com/Kinovarobotics/ros2_kortex/blob/main/kortex_description/arms/gen3/7dof/urdf/gen3_macro.xacro#L131) and RQT uses the values +- 3.14 for the limits).
image

I'm okay with leaving the defaults as is if we document what the values do (position limits scale the slider bar and velocity settings have no meaning).

christophfroehlich: updated

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -131,13 +157,32 @@ def parse_joint_limits(urdf_string, joints_names, use_smallest_joint_limits=True
             if joint.type == "continuous":
                 minval = -pi
```
christophfroehlich, line 158: @MarqRazz the position limits for continuous joints were not changed here. The parser raises an exception if no velocity limit is set, so I hope the position limit is a useful default here?

I realize now before this PR the continuous joints were treated like revolute ones, but with the following lines they will be deactivated:
https://github.com/ros-controls/ros2_controllers/pull/2404/changes#diff-1fe0d2fd8e18b20c3df84a8d9060e7661230e1f8aa0329f7d6f53aa8ce14dedaR291

I think we should change that and set has_position_limits true for continuous joints (the variable name does not fit well, but the behavior remains the same). 

What do you think?

MarqRazz: Ya we could set `has_position_limits = True` here. We are assigning limit values even though it doesn't have any and the RQT widget doesn't support jogging a joint beyond those values currently so we might as well be consistent.

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
```diff
@@ -113,6 +123,22 @@ def parse_joint_limits(urdf_string, joints_names, use_smallest_joint_limits=True
 
         # No <limit> element at all means urdf_parser_py sets joint.limit to None
         if joint.limit is None:
+            if name in joints_names and allow_incomplete_joints:
```
christophfroehlich, line 126: btw urdf_parser from robot_state_publisher will never run without a limit tag
```
[robot_state_publisher-2] Error:   Joint [joint1] is of type REVOLUTE but it does not specify limits
[robot_state_publisher-2]          at line 607 in ./urdf_parser/src/joint.cpp
[robot_state_publisher-2] Error:   joint xml is not initialized correctly
[robot_state_publisher-2]          at line 243 in ./urdf_parser/src/model.cpp
[robot_state_publisher-2] [WARN] [1781641700.051294975] [urdf]: Failed to parse robot description using: urdf_xml_parser/URDFXMLParser
[robot_state_publisher-2] 
[robot_state_publisher-2] terminate called after throwing an instance of 'std::runtime_error'
[robot_state_publisher-2]   what():  Unable to initialize urdf::model from robot description
```
but it seems that urdf_parser_py allows this, see `test_missing_limit_tag_for_unrequired_joint_skipped_silently`

rqt_joint_trajectory_controller/rqt_joint_trajectory_controller/joint_limits_urdf.py
christophfroehlich: done

christophfroehlich:
@MarqRazz can you please have another look here?

MarqRazz (approved):
Looks good. Thanks for adding a test for continuous joint that have position_limits = true.

PR #2422, opened 2026-06-17
Deliver abort action result before destroying goal handle on preemption

vedh1234:
## Description

preempt_active_goal() called `setCanceled()`, which only sets an in-memory flag, then immediately destroyed the `goal_handle_timer_` before `runNonRealtime()` could fire. The result message was never sent over the network, leaving the action client waiting forever. Fix: call `runNonRealtime()` explicitly before the timer is reset to guarantee result delivery. Also corrects the result code from CANCELED (client-requested) to ABORTED (server-decided preemption)

Yes. Action clients whose goal is preempted by a new incoming trajectory now receive an ABORTED result immediately. 

Yes, Cluade for the test case. 

## TODOs

- [x] Fork the repository.

### DIFF
--- a/joint_trajectory_controller/doc/userdoc.rst
+++ b/joint_trajectory_controller/doc/userdoc.rst
@@ -116,7 +116,7 @@ Preemption policy [#f1]_
 
 Only one action goal can be active at any moment, or none if the topic interface is used. Path and goal tolerances are checked only for the trajectory segments of the active goal.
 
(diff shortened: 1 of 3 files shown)


christophfroehlich (approved):
LGTM, thank you!

urfeex (commented):
that looks good :-)

saikishor (approved):

PR #2442, opened 2026-06-26
Enhance: Added joint limit loader plugin

### LINKED ISSUE #2434: Jerk Limiting in the Joint Limits and Integrated to JTC
We have issue where jerk is not limited on the trajectories executed on the hardware. This leads to the difficulties in some cases. Especially where JTC is used for trajectory interpolation it is not sufficient to have limiters towards the hardware, but also the limiter has to be integrated within the JTC itself to properly calculate the next interpolated position.

Few years ago we were working on this but never finish up. There are two main tasks to do:

- [ ] extend the [`SaturationLimiter` in joint_limits](https://github.com/ros-controls/ros2_control/tree/master/joint_limits) package to support the jerk limiting. For this is quite important to consider all 7 dynamic cases as the limited jerk propgates back to the commanded positions.
- [ ] extend JTC to support joint limiter plugins.

The code has to be exceptionally well tested. This is a perfect task for the Test driven development. Therefore, it is recommended to first write the tests with the evaluation and then start writing code that should satisfy those tests.

First part is straightforward with extending the limiter with accepting jerk limits from the `.yaml` file as done for other limits and add maths for it to it. Here is extremely indispensable to have clan, understandable code and concepts with literally every single line tested. This means that clear test cases with comprehensive documentation about their values should be added. The target code should come in `SturationLimiter` class. Also, when the current code is analysed the checkboxes in [this issue](https://github.com/ros-controls/ros2_control/issues/340) should be marked as finished.

The JTC integration of the limiters should be plugin based as done on the [`jtc-add-joint-limiter` on b»robotized fork](https://github.com/b-robotized-forks/ros2_controllers/tree/jtc-add-joint-limiter). This is the simple part, the complex part is to integrate this with the interpolation defined in JTC. The referenced branch also links to the different versions of interpolation adding new object for this, but now with time-distance, I am not convinced that should be done any more (see `interpolation_mkethods.hpp` file).

**Describe alternatives you've considered**
We could use external libraries for this like [Ruckig](https://github.com/destogl/ros2_control/tree/add-ruckig-joint-limiter) of Reflexxes, but they generally make things more complex that is required. Therefore, we should avoid doing that.

sachinkum0009:
## Description

This PR integrates the Joint Limiter using Plugin Loader to use Joint Saturation Limiter. 

> Summary
- Added Pluginlib to load the JointSaturationLimit plugin
- Added Tests to validate the plugin for Pos, Vel, Vel + Acc and Pos + Vel + Acc 

Fixes #2434 

Yes, with this PR, the JTC uses Joint Saturation Limiter to limit the pos, vel, acc, or jerk based on the configuration.

Used OpenCode to understand the code and write tests.

## TODOs

- [X] Fork the repository.
- [X] Send a pull request, answering any default questions in the pull request interface.
- [X] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.


sachinkum0009:
# Test Integrated

|Test Name|Implemented|
|---|---|
|Joint limiter failed to configure|✅|
|Joint limiter config successfully|✅|
|joint limiter failed to activate|✅|
|Joint limiter activate successfully|✅|
||✅|
|Limits Position|✅|
|Limits Velocity|✅|
|Limits Acceleration|✅|
|Limits Pos, Vel, Acc|✅|
|Short Sample Traj within limits|✅|
|Short Sample Traj on limits|✅|
|Short Sample Traj over limits|✅|
|Long Sample Traj within limits|✅|
|Long Sample Traj on limits|✅|
|Long Sample Traj over limits|✅|

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -374,6 +395,12 @@ controller_interface::return_type JointTrajectoryController::update(
 
     if (valid_point)
     {
+      // Apply joint limiter to the next commanded state
+      if (joint_limiter_)
+      {
+        joint_limiter_->enforce(state_current_, command_next_, update_period_);
```
destogl, line 377: why is here the right place to enforce limits? Shouldn't we do this in before and after interpolation?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -838,6 +865,11 @@ void JointTrajectoryController::query_state_service(
       RCLCPP_ERROR(
         logger, "Requested sample time is earlier than the current trajectory start time.");
     }
+    // Apply joint limiter to the requested state
+    if (joint_limiter_)
+    {
+      joint_limiter_->enforce(state_current_, state_requested, update_period_);
```
destogl, line 836: Why this second call in the query state service?

joint_trajectory_controller/test/test_joint_limiter_jtc.cpp
```diff
@@ -0,0 +1,585 @@
...
+  EXPECT_NEAR(cmd_next.positions[2], INITIAL_POS_JOINT3 + 0.5 * 0.01, COMMON_THRESHOLD);
+  EXPECT_NEAR(cmd_next.velocities[0], 0.5, COMMON_THRESHOLD);
+  EXPECT_NEAR(cmd_next.velocities[1], 0.5, COMMON_THRESHOLD);
+  EXPECT_NEAR(cmd_next.velocities[2], 0.5, COMMON_THRESHOLD);
+
+  // Hardware command interfaces also reflect the limited values
+  EXPECT_NEAR(joint_vel_[0], 0.5, COMMON_THRESHOLD);
+  EXPECT_NEAR(joint_vel_[1], 0.5, COMMON_THRESHOLD);
+  EXPECT_NEAR(joint_vel_[2], 0.5, COMMON_THRESHOLD);
+  EXPECT_NEAR(joint_pos_[0], INITIAL_POS_JOINT1 + 0.5 * 0.01, COMMON_THRESHOLD);
```
destogl: I am not sure whether this is correct, as you are testing the limiter. What happens with interpolation in this case? Are we breaking it?

destogl (commented):
Please comment. It seems a bit trivial implemenation. Explain why you think it is correct.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
sachinkum0009: I think the best place to enforce limits is after the interpolation. If appied limits before interpolation, it can produce inconsistent interpolation results.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
sachinkum0009: As the `~/query_state` service (control_msgs::srv::QueryTrajectoryState) allows querying the expected state at a future time. I think if we have joint limiter plugin enabled then we should return the response with limits applied.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
destogl: What happens than with the interpolation? Will we still have interpolated values?

sachinkum0009: Thanks. Yes, we still have the interpolated values, as the joint limiter is applied before publishing the next cmd to hardware interface.

joint_trajectory_controller/test/test_joint_limiter_jtc.cpp
sachinkum0009: Thanks, I have updated the tests to validate the interpolated values and make sure the position, velocity and acceleration remains within the limits.

sachinkum0009:
I have plotted few test cases for JTC with joint saturation limiter plugin.

### Trajector on the limits
trajectory_plot1

### Trajectory over the limits
trajectory_plot2

Pls LMK if there are some more suggestions or feedback. I would like to merge this ASAP. 🙏

PR #2445, opened 2026-06-30
fix: correct test_load_controller tests for motion_primitives and pid_controller

### LINKED ISSUE #1258: test_load_controller tests always return true

SouriRishik:
Fixes #1258

## Problem
The `test_load_controller` tests in `motion_primitives_controllers` and `pid_controller` were using `ASSERT_NO_THROW` to check that loading the controller does not throw an exception, but never verified that the controller actually loaded successfully. Additionally, neither test was setting a `params_file` or `type` parameter via `set_parameter`, so the controllers were not being validated against their expected configuration. This meant the tests would pass even if the controller failed to load and returned `nullptr`.

## Fix
Updated both tests to set the `params_file` and `type` parameters explicitly using `cm.set_parameter(...)`, then changed the assertion to `ASSERT_NE(cm.load_controller(...), nullptr)` so the test correctly verifies that the controller loads successfully with the provided configuration. This follows the same pattern already used in `admittance_controller` and `range_sensor_broadcaster`.

no

## Did you use Generative AI?
no

### DIFF
--- a/motion_primitives_controllers/CMakeLists.txt
+++ b/motion_primitives_controllers/CMakeLists.txt
@@ -84,6 +84,8 @@ if(BUILD_TESTING)
   find_package(hardware_interface REQUIRED)
   find_package(ros2_control_test_assets REQUIRED)
 
+  add_definitions(-DTEST_FILES_DIRECTORY="${CMAKE_CURRENT_SOURCE_DIR}/test")
+
   ament_add_gmock(
     test_load_motion_primitives_forward_controller
     test/test_load_motion_primitives_forward_controller.cpp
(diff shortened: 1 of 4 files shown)


christophfroehlich (changes requested):
Thanks, but please verify that your code builds before you open a PR.

`error: use of undeclared identifier 'TEST_FILES_DIRECTORY'`
this has to be set from the CMake file.

bmagyar (approved):
thank you!

PR #2409, opened 2026-06-12
docs: update dead control.ros.org master routes to rolling

Ishan1923:
Updates hardcoded control.ros.org/master documentation URLs to rolling in:
- ros2_controllers/doc/index.rst (16 controller documentation links)
- README.md (Rolling and Lyrical distro rows)

Split from #2398

PR #2460, opened 2026-07-05
docs: update dead control.ros.org master routes to rolling/lyrical in README

Ishan1923:
Split out from #2409 per review feedback.

Updates the hardcoded control.ros.org/master/... documentation links in the Build status table:

Rolling row:

Before: https://control.ros.org/master/doc/ros2_controllers/doc/controllers_index.html
After: https://control.ros.org/rolling/doc/ros2_controllers/doc/controllers_index.html

Lyrical row:

Before: https://control.ros.org/master/doc/ros2_controllers/doc/controllers_index.html
After: https://control.ros.org/lyrical/doc/ros2_controllers/doc/controllers_index.html

These links were pointing to a dead/incorrect master path; this aligns them with the correct versioned docs paths, consistent with how Kilted/Jazzy/Humble rows already reference their own distro paths.
Should be backported to other distro branches (kilted, jazzy, humble) as needed.


christophfroehlich (approved):
Thanks

PR #2471, opened 2026-07-13
Add option to timestamp joint_states from hardware measurement-time interfaces

JenniferBuehler:
## Description

The current state of the joint_state_broadcaster uses `now()` to stamp the joint states. Some hardware however provides encoder measurement timestamps which the user may want to use instead, to get the true measurement time. This PR adds an option to reach such measurement time from the state interfaces *optionally*: if the hardware interface doesn't declare and write it, it falls back to the current behavior, so this wouldn't break any existing use of the joint state broadcaster. 

> [!NOTE]
> Adding the measurement state interface to the hardware is a statement of the user that they want measurement time _and nothing else_. The fallback to controller time is at configuration time, by not adding the state interfaces. Once it's declared, it's expected - if no timestamp is set (reads an all-zero value), then no joint state is published. This is to avoid pitfalls of expecting measurement time and not noticing that the fallback became active.

It's an optional feature, while it's user facing (the user being the one writing the hardware interfaces), it's fully backwards compatible.

Claude Opus 4.8 was used to execute an implementation plan that was done by the human (myself) and then questioned **a lot** and iterated on. Finally it was edited, verified and tested by the human (also myself).

#### 1. Unit tests

New cases cover the feature (should pass in CI). Please see inline comments.

#### 2. Interactive test with mock hardware [optional]

If the unit tests are considered insufficient, here is a suggestion on how this can be tested interactively with `mock_components/GenericSystem` to expose the two measurement-time state interfaces with fixed values.

`jsb_timestamp_test.urdf`

```xml
<?xml version="1.0"?>                                                                               
<robot name="TimestampTestRobot">                                                                   
  <link name="base_link"/>                                                                          
  <joint name="joint1" type="revolute">                                                             
    <parent link="base_link"/><child link="link1"/>                                                 
    <axis xyz="0 0 1"/>                                                                             
    <limit effort="1" velocity="1" lower="-3.14" upper="3.14"/>                                     
  </joint>                                                                                          
  <link name="link1"/>                                                                              
                                                                                                    
  <ros2_control name="TimestampTestSystem" type="system">                                           
    <hardware>                                                                                      
      <plugin>mock_components/GenericSystem</plugin>                                                
    </hardware>                                                                                     
    <joint name="joint1">                                                                           
      <command_interface name="position"/>                                                          
      <state_interface name="position"><param name="initial_value">0.5</param></state_interface>    
      <state_interface name="velocity"/>                                                            
    </joint>                                                                                        
    <!-- Two constant state interfaces standing in for a hardware measurement clock -->             
    <sensor name="measurement_clock">                                                               
      <state_interface name="measurement_time_sec" data_type="uint32"><param name="initial_value">1234</param></state_interface>
      <state_interface name="measurement_time_nsec" data_type="uint32"><param name="initial_value">567000000</param></state_interface>
    </sensor>                                                                                       
  </ros2_control>                                                                                   
</robot>        
```

`jsb_timestamp_test.yaml`

```yaml
controller_manager:
  ros__parameters:
    update_rate: 100
    joint_state_broadcaster:
      type: joint_state_broadcaster/JointStateBroadcaster

joint_state_broadcaster:
  ros__parameters:
    timestamp_state_interfaces:
      sec: measurement_clock/measurement_time_sec
      nsec: measurement_clock/measurement_time_nsec
```

Run it (three terminals with the workspace sourced):

1) publish the URDF for robot_description
```bash
ros2 run robot_state_publisher robot_state_publisher \
    --ros-args -p robot_description:="$(cat src/ros2_controllers/joint_state_broadcaster/jsb_timestamp_test.urdf)"
```

2) start the controller manager with the mock hardware

```bash
ros2 run controller_manager ros2_control_node \
    --ros-args --params-file src/ros2_controllers/joint_state_broadcaster/jsb_timestamp_test.yaml
```

3) load + activate the broadcaster, then inspect the stamp
```bash
ros2 run controller_manager spawner joint_state_broadcaster \
    --param-file src/ros2_controllers/joint_state_broadcaster/jsb_timestamp_test.yaml
```
```bash
ros2 topic echo /joint_states --once
```

Expected: the message header shows the hardware-provided stamp:
```yaml
  header:
    stamp:
      sec: 1234
      nanosec: 567000000
```

Negative check (default behaviour preserved):
Remove the entire top-level `joint_state_broadcaster:` block, leaving only the `controller_manager:` section from the yaml:
```yaml
  controller_manager:
    ros__parameters:
      update_rate: 100
      joint_state_broadcaster:
        type: joint_state_broadcaster/JointStateBroadcaster
```
`header.stamp` now shows the controller-manager clock (a current time)

Absent check (no measurement time means nothing is published):
An all-zero stamp on the interfaces means the hardware has no measurement time.
In that case the joint state is not published at all, since no stamp would describe its data correctly: 
there is no fallback to the controller time and no reuse of an older measurement time.
It would not be a correct state and could slip through unnoticed or create bugs hard to find.
An absent joint state instead will be clearly noticed.

Keep the yaml as-is and set both `initial_value` entries in the URDF's
sensor block to zero (or skp the entire `initial-value` param):

```xml
    <sensor name="measurement_clock">
      <state_interface name="measurement_time_sec" data_type="uint32"><param name="initial_value">0</param></state_interface>
      <state_interface name="measurement_time_nsec" data_type="uint32"><param name="initial_value">0</param></state_interface>
    </sensor>
```

`ros2 topic echo /joint_states --once` now produces no output (interrupt it),
and the controller manager repeats about once a second:

```
State interfaces 'measurement_clock/measurement_time_sec' and 'measurement_clock/measurement_time_nsec' carry no valid measurement time. Not publishing joint states until they do.
```

### DIFF
--- a/joint_state_broadcaster/src/joint_state_broadcaster.cpp
+++ b/joint_state_broadcaster/src/joint_state_broadcaster.cpp
@@ -23,6 +23,9 @@
 
 #include "hardware_interface/types/hardware_interface_type_values.hpp"
 #include "rclcpp/qos.hpp"
+#include <cmath>
+#include <cstdint>
+
 #include "rclcpp/time.hpp"
 #include "std_msgs/msg/header.hpp"
 
@@ -94,6 +97,14 @@ controller_interface::InterfaceConfiguration JointStateBroadcaster::state_interf
       {
         state_interfaces_config.names.push_back(joint + "/" + interface);
       }
+    }
+    // Also claim the optional measurement-time interfaces (if configured). In ALL mode they are
+    // already included; here (INDIVIDUAL) they must be requested explicitly.
+    if (!params_.timestamp_state_interfaces.sec.empty() &&
+      !params_.timestamp_state_interfaces.nsec.empty())
+    {
+      state_interfaces_config.names.push_back(params_.timestamp_state_interfaces.sec);
+      state_interfaces_config.names.push_back(params_.timestamp_state_interfaces.nsec);
     }
   }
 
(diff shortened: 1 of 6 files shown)


joint_state_broadcaster/src/joint_state_broadcaster.cpp
JenniferBuehler, line 615: Info: This (and two lines below) is the reason that we use double as type. Adding the other types that are now supported is considered no part of this PR.

JenniferBuehler:
Update: Through testing and updating I noticed that the original reason to need `double` interfaces expired. So will now apply simplifications to the PR. Updates coming - and this staying in draft for now.

joint_state_broadcaster/src/joint_state_broadcaster.cpp
JenniferBuehler: Superflous after recent changes.

PR #2133, opened 2026-01-29
refactor(jtc): improve jtc tolerances

thedevmystic:
Hello respected maintainers and reviewers!
This is Surya!

This PR addresses the following:
  - Seperate declaration from definition for better compilation speed and readability.
  - Refactor functions for better readability.
  - Add new function (create_error_trajectory_point).
  - Add extensive doxygen-style comments.
  - Add tests for new function (create_error_trajectory_point).

---

1. If you have question, why I moved (or seperated) it to tolerances.cpp it is because the file was getting way too long for no reason. It would both increase compile time and future maintainance.

2. If you have question, why I added `create_error_trajectory_point` it is because we can calculate the current error point and then compare it with tolerances with ease!

---

**A note**: This is a continuation of the previous PR #2048 (closed now). Instead of doing new features (new tolerances option) and refactoring in one PR. I split them into 2 PRs one doing refactoring (current) and another future PR adding feature.

PR #2478, opened 2026-07-15
Link JTC validation library instead of header-only

thedevmystic:
Hello, respected maintainers and reviewers! This is Surya!

So, this PR concerns with loading `validate_jtc_parameters.hpp`as `.cpp`. If you may remember I did this similar PR previously but it's changes was reverted. This is important for my another PR which is #2133. Without this specific change we get mult. def. errors.

Changes in this PR:
- Seperated definitions of `validate_jtc_parameters.hpp` into `validate_jtc_parameters.cpp`.

A question: While I did a traditional `.hpp/.cpp` approach we can also just inline it to satisfy ODR. If we want that then I can rewrite it as inline.


thedevmystic:
Sorry for tagging reviewers but it has been a month without any activities or followups.

@aprotyas @christophfroehlich @destogl @mamueluth @Maverobot

PR #2488, opened 2026-07-26
feat: warn when ForwardCommandController receives commands while inactive

### LINKED ISSUE #1760: ForwardCommandController: It should at least warn when receives messages while not active
JTC for instance error log when it is not active: https://github.com/ros-controls/ros2_controllers/blob/1dce1cc7bb50b78bad258183e1d53b8a8bad4db8/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L652

I would like something equivalent for ForwardCommandController

I've tried some solutions, but not quite sure why this does not work
```c++
joints_command_subscriber_ = this->get_node()->create_subscription<
    std_msgs::msg::Float64MultiArray>(
    "commands", rclcpp::SystemDefaultsQoS(),
    [this](const forward_command_controller::CmdType::SharedPtr msg) {
      auto current_state = this->get_node()->get_current_state().id();
      if (current_state != lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE) {
        RCLCPP_WARN(this->get_node()->get_logger(),
          "Received command while controller is not active");
        return;
      }

      rt_command_ptr_.writeFromNonRT(msg);
    });
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Rudyy75:
## Description

This PR fixes #1760 where `ForwardCommandController` silently ignores commands when it is in an inactive lifecycle state. 
**Changes made:**
- Added a lifecycle state guard (`get_node()->get_current_state()`) to the subscription callback in `ForwardControllersBase`.
- It now logs a throttled warning (`RCLCPP_WARN_THROTTLE`) and drops the message if the controller is not in `PRIMARY_STATE_ACTIVE`.
- Since it's implemented in the base class, it fixes this behavior for both `ForwardCommandController` and `MultiInterfaceForwardCommandController`.

Fixes #1760 

Yes, controllers will now log a warning when commands are received while inactive, instead of silently ignoring them.

Yes, Claude Sonnet 4.6

## TODOs

- [x] Fork the repository.

### DIFF
--- a/forward_command_controller/src/forward_controllers_base.cpp
+++ b/forward_command_controller/src/forward_controllers_base.cpp
@@ -22,6 +22,7 @@
 
 #include "controller_interface/helpers.hpp"
 #include "hardware_interface/loaned_command_interface.hpp"
+#include "lifecycle_msgs/msg/state.hpp"
 #include "rclcpp/logging.hpp"
 #include "rclcpp/qos.hpp"
 
@@ -74,6 +75,14 @@ controller_interface::CallbackReturn ForwardControllersBase::on_configure(
     "~/commands", rclcpp::SystemDefaultsQoS(),
     [this](const CmdType::SharedPtr msg)
     {
+      if (get_node()->get_current_state().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+      {
+        RCLCPP_WARN_THROTTLE(
+          get_node()->get_logger(), *(get_node()->get_clock()), 1000,
+          "Can't accept new commands. Controller is not active.");
+        return;
+      }
+
       const auto cmd = *msg;
 
       if (!std::all_of(


forward_command_controller/src/forward_controllers_base.cpp
```diff
@@ -74,6 +75,14 @@ controller_interface::CallbackReturn ForwardControllersBase::on_configure(
     "~/commands", rclcpp::SystemDefaultsQoS(),
     [this](const CmdType::SharedPtr msg)
     {
+      if (get_node()->get_current_state().id() != lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
```
christophfroehlich, line 78: IMHO this is not a problem as this subscriber does not run in the RT loop, but maybe this is faster:
```suggestion
      if (get_lifecycle_id() != lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
```
For reference https://github.com/ros-controls/ros2_control/pull/2884

@saikishor what is best practice here?

fmauch: I guess, the `get_lifecycle_id()` version is simpler and therefore alone preferable.

fmauch (commented):
This seems like a low-hanging fruit to finish and merge. Only a decision on the function to call for getting the lifecycle state is missing.

fmauch (approved):
Looks good to me

christophfroehlich (approved):
LGTM

PR #2491, opened 2026-07-28
[JTC] Add positions_upsampling feature for positions-only trajectories

vedh1234:
## Description

Add `positions_upsampling` feature to `JointTrajectoryController` to upsample sparse, positions-only trajectory chunks into smooth global C2 cubic-spline trajectories.

### Background:

AI / learned policies (e.g. diffusion policy, ACT) emit **action chunks**: short `JointTrajectory` messages containing only positions at the policy's output rate (e.g. 10–50 Hz), with no velocities or accelerations. Fed positions-only, JTC interpolates linearly (C0) producing velocity discontinuities and jerk spikes at every knot. This is unsuitable for real hardware.

### Approach

The controller intercepts incoming `~/joint_trajectory` messages via `topic_callback` and, if `positions_upsampling.enable` is true and it receives positions-only chunks:

1. **Synthesises timing** from a configurable `positions_upsampling.policy_frequency` parameter when `time_from_start` is absent.
2. **Solves knot velocities** via `fill_cubic_spline_velocities()`, an O(n) tridiagonal Thomas solve with rest boundary conditions (v₀ = vₙ₋₁ = 0) and writes them into `points[i].velocities`.
3. **Passes through to JTC validation**, where the existing cubic-Hermite sampler reproduces the global C2 spline with no changes to the update loop or output path.

Chunks that already carry velocities pass through unchanged, meaning the feature is a strict superset of existing JTC behavior and operates as a safe no-op when disabled.

Yes, as an opt-in feature. Users can now enable `positions_upsampling.enable` (default: `false`) in their JTC configuration. When enabled, positions-only trajectory messages on the topic interface are automatically upsampled to smooth C2 cubic splines. The `positions_upsampling.policy_frequency` parameter (Hz) controls the synthesised timing for chunks that arrive without `time_from_start`. Existing trajectories with velocities are unaffected, and because it is off by default, there is no breaking change for existing users.

Yes, Google Gemini and Claude were used to make improvements in manually written code, architectural analysis, edge-case review. All generated code was reviewed, modified, and verified by the author.

### Additional Information

**Known limitations / future work:**
- Each chunk uses rest boundary conditions (v₀ = vₙ₋₁ = 0), so streaming chunks decelerate to a stop at each boundary. Cross-chunk C2 continuity feature is deferred and yet to be added. 
- Only the topic path (`~/joint_trajectory`) is upsampled. `FollowJointTrajectory` action goals bypass `topic_callback` and are not upsampled.

## TODOs

- [x] Fork the repository.

## Test Scenario: 
Two position only motion profiles were sent on the using the `~/joint_trajectory`. As we can see, with upsampling enabled and being C2 continuous, the staircase like discontinuity is not observed in velocity plot in both motion profiles. The sharp corners we observe are due to the lack of cross chunk continuity.  

<table>
    <tr>
      <td align="center"><b>Pick &amp; place</b></td>
      <td align="center"><b>Sine Wave</b></td>
    </tr>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </table>

<p align="center">
  pick_and_place
</p>


joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -92,7 +92,7 @@ struct SegmentTolerances
  * \param params The ROS Parameters
  * \return Trajectory segment tolerances.
  */
-SegmentTolerances get_segment_tolerances(rclcpp::Logger & jtc_logger, const Params & params)
+inline SegmentTolerances get_segment_tolerances(rclcpp::Logger & jtc_logger, const Params & params)
```
saikishor:
```suggestion
SegmentTolerances get_segment_tolerances(rclcpp::Logger & jtc_logger, const Params & params)
```

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -126,7 +126,7 @@ SegmentTolerances get_segment_tolerances(rclcpp::Logger & jtc_logger, const Para
   return tolerances;
 }
 
-double resolve_tolerance_source(const double default_value, const double goal_value)
+inline double resolve_tolerance_source(const double default_value, const double goal_value)
```
saikishor:
```suggestion
double resolve_tolerance_source(const double default_value, const double goal_value)
```

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -162,7 +162,7 @@ double resolve_tolerance_source(const double default_value, const double goal_va
  * \param joints The joints configured by ROS parameters
  * \return Trajectory segment tolerances.
  */
-SegmentTolerances get_segment_tolerances(
+inline SegmentTolerances get_segment_tolerances(
```
saikishor:
```suggestion
SegmentTolerances get_segment_tolerances(
```

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -211,7 +211,15 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   rclcpp::Duration action_monitor_period_ = rclcpp::Duration(50ms);
 
   // callback for topic interface
-  void topic_callback(const std::shared_ptr<trajectory_msgs::msg::JointTrajectory> msg);
+  virtual void topic_callback(const std::shared_ptr<trajectory_msgs::msg::JointTrajectory> msg);
```
saikishor:
```suggestion
  void topic_callback(const std::shared_ptr<trajectory_msgs::msg::JointTrajectory> msg);
```

joint_trajectory_controller/include/joint_trajectory_controller/validate_jtc_parameters.hpp
```diff
@@ -24,7 +24,7 @@
 
 namespace joint_trajectory_controller
 {
-tl::expected<void, std::string> command_interface_type_combinations(
+inline tl::expected<void, std::string> command_interface_type_combinations(
```
saikishor:
```suggestion
tl::expected<void, std::string> command_interface_type_combinations(
```

joint_trajectory_controller/include/joint_trajectory_controller/validate_jtc_parameters.hpp
```diff
@@ -68,7 +68,7 @@ tl::expected<void, std::string> command_interface_type_combinations(
   return {};
 }
 
-tl::expected<void, std::string> state_interface_type_combinations(
+inline tl::expected<void, std::string> state_interface_type_combinations(
```
saikishor:
```suggestion
tl::expected<void, std::string> state_interface_type_combinations(
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -1039,6 +1039,24 @@ controller_interface::CallbackReturn JointTrajectoryController::on_configure(
...
+    {
+      RCLCPP_WARN(logger, "spline_upsampling has no effect when interpolation_method is 'none'.");
+    }
+    if (params_.spline_upsampling.policy_frequency == 0.0)
+    {
+      RCLCPP_WARN(
+        logger,
+        "spline_upsampling.policy_frequency is 0: chunks without their own timing will be "
+        "rejected.");
+    }
```
saikishor: Use WARN_EXPRESSION instead

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
```diff
@@ -130,6 +130,27 @@ joint_trajectory_controller:
       one_of<>: [["splines", "none"]],
     }
   }
+  spline_upsampling:
```
saikishor: Did we say to comeup with a different name to this?

joint_trajectory_controller/src/trajectory.cpp
```diff
@@ -98,6 +98,96 @@ void wraparound_joint(
...
+      const double duration_prev = times[i] - times[i - 1];
+      const double duration_next = times[i + 1] - times[i];
+      const double pos_prev = traj.points[i - 1].positions[joint];
+      const double pos_curr = traj.points[i].positions[joint];
+      const double pos_next = traj.points[i + 1].positions[joint];
+
+      lower_diagonal[i] = 1.0 / duration_prev;
+      diagonal[i] = 2.0 * (1.0 / duration_prev + 1.0 / duration_next);
+      upper_diagonal[i] = 1.0 / duration_next;
+      rhs[i] = 3.0 * ((pos_next - pos_curr) / (duration_next * duration_next) +
```
saikishor: Assign to variables if possible the 2.0 and 3.0

saikishor (changes requested):
Rest LGTM

joint_trajectory_controller/src/joint_trajectory_controller.cpp
vedh1234: Okay works

joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml
vedh1234: Yes I will update it here and in example 19 both.

joint_trajectory_controller/src/trajectory.cpp
vedh1234: will add variables.

saikishor (dismissed):
LGTM

Thanks for the changes

saikishor (dismissed):

doc/release_notes.rst
christophfroehlich, line 1: I tried to fix the merge conflicts, but the links are not referring to this PR. Please check

vedh1234: I'll look into it and update

vedh1234: I have updated it + added a small fix on waypoint timing index and blend logic.

saikishor:
@vedh1234 do you have some plots for this one step ahead blending commit?. Just curious if it changes any behaviour.

vedh1234:
Yes It changes the behavior: I looked at some policies and docs, they predict future actions based on current observations. If we stamp their first waypoint with t=0, we force JTC to execute a future target instantly. Because there is zero time between "now" and the target, the controller's lead-in logic which is part of the blend/bridge never fires. So by shifting the first waypoint to t=dt (one step ahead), we give the controller exactly one action period to absorb that mismatch. It also stops the bridge from being dropped, which is why the blending tests only pass with this included. Without that smoothing window, any slight mismatch between the outgoing chunk and the incoming chunk positioncauses a violent, single-cycle jump. 

timing_without_continuity

In above plot I added a position mismatch at chunk arrival, meaning first waypoint doesn't match the current state and because it's t=0 you can see the spike which is not there in the t=dt case. 

Also now the small oscillation that you observe in the start of every chunk because of V0 being 0, is now solved with the cross chunk continuity with this [PR](https://github.com/vedh1234/ros2_controllers/pull/4)

saikishor:
@vedh1234 Good job with the plots

saikishor (dismissed):
LGTM

joint_trajectory_controller/doc/trajectory.rst
christophfroehlich, line 1: does it make sense to update the plots here? They come from a rather old script in https://github.com/christophfroehlich/test_jtc_trajectories

christophfroehlich (commented):
Changes LGTM, but maybe the docs could be improved and/or yours added for the new behavior.

joint_trajectory_controller/doc/trajectory.rst
vedh1234: Okay I'll add a few separate plot on what happens with this feature enabled, but should we update all the plots ? because the old once are still applicable with the feature disabled. Rather I was thinking if we should add similar plot to old one to maintain the pattern. What do you think ?

christophfroehlich: at least the new feature should be captured here.
In times of LLMs someone could refurbish all the plots in the end, this could be a new good-second-issue. Maybe you can create one, and add info/data how you generated your plot

vedh1234: Okay agreed, that would be better. 
What I'll do is add a plot on this new feature in this PR itself . Then I'll create a separate issue to cover and update plots of both blending and spline, also to update docs accordingly with reference to new plots.

vedh1234: I have updated the plot, thanks to your scripts. I used same script as yours to keep the same plotting style, just updated the call for upsampling. Also updated the docs.

vedh1234: unrelated, just curious if we should be creating the new plots using a probe by adding a temporary test that stores the data in csv and then generate the plot (this is how I generated earlier plots). In your scripts you are directly calling underlying functions in the `trajectory.cpp`,  but to be sure how the actual traj would be executed can only be confirmed by running the update loop because in between there are some parameters like `interpolate_from_desired_state` or now the blending logic, effect of which otherwise doesn't appear in plots. what do you think ? If you agree I'll mention this approach in the new issue.

christophfroehlich: sure, add this to the issue. this is then an implementation detail I'd say

christophfroehlich (dismissed):
Thanks a lot!

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -214,6 +214,14 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   // callback for topic interface
   void topic_callback(const std::shared_ptr<trajectory_msgs::msg::JointTrajectory> msg);
 
+  // Non-RT hook run on every incoming trajectory before validation. Currently upsamples
+  // positions-only chunks to a C2 spline when positions_upsampling is enabled; no-op otherwise.
```
bmagyar:
```suggestion
  // Non-RT hook run on every incoming trajectory before validation.
```
removing implementation details

bmagyar (approved):
thank you!

PR #2503, opened 2026-07-29
Allow disabling diff drive command timeout

### LINKED ISSUE #440: Disable cmd_vel_timeout
Hey !

Is there any way to disable the cmd_vel_timeout for the diff drive controller? It is not very apparent how to do this in the documentation. I have also checked the source code as well and it does not seem possible in its current implementation. 

I am happy to create a PR if needed.

See this https://github.com/cyberbotics/webots_ros2/issues/468 for more context.

dlanov:
## Summary

Allow the diff drive controller command timeout to be disabled by setting `cmd_vel_timeout` to `0.0`.

Previously, every command with a positive age was considered stale when the timeout was zero, causing the controller to replace valid velocity commands with zero.

## Changes

- Skip stale-command braking when `cmd_vel_timeout` is `0.0`.
- Preserve the existing timeout behavior for all positive timeout values.
- Keep the NaN warning throttle finite when the command timeout is disabled.
- Document that `0.0` disables the timeout.
- Add a deterministic regression test using an explicitly old command timestamp.
- Preserve the existing runtime parameter-update behavior.

## Testing

- `colcon build --symlink-install --packages-select diff_drive_controller --cmake-args -DBUILD_TESTING=ON`
- `colcon test --packages-select diff_drive_controller --event-handlers console_cohesion+`
- `colcon test-result --test-result-base build/diff_drive_controller --verbose`
- `pre-commit run --files diff_drive_controller/src/diff_drive_controller.cpp diff_drive_controller/src/diff_drive_controller_parameter.yaml diff_drive_controller/test/test_diff_drive_controller.cpp`
- `git diff --check`

Results:

- 42 tests
- 0 errors
- 0 failures
- 0 skipped
- All applicable pre-commit checks passed

Closes #440

## AI assistance

Claude Code assisted with analyzing the issue and preparing the narrowly scoped implementation and regression test. I manually reviewed the complete diff and verified the build, tests, formatting, lint, and whitespace-check results.

### DIFF
--- a/diff_drive_controller/src/diff_drive_controller.cpp
+++ b/diff_drive_controller/src/diff_drive_controller.cpp
@@ -118,12 +118,19 @@ controller_interface::return_type DiffDriveController::update_reference_from_sub
     command_msg_ = current_ref_op.value();
   }
 
+  constexpr rcutils_duration_value_t warning_throttle_ms = 1000;
+
   const auto age_of_last_command = time - command_msg_.header.stamp;
-  // Brake if cmd_vel has timeout, override the stored command
-  if (age_of_last_command > cmd_vel_timeout_)
+  const bool cmd_vel_timeout_disabled = cmd_vel_timeout_ == rclcpp::Duration::from_seconds(0.0);
+  // Brake if cmd_vel has timeout, override the stored command.
+  // A cmd_vel_timeout of 0.0 disables the timeout.
+  if (!cmd_vel_timeout_disabled && age_of_last_command > cmd_vel_timeout_)
   {
     ordered_exported_reference_interfaces_[0]->set_value(0.0);
     ordered_exported_reference_interfaces_[1]->set_value(0.0);
+    RCLCPP_WARN_THROTTLE(
+      logger, *get_node()->get_clock(), warning_throttle_ms,
+      "Velocity command timed out. Braking.");
   }
   else if (
     std::isfinite(command_msg_.twist.linear.x) && std::isfinite(command_msg_.twist.angular.z))
@@ -134,8 +141,7 @@ controller_interface::return_type DiffDriveController::update_reference_from_sub
   else
   {
     RCLCPP_WARN_SKIPFIRST_THROTTLE(
-      logger, *get_node()->get_clock(),
-      static_cast<rcutils_duration_value_t>(cmd_vel_timeout_.seconds() * 1000),
+      logger, *get_node()->get_clock(), warning_throttle_ms,
       "Command message contains NaNs. Not updating reference interfaces.");
   }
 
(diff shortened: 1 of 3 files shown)


diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -119,8 +119,10 @@ controller_interface::return_type DiffDriveController::update_reference_from_sub
   }
 
   const auto age_of_last_command = time - command_msg_.header.stamp;
-  // Brake if cmd_vel has timeout, override the stored command
-  if (age_of_last_command > cmd_vel_timeout_)
+  const bool cmd_vel_timeout_disabled = cmd_vel_timeout_ == rclcpp::Duration::from_seconds(0.0);
+  // Brake if cmd_vel has timeout, override the stored command.
+  // A cmd_vel_timeout of 0.0 disables the timeout.
+  if (!cmd_vel_timeout_disabled && age_of_last_command > cmd_vel_timeout_)
```
christophfroehlich, line 127: should we add a warning here if the timeout is hit?

diff_drive_controller/src/diff_drive_controller.cpp
```diff
@@ -133,9 +135,13 @@ controller_interface::return_type DiffDriveController::update_reference_from_sub
   }
   else
   {
+    // the timeout doubles as the warning throttle period; keep it finite when disabled
```
christophfroehlich: what do you mean with _doubles_? the throttle is just not effective if cmd_vel_timeout_ is zero, right?

diff_drive_controller/src/diff_drive_controller.cpp
dlanov: Good point. I've just added a throttled warning when the velocity command times out and the controller applies braking. It uses a fixed one-second throttle to avoid logging on every controller update while the command remains expired.

diff_drive_controller/src/diff_drive_controller.cpp
dlanov: You’re right—the wording was unclear. The timeout value was being reused as the NaN-warning throttle period, which makes the throttle ineffective when the timeout is zero. I removed that coupling and now use a fixed one-second throttle for both the timeout and NaN warnings.

christophfroehlich (commented):
please never do force pushes on already reviewed PRs, this will break github UI and just needs more effort for re-reviewing.

dlanov:
Understood—sorry about that. I amended the commit to keep the PR history compact, but I see how that disrupted the existing review. I won’t force-push reviewed PRs again; I’ll add follow-up commits and push them normally.

christophfroehlich (approved):
LGTM, thank you!

bmagyar (approved):
please use the PR template next time though, using a bot is not an excuse

PR #2514, opened 2026-08-01
Fix admittance position state updates

### LINKED ISSUE #594: admittance_position not updating at runtime for admittance controller
**Describe the bug**
For the admittance controller, the `admittance_state_.admittance_position` is never updated. This doesn't affect the behavior of the controller, but would nice to be able to visualize for debugging (it currently incorrectly shows 0s and nans).

**To Reproduce**
1. Run an instance that includes an admittance controller and activate it
2. run `ros2 topic echo /admittance_controller/status --field admittance_position`
3. admittance position values show uninitialized values for position, and nans for rotation

**Expected behavior**
`admittance_position` should be updated with the most recent position values.  I believe the velocity and acceleration values are updated [here](https://github.com/ros-controls/ros2_controllers/blob/humble/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp#L293) and it would make sense to add that here as well.

**Screenshots**
image

dlanov:
## Description

Fixes #594.

The admittance controller publishes `admittance_position` from its internal state, but the transform was neither initialized nor updated during the admittance calculation. This could result in invalid or stale values in `AdmittanceControllerState`.

This change:

* Initializes `admittance_position` to the identity transform.
* Updates its translation using the Cartesian offset already calculated as the desired position minus the reference position.
* Updates its rotation using `R_desired * R_reference.transpose()`.
* Adds tests for identity initialization and a nonidentity relative transform.

YES. `AdmittanceControllerState.admittance_position` now contains a valid, updated Cartesian admittance offset instead of an uninitialized or stale transform.

YES, for implementation assistance and test design.

### Additional Information

The regression test uses different nonidentity reference and desired transforms so that it distinguishes the relative transform from the absolute desired transform.

## Testing

```text
pre-commit run --files \
  admittance_controller/include/admittance_controller/admittance_rule.hpp \
  admittance_controller/include/admittance_controller/admittance_rule_impl.hpp \
  admittance_controller/test/test_admittance_rule_impl.cpp

colcon build --packages-select admittance_controller --symlink-install

colcon test --packages-select admittance_controller

Summary: 47 tests, 0 errors, 0 failures, 0 skipped
```

## TODOs

* [x] Fork the repository.
* [x] Keep the change focused on the reported issue.
* [x] Ensure local tests pass.
* [x] Run pre-commit checks.
* [x] Commit with a clear message and DCO sign-off.
* [x] Remain available to address CI or review feedback.

### DIFF
--- a/admittance_controller/include/admittance_controller/admittance_rule.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule.hpp
@@ -38,6 +38,7 @@ struct AdmittanceState
   {
     admittance_velocity.setZero();
     admittance_acceleration.setZero();
+    admittance_position.setIdentity();
     damping.setZero();
     mass.setOnes();
     mass_inv.setZero();
--- a/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
+++ b/admittance_controller/include/admittance_controller/admittance_rule_impl.hpp
@@ -255,6 +255,11 @@ bool AdmittanceRule::calculate_admittance_rule(AdmittanceState & admittance_stat
   auto R_ref = admittance_state.ref_trans_base_ft.rotation();
   auto R_desired = desired_trans_base_ft.rotation();
   auto R = R_desired * R_ref.transpose();
+
+  admittance_state.admittance_position.setIdentity();
+  admittance_state.admittance_position.translation() = X.head<3>();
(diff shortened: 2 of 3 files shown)


christophfroehlich (approved):
LGTM, thank you!

saikishor (approved):

PR #2523, opened 2026-08-03
fix: prevent JTC segfault when continuous joint has no URDF limits (#2480)

### LINKED ISSUE #2480: JointTrajectoryController::on_init() segfaults on a continuous joint with no   <limit> tag
**Describe the bug**
  `JointTrajectoryController::on_init()` segfaults (SIGSEGV) during controller
  loading.  This is valid URDF (continuous joints aren't required to declare limits), but the controller
  doesn't guard against it.

  In `joint_trajectory_controller.cpp`, `on_init()` checks that the joint exists
  in the parsed URDF model, but not that `limits` is non-null before dereferencing it:

  https://github.com/ros-controls/ros2_controllers/blob/jazzy/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L95-L99
  
  ```cpp
  auto urdf_joint = model.getJoint(params_.joints[i]);
  if (urdf_joint)
  {
    max_joint_vel[i] = urdf_joint->limits->velocity;  // <-- urdf_joint->limits
  can be null
  }

  For a continuous joint with no <limit> tag, urdf_joint is non-null but urdf_joint->limits is null, so this dereferences a null pointer.


**To Reproduce**
Steps to reproduce the behavior:
  1. Define a URDF joint as type="continuous" with no <limit> child element.
  2. Configure a JointTrajectoryController that includes that joint.
  3. Launch ros2_control_node and load the controller.
  4. ros2_control_node crashes with SIGSEGV inside JointTrajectoryController::on_init().


**Expected behavior**
 Expected the node to not crash. 

**Screenshots**
If applicable, add screenshots to help explain your problem.

 - ROS 2 Jazzy, Ubuntu 24.04
  - ros-jazzy-joint-trajectory-controller 4.40.1 (apt), also reproduced building from the jazzy branch HEAD from source — not specific to the packaged binary.

**Suggested fix**

  auto urdf_joint = model.getJoint(params_.joints[i]);
  if (urdf_joint && urdf_joint->limits)
  {
    max_joint_vel[i] = urdf_joint->limits->velocity;
  }

  I've applied and tested this locally (rebuilt joint_trajectory_controller from source with
  this one-line change) and confirmed it resolves the crash — the controller now loads and activates normally against the same URDF.


**Additional context**

trippyogi:
﻿## Summary
- Guard null `urdf_joint->limits` in `JointTrajectoryController::on_init()` so continuous joints without a URDF `<limit>` tag no longer segfault.
- Leave velocity at the existing 0.0 default and warn; do not invent limits.
- Add continuous-no-limits URDF fixture + regression test.

Fixes #2480

## Test plan
- [ ] `colcon test --packages-select joint_trajectory_controller`
- [ ] Confirm `continuous_joint_without_urdf_limits` passes

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -100,7 +100,9 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
         }
         else if (urdf_joint)
         {
-          RCLCPP_WARN(
+          // Continuous joints may omit <limit>; velocity is only needed for
+          // decelerate_on_cancel (warned later if that feature cannot run).
+          RCLCPP_DEBUG(
             get_node()->get_logger(),
             "Joint '%s' has no <limit> in the URDF; velocity limit unavailable (using 0.0).",
             params_.joints[i].c_str());
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -93,9 +93,17 @@ controller_interface::CallbackReturn JointTrajectoryController::on_init()
       for (size_t i = 0; i < params_.joints.size(); ++i)
       {
         auto urdf_joint = model.getJoint(params_.joints[i]);
-        if (urdf_joint)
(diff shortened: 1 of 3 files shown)


christophfroehlich (approved):
LGTM, thanks for the fix!

saikishor (approved):

PR #2484, opened 2026-07-23
fix: migrate active goal and tolerances to thread safe box in JTC

SouriRishik:
## Description

This PR migrates the JointTrajectoryController (JTC) to the modern RealtimeThreadSafeBox API to replace the deprecated RealtimeBuffer for the active goal handle and tolerances. 

The migration from stale PR #1480 has been completed and brought up to date with the following fixes:

1. Fixed a bug in all five `try_set()` calls where the lambda parameter was passed by value (`[](auto goal)`). This caused the compiler to copy the smart pointer, resulting in the active goal handle never being cleared from the box and causing subsequent actions/tests to fail or timeout.
2. Resolved the merge conflict in `preempt_active_goal()` by merging the new box `.set()` API with the master branch's preemption abort-result delivery logic (`active_goal->runNonRealtime()`).
3. Added a clean target fallback check in `CMakeLists.txt` to define the `ros2_control_test_assets::ros2_control_test_assets` target if it is not exported, ensuring compatibility with older ROS environments (such as Humble).

This finishes the stale PR #1480 and resolves the test failures.

Finally resolves #168 

## Is this user-facing behavior change?

No

## Did you use Generative AI?

No

## Additional Information

Verified locally that the pre-commit checks run and pass successfully:

```bash
pre-commit run --files joint_trajectory_controller/src/joint_trajectory_controller.cpp joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp joint_trajectory_controller/test/test_trajectory_controller_utils.hpp

PR #2564, opened 2026-08-14
[JTC] Add thread-safe snapshot of last commanded state

vedh1234:
## Description

Adds `rt_last_commanded_state_`, a `RealtimeThreadSafeBox` copy of `last_commanded_state_` that can be read safely from outside the update loop. `update()` publishes to it with `try_set`, so the control loop never blocks. `last_commanded_state_` itself is unchanged and stays the update loop's working copy, so no existing line is modified. This mirrors how other controllers pair a plain member with a box (`diff_drive`, `admittance`, `pid`), only in the opposite direction: here the update loop writes and other threads read.

Needed by controllers that build a trajectory outside the update loop and want to continue from the last commanded state instead of restarting from the measured one.

No. Nothing reads the new member yet, and no existing behaviour changes.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
@@ -41,6 +41,7 @@
 #include "realtime_tools/realtime_buffer.hpp"
 #include "realtime_tools/realtime_publisher.hpp"
 #include "realtime_tools/realtime_server_goal_handle.hpp"
+#include "realtime_tools/realtime_thread_safe_box.hpp"
 #include "trajectory_msgs/msg/joint_trajectory.hpp"
 #include "trajectory_msgs/msg/joint_trajectory_point.hpp"
 
@@ -113,6 +114,9 @@ protected:
   // variables for storing internal data for open-loop control
   trajectory_msgs::msg::JointTrajectoryPoint last_commanded_state_;
   rclcpp::Time last_commanded_time_;
+  // Thread-safe snapshot of last_commanded_state_ for non-RT readers.
+  realtime_tools::RealtimeThreadSafeBox<trajectory_msgs::msg::JointTrajectoryPoint>
+    rt_last_commanded_state_;
   /// Specify interpolation method. Default to splines.
   interpolation_methods::InterpolationMethod interpolation_method_{
     interpolation_methods::DEFAULT_INTERPOLATION};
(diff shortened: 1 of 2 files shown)


saikishor (commented):
@vedh1234 should we remove the non rt one and replace all instances with this one?

What do you think? @christophfroehlich @bmagyar

christophfroehlich:
> @vedh1234 should we remove the non rt one and replace all instances with this one?
> 
> What do you think? @christophfroehlich @bmagyar

If it is used only in update(), why to replace them with the RT Box?

christophfroehlich (dismissed):
LGTM

joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller.hpp
```diff
@@ -113,6 +114,9 @@ class JointTrajectoryController : public controller_interface::ControllerInterfa
   // variables for storing internal data for open-loop control
   trajectory_msgs::msg::JointTrajectoryPoint last_commanded_state_;
   rclcpp::Time last_commanded_time_;
+  // Thread-safe snapshot of last_commanded_state_ for non-RT readers.
```
bmagyar:
```suggestion
```
the type + variable name states this already

bmagyar (approved):

vedh1234:
> > @vedh1234 should we remove the non rt one and replace all instances with this one?
> > What do you think? @christophfroehlich @bmagyar
> 
> If it is used only in update(), why to replace them with the RT Box?

I am adding this for the cross chunk continuity, as in non-RT I need to access the `last_commanded_state_` to get the current state (which I will be adding in separate PR). 
Also wanted to ask in this [PR#1721](https://github.com/ros-controls/ros2_controllers/pull/1721), you moved it to RT box for all the other controllers replacing the buffers except for JTC was there any specific reason that we did not choose to move to RT box for JTC?

christophfroehlich:
> > > @vedh1234 should we remove the non rt one and replace all instances with this one?
> > > What do you think? @christophfroehlich @bmagyar
> > 
> > 
> > If it is used only in update(), why to replace them with the RT Box?
> 
> I am adding this for the cross chunk continuity, as in non-RT I need to access the `last_commanded_state_` to get the current state (which I will be adding in separate PR). Also wanted to ask in this [PR#1721](https://github.com/ros-controls/ros2_controllers/pull/1721), you moved it to RT box for all the other controllers replacing the buffers except for JTC was there any specific reason that we did not choose to move to RT box for JTC?

Because I had failing tests which I was not able to solve back then. Should be implemented now with 
https://github.com/ros-controls/ros2_controllers/pull/2484, could you please leave a review there?

vedh1234:
> Because I had failing tests which I was not able to solve back then. Should be implemented now with #2484, could you please leave a review there?

Okay sure, I'll review it. 
Thanks!

saikishor (approved):

bmagyar:
I'd leave the backports out for now ;)

PR #2567, opened 2026-08-17
Add closed-form steering integration tests

Dylan-Gallagher:
## Summary

- add closed-form pose expectations for the straight Runge-Kutta path
- add closed-form pose expectations for the nonzero-curvature exact path
- run the same assertions for bicycle, tricycle, and Ackermann configurations
- move the former Ackermann-only open-loop straight/left/right cases into the
  fixture, retaining their inputs while strengthening their pose assertions
- add nonzero-pose cases so heading-frame mistakes are observable

This is the first small piece of #930, split from the broader test proposal as
requested in the issue. Feedback-overload and state/error-path tests are left
for separate follow-up PRs.

## Validation

- full `steering_controllers_library` suite: 32/32 tests pass on ROS Rolling
- repository pre-commit suite passes for the changed file, including
  clang-format, cppcheck, cpplint, copyright, and codespell
- a temporary mutation halving the exact-turn radius made all three curved
  parameter rows fail; production source was then restored byte-for-byte
- test-only diff; no production or public API change

Part of #930.

Generated-by: OpenAI Codex (GPT-5; accessed 2026-08-17)

### DIFF
--- a/steering_controllers_library/test/test_steering_odometry.cpp
+++ b/steering_controllers_library/test/test_steering_odometry.cpp
@@ -57,11 +57,13 @@ Pose integrate_twist(
     final_heading};
 }
 
-void expect_pose(const steering_kinematics::SteeringKinematics & odom, const Pose & expected)
+auto PoseNear(const Pose & expected)
 {
-  EXPECT_NEAR(odom.get_x(), expected.x, TOLERANCE);
-  EXPECT_NEAR(odom.get_y(), expected.y, TOLERANCE);
-  EXPECT_NEAR(odom.get_heading(), expected.heading, TOLERANCE);
+  return ::testing::AllOf(
+    ::testing::Field("x", &Pose::x, ::testing::DoubleNear(expected.x, TOLERANCE)),
+    ::testing::Field("y", &Pose::y, ::testing::DoubleNear(expected.y, TOLERANCE)),
+    ::testing::Field(
+      "heading", &Pose::heading, ::testing::DoubleNear(expected.heading, TOLERANCE)));
 }
 
 class SteeringOdometryIntegratorTest : public ::testing::TestWithParam<SteeringConfiguration>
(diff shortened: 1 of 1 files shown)


steering_controllers_library/test/test_steering_odometry.cpp
```diff
@@ -16,9 +17,111 @@
...
+
+  const double final_heading = initial.heading + angular * dt;
+  const double turning_radius = linear / angular;
+  return {
+    initial.x + turning_radius * (std::sin(final_heading) - std::sin(initial.heading)),
+    initial.y - turning_radius * (std::cos(final_heading) - std::cos(initial.heading)),
+    final_heading};
+}
+
+void expect_pose(const steering_kinematics::SteeringKinematics & odom, const Pose & expected)
```
christophfroehlich: maybe this can be refactored as a custom matcher, otherwise it is hard to find the failing part in the test file.

christophfroehlich (changes requested):
Thank you. Can some of the existing tests already be handled by the new fixture instead?

Dylan-Gallagher:
Yes — I moved the existing Ackermann-only open-loop straight, left-turn, and right-turn cases into the parameterized fixture in 077bb7d. Their original inputs are retained, their sign-only turn checks are strengthened with the independent closed-form pose oracle, and the same cases now run for Bicycle and Tricycle too. The full package suite remains green locally (32/32), along with the changed-file pre-commit checks.

Dylan-Gallagher:
CI context for review: the changed steering test and the steering controller suites pass in the hosted Jazzy job. That job is red only because the unrelated `force_torque_sensor_broadcaster/test_wrench_transformer` did not generate its result file. The two deprecation jobs fail while building upstream `ros2_control/hardware_interface` (`test_handle.cpp` deprecated constructors), before this repository's changed test file is involved. The remaining hosted matrix and Codecov checks pass, so I have not added unrelated CI workarounds to this one-file PR.

steering_controllers_library/test/test_steering_odometry.cpp
Dylan-Gallagher: Addressed in ca3b643f: the pose assertion now uses a composed field matcher, so failures identify the exact `x`, `y`, or `heading` component and show its expected tolerance. The full 32-test package suite and all applicable changed-file pre-commit hooks pass.

christophfroehlich (approved):
Thank you!

PR #2575, opened 2026-08-24
Add option of using underlying forward and inverse kinematics in motion primitive forward controller

URJala:
## Description
This PR introduces a new parameter in the `motion_primitives_forward_controller` called `has_kinematics`. This parameter can be used if the underlying hardware interface is capable of doing both forward and inverse kinematics, where it will enable the controller to transfer joint angles with `LINEAR_CARTESIAN` commands or poses with `LINEAR_JOINT` commands. It is then up to the hardware interface to compute the kinematics and use the results for movement. 
To facilitate this change the `motion_primitives_base_controller`'s `set_command_interfaces` has been changed to also reset all unused command interfaces, such that the hardware interface can infer whether to use joint angles or poses with a motion primitive command depending on what is available.
**I would also like to enable `CIRCULAR_CARTESIAN` commands to be able to use joint angles if the parameter is set, but that would require the command message to use something other than a raw vector of doubles, and I would like some guidance on what that might look like, perhaps a vector of vector6D or `JointCommand`s?**
The motivation for this PR is that both the UR and KUKA robots support this behavior (and I would imagine many other manufacturers as well)

Fixes # (issue)

Yes, but default behavior is unchanged. The default value of the parameter is `false`, which follows the same accept or reject rules as the current implementation.

Yes, tests were written with AI: Github Copilot, GPT-5.5

## TODOs

- [x ] Fork the repository.
- [ x] Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
- [x ] Ensure local tests pass. (`colcon test` and `pre-commit run` (requires you to install pre-commit by `pip3 install pre-commit`)
- [ x] Commit to your fork using clear commit messages.
- [ x] Send a pull request, answering any default questions in the pull request interface.


motion_primitives_controllers/src/motion_primitives_forward_controller_parameter.yaml
```diff
@@ -5,3 +5,9 @@ motion_primitives_forward_controller:
     description: "All interfaces use the naming scheme `tf_prefix_ + motion_primitive/<interface name>`",
     read_only: true,
   }
+  has_kinematics: {
```
christophfroehlich: I think the parameter could be more descriptive, something like
```suggestion
  hardware_solves_kinematics: {
```

motion_primitives_controllers/test/test_motion_primitives_forward_controller.cpp
```diff
@@ -154,6 +154,174 @@ TEST_F(MotionPrimitivesForwardControllerTest, receive_single_action_goal)
...
+{
+  SetUpController(true);
+
+  ASSERT_TRUE(configure_succeeds(controller_));
+  ASSERT_TRUE(activate_succeeds(controller_));
+
+  MotionPrimitive primitive;
+  primitive.type =
+    static_cast<uint8_t>(motion_primitives_controllers::MotionType::LINEAR_CARTESIAN);
+  primitive.joint_positions = {1.1, 1.2, 1.3, 1.4, 1.5, 1.6};
```
christophfroehlich: why not using your helper `make_linear_joint_primitive` here?

christophfroehlich (changes requested):
Unfortunately, I don't have experience with either of these robots. Anyone else has to review this.

motion_primitives_controllers/test/test_motion_primitives_forward_controller.cpp
URJala: Because I want to make a `LINEAR_CARTESIAN` move, but use joint angles. I could make a `LINEAR_JOINT` primitive with the helper, and then change the type afterwards, if that is more clear?
Edit: I see what you mean, I do that in other places. I will change it to be like the other places.

christophfroehlich: this is what you do with accepts_linear_joint_with_pose_when_has_kinematics, right? Better use the same pattern here and there, no matter which one you decide

motion_primitives_controllers/doc/README.md
```diff
@@ -33,6 +33,9 @@ This controller provides an interface for sending motion primitives to an indust
   - `LINEAR_CARTESIAN`
   - `CIRCULAR_CARTESIAN`
 
+### Forward and inverse kinematics
+The controller has a parameter called `hardware_solves_kinematics`, which enables sending joint angles with `LINEAR_CARTESIAN` commands or sending poses with `LINEAR_JOINT` commands. This should only be set to `true` if the underlying hardware interface is capable of doing both forward and inverse kinematics.
```
fmauch, line 37: It's actually the hardware (as in "the robot", not the hardware interface as in "ros-hardware-interface"). 
```suggestion
The controller has a parameter called `hardware_solves_kinematics`, which enables sending joint angles with `LINEAR_CARTESIAN` commands or sending poses with `LINEAR_JOINT` commands. This should only be set to `true` if the underlying hardware is capable of doing both forward and inverse kinematics.
```

motion_primitives_controllers/src/motion_primitives_base_controller.cpp
```diff
@@ -173,30 +190,53 @@ bool MotionPrimitivesBaseController::set_command_interfaces()
...
       std::ignore = command_interfaces_[24].set_value(arg.value);
     }
     else
     {
       RCLCPP_WARN(get_node()->get_logger(), "Unknown additional argument: %s", arg.name.c_str());
     }
   }
+
+  // Reset unused additional arguments
+  for (size_t i = 0; i < add_args_used.size(); ++i)
```
fmauch, line 233: Instead of creating the `add_args_used` helper, you could set all three to NaN just before.

motion_primitives_controllers/src/motion_primitives_forward_controller.cpp
```diff
@@ -285,23 +286,68 @@ rclcpp_action::GoalResponse MotionPrimitivesForwardController::goal_received_cal
...
+            return rclcpp_action::GoalResponse::REJECT;
+          }
+        }
+        else
+        {
+          if (primitive.joint_positions.empty())
+          {
+            RCLCPP_ERROR(
+              get_node()->get_logger(),
+              "Primitive %zu invalid: LINEAR_JOINT requires joint_positions.", i);
```
fmauch, line 314: With this check flow feedback could potentially be unclear. If `hardware_solves_kinematics == false` and a user sends poses, the output will be "Primitive %zu invalid: LINEAR_JOINT requires joint_positions.". This is misleading if the user intentionally sent a pose, but forgot to set `hardware_solves_kinematics=true`. Maybe adding a sentence that if motion to a Cartesian path was intended (and the hardware supports that) setting `hardware_solves_kinematics` to true is required.

motion_primitives_controllers/src/motion_primitives_forward_controller.cpp
```diff
@@ -285,23 +286,68 @@ rclcpp_action::GoalResponse MotionPrimitivesForwardController::goal_received_cal
...
+              i);
+            return rclcpp_action::GoalResponse::REJECT;
+          }
+        }
+        else
+        {
+          if (primitive.poses.empty())
+          {
+            RCLCPP_ERROR(
+              get_node()->get_logger(), "Primitive %zu invalid: LINEAR_CARTESIAN requires poses.",
```
fmauch, line 347: same as above.

motion_primitives_controllers/src/motion_primitives_forward_controller_parameter.yaml
```diff
@@ -5,3 +5,9 @@ motion_primitives_forward_controller:
     description: "All interfaces use the naming scheme `tf_prefix_ + motion_primitive/<interface name>`",
     read_only: true,
   }
+  hardware_solves_kinematics: {
+    type: bool,
+    default_value: false,
+    description: "Whether or not the hardware interface that is used with this controller can do both forward and inverse kinematics. Enables sending poses for LINEAR_JOINT commands and joint angles for LINEAR_CARTESIAN commands.",
```
fmauch, line 11:
```suggestion
    description: "Whether or not the hardware that is used with this controller can do both forward and inverse kinematics. Enables sending poses for LINEAR_JOINT commands and joint angles for LINEAR_CARTESIAN commands.",
```

fmauch (changes requested):

URJala:
I would also still like to have it clarified whether it is desirable to have a circular movement be defined with joint positions. It is not possible (not nicely at least) with the current message format, where the joint positions are just a vector of numbers. I don't know if we could make a second joint positions vector, or make a vector of vector_6d? Also requires more command interfaces of course.

fmauch:
> I would also still like to have it clarified whether it is desirable to have a circular movement be defined with joint positions. It is not possible (not nicely at least) with the current message format, where the joint positions are just a vector of numbers. I don't know if we could make a second joint positions vector, or make a vector of vector_6d? Also requires more command interfaces of course.

I think, it's fair to keep that out of this PR for now.

fmauch (approved):
From my perspective this looks good. I've also tested it together with https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1942

PR #2576, opened 2026-08-24
feat: migrate diff_drive_controller and joint_trajectory_controller to new RealtimePublisher constructor

SouriRishik:
## Description

This PR migrates `diff_drive_controller` and `joint_trajectory_controller` to use the new `RealtimePublisher` constructor that creates publishers internally.

This simplifies using `RealtimePublisher` in the controllers by removing the need to manually instantiate and pass in `rclcpp::Publisher` objects, and removes redundant publisher member variables from the classes.

Note: This is the initial subset of controllers migrated.
Depends on https://github.com/ros-controls/realtime_tools/pull/573

No.

## Did you use Generative AI?

No.

## Additional Information

Verified locally that the pre-commit checks run and pass successfully.


christophfroehlich (approved):
LGTM, I'll tag the rolling version soon have broader feedback from downstream packages.

bmagyar (approved):
thank you!

PR #2578, opened 2026-08-27
Setting command interfaces to 0 instead of NaN on deactivate (mecanum_drive_controller)

nathandunk:
## Description

On the mecanum_drive_controller, on deactivate, the controller sets the command interfaces to `std::numeric_limits<double>::quiet_NaN()`. On at least one HWI (clearpath ridgeback), this results in commands being sent to the motors that run at full speed. On a different version of this controller provided by clearpath (which seems like it might have been a fork of an earlier version of this controller), [this same change](https://github.com/clearpathrobotics/clearpath_mecanum_drive_controller/pull/4/changes) was made by [my suggestion](https://github.com/clearpathrobotics/clearpath_mecanum_drive_controller/issues/3) and it resolved this issue. 

Hopefully should make robots more safe, but if some HWI was relying on this value being NaN, then could change behavior.

No

## TODOs

- [x] Fork the repository.

### DIFF
--- a/mecanum_drive_controller/src/mecanum_drive_controller.cpp
+++ b/mecanum_drive_controller/src/mecanum_drive_controller.cpp
@@ -418,8 +418,7 @@ controller_interface::CallbackReturn MecanumDriveController::on_deactivate(
   bool value_set_no_error = true;
   for (size_t i = 0; i < NR_CMD_ITFS; ++i)
   {
-    value_set_no_error &=
-      command_interfaces_[i].set_value(std::numeric_limits<double>::quiet_NaN());
+    value_set_no_error &= command_interfaces_[i].set_value(0.0);
   }
   if (!value_set_no_error)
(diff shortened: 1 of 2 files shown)


saikishor (commented):
Can you add a test for this?

I believe it is velocity controlled, so I think setting zero is fine

nathandunk:
Actually found that there was a test that addressed values after deactivation/reactivation, but it was set for checking for NaNs so I changed that to 0s. Was able to successfully run tests on the equivalent version on humble and checked that this change worked, but don't have a workflow up to run tests on rolling.

christophfroehlich (approved):
LGTM, I don't see any safety issues when setting that to zero. Any implicit contracts relying on the NaN value to auto-deactivate something seems to be a bad design too, so there is no issue in merging/backporting this IMHO.

PR #2586, opened 2026-08-31
Motion primitives forward controller error handling

URJala:
## Description
Abort active goal when execution state is `ERROR`. Currently, if a hardware interface changes the `ExecutionState` to `ERROR`, the current action will hang in limbo, and the controller will also stay in `ERROR`, unless the hardware interface changes it again. With this an active goal will be aborted, and the command interface will be set to `STOP_MOTION`, such that the controller will run through the same steps as if the user canceled the action.

Edit: I have introduced another change in this PR, as it is in the same vein of error handling. As this has had no reviews yet, I hope it is okay. 
I have added a check to see if the action is actually canceling, when the controller reaches `ExecutionState::STOPPED`. If the action is canceling, the action will be canceled, otherwise the action will be aborted. The current behavior results in the action attempting an illegal transition if the controller reaches `ExecutionState::STOPPED` without a user having requested a cancel.

Fixes # (issue)

Yes, but I believe the current behavior is unintended.

Yes, Github copilot: Claude Sonnet 5

## TODOs

- [x] Fork the repository.


motion_primitives_controllers/src/motion_primitives_forward_controller.cpp
```diff
@@ -169,6 +177,29 @@ controller_interface::return_type MotionPrimitivesForwardController::update(
...
+
+        if (has_active_goal_)
+        {
+          rt_goal_handle_.try_get(
+            [&](const std::shared_ptr<RealtimeGoalHandle> & goal_handle)
+            {
+              auto result = std::make_shared<ExecuteMotionAction::Result>();
+              // -2 is not part of the action's enumerated error codes, used here since no
+              // dedicated "execution failed" code exists.
+              result->error_code = -2;
```
urfeex: Maybe, it would be good to suggest an error code constant in the action definition.

motion_primitives_controllers/src/motion_primitives_forward_controller.cpp
```diff
@@ -169,6 +177,29 @@ controller_interface::return_type MotionPrimitivesForwardController::update(
...
+              RCLCPP_ERROR(get_node()->get_logger(), "Motion primitives execution aborted.");
+            });
+        }
+        reset_command_interfaces();
+        // cancel commands already transferred to the hw-interface
+        std::ignore =
+          command_interfaces_[0].set_value(static_cast<double>(MotionHelperType::STOP_MOTION));
+        // clear the queue (ignore return value)
+        static_cast<void>(moprim_queue_.get_latest(current_moprim_));
+        robot_stop_requested_ = true;
```
urfeex: Since this is basically a copy from code above, it might make sense to factor this out into a `cleanup_motion()` method or similar.

urfeex (commented):
In general, this looks good and makes sense. I'd suggest refactoring the `cleanup_motion()` and make a PR to control_msgs for the constant definition.

URJala:
A PR for constants in the action definition has been opened [here](https://github.com/ros-controls/control_msgs/pull/331)

PR #2587, opened 2026-08-31
Fix/jtc cancel anchor scoped

MarqRazz:
On cancel, JTC anchors its stop to `state_current_` (feedback), but the command stream is at `command_next_`. On any robot with a following error the commanded position steps backwards by exactly that error in one control period. Measured on a KUKA LBR iisy 15 R930 over RSI at 250 Hz: a −0.005568 rad step on j2 against a −0.005597 tracking error — −397 rad/s² over one 4 ms cycle, ~300× the joint's 1.32 limit. The KRC rejects the setpoint and drops the drives. Not tunable: `max_deceleration_on_cancel` and scaling shape the ramp after the join, not the step at it. `decelerate_on_cancel: false` fails identically.

## Changes
1. Don't command NaN when a state interface is exported but never written — NaN propagates silently (`std::max(0.0, NaN)` is 0.0) and fills the whole stop trajectory. Pre-existing, independent.
2. Anchor the stop to the last command **on the cancel path only**. The five fault paths and `on_activate` keep the measured state: after a tolerance violation the robot has demonstrably failed to track, so latching an unreachable command would sustain the error.
3. Name the offending joints in the diagnostics.
## Testing
A regression test per fix, each verified to fail without the change. Full JTC suite passes.

### DIFF
--- a/joint_trajectory_controller/src/joint_trajectory_controller.cpp
+++ b/joint_trajectory_controller/src/joint_trajectory_controller.cpp
@@ -1543,15 +1543,17 @@ rclcpp_action::CancelResponse JointTrajectoryController::goal_cancelled_callback
     active_goal->setCanceled(action_res);
     rt_active_goal_.writeFromNonRT(RealtimeGoalHandlePtr());
 
+    // The robot was tracking when the cancel arrived, so anchor the stop to the last commanded
+    // point to keep the command stream continuous. Fault paths deliberately do not do this.
     if (should_decelerate_on_cancel_)
     {
       // calculate stopping position based on max deceleration
-      add_new_trajectory_msg(decelerate_to_hold_position());
+      add_new_trajectory_msg(decelerate_to_hold_position(true));
     }
     else
     {
-      // hold current position
-      add_new_trajectory_msg(set_hold_position());
+      // hold last commanded position
+      add_new_trajectory_msg(set_hold_position(true));
     }
   }
   return rclcpp_action::CancelResponse::ACCEPT;
@@ -2133,12 +2135,30 @@ void JointTrajectoryController::preempt_active_goal()
   }
 }
 
+const trajectory_msgs::msg::JointTrajectoryPoint & JointTrajectoryController::select_hold_anchor(
+  const bool from_last_command) const
+{
+  // state_current_ is feedback and lags the command by the following error. Anchoring a stop to it
+  // steps the command stream back by that error in one cycle, which downstream reads as a huge
+  // acceleration. last_commanded_state_ is what was last written, so it keeps the stream continuous.
+  if (
+    from_last_command && last_commanded_state_.positions.size() >= num_cmd_joints_ &&
+    std::all_of(
+      last_commanded_state_.positions.cbegin(),
+      last_commanded_state_.positions.cbegin() + static_cast<std::ptrdiff_t>(num_cmd_joints_),
+      [](double x) { return std::isfinite(x); }))
+  {
+    return last_commanded_state_;
+  }
+  return state_current_;
+}
+
 std::shared_ptr<trajectory_msgs::msg::JointTrajectory>
-JointTrajectoryController::set_hold_position()
+JointTrajectoryController::set_hold_position(const bool from_last_command)
 {
   // Command to stay at current position. Never latch a non-finite position -- it would be written
   // straight to the command interfaces; keep the previous target for those joints instead.
-  const auto & source = state_current_.positions;
+  const auto & source = select_hold_anchor(from_last_command).positions;
   auto & hold = hold_position_msg_ptr_->points[0].positions;
   if (hold.size() != source.size())
   {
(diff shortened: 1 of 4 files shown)


christophfroehlich:
Claude can you fill the pull request template please? :stuck_out_tongue_closed_eyes:

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -2131,11 +2136,67 @@ void JointTrajectoryController::preempt_active_goal()
...
+    if (offenders_len < sizeof(offenders) - 1)
+    {
+      const int written = snprintf(
+        offenders + offenders_len, sizeof(offenders) - offenders_len, "%s%s",
+        (offenders_len > 0) ? ", " : "", params_.joints[i].c_str());
+      offenders_len =
+        (written > 0)
+          ? std::min(offenders_len + static_cast<size_t>(written), sizeof(offenders) - 1)
+          : sizeof(offenders) - 1;
+    }
```
christophfroehlich, line 2190: should we use fmt instead?
```suggestion
    if (offenders_len >= sizeof(offenders) - 1)
    {
      continue;
    }

    const auto result = fmt::format_to_n(
      offenders + offenders_len,
      sizeof(offenders) - 1 - offenders_len,
      "{}{}",
      offenders_len > 0 ? ", " : "",
      params_.joints[i]);

    offenders_len += result.size;
```

christophfroehlich (changes requested):
Is the NaN handling really necessary? It just complicated the code a lot.

I remember a discussion about the hold position being the current state or last commanded state, but can't find it anymore. Anyone else can remember that?

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -2144,11 +2205,38 @@ JointTrajectoryController::set_hold_position()
...
+  // every `t < stop_time_[i]` is false, and the whole ramp fills with a NaN hold position.
+  const auto finite_prefix = [this](const std::vector<double> & v)
+  {
+    return v.size() >= num_cmd_joints_ &&
+           std::all_of(
+             v.cbegin(), v.cbegin() + static_cast<std::ptrdiff_t>(num_cmd_joints_),
+             [](double x) { return std::isfinite(x); });
+  };
+  const bool positions_ok = finite_prefix(p0);
+  const bool velocities_ok = finite_prefix(v0);
```
bmagyar, line 2228:
```suggestion
  const bool positions_ok = isfinite(p0);
  const bool velocities_ok = isfinite(v0);
```

good opportunity to define a *non-lambda* helper function for this, we probably could use it elsewhere in the same codebase.

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -2144,11 +2205,38 @@ JointTrajectoryController::set_hold_position()
...
+  if (!positions_ok || !velocities_ok)
+  {
+    RCLCPP_ERROR_THROTTLE(
+      get_node()->get_logger(), *get_node()->get_clock(), 1000,
+      "Cannot compute a deceleration ramp: %s non-finite or wrongly sized. Does the hardware "
+      "write to every state interface it exports? Holding position instead.",
+      (!positions_ok && !velocities_ok) ? "positions and velocities are"
+                                        : (!positions_ok ? "positions are" : "velocities are"));
```
bmagyar, line 2236: It'd be friendlier to users to make the checks separately so we can report the exact error

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -2144,11 +2205,38 @@ JointTrajectoryController::set_hold_position()
...
 
 std::shared_ptr<trajectory_msgs::msg::JointTrajectory>
-JointTrajectoryController::decelerate_to_hold_position()
+JointTrajectoryController::decelerate_to_hold_position(const bool from_last_command)
 {
   double max_t_stop = 0.0;
-  const auto & p0 = state_current_.positions;
-  const auto & v0 = state_current_.velocities;
+  // Take p0 and v0 from the same point: a commanded p0 with a measured v0 leaves a slope
+  // discontinuity at the join, which is the same defect one derivative up.
```
bmagyar, line 2212:
```suggestion
```

joint_trajectory_controller/src/joint_trajectory_controller.cpp
```diff
@@ -2131,11 +2136,67 @@ void JointTrajectoryController::preempt_active_goal()
...
+  // state_current_ is feedback and lags the command by the following error. Anchoring a stop to it
+  // steps the command stream back by that error in one cycle, which downstream reads as a huge
+  // acceleration. last_commanded_state_ is what was last written, so it keeps the stream
+  // continuous.
+  if (
+    from_last_command && last_commanded_state_.positions.size() >= num_cmd_joints_ &&
+    std::all_of(
+      last_commanded_state_.positions.cbegin(),
+      last_commanded_state_.positions.cbegin() + static_cast<std::ptrdiff_t>(num_cmd_joints_),
+      [](double x) { return std::isfinite(x); }))
```
bmagyar, line 2151: good place to re-use the helper defined down below, if it wasn't an amnesiac model writing this

bmagyar (changes requested):
and please vet the narration...

PR #2592, opened 2026-09-01
[JTC] Redefine constraints.goal_time semantics and default to 10s

### LINKED ISSUE #2582: JTC: Change behavior of goal_time=0
> However, the current documented and intended behavior of goal_time = 0 is: "If set to zero, the controller will wait a potentially infinite amount of time." Some users may explicitly rely on this behavior because the action is considered complete only after all joints satisfy their configured goal tolerances, regardless of how long that takes. Changing the semantics of goal_time = 0 would therefore be a breaking change for anyone who expects the controller to keep waiting until the goal tolerances are met.

I don't think waiting an infinite amount of time is the right approach. I don't think we should allow it. For the action, the user will have to cancel it himself right?

_Originally posted by @saikishor in https://github.com/ros-controls/ros2_controllers/issues/2360#issuecomment-4645708468_

kamal2730:
## Description

Previously, `constraints.goal_time: 0.0` (the default) meant the controller could wait indefinitely for the goal to be reached. As a result, a goal that could never be reached would cause the action to hang forever.

This PR normalizes the `goal_time` semantics and changes the default to a finite value.

## Changes

- Default changed from `0.0` to `10.0` seconds.
- New semantics:
  - `goal_time < 0` → wait indefinitely (the old `0.0` behavior, now explicit).
  - `goal_time == 0` → no grace period; the goal must be reached on time.
  - `goal_time > 0` → wait for the specified grace period after the last trajectory point.
- The runtime check now uses `std::isfinite(...)` instead of `!= 0.0`, so negative values are handled correctly.
- Updated documentation (`userdoc`, `decelerate_on_cancel`, and the migration guide).
- Updated tests accordingly.

## Migration

Configurations relying on the old wait-forever default must now explicitly set:

```yaml
constraints:
  goal_time: -1.0  # wait indefinitely
```

**Note:** `cmd_timeout` only activates when `cmd_timeout > constraints.goal_time`. Therefore, `cmd_timeout` tests now explicitly set `goal_time` to a small positive value, e.g. `0.001`.

Fixes #2582

YES

NO

## TODOs

- [x] Fork the repository.

### DIFF
--- a/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
@@ -79,7 +79,7 @@ struct SegmentTolerances
  *
  * \code
  * constraints:
- *  goal_time: 1.0                   # Defaults to zero
+ *  goal_time: 1.0                   # Defaults to 10.0, negative waits indefinitely
  *  stopped_velocity_tolerance: 0.02 # Defaults to 0.01
  *  foo_joint:
  *    trajectory: 0.05               # Defaults to zero (ie. the tolerance is not enforced)
--- a/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
+++ b/joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
@@ -98,7 +98,9 @@ SegmentTolerances get_segment_tolerances(rclcpp::Logger & jtc_logger, const Para
   auto const n_joints = params.joints.size();
 
   SegmentTolerances tolerances;
-  tolerances.goal_time_tolerance = constraints.goal_time;
+  // Negative goal_time means wait indefinitely for the goal to be reached
+  tolerances.goal_time_tolerance =
+    constraints.goal_time < 0.0 ? std::numeric_limits<double>::infinity() : constraints.goal_time;
   static auto logger = jtc_logger.get_child("tolerance");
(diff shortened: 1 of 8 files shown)


christophfroehlich (commented):
@kamal2730 Please add a note in the release note that the action_execution_timeout parameter is not existing anymore (for example, users upgrading from jazzy to lyrical in the future)
@saikishor you've brought up that we never should wait for infinite time, but this proposal here might make sense to opt-in that behavior?

christophfroehlich (dismissed):

saikishor (approved):
I think it is okay to have this negative value to have infinity goal wait time, for sure some users get benifited from this

joint_trajectory_controller/include/joint_trajectory_controller/tolerances.hpp
```diff
@@ -98,7 +98,9 @@ SegmentTolerances get_segment_tolerances(rclcpp::Logger & jtc_logger, const Para
   auto const n_joints = params.joints.size();
 
   SegmentTolerances tolerances;
-  tolerances.goal_time_tolerance = constraints.goal_time;
+  // Negative goal_time means wait indefinitely for the goal to be reached
+  tolerances.goal_time_tolerance =
+    constraints.goal_time < 0.0 ? std::numeric_limits<double>::infinity() : constraints.goal_time;
```
christophfroehlich, line 103: setting goal_time to nan will work also here, but this is fine I guess

christophfroehlich (approved):
Thanks

PR #2602, opened 2026-09-11
feature(pid_controller): Optionally reset command interfaces at deactivation

christophfroehlich:
## Description
With velocity commands it is dangerous not to reset the command interfaces at deactivation, while it does make sense to keep the old command for effort or position controller.

no

GPT 5.6 for the tests


saikishor (approved):
LGTM

PR #2620, opened 2026-09-20
Warn once per velocity command timeout in diff_drive_controller

### LINKED ISSUE #2619: diff_drive_controller keeps warning about a timed-out velocity command while the robot is parked
## Bug description

Since #2503, `DiffDriveController::update_reference_from_subscribers()` logs a throttled warning whenever the last received command is older than `cmd_vel_timeout`:

```cpp
const auto age_of_last_command = time - command_msg_.header.stamp;
const bool cmd_vel_timeout_disabled = cmd_vel_timeout_ == rclcpp::Duration::from_seconds(0.0);
if (!cmd_vel_timeout_disabled && age_of_last_command > cmd_vel_timeout_)
{
  ordered_exported_reference_interfaces_[0]->set_value(0.0);
  ordered_exported_reference_interfaces_[1]->set_value(0.0);
  RCLCPP_WARN_THROTTLE(
    logger, *get_node()->get_clock(), warning_throttle_ms,
    "Velocity command timed out. Braking.");
}
```

The condition is level-triggered, so it stays true for as long as no new command arrives. A robot that is simply parked - idle, docked or charging - publishes nothing on `cmd_vel`, so an active controller keeps warning once per second indefinitely even though nothing is wrong.

On our robot (`cmd_vel_timeout: 0.5`, controller active, no navigation goal running) the current system log holds 1135 of these lines and keeps growing at one per second.

The braking behaviour itself is correct and unchanged by #2503; only the logging is new.

## Expected behavior

The warning should be edge-triggered: logged once when the command stream goes stale, and again only after a fresh command has been received and the timeout is reached anew. A parked robot should not produce an unbounded stream of warnings.

## Affected code

`diff_drive_controller/src/diff_drive_controller.cpp`

Function: `DiffDriveController::update_reference_from_subscribers()`

Introduced in #2503.

## Suggested fix

Remember whether the timeout has already been reported and log only on the transition:

```cpp
if (!cmd_vel_timeout_disabled && age_of_last_command > cmd_vel_timeout_)
{
  ordered_exported_reference_interfaces_[0]->set_value(0.0);
  ordered_exported_reference_interfaces_[1]->set_value(0.0);
  if (!command_timed_out_)
  {
    command_timed_out_ = true;
    RCLCPP_WARN(logger, "Velocity command timed out. Braking.");
  }
}
else
{
  command_timed_out_ = false;
  ...
}
```

Happy to send a PR along these lines.

## Environment

- ROS 2 Jazzy
- `diff_drive_controller` 4.42.1

otamachan:
## Description

`diff_drive_controller` warns once per velocity command timeout instead of once per throttle period for as long as the timeout lasts.

The timeout condition is level-triggered: it holds for as long as no new command arrives. A robot that is parked - idle, docked or charging - publishes nothing on `cmd_vel`, so an active controller kept logging `Velocity command timed out. Braking.` every second indefinitely. On our robot that filled the system log with thousands of identical lines while nothing was wrong.

- Log the braking warning on the transition into the timeout instead of on every update while it lasts.
- Clear that state when a usable command is applied again, and on `reset()`.

The braking behaviour itself is unchanged, so the existing tests cover it. I did try to add a test asserting the number of warnings, but a custom `rcutils` output handler is never invoked from these tests and there is no precedent for log assertions in this repository, so I left it out rather than carry the machinery. Happy to add one if you know the trick.

Fixes #2619

Yes, in the log output only. The warning is emitted once when the command stream goes stale, instead of once per second for as long as it stays stale. The braking behaviour is unchanged.

Yes. The change and its regression test were written with Claude Code (Opus), reviewed by me.

### Additional Information

Found on a robot parked on its charger: the controller stays active while nothing publishes `cmd_vel`, so the warning repeated once per second for hours.

Checked in a `ros:rolling-ros-base` container: `colcon build` and `colcon test` on `diff_drive_controller` pass (31 tests), and `pre-commit run` on the two changed files passes.

## TODOs

- [x] Fork the repository.


fmauch (approved):
seems reasonable

christophfroehlich (approved):
LGTM, thanks!

PR #2622, opened 2026-09-21
fix(mecanum_drive_controller): zero all four wheels on NaN reference …

### LINKED ISSUE #2616: MecanumDriveController may retain stale reference commands when receiving invalid values
## Bug description

`MecanumDriveController::update_reference_from_subscribers()` does not clear the exported reference interfaces when the current reference message contains one or more `NaN` values.

In this branch, no values are written:

```cpp
if (
  !std::isnan(current_ref_.twist.linear.x) && !std::isnan(current_ref_.twist.linear.y) &&
  !std::isnan(current_ref_.twist.angular.z))
{
  ordered_exported_reference_interfaces_[0]->set_value(current_ref_.twist.linear.x);
  ordered_exported_reference_interfaces_[1]->set_value(current_ref_.twist.linear.y);
  ordered_exported_reference_interfaces_[2]->set_value(current_ref_.twist.angular.z);
}
```

As a result, previously exported valid commands can remain active after an invalid or partially populated reference is received. This can cause downstream controllers in chained mode to continue using stale velocity commands.

The issue is also relevant after `reset_controller_reference_msg()`, which resets the reference fields to `NaN`, while the exported reference interfaces are not explicitly cleared by `update_reference_from_subscribers()`.

## Expected behavior

When the current reference is invalid or contains `NaN`, all exported reference interfaces should be cleared, preferably by setting them to `NaN` or another value that prevents downstream command use.

## Affected code

`mecanum_drive_controller/src/mecanum_drive_controller.cpp`

Function:

`MecanumDriveController::update_reference_from_subscribers()`

## Suggested fix

Explicitly reset all exported reference interfaces in the invalid-reference branch:

```cpp
ordered_exported_reference_interfaces_[0]->set_value(
ordered_exported_reference_interfaces_[1]->set_value(
ordered_exported_reference_interfaces_[2]->set_value(
```

A regression test should verify that a valid command is not retained after an invalid reference is received.

mibrahim-cpr:
## Description

The stop path in `update_and_write_commands` used a `||`-chain of `set_value(0.0, std::numeric_limits<unsigned int>::max())` calls to write zero to every wheel command interface:

    const bool value_set_error =
        command_interfaces_[FRONT_LEFT ].set_value(0.0, UINT_MAX) ||
        command_interfaces_[FRONT_RIGHT].set_value(0.0, UINT_MAX) ||
        command_interfaces_[REAR_RIGHT ].set_value(0.0, UINT_MAX) ||
        command_interfaces_[REAR_LEFT  ].set_value(0.0, UINT_MAX);

Because `set_value` is invoked with `UINT_MAX` as its `max_tries` cap, the retry loop is effectively infinite and the call always returns `true`; combined with `||`, the very first call short-circuits the chain and `FRONT_RIGHT`, `REAR_RIGHT`, `REAR_LEFT` are never written. Those three `hw_commands_[]` slots retain whatever the IK branch wrote to them on the previous tick, so on any tick where the reference interfaces are NaN (chained mode with the upstream controller no longer writing, or any tick after `update_reference_from_subscribers` reset the refs to NaN without immediately re-populating them) the observable at the hardware interface is `[FL=0, FR=X, RR=X, RL=X]` with X = the last IK-computed value. The `RCLCPP_ERROR_EXPRESSION(!value_set_error, ...)` guard also never fires because `value_set_error` is `true` on the success path, making the failure silent.

Replace the `||`-chain with the same non-short-circuiting `&=` accumulator pattern already used in `on_deactivate`, and drop the `UINT_MAX` `max_tries` argument (only needed to paper over the wrong operator). Every wheel command interface is now written on every tick regardless of the outcome of the previous writes, and the log fires iff any single `set_value` actually fails.

Add a regression test `when_reference_is_nan_in_chained_mode_expect_all_wheels_zeroed` that reproduces the failure via chained mode: tick 1 primes the four wheel commands with non-symmetric non-zero IK outputs, tick 2 lets the auto-reset NaN reference interfaces drive the else branch and asserts all four wheel command interfaces are exactly `0.0`. With the buggy `||`-chain the test fails with `FL=0, FR=3.5, RR=1.5, RL=2.5`; with the fix it passes.

Confirmed on hardware: on a Clearpath DO150 mecanum robot the pre-fix behavior produced `platform/motors/cmd = [0.0, X, X, X]` immediately after the teleop deadman was released and persisted until a subsequent Twist happened to send the controller through the IK branch again.

Fixes #2616 

Yes. Before this change, if the reference interfaces of an active mecanum controller went to NaN — the intended "no command / stop" state — three of the four wheels would silently continue writing their last inverse-kinematics velocity to the hardware interface, causing the robot to keep driving after teleop, autonomy, or the upstream stage of a chained controller stopped publishing. After this change, all four wheel command interfaces are written to 0.0 in that condition, matching the documented behavior.

Confirmed on hardware: on a Clearpath DO150 mecanum robot the pre-fix behavior produced platform/motors/cmd = [0.0, 1.003, 1.003, 1.003] on every tick immediately after the teleop deadman was released, and persisted until a subsequent Twist happened to send the controller through the IK branch again with a zero reference (the "self-heal" tick observed at t = 17.744 s in the diagnostic rosbag).

Yes — GitHub Copilot (Claude Sonnet 4.5) assisted with the diff to mecanum_drive_controller.cpp, the regression test in test_mecanum_drive_controller.cpp, and the corresponding FRIEND_TEST declaration in test_mecanum_drive_controller.hpp. Root-cause analysis (rosbag inspection, comparison against the packaged 4.40.1 and apt-get source 4.42.1) and the on-hardware validation were performed by the author.

### Additional Information

The on_deactivate method in this same file already uses the &= accumulator pattern; this change brings the hot-path stop code into consistency with it.
The IK branch (line ~610) still uses && and would short-circuit if FRONT_LEFT.set_value(v) ever returned false. In practice the default max_tries = 10 on the non-UINT_MAX overload of set_value makes that outcome extremely unlikely, and — unlike the else branch — the && chain does at least fail loudly via the RCLCPP_ERROR_EXPRESSION guard when it does short-circuit. I've left it alone in this PR to keep the scope tight; happy to convert it to the same &= pattern in a follow-up if reviewers prefer.
No public API or parameter changes.

## TODOs

- [x] Fork the repository.


mibrahim-cpr:
Hello, I would prefer to have the changes on the clearpathrobotics fork reviewed https://github.com/ros-controls/ros2_controllers/pull/2624

PR #2624, opened 2026-09-21
Fix/mecanum zero all wheels

### LINKED ISSUE #2616: MecanumDriveController may retain stale reference commands when receiving invalid values

mibrahim-cpr:
## Description

In chainable mode, the previous update tick resets reference_interfaces_ to NaN, so the next tick takes the else branch of update_and_write_commands and is expected to write 0.0 to every wheel command interface. The old implementation chained the four set_value(0.0) calls with `&&`:

    const bool value_set_error =
      command_interfaces_[FRONT_LEFT].set_value(0.0)  &&
      command_interfaces_[FRONT_RIGHT].set_value(0.0) &&
      command_interfaces_[REAR_RIGHT].set_value(0.0)  &&
      command_interfaces_[REAR_LEFT].set_value(0.0);

`&&` short-circuits on the first false, so if set_value on FRONT_LEFT ever returns false (its bounded CAS retry loop exhausts under contention or during a hardware-side stall), the remaining three wheels are never written and retain their last non-zero inverse-kinematics command. On a mecanum robot the operator sees three wheels keep spinning after the commanded twist is cleared.

Replace the `&&`-chain with an explicit `&=` accumulator so each wheel is written independently and the log line still fires if any set_value failed. This matches the corresponding upstream fix in mecanum_drive_controller for the jazzy `||`+UINT_MAX variant, which had the same class of defect (short-circuit skipping wheel writes) with different short-circuit semantics.

Add a regression test, when_reference_is_nan_in_chained_mode_expect_ all_wheels_zeroed, that:
  1. Activates the controller in chained mode.
  2. Tick 1: writes non-zero reference_interfaces_ and asserts all four wheel command interfaces are non-zero (IK ran).
  3. Tick 2: relies on update_and_write_commands having reset the references to NaN and asserts every wheel is commanded to 0.0.

Before the fix the test fails on tick 2 (three wheels retain their IK values); after the fix it passes.

Fixes #2616 

Yes. This is a runtime-behavior fix, not a refactor.

Before this patch, on any mecanum robot running this controller (Clearpath DO100, DO150, R100, and A300 configured with drivetrain.control: omni_4wd), if the CAS retry loop inside LoanedCommandInterface::set_value ever exhausted its bounded attempts on the front-left wheel while the controller was in the "no valid reference" branch, the remaining three wheels would keep whatever inverse-kinematics command they had on the previous tick. In practice this shows up as three wheels continuing to spin after the operator releases the deadman: reference_interfaces_ get reset to NaN, the else branch runs, command_interfaces_[FRONT_LEFT].set_value(0.0) returns false, and the &&-chain short-circuits before touching FRONT_RIGHT / REAR_RIGHT / REAR_LEFT.

After this patch, every wheel is written independently, so a failed set_value on one wheel no longer skips the others. Operator-visible behavior on a stuck-wheel event goes from "three wheels keep driving" to "all four wheels stop and one wheel logs a set_value failure". No API, message, parameter, or topic change.

Yes. GitHub Copilot (Claude Opus) was used as a pair-programming assistant to:

* localize the bug from a rosbag capture (matching the [0, non-zero, non-zero, non-zero] wheel-command pattern to the short-circuit in update_and_write_commands),
* draft the fix and the accompanying regression test,
* draft this PR description.

All code was reviewed and validated locally: the fix and test were compiled with colcon build --packages-select mecanum_drive_controller --cmake-args -DBUILD_TESTING=ON and run via ctest -R "test_mecanum_drive_controller$" -V. The regression test was also verified to fail against the unpatched &&-chain baseline (reproduces the bug) and pass after the patch.

### Additional Information

Upstream context. This fork's mecanum_drive_controller.cpp at tag 6.5.0 carries a &&-chain of set_value(0.0) calls in the else branch of update_and_write_commands. The corresponding upstream ros-controls/ros2_controllers jazzy branch (shipped as ros-jazzy-mecanum-drive-controller >= 4.40.1) carries a functionally worse ||-chain of set_value(0.0, std::numeric_limits<unsigned int>::max()) calls with the same short-circuit-skip class of defect — that one skips wheels on success rather than on failure and is what actually reproduces in the field with default parameters. Both variants are fixed by replacing the chain with an explicit &= accumulator; the corresponding upstream PR is tracked at <upstream PR link> and closes Fixes #<upstream issue>.

## TODOs

- [x] Fork the repository.


mecanum_drive_controller/src/mecanum_drive_controller.cpp
```diff
@@ -673,13 +673,24 @@ controller_interface::return_type MecanumDriveController::update_and_write_comma
...
-    const bool value_set_error =
-      command_interfaces_[FRONT_LEFT].set_value(0.0, std::numeric_limits<unsigned int>::max()) ||
-      command_interfaces_[FRONT_RIGHT].set_value(0.0, std::numeric_limits<unsigned int>::max()) ||
-      command_interfaces_[REAR_RIGHT].set_value(0.0, std::numeric_limits<unsigned int>::max()) ||
-      command_interfaces_[REAR_LEFT].set_value(0.0, std::numeric_limits<unsigned int>::max());
+    // Reset the rate-limiter history so that when a fresh non-NaN reference
+    // resumes, `limiter->limit()` starts from rest instead of slewing from the
+    // stale pre-NaN command back to the new target. Without this, releasing
+    // and re-enabling the deadman while the joystick is centered can produce
+    // a spurious wheel burst.
```
bmagyar:
```suggestion
```
let's not narrate

mecanum_drive_controller/src/mecanum_drive_controller.cpp
```diff
@@ -673,13 +673,24 @@ controller_interface::return_type MecanumDriveController::update_and_write_comma
...
+    previous_two_commands_ = std::queue<std::array<double, 3>>(
+      std::deque<std::array<double, 3>>{{{0.0, 0.0, 0.0}}, {{0.0, 0.0, 0.0}}});
+
+    // Use `&=` (not `&&`) so every wheel is actually written; an `&&`-chain
+    // short-circuits on the first failed set_value and leaves the remaining
+    // wheels at their last (non-zero) inverse-kinematics value.
```
bmagyar:
```suggestion
```

bmagyar (dismissed):

bmagyar (approved):
thank you!

